Issue #1773, opened 2025-09-27
device/telemetry: submit partitions of samples in parallel


snormore:
Similar to what we did in https://github.com/malbeclabs/doublezero/pull/1772 but for device-link telemetry, and with a lower concurrency limit to keep it more contained on the DZDs.

PR #1828, opened 2025-10-01
device/telemetry: submit partitions in parallel

snormore:
## Summary of Changes
- Update device telemetry agent to submit partitions of samples in parallel instead of sequentially, with a max concurrency of 10, similar to https://github.com/malbeclabs/doublezero/pull/1772
- Simplify and fix flush on close logic, pass a new context with 30s timeout
- Resolves https://github.com/malbeclabs/doublezero/issues/1773

## Testing Verification
- Updated existing test coverage to match changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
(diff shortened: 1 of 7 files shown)


snormore:
Before and after on devnet:

Screenshot 2025-10-06 at 11 53 38 AM

Should be more visible on testnet/mainnet too, since there are more circuits per device there which we can see in the spikiness on the samples submission chart.

PR #1833, opened 2025-10-02
activator: non-blocking

### LINKED ISSUE #1215: Activator ignores SIGTERM/SIGINT and never exits
The activator process does not terminate on SIGTERM or SIGINT (Ctrl+C). It keeps running until forced with SIGKILL (kill -9). This suggests at least one thread/task is blocking shutdown (e.g., waiting on a channel/condvar, a blocking I/O, or a join() that never returns).

Steps to Reproduce

1. Start the activator normally.
2. Send SIGTERM (kill -TERM <pid>) or press Ctrl+C in the foreground.
3. Observe that the process remains alive indefinitely.
4. Only kill -9 terminates it.

Expected Behavior

- On SIGTERM/SIGINT, activator begins graceful shutdown:
   - Broadcasts a shutdown signal to workers.
   - Stops background tasks and closes sockets.
   - Exits cleanly within 5–10s with code 0 (or a specific non-zero if failures occur).

Actual Behavior
- No clean shutdown; process stays running
- Requires kill -9.

elitegreg:
## Summary of Changes
* Replace blocking threads with async tasks
* Fixes shutdown freeze caused by blocking threads. This allows systemd to restart the process on catastropic failure.

Closes #1215

## Testing Verification
* Existing tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,8 @@ All notable changes to this project will be documented in this file.
   - Fix ProgramConfig resize during global state initialization.
 - QA
   - Traceroute when packet loss is detected
-
+- Tools
+  - Add `solana-tpu-quic-ping` tool for testing Solana TPU-QUIC connections with stats emitted periodically
 
 ## [v0.7.1](https://github.com/malbeclabs/doublezero/compare/client/v0.7.0...client/v0.7.1) – 2025-11-18
 
--- a/client/doublezero/src/command/connect.rs
+++ b/client/doublezero/src/command/connect.rs
@@ -29,7 +29,7 @@ use indicatif::ProgressBar;
 use solana_sdk::pubkey::Pubkey;
 use std::{collections::HashMap, net::Ipv4Addr, str::FromStr, time::Duration};
 
-#[derive(Clone, Debug, ValueEnum)]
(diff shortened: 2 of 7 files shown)


activator/src/activator.rs
```diff
@@ -116,18 +106,55 @@ pub async fn get_snapshot_poll(
...
+            Ok(async_client) => match async_client.subscribe().await {
+                Ok((mut subscription, unsubscribe)) => {
+                    info!("Websocket subscription established");
+                    while !stop_signal.load(std::sync::atomic::Ordering::Relaxed) {
+                        if let Some(msg) = subscription.next().await {
+                            let keyed_account = msg.value;
+                            let pubkey = keyed_account.pubkey.clone();
+                            match rpckeyedaccount_decode(keyed_account) {
+                                Ok(Some((pubkey, account))) => {
+                                    tx.send((pubkey, account)).await.unwrap();
```
vihu, line 130: You could do something like:

```rust
if let Err(e) = tx.send(...).await {
    print_some_error_or_whatever
    break;
}
```

The unwrap would crash the entire activator process otherwise

bgm-malbeclabs:
I still don't see how this is going to work if the existing tests passed and we're seeing deadlocks in testnet. Perhaps I'm missing something though.

PR #1840, opened 2025-10-04
e2e: qa test now ensures dz is disconnected and user is removed from serviceability before attempting multicast connection

nikw9944:
## Summary of Changes
* Update e2e/qa_test.go to ensure the client is disconnected before attempting a multicast connection
* Also wait for the qaagent's client IP to be removed from the serviceability smartcontract before proceeding
* This should prevent "User with different type already exists" errors

### DIFF
--- a/e2e/qa_test.go
+++ b/e2e/qa_test.go
@@ -74,7 +74,7 @@ func TestMain(m *testing.M) {
 	}
 	serviceabilityClient = serviceability.New(rpc.New(networkConfig.LedgerPublicRPCURL), networkConfig.ServiceabilityProgramID)
 
-	// Get all devices from on-chain data
+	// Get all devices from onchain data
 	ctx := context.Background()
 	data, err := serviceabilityClient.GetProgramData(ctx)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


snormore (approved):

packethog:
This error isn't from the local tunnel still being configured. It's  because the activator has yet to delete the user onchain when the connect command is issued for multicast. If we want to fix this, we should check to make sure there isn't a user onchain. Also, Juan has a recent PR that will wait for the user to be fully deleted before the disconnect command exits.

nikw9944:
I guess we need a way to find the client IP to check if the user is onchain, and look, you just added that to qaagent - https://github.com/malbeclabs/doublezero/pull/1831. Should I cut a qaagent release and use that new call to wait for the user to be removed? Juan's change looks pretty hairy, not sure we want to wait for it.

packethog:
> I guess we need a way to find the client IP to check if the user is onchain, and look, you just added that to qaagent - https://github.com/malbeclabs/doublezero/pull/1831. Should I cut a qaagent release and use that new call to wait for the user to be removed? Juan's change looks pretty hairy, not sure we want to wait for it.

That's grabbing the public IP of the QA agent because Juan needs to include it in a CLI arg. We would need to use the serviceability RPC and make sure the unicast user is completely removed onchain, post unicast disconnect.

packethog (approved):

PR #1932, opened 2025-10-16
smartcontract: check whether mgroup is_authorized correctly

vihu:
Summary
----

This change updates the authorization logic for multicast group allowlist operations (add/remove publisher and subscriber) to properly check both the multicast group owner AND the foundation allowlist from global state. Previously, only the owner could modify allowlists; now foundation-allowlisted accounts can also perform these operations.

## Changes

- Add globalstate account parameter to 4 multicast group allowlist processors:
    - add_multicastgroup_pub_allowlist
    - remove_multicast_pub_allowlist
    - add_multicastgroup_sub_allowlist
    - remove_multicast_sub_allowlist
- Allow if payer is the multicast group owner, or Allow if payer is in the foundation allowlist from global state


juan-malbeclabs (approved):
LGTM

PR #1949, opened 2025-10-18
programs: backward compatible instruction args deserialization

snormore:
## Summary of Changes
- Add a [new crate](https://github.com/malbeclabs/borsh-incremental-rs) providing the `BorshDeserializeIncremental` derive macro for incremental deserialization of structs with per-field defaults.
  - Enables decoding to succeed when new fields are added, allowing backward-compatible evolution of instruction argument types and onchain state layouts, without breaking interaction of existing clients and components.
  - Generalizes the manual partial-decode pattern used for [state objects](https://github.com/malbeclabs/doublezero/blob/main/smartcontract/programs/doublezero-serviceability/src/state/accesspass.rs#L184-L199), so we no longer have to hand-write incremental deserialization for each type.
- Update serviceability and telemetry program instruction args to use the derive.
- Resolves https://github.com/malbeclabs/doublezero/issues/1950

## Testing Verification
- Added test coverage for the new derive macro, including default fallbacks and partial input handling.

### DIFF
--- a/smartcontract/programs/doublezero-telemetry/src/instructions.rs
+++ b/smartcontract/programs/doublezero-telemetry/src/instructions.rs
@@ -69,3 +69,42 @@ impl TelemetryInstruction {
         Ok(instruction)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn test_instruction(instruction: TelemetryInstruction) {
+        let unpacked = TelemetryInstruction::unpack(&instruction.pack().unwrap()).unwrap();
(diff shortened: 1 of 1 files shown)


crates/borsh-incremental/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "borsh-incremental"
```
snormore: Should this live in it's own repo?

karl-dz (approved):

elitegreg (approved):

PR #1951, opened 2025-10-19
Add auto-assignment to CLI and validation to smartcontract for exchange.bgp_community

nikw9944:
NOTE: This PR is now deprecated in favor of https://github.com/malbeclabs/doublezero/pull/1971

## Summary of Changes
- Add auto-assignment and validation to exchange.bgp_community
- cli: `doublezero exchange create` now scans all exchanges and assigns the next available unused value in the valid range. `create` does not accept --bgp-community, but `update` does, so we can mitigate in case a bug leads to a duplicate value.
- serviceability: validates that the value of bgp.community is between BGP_COMMUNITY_MIN=10000 and BGP_COMMUNITY_MAX=10999
- Next steps in separate PRs:
   1. monitor: raise an alert on duplicate or out-of-range bgp_community;
   2. controller: add exchange routing policy configuration

## Testing Verification
* Unit tests updated/created
* Ad-hoc testing in devcontainernet


smartcontract/cli/src/exchange/create.rs
```diff
@@ -21,22 +24,35 @@ pub struct CreateExchangeCliCommand {
...
-    pub bgp_community: Option<u16>,
 }
 
 impl CreateExchangeCliCommand {
     pub fn execute<C: CliCommand, W: Write>(self, client: &C, out: &mut W) -> eyre::Result<()> {
         // Check requirements
         client.check_requirements(CHECK_ID_JSON | CHECK_BALANCE)?;
 
+        let exchanges = client.list_exchange(ListExchangeCommand)?;
+        let used_communities: HashSet<u16> = exchanges
```
elitegreg: Just in terms of algorithm here, we need to find 1 element in the set. As written, the returned exchanges is traversed once, then we do a hash lookup. But all we need to do is traverse once, so I think building the HashSet is not needed.

Issue #1960, opened 2025-10-20
client/probing: ping implementation


snormore:
ICMP ping implementation for liveness probing over doublezero interfaces, even when certain routes are not in the the kernel routing table.

Issue #1887, opened 2025-10-09
client: implement initial support for active probing


packethog:

snormore:
Tracking of remaining work via this milestone: https://github.com/malbeclabs/doublezero/milestone/29

PR #1968, opened 2025-10-20
tools/uping: initial implementation

snormore:
## Summary of Changes
- Adds `uping` tool, a minimal Linux-only raw ICMP echo library and toolset for user-space liveness probing over doublezero interfaces, even when certain routes are not in the the kernel routing table.
- Resolves https://github.com/malbeclabs/doublezero/issues/1960
- Related to https://github.com/malbeclabs/doublezero/issues/1887

## Testing Verification
- Unit tests for sender and listener implementations.
- Manual verification with `uping-send` and `uping-recv` in local devnet.

### DIFF
--- a/tools/uping/pkg/uping/listener.go
+++ b/tools/uping/pkg/uping/listener.go
@@ -4,16 +4,14 @@ package uping
 
 import (
 	"context"
-	"encoding/binary"
 	"fmt"
 	"log/slog"
 	"net"
 	"os"
-	"sync/atomic"
 	"time"
-	"unsafe"
 
-	"golang.org/x/sys/unix"
+	"golang.org/x/net/icmp"
+	"golang.org/x/net/ipv4"
 )
 
 const defaultListenerTimeout = 1 * time.Second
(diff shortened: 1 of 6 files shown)


tools/uping/pkg/uping/listener.go
```diff
@@ -0,0 +1,282 @@
...
+	if err := unix.SetsockoptInt(fd, unix.IPPROTO_IP, unix.IP_PKTINFO, 1); err != nil {
+		return fmt.Errorf("setsockopt IP_PKTINFO: %w", err)
+	}
+	// We will craft the IPv4 header ourselves to force source/dst and keep TX on the interface.
+	if err := unix.SetsockoptInt(fd, unix.IPPROTO_IP, unix.IP_HDRINCL, 1); err != nil {
+		return fmt.Errorf("setsockopt IP_HDRINCL: %w", err)
+	}
+	if err := unix.SetNonblock(fd, true); err != nil {
+		return fmt.Errorf("set nonblock: %w", err)
+	}
```
packethog: Are we able to use the listening packet conn from `ListenPacket` in `golang.org/x/net/icmp`: https://pkg.go.dev/golang.org/x/net/icmp#ListenPacket

Then use the `Control` func off the raw conn to set the appropriate socket options. It seems more idiomatic than working with the unix package directly.

PR #1971, opened 2025-10-20
Add auto-assignment and validation to smartcontract for exchange.bgp_community

nikw9944:
## Summary of Changes
- On create, serviceability sets exchange.bgp_community=globalstate.next_bgp_community and increments next_bgp_community. The system always auto-assigns so `doublezero exchange create` has no --bgp-community flag
- Bounds check (10000 - 10999) in serviceability
- CLI update flag for setting bgp_community in case a bug leads to a duplicate
- Add next_bgp_community to globalconfig with default of 10000
- Add `doublezero global-config set --next-bgp-community` so authorized users can set the value. Required due to testnet and mainnet-beta environments having existing exchanges onchain
- Next steps (in separate PRs):
   1. monitor: raise an alert on duplicate or out-of-range bgp_community;
   2. controller: add exchange routing policy configuration

## Testing Verification
* Created new tests verifying auto-assignment and bound checking

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/processors/exchange/update.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/exchange/update.rs
@@ -88,7 +88,6 @@ pub fn process_update_exchange(
 
     let mut exchange: Exchange = Exchange::try_from(exchange_account)?;
 
-    let mut globalconfig_updated = false;
     if let Some(ref code) = value.code {
         exchange.code =
             validate_account_code(code).map_err(|_| DoubleZeroError::InvalidAccountCode)?;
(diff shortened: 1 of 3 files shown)


smartcontract/programs/doublezero-serviceability/src/processors/exchange/update.rs
```diff
@@ -82,11 +102,21 @@ pub fn process_update_exchange(
...
     }
-    if let Some(ref bgp_community) = value.bgp_community {
-        exchange.bgp_community = *bgp_community;
+    if let Some(_bgp_community) = value.bgp_community {
+        exchange.bgp_community = assign_bgp_community(&mut globalconfig);
+        globalconfig_updated = true;
     }
 
     account_write(exchange_account, &exchange, payer_account, system_program)?;
+    if globalconfig_updated {
```
elitegreg: Not a big deal, but you could move the if block contents to where you assign globalconfig_updated to true above. Then no need to set a bool and check it.

elitegreg (approved):

smartcontract/programs/doublezero-serviceability/src/processors/exchange/update.rs
nikw9944: Nice catch, fixed

Issue #1509, opened 2025-09-02
controller: deprecate enable-interfaces-and-peers flag


snormore:
Deprecate the transitionary feature flag in the controller.

PR #1976, opened 2025-10-21
controller: intra-exchange routing policy

nikw9944:
## Summary of Changes
* Retrieve exchange.bgp_community from serviceability and add it to state cache
* Update config template to implement routing policy described in rfc6
* This change causes traffic between users in the local exchange to route over the internet instead of DoubleZero, saving device bandwidth and reducing latency for users

## Testing Verification
* controller unit tests and fixtures updated
* e2e fixture updates
* e2e test cases validating that two devices connected to the same exchange don't exchange routes via doublezero
* Added a 2nd device to e2e tests that require connectivity between two clients: ulti_client_test.go and user_ban_test.go
* Removed e2e/no_ifaces_peers_test.go, which conflicts with this change, which should be fine since the controller's `-enable-interfaces-and-peers` flag is [deprecated](https://github.com/malbeclabs/doublezero/issues/1509)

### DIFF
--- a/controlplane/controller/internal/controller/server.go
+++ b/controlplane/controller/internal/controller/server.go
@@ -32,6 +32,9 @@ const (
 	ISISAreaNumber      = "0000"
 	ISISSystemIDPadding = "0000"
 	ISISNSelector       = "00"
+
+	bgpCommunityMinValid = 10000
+	bgpCommunityMaxValid = 10999
 )
 
 var (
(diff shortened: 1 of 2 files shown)


controlplane/controller/internal/controller/server.go
```diff
@@ -265,6 +265,29 @@ func (c *Controller) updateStateCache(ctx context.Context) error {
...
+
+		if contributor, ok := contributorMap[device.ContributorPubKey]; ok {
+			d.ContributorCode = contributor.Code
+		} else {
+			d.ContributorCode = "unknown"
+		}
+
+		if exchange, ok := exchangeMap[device.ExchangePubKey]; ok {
+			d.ExchangeCode = exchange.Code
+			d.BgpCommunity = exchange.BgpCommunity
```
packethog: Do we need to check to make sure this is a valid community value? I thought you had some bounds checking on a different PR.

packethog (approved):

PR #1981, opened 2025-10-22
tools/uping: sender use icmp ping datagram socket with echo id demux

snormore:
## Summary of Changes
- Update uping sender to use ICMP ping datagram socket (instead of raw `ip4:icmp`) with echo ID for kernel demux
- Sender only accepts replies from given interface
- Include `Src: l.src4` in `ipv4.ControlMessage` on replies to ensure source IP is explicit and stays on the bound iface.
- https://github.com/malbeclabs/doublezero/issues/1960

## Testing Verification
- Add test coverage around expectations of accepting replies only on the right interface and demux behavior

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -35,7 +35,7 @@ jobs:
         with:
           go-version-file: go.mod
           cache: true
-      - run: make go-test nocontainertest
+      - run: sudo sysctl -w net.ipv4.ping_group_range="0 0"
   go-container-test:
     runs-on: ubuntu-latest
     needs: go-build


tools/uping/pkg/uping/sender.go
```diff
@@ -471,3 +491,26 @@ func transientSendRetryable(err error) bool {
 	return errors.Is(err, net.ErrClosed) ||
 		errors.Is(err, unix.EBADF) || errors.Is(err, unix.ENODEV) || errors.Is(err, unix.ENETDOWN)
 }
+
+// listenICMPDatagram creates an ICMP “ping” datagram socket bound to sip with sin_port=echoID.
+func listenICMPDatagram(sip net.IP, echoID uint16) (net.PacketConn, error) {
+	fd, err := unix.Socket(unix.AF_INET, unix.SOCK_DGRAM|unix.SOCK_CLOEXEC, unix.IPPROTO_ICMP)
```
packethog: Why do we need to set `SOCK_CLOEXEC` here?

snormore: Not really needed tbf, just defensive to avoid reuse across exec. Can remove if you want

PR #1984, opened 2025-10-22
monitor: publish prometheus metrics for exchange.bgp_community values that are duplicates or out-of-range

nikw9944:
## Summary of Changes
* Add two new metrics to controlplane/monitor:
    * doublezero_monitor_exchange_bgp_community_duplicates 
    * doublezero_monitor_exchange_bgp_community_out_of_range
* These metrics can be used to alert when exchange.bgp_communities are out of range or duplicated onchain


controlplane/monitor/internal/serviceability/audits.go
```diff
@@ -32,3 +40,60 @@ func checkUnlinkedInterfaces(device serviceability.Device, iface serviceability.
...
+			pubkey,
+			exchange.Code,
+			bgpCommunity,
+		).Inc()
+
+		log.Warn("exchange BGP community out of range",
+			"exchange_pubkey", pubkey,
+			"exchange_code", exchange.Code,
+			"bgp_community", exchange.BgpCommunity,
+			"valid_range", "10000-10999",
```
snormore:
```suggestion
			"valid_range", fmt.Sprintf("%d-%d", bgpCommunityMinValid, bgpCommunityMaxValid),
```

PR #2014, opened 2025-10-28
e2e: qa: wait for multicast tunnel to be up before starting connectivity test

nikw9944:
## Summary of Changes
* Update qaagent ConnectMulticast to wait for the tunnel to come up before returning to avoid connectivity test failures
* Add device code to logging output to assist with troubleshooting

### DIFF
--- a/e2e/qa_test.go
+++ b/e2e/qa_test.go
@@ -788,25 +788,29 @@ func connectHosts(t *testing.T, hosts []string, device *Device) (map[string]stri
 			return nil, nil, fmt.Errorf("failed to get status for %s: %w", host, err)
 		}
 
-		ip := getIPFromStatus(status)
-		if ip == "" {
+		if len(status.Status) == 0 {
+			return nil, nil, fmt.Errorf("no status entries returned for %s", host)
+		}
+
+		s := status.Status[0]
+		if s.DoubleZeroIp == "" {
 			return nil, nil, fmt.Errorf("failed to get IP for %s", host)
 		}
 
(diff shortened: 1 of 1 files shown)


nikw9944:
https://github.com/malbeclabs/doublezero/actions/runs/18891427991

e2e/qa_test.go
```diff
@@ -1001,53 +995,54 @@ func performPingWithRetries(t *testing.T, client pb.QAAgentServiceClient, source
...
-// connected to the same exchange, `ping -I doublezero0` will fail. This check lets us avoid that.
-func shouldUseSourceIfaceSimple(sourceDevice, targetDevice *Device) bool {
-	return sourceDevice.ExchangeCode != targetDevice.ExchangeCode
-}
-
-func getIPFromStatus(resp *pb.StatusResponse) string {
-	for _, status := range resp.Status {
-		if (status.UserType == "IBRL" || status.UserType == "IBRLWithAllocatedIP") && status.DoubleZeroIp != "" {
-			return status.DoubleZeroIp
+func waitForTunnelUp(t *testing.T, ctx context.Context, client pb.QAAgentServiceClient, host string) *pb.StatusResponse {
```
packethog: Don't we already do this in the agent in the connect unicast rpc? I don't think it returns until the tunnel status is up else it returns an error.

nikw9944: D'oh! We sure do. Now updated to do the same for qaagent's ConnectMulticast().

PR #2076, opened 2025-11-05
client/daemon: route config

bgm-malbeclabs:
## Summary of Changes
This PR introduces an unstable feature that allows a user to remove a particular set of doublezero routes. The routes to exclude are in a file called `route-config.json` and routes to be excluded are of the format `{ "exclude": ["1.2.3.3", etc]}`.

## Testing Verification
This was deployed to both devnet and testnet. It was tested by starting with no excluded routes, then adding some, then restarting the daemon and verifying that the routes were no longer in `ip routes`. Then the file was removed, the daemon was restarted and the routes were once again present in `ip routes`. Then we tested with an empty file, no file, empty json which all worked, and invalid IPs which returned the expected error.

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/processors/link/update.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/link/update.rs
@@ -24,8 +24,8 @@ pub struct LinkUpdateArgs {
     pub mtu: Option<u32>,
     pub delay_ns: Option<u64>,
     pub jitter_ns: Option<u64>,
-    pub delay_override_ns: Option<u64>,
     pub status: Option<LinkStatus>,
+    pub delay_override_ns: Option<u64>,
 }
 
 impl fmt::Debug for LinkUpdateArgs {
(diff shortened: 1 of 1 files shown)


packethog (changes requested):

packethog (dismissed):

client/doublezerod/internal/routing/config.go
```diff
@@ -0,0 +1,78 @@
...
+	if ok {
+		c.log.Info("routes: excluding configured route", "route", r.String())
+		return nil
+	}
+	return c.nlr.RouteAdd(r)
+}
+
+func (c *ConfiguredRouteReaderWriter) RouteDelete(r *Route) error {
+	return c.nlr.RouteDelete(r)
+}
```
packethog: Did we test what happens with route withdraws when a route is added to the exception list? If we never write it on addition, then try to delete it here, I think we're going to raise an error. It seems like we never want to add or delete if it's in the exception list, unless I'm missing a corner case.

packethog (approved):

PR #2122, opened 2025-11-11
Refactor device interface handling and introduce InterfaceSubType

juan-malbeclabs:
This PR introduces support for CYOA and DIA interface sub-types and updates the interface module structure accordingly. The goal is to provide a more explicit classification of interface behavior while simplifying validation and lifecycle handling in the DZD.

These changes prepare the groundwork for treating CYOA/DIA interfaces as first-class physical (or virtual) interfaces within the provisioning and activation workflows.

## Testing Verification
* test result: ok. 98 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.01s

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,14 +3,17 @@
 All notable changes to this project will be documented in this file.
 
 ## Unreleased
+
 ### Added
+
 - Validate AccessPass before client connection (CLI) ([#1356](https://github.com/malbeclabs/doublezero/issues/1356))
 
 - Onchain programs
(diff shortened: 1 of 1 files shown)


smartcontract/programs/doublezero-serviceability/src/processors/device/interface/update.rs
```diff
@@ -102,6 +108,18 @@ pub fn process_update_device_interface(
     let mut iface = device.interfaces[idx].into_current_version();
     iface.name = validate_iface(&value.name).map_err(|_| DoubleZeroError::InvalidInterfaceName)?;
 
+    if let Some(interface_sub_type) = value.interface_sub_type {
```
snormore: Want to make this a match statement instead of if/else?

PR #2169, opened 2025-11-17
doublezerod: return errors as json

nikw9944:
## Summary of Changes
* Fixed a few cases where doublezerod returned errors that were not json-formatted
* This led to swallowing errors in failed QA tests, with the QA test logging `failed to unmarshal status response: invalid character 'E' looking for beginning of value`

## Testing Verification
* No tests were updated because the conditions leading to these error cases are difficult to reproduce under test

### DIFF
--- a/client/doublezerod/internal/manager/http.go
+++ b/client/doublezerod/internal/manager/http.go
@@ -82,8 +82,9 @@ func (n *NetlinkManager) ServeStatus(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 	status, err := n.Status()
 	if err != nil {
-		w.WriteHeader(http.StatusInternalServerError)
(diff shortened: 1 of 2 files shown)


client/doublezerod/internal/manager/http.go
```diff
@@ -82,8 +82,9 @@ func (n *NetlinkManager) ServeStatus(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 	status, err := n.Status()
 	if err != nil {
-		w.WriteHeader(http.StatusInternalServerError)
-		_, _ = w.Write([]byte(fmt.Sprintf(`{"status": "error", "description": "error while getting status: %v"}`, err)))
+		slog.Error("error while getting status", "error", err)
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte(`[{"doublezero_status": {"session_status": "error"}}]`))
```
packethog: Why wasn't the previous response valid json? We're also losing context by not returning the actual error.

PR #2189, opened 2025-11-19
qa: ping loss threshold and check if routes were uninstalled

snormore:
## Summary of Changes
- Reduce per-probe ping timeout to 5s
- Add loss threshold where the test completely fails if over the threshold
- In QA agent, log if packet loss detected for visibility/tracking separate from full QA failures
- On ping failure, check installed routes and if target no longer there then log error for visibility
- Closes https://github.com/malbeclabs/infra/issues/379

## Testing Verification
- Executed these tests on devnet and testnet

### DIFF
--- a/e2e/internal/qa/client_unicast.go
+++ b/e2e/internal/qa/client_unicast.go
@@ -9,8 +9,12 @@ import (
 )
 
 const (
-	connectUnicastTimeout = 90 * time.Second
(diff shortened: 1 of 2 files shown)


ben-dz (approved):
I think this is okay, but I wonder if we should count total ping failures across all tests- My instinct says it's not concerning if we loose one ping to over one route, but if we loose one ping over every route that would warrant investigation?

PR #2290, opened 2025-11-25
fix: strict equality comparison imposed

### LINKED ISSUE #2235: In `find_interface`, using `eq_ignore_ascii_case` may incorrectly match an unrelated interface. Consider using a strict equality comparison instead

ANISH-SR:
## Summary of Changes
* Device::find_interface: Switched lookup from eq_ignore_ascii_case to strict iface.name == name comparison.
* ll interface-related processors now normalize value.name through validate_iface before calling find_interface, ensuring strict matching works consistently the codebase.
* Callers may still pass mixed-case names eg. ethernet1/1 because validate_iface canonicalizes them to the stored format eg. Ethernet1/1
* Prevents accidental case-insensitive matches while avoiding InterfaceNotFound for valid inputs.

## Testing Verification
* all rust tests green.
* all lint checks green.

Fixes #2235

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/processors/device/interface/activate.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/device/interface/activate.rs
@@ -8,7 +8,7 @@ use crate::{
 };
 use borsh::BorshSerialize;
 use borsh_incremental::BorshDeserializeIncremental;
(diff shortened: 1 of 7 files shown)


karl-dz:
Hi @ANISH-SR, we discussed this finding with Ottersec and we actually do not need to address it. We will be closing the issue. Thank you

PR #2293, opened 2025-11-25
The device interface list command is modified so that it shows all in…

juan-malbeclabs:
This pull request updates the device interface listing CLI command to support listing interfaces for all devices when no device is specified, and improves output clarity by including the device code in the display. The changes also update the tests to reflect the new output format.

### CLI Command Improvements

* The `device` argument in `ListDeviceInterfaceCliCommand` is now optional, allowing users to list interfaces for all devices if no device is specified.
* The command logic is updated to handle both single-device and all-devices listing, using `list_device` for the latter.

### Output Format Enhancements

* The `DeviceInterfaceDisplay` struct now includes a `device` field, so the output shows which device each interface belongs to. [[1]](diffhunk://#diff-1202e1ef0267b835d18d21f72877848420aea034a2e97b7a2012ef22c82e9833R30) [[2]](diffhunk://#diff-1202e1ef0267b835d18d21f72877848420aea034a2e97b7a2012ef22c82e9833R49-R62)
* Test assertions are updated to expect the new output format, which includes the device code in both table and JSON outputs.

### Dependency Updates

* Imports are updated to include `ListDeviceCommand` from the SDK, enabling listing interfaces for all devices.

## Testing Verification
* test result: ok. 102 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.01s

### DIFF
--- a/smartcontract/cli/src/device/interface/list.rs
+++ b/smartcontract/cli/src/device/interface/list.rs
@@ -1,7 +1,10 @@
 use crate::{doublezerocommand::CliCommand, validators::validate_pubkey_or_code};
 use clap::Args;
 use doublezero_program_common::types::NetworkV4;
-use doublezero_sdk::{commands::device::get::GetDeviceCommand, InterfaceType};
+use doublezero_sdk::{
+    commands::device::{get::GetDeviceCommand, list::ListDeviceCommand},
+    CurrentInterfaceVersion, InterfaceType,
+};
 use doublezero_serviceability::state::interface::{
     InterfaceCYOA, InterfaceDIA, LoopbackType, RoutingMode,
 };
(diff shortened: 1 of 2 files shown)


smartcontract/cli/src/device/interface/list.rs
```diff
@@ -88,6 +88,26 @@ impl ListDeviceInterfaceCliCommand {
     }
 }
 
+fn map_interface(iface: &CurrentInterfaceVersion, device_code: &str) -> DeviceInterfaceDisplay {
```
snormore: Want to call this something like `build_display`? It would make the code above a bit easier to read

juan-malbeclabs: Done!

PR #2304, opened 2025-11-26
RFC-27: IP Ownership Verification Service for user connection

juan-malbeclabs:
Adds **RFC-27**, which specifies an IP Ownership Verification step for DoubleZero user creation: an external IP Verification Service issues a signed `IpOwnershipProof` that the serviceability program validates **onchain** before binding a `client_ip`.

RFC: [`rfcs/rfc27-ip-verification.md`](rfcs/rfc27-ip-verification.md)

## Summary

- **Targets a concrete gap:** for wildcard access passes (`0.0.0.0` / `allow_multiple_ip`, including the `EdgeSeat` passes issued by the shred-oracle) the program accepts any globally-routable `client_ip` without verifying control of it (`create_core.rs:186-192`; `is_global` only checks routability). The proof re-introduces the missing per-IP control, closing IP squatting/DoS (User PDA is keyed by `(client_ip, user_type)`) and traffic-misdirection risks.
- **Control lives onchain, so it is not bypassable:** signature validated via the native Ed25519 precompile + Instructions sysvar introspection; an attacker cannot obtain a proof for an IP they can't originate traffic from.
- **Complements, not replaces, AccessPass:** AccessPass still gates *who* connects; the proof governs *which IP* may be bound.
- **Honest scope:** documents origin-vs-ownership limits (NAT/CGNAT/shared egress), source-IP consistency on multi-homed hosts, the centralized verifier trust root + rotation, and a legacy compatibility window.

## Changes vs. previous revision

- Renumbered RFC 12 → **RFC-27** (file rename + CHANGELOG entry). 22 was claimed on main by `rfc22-client-pim-register-beacon.md`; 27 is the next free number after RFC-26.
- Rewrote the design around an **onchain-validated** proof and reconciled it with the existing AccessPass / dynamic-IP / EdgeSeat mechanisms (previously unmentioned).
- Specified the Solana Ed25519 verification mechanism (precompile + sysvar), fixed inconsistencies (POST vs GET, removed the unverified "payer signature", `client_ip: u32` → `Ipv4Addr`), and filled the empty template sections (New Terminology, Alternatives Considered, Impact).
- Rebased onto main; the CHANGELOG entry now sits under `## Unreleased` (it had been left behind in an already-released section).

## Testing Verification

- Documentation-only change. Verified the file follows `rfcs/rfc0-template.md` (all sections present, none empty), the mermaid diagram is well-formed, and no stale "RFC 11/12/22" references to this RFC remain in `rfcs/` or `CHANGELOG.md`.
- Technical claims cross-checked against `create_core.rs` (wildcard/`allow_multiple_ip` acceptance path), `state/user.rs`, `processors/accesspass/set.rs`, the CLI `connect`/`check_accesspass` path, and the `shred-oracle` EdgeSeat flow in `doublezero-shreds`.


elitegreg (approved):
Approving. The design is sound: precompile plus Instructions-sysvar introspection is the right mechanism, the required-checks and rejection lists are complete, putting the verifier key in global state so it rotates without a program upgrade is the right call, and the Security Considerations are honest about origin-versus-ownership under NAT/CGNAT and about the centralized trust root. The gap it targets is real and I verified it against current main: the wildcard bypass in `create_user_core` is there, and `is_global` is the only validation applied to a caller-supplied `client_ip`.

Five things to address, the first two before merge.

**1. The `check_accesspass` claim is no longer true.** The Impact section says:

> Separately, `check_accesspass` should also probe the dynamic (`0.0.0.0`) AccessPass PDA, which it does not today (`smartcontract/cli/src/requirements.rs`).

It does today. `check_accesspass` calls `GetAccessPassCommand`, which prefers the UNSPECIFIED PDA and falls back to the exact-IP one. That landed in #3860 on 2026-06-09, about six months after this PR was opened. Suggest removing the sentence.

**2. The code citations have drifted.** Against current main: `create_core.rs:173-183` is now 186-197, `create_core.rs:125` is now 141, and `state/user.rs:381` (cited for `is_global`) is now 427, where 381 is unrelated deserializer code. Same cause as (1): this PR was opened 2025-11-26 and main has moved under it. Rather than patch the three line numbers, it is probably worth re-verifying the whole doc against current main, since nine months of drift may have moved more than I spot-checked.

**3. The compatibility window reopens the gap it closes.** While both the legacy no-proof flow and the new flow are accepted, a squatter simply uses the legacy path, which is the behavior the RFC exists to prevent. Suggest requiring proofs on the wildcard and `EdgeSeat` flows from day one, and applying the legacy window only to specific-IP passes, where the pass already binds the address onchain and the proof is redundant anyway.

**4. Worth resolving the replay open question here rather than deferring it.** "Should the proof bind to the specific user account (or a nonce)?" Epoch-granularity freshness is a wide window. If a proof is ever used to authorize something destructive, such as releasing an existing user account whose IP has changed hands (#4190), a proof obtained for a routine connect could be replayed within the same epoch. Binding to the target account or a nonce closes that, and the answer shapes work that depends on this RFC.

**5. The verification service has no stated owner.** The RFC introduces an off-chain component that gates every new connection, so its availability becomes a connect-path dependency, and it has no named owner, deployment story, or SLO. Worth a line in Impact.

One non-blocking note: the PR title and body say "RFC-22", but the file, the document header, and the CHANGELOG all say 27. Twenty-seven is correct (`rfc22-client-pim-register-beacon.md` already exists and the highest is 26), so only the PR text needs updating.

Also worth flagging for the implementation PR: this would be the program's first onchain Ed25519 verification. Sysvar introspection is easy to get subtly wrong around instruction-index assumptions, multiple Ed25519 instructions in one transaction, and offset parsing. That PR should get a careful security pass.

PR #2318, opened 2025-11-29
feat(verify): strict verify iface status is pending

### LINKED ISSUE #2231: In `RejectDeviceInterface`, verify that `iface.status == Pending`

ANISH-SR:
## Summary of Changes
* Added a status guard to allow rejection only when iface.status == Pending.
* Returns DoubleZeroError::InvalidStatus for all other states.
* If the interface is Pending, the existing flow remains unchanged:
  * Set status to Rejected. 
  * Write updated interface back to the device account.
* No other logic or account checks were modified.

## Testing Verification
* All rust checks green.
* All linting checks green.

Closes #2231


karl-dz:
Looks like we added a changelog check to CI since you've added this change. Could you update the changelog? Thank you

PR #2332, opened 2025-12-02
Introduce IP and Connection-Type PDA Generation to Remove Global Index Race Condition

juan-malbeclabs:
This pull request introduces a new mechanism for generating user PDAs, making them fully derived from the user’s IP address and connection type. This replaces the previous global index–based PDA computation, eliminating a race condition that occurred when multiple users were created concurrently. The new approach provides deterministic, collision-free PDA generation and aligns the account model with the upcoming IP verification and allocation workflows.

## User PDA Derivation Update

- Implemented a new PDA generation strategy where user accounts are derived from (ip, connection_type) instead of relying on a global index counter. This removes dependency on a mutable on-chain index, which previously introduced a race condition when two user creations happened at the same time.
With the new scheme, PDAs are stable, reproducible, and fully determined by user attributes, ensuring safe account creation and simplifying future migrations and lookups.

- Updated all PDA helper functions, CLI modules, and smart contract instruction handlers to use the new derivation format, replacing calls to the old index-based PDA logic. All instructions that previously relied on reading or incrementing the global index were refactored to use the new deterministic PDA seeds.

## Race Condition Elimination

- Removed the use of GlobalState.index for PDA generation and eliminated all logic that incremented or validated this value during user creation.
- The removal of the global index resolves a concurrency issue where simultaneous user creations could lead to one transaction succeeding and the other failing due to mismatched index values.
- The new deterministic PDA scheme is inherently race-free, since account derivation no longer depends on shared mutable state.

## Compatibility and Migration

- Added support for deriving both the old and new PDAs, enabling the migration process to locate and upgrade existing user accounts that were previously created using the index-based scheme.
- Test utilities were updated to use get_user_old_pda when needed, ensuring full compatibility during the migration phase.

## Integration With Migration Framework

- Migration flows detect existing users under the old PDA scheme, create their new PDAs using (IP, connection_type) seeds, and safely transition on-chain state.

## Testing Verification
* All tests were executed, and a local migration test was also performed by creating users with the previous version and running the migration command.

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/helper.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/helper.rs
@@ -72,6 +72,52 @@ where
     Ok(())
 }
 
+pub fn account_create_with_seed<'a, T>(
+    account: &AccountInfo<'a>,
+    instance: &T,
+    payer_account: &AccountInfo<'a>,
+    system_program: &AccountInfo<'a>,
+    program_id: &Pubkey,
+    new_account_signer_seeds: &[&[u8]],
+) -> ProgramResult
+where
+    T: AccountTypeInfo + BorshSerialize + Validate + Debug,
+{
+    // Validate the instance
+    instance.validate()?;
+
+    let account_space = AccountTypeInfo::size(instance);
+
+    #[cfg(test)]
+    {
+        let rent = Rent::get().expect("Unable to get rent");
+        let required_lamports = rent.minimum_balance(account_space);
+        msg!("Rent: {}", required_lamports);
+    }
+    // Create the index account
+    try_create_account(
+        payer_account.key,  // Account paying for the new account
+        account.key,        // Account to be created
+        account.lamports(), // Current amount of lamports on the new account
+        account_space,      // Size in bytes to allocate for the data field
+        program_id,         // Set program owner to our program
+        &[
+            account.clone(),
+            payer_account.clone(),
+            system_program.clone(),
+        ],
+        new_account_signer_seeds,
+    )?;
+
+    let mut account_data = &mut account.data.borrow_mut()[..];
+    instance.serialize(&mut account_data).unwrap();
+
+    #[cfg(test)]
+    msg!("Created: {:?}", instance);
+
+    Ok(())
+}
+
 pub fn account_write<'a, T>(
     account: &AccountInfo<'a>,
     instance: &T,
--- a/smartcontract/programs/doublezero-serviceability/src/pda.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/pda.rs
@@ -2,9 +2,13 @@ use std::net::Ipv4Addr;
 
 use solana_program::pubkey::Pubkey;
 
-use crate::seeds::{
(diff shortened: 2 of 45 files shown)


smartcontract/programs/doublezero-serviceability/src/processors/user/create.rs
```diff
@@ -75,11 +82,24 @@ pub fn process_create_user(
...
+    let (expected_old_pda_account, bump_old_seed) =
+        get_user_old_pda(program_id, globalstate.account_index);
+    let (expected_pda_account, bump_seed) =
+        get_user_pda(program_id, &value.client_ip, value.user_type);
+
+    let pda_ver = if user_account.key == &expected_pda_account {
+        PDAVersion::V2
+    } else if user_account.key == &expected_old_pda_account {
+        PDAVersion::V1
+    } else {
```
karl-dz: Why does this and the other create user instruction support both PDAs? Shouldn't this change only support the new PDA? Or is this support for folks that have the older client until the next minor version release?

smartcontract/programs/doublezero-serviceability/src/helper.rs
```diff
@@ -132,6 +178,62 @@ where
...
+                    payer_account.clone(),
+                    system_program.clone(),
+                ],
+                signers_seeds,
+            )?;
+        }
+    }
+
+    Ok(())
+}
```
karl-dz: Why is this method here?

smartcontract/programs/doublezero-serviceability/src/helper.rs
```diff
@@ -72,6 +72,52 @@ where
...
+    )?;
+
+    let mut account_data = &mut account.data.borrow_mut()[..];
+    instance.serialize(&mut account_data).unwrap();
+
+    #[cfg(test)]
+    msg!("Created: {:?}", instance);
+
+    Ok(())
+}
```
karl-dz: How many account creation methods are there? Can we consolidate them?

smartcontract/programs/doublezero-serviceability/src/pda.rs
```diff
@@ -44,10 +48,17 @@ pub fn get_link_pda(program_id: &Pubkey, index: u128) -> (Pubkey, u8) {
     Pubkey::find_program_address(&[SEED_PREFIX, SEED_LINK, &index.to_le_bytes()], program_id)
 }
 
-pub fn get_user_pda(program_id: &Pubkey, index: u128) -> (Pubkey, u8) {
+pub fn get_user_old_pda(program_id: &Pubkey, index: u128) -> (Pubkey, u8) {
```
karl-dz: Should this be marked for deprecation?

karl-dz (dismissed):
Also, even though it seems unlikely that this would happen, we should add a revert if actual key == new key for the migrate instruction to avoid a no-op if someone were to pass in the same account for each of these accounts

smartcontract/programs/doublezero-serviceability/src/helper.rs
juan-malbeclabs: It could be removed; it ultimately wasn’t used.
I extended both the create and write methods so that we can define the seed, which changes depending on the PDA version.

PR #2428, opened 2025-12-11
serviceability: enforce Activated status on suspend handlers

RedaRahmani:
# Title

Enforce Activated status before suspend + add negative tests
# Summary

This PR makes the suspend logic in the serviceability program a bit stricter and more consistent.

Right now, some *_suspend handlers don’t check the current status of the entity before allowing a suspend. That means it’s possible to try to suspend something that is still Pending or already Suspended, while link/suspend.rs already enforces that only Activated entities can be suspended.

Here I align all suspend handlers with that same rule and add one negative test per entity type to make sure we don’t regress in the future.

Closes : #2221.

## What changed
1. Enforce Activated before suspend

For the following handlers:

`location/suspend.rs`

`exchange/suspend.rs`

`contributor/suspend.rs`

`device/suspend.rs`

`user/suspend.rs`

`multicastgroup/suspend.rs`

I now:

Deserialize the entity

Check that `entity.status == EntityStatus::Activated`

Return` DoubleZeroError::InvalidStatus` otherwise (with a small debug log under `#[cfg(test)]`)

The check follows the same pattern already used in link/suspend.rs:
```
if entity.status != EntityStatus::Activated {
    #[cfg(test)]
    msg!("{:?}", entity);
}
```

This doesn’t change behaviour for valid suspends (entities already Activated). It just tightens validation when the status is not valid for a suspend.

2. New negative tests

I added one negative test per entity type to make the new checks explicit and guarded by tests:

`test_suspend_location_from_suspended_fails`

`test_suspend_exchange_from_suspended_fails`

`test_suspend_contributor_from_suspended_fails`

`test_suspend_device_from_pending_fails`

`test_suspend_user_from_suspended_fails`

`test_suspend_multicastgroup_from_pending_fails`

Each test:

Puts the entity in an invalid state for suspend (Pending or Suspended depending on the case),

Calls the corresponding *_suspend processor,

Asserts that it fails with `DoubleZeroError::InvalidStatus`.

This way each handler’s status check is covered and future changes will be forced to keep this behaviour.

3. Fix in device activation test setup

While working on the tests, I also fixed the account list passed to device activation.

device/activate expects:

`device`

`globalstate`

`payer`

`system_program`

The previous test setup was missing some of these, so it didn’t really mirror the on-chain invocation. I updated the test to provide the full, correct account list.

This doesn’t change behaviour, but it makes the test more realistic and closer to what actually happens at runtime.

Testing

From the smartcontract workspace:
```
# During development: specific suspend tests
cargo test -p doublezero-serviceability --test user_tests test_suspend --test-threads=1

# All suspend-related negative tests
cargo test -p doublezero-serviceability "test_suspend.*fails" --test-threads=1

# Full test suite for the serviceability crate
cargo test -p doublezero-serviceability

# Lint
cargo clippy -p doublezero-serviceability
```

# Results:

All 6 new negative tests pass

All existing tests in doublezero-serviceability pass

cargo clippy clean (no new warnings)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 - CLI
   - Remove log noise on resolve route
 - Onchain programs
+  - Enforce Activated status check before suspending contributor, exchange, location, and multicastgroup accounts
   - Removed device and user allowlist functionality, updating the global state, initialization flow, tests, and processors accordingly, and cleaning up unused account checks.
   - Serviceability: require DeactivateMulticastGroup to only close multicast group accounts when both `publisher_count` and `subscriber_count` are zero, preventing deletion of groups that still have active publishers or subscribers.
   - Deprecated the user suspend status, as it is no longer used.
--- a/smartcontract/programs/doublezero-serviceability/src/processors/contributor/suspend.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/contributor/suspend.rs
@@ -72,6 +72,13 @@ pub fn process_suspend_contributor(
     }
 
     let mut contributor: Contributor = Contributor::try_from(contributor_account)?;
+
+    if contributor.status != ContributorStatus::Activated {
+        #[cfg(test)]
+        msg!("{:?}", contributor);
+        return Err(DoubleZeroError::InvalidStatus.into());
+    }
+
     contributor.status = ContributorStatus::Suspended;
 
     try_acc_write(&contributor, contributor_account, payer_account, accounts)?;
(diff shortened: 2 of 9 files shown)


RedaRahmani:
Hi @martinsander00 , sorry for the confusion earlier. I’ve now squashed everything into one commit and re-pushed with a signed (Verified) commit.

martinsander00:
@RedaRahmani looks like everything is good to go. I missed something and that is that your CHANGELOG entry is in the 0.8.1 release section. It should be in the unreleased section. That is my bad for missing it

PR #2448, opened 2025-12-16
telemetry: add serviceability data to flow-enricher

### LINKED ISSUE #2447: enricher: fetch/cache serviceability data

packethog:
## Summary of Changes
This adds the periodic fetching and caching of serviceability data to the flow enricher which will allow us to add metadata to incoming flow data. Flow annotation will be done in an upcoming PR. Without this change, flow data is not as useful as it could be since we wouldn't know user information, exchange, region, exporting device for the flow. 

## Testing Verification
There are unit tests attached to demonstrate initial/periodic fetching of serviceability data using an interface stub.

closes #2447

### DIFF
--- a/telemetry/flow-enricher/cmd/flow-enricher/main.go
+++ b/telemetry/flow-enricher/cmd/flow-enricher/main.go
@@ -12,10 +12,10 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/gagliardetto/solana-go/rpc"
 	"github.com/malbeclabs/doublezero/config"
 	"github.com/malbeclabs/doublezero/smartcontract/sdk/go/serviceability"
 	enricher "github.com/malbeclabs/doublezero/telemetry/flow-enricher/internal/flow-enricher"
+	"github.com/malbeclabs/doublezero/tools/solana/pkg/rpc"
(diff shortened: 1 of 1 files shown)


telemetry/flow-enricher/cmd/flow-enricher/main.go
```diff
@@ -88,11 +92,21 @@ func main() {
 		os.Exit(1)
 	}
 
+	networkConfig, err := config.NetworkConfigForEnv(os.Getenv("DZ_ENV"))
+	if err != nil {
+		logger.Error("error getting network config", "error", err)
+		os.Exit(1)
+	}
+
+	serviceabilityClient := serviceability.New(rpc.New(networkConfig.LedgerPublicRPCURL), networkConfig.ServiceabilityProgramID)
```
snormore: I added a wrapper around the RPC client that includes a few retries with backoff if you want to use that here.

Something like (note `rpc` package is a different import, from our `tools/solana/pkg/rpc` dir):
```go
rpcClient := rpc.NewWithRetries(networkConfig.LedgerPublicRPCURL, nil)
serviceabilityClient := serviceability.New(rpcClient, networkConfig.ServiceabilityProgramID)
```

Issue #2465, opened 2025-12-18
programs/svc: allow setting ops manager key by contributor owner


snormore:
Change access-control in [contributor update instruction](https://github.com/malbeclabs/doublezero/blob/main/smartcontract/programs/doublezero-serviceability/src/processors/contributor/update.rs#L69-L72) to allow setting ops manager key by the `contributor.owner`, where currently the whole instruction only allows foundation allowlist member.

PR #2489, opened 2025-12-18
programs/svc: allow contributor owner to update ops manager key

snormore:
## Summary of Changes
- Allow contributor owner to update ops manager key, where before only foundation allowlist members could
- Closes https://github.com/malbeclabs/doublezero/issues/2465

## Testing Verification
- Added test coverage for the change

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/processors/contributor/update.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/contributor/update.rs
@@ -66,7 +66,7 @@ pub fn process_update_contributor(
         "PDA Account is not writable"
     );
 
-    let mut contributor: Contributor = Contributor::try_from(contributor_account)?;
+    let mut contributor = Contributor::try_from(contributor_account)?;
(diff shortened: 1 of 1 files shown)


smartcontract/programs/doublezero-serviceability/src/processors/contributor/update.rs
```diff
@@ -65,15 +65,26 @@ pub fn process_update_contributor(
         contributor_account.is_writable,
         "PDA Account is not writable"
     );
-    // Parse the global state account & check if the payer is in the allowlist
+
+    let mut contributor: Contributor = Contributor::try_from(contributor_account)?;
```
karl-dz: Nitpicking, but the type doesn't have to be defined here. Should be able to either:
```rust
let mut contributor = Contributor::try_from(contributor_account)?;
```

Or
```rust
let mut contributor: Contributor = TryFrom::try_from(contributor_account)?;
```

karl-dz (approved):

PR #2494, opened 2025-12-19
rfc: QUIC Changes

ben-dz:
## Summary of Changes
RFC defines changes required to Solana validators and transaction senders' use of QUIC to support edge filtering.

### DIFF
--- a/rfcs/rfcx-quic-changes-for-edge-filtering.md
+++ b/rfcs/rfcx-quic-changes-for-edge-filtering.md
@@ -1,0 +1,299 @@
+# DoubleZero QUIC Changes for Edge Filtering
+
+## Summary
+
+**Status: Draft**
+
+Some modifications to QUIC will be required to allow an FPGA in the DoubleZero network to provide edge filtering services for Solana validators. The overarching goal is to make minimal changes to Solana Validator and QUIC library code, as compared to Version 1 of QUIC currently used. Additionally, as much as possible to make changes in such a way that client-side (such as RPCs) code requires little to no change, or changes can be phased in. This RFC proposes making changes or restrictions in three areas of QUIC: Encryption, Flow Control, and Packet Formatting.
+
+<br>
+
+## Motivation
+
+Edge Filtration is one of the key things that DoubleZero is uniquely positioned to provide for Solana. The DoubleZero network architecture allows expensive FPGAs to provide filtering services for multiple validators at once, effectively sharing the cost of these high-priced and flexible components. In the time since the proof-of-concepts for the edge filtering were done, Solana transitioned from using straight UDP to QUIC for its TPU connections. QUIC transactions now make up >95% of Solana TPU traffic, so any effective edge filtering must be able to filter QUIC traffic. There are three distinct challenges that are imposed by QUIC as compared to UDP.
+
+1. Encryption
+    
+    By design QUIC encrypts as much of the packet and header as possible to prevent MitM attacks and network snooping. The FPGA cannot parse encrypted data without the decryption keys.
+    
+2. Flow Control
+
+    QUIC connections are "reliable" like TCP in such a way that dropped packets would cause the sender to continuously re-send dropped data. Even if that is worked-around by faking an ACK of dropped packet back towards the sender, the sender would also eventually use up all their data credits (reaching the `MAX_DATA` limit provided by the server) and stall the connection. The server only advances the `MAX_DATA` allowance once an application has retrieved the data from the protocol layer. If the data never arrives, the server will not increase `MAX_DATA` and the client will not be able to send more data.
+    
+3. Packet Formatting
+    
+    The structure and contents of packets are incredibly variable, and a parser must understand every previous frame in a packet to process the next frame. This makes them difficult to parse in an FPGA. Additionally, stream data can be fragmented arbitrarily, making it hard to analyze the stream data.
+
+<br>
+
+## New Terminology
+
+- **Client** - Transaction Sender (e.g. RPC)
+- **Server** - Validator 
+<br>
+
+## Alternatives Considered
+
+- Only filter UDP TPU traffic
+- Create a new version of QUIC that is more adapted for filtering
+- Build an all-new protocol to use for TPU traffic
+- Don't provide filtering services that requires deep packet inspection
+- Provide deep packet inspection, but close connections as the filtering method
+
+<br>
+
+## Scope
+
+This RFC does not consider _how_ Wiredancer will process packets. One way to think about the scope is: This RFC covers what would be required for the FPGA to be able to drop all streams with even numbers.
+
+<br>
+
+## Detailed Design
+
+### Architecture
+
+```
+                   ┌─────────────────────────────────────────────────────────────────┐
+                   │                                FPGA                             │
+                   │                                                                 │
+                   │     ┌────────────────────────────────────────┐                  │
+┌──────────────┐   │     │                QUIC PARSER             │                  │
+│              │   │     │     ┌─────────────┐                    │                  │
+│ RPC / Client │───│───┬─│────>|   Decrypt   │––– have keys ––––––│––––––––┐         │
+│              │<──│─┬─│─│───┐ |             |––– no keys –┐      │        │         │
+└──────────────┘   │ │ │ │   │ └─────────────┘             │      │        │         │
+                   │ │ │ │   │         ^                   │      │        │         │
+                   │ │ ↓ │   │         │                   │      │        │         │
+                   │  N  │   │    ┌────┴──────┐            │      │        │         │
+                   │  o  │   │    │ Key Store │––––┐       │      │ ┌──────────────┐ │
+                   │  t  │   │    └───────────┘    │       │      │ │              │ │
+                   │     │   │             ^       │       │      │ │              │ │
+                   │  Q  │   └ Data ┐     Keys     │       │      │ │  WireDancer  │ │
+                   │  U  │          |      |       │       │      │ │ (Dedup/Sigv) │ │
+                   │  I  │   ┌──────────────────┐  │       │      │ │              │ │
+                   │  C  │   │ Key Extract from │  │       │      │ │              │ │
+                   │ ↑ │ │   │ HANDSHAKE_DONE   │  │       │      │ │              │ │
+                   │ │ │ │   └──────────────────┘  │       │      │ │   Drop Pass  │ │
+                   │ │ │ │          ^              │       │      │ └──────────────┘ │
+                   │ │ │ │ ┌────────┘      ┌───────┘       │      │      │    │      │
+┌──────────────┐   │ │ │ │ │  ┌─────────── V ──────────────┘      │      │    │      │
+│  Validator / │───│─┴─│─│─┘  │ ┌────────────┐    ┌────────────┐  │      │    │      │
+│    Server    │<––│───┴─│────┴─│ Re-Encrypt │<─┬─│ Frame Swap │<─│──────┘    │      │
+└──────────────┘   │     │      └────────────┘  | └────────────┘  │           │      │
+                   │     │                      |                 │           │      │
+                   │     │                      └–– Approve ──────│───────────┘      │
+                   │     └───────────────────────────────────────–┘                  │
+                   │                                                                 │
+                   │                                                                 │
+                   └─────────────────────────────────────────────────────────────––––┘
+```
+
+This initial architecture passes through any non-QUIC traffic, and any traffic for which the FPGA does not have the keys. A future improvement may add an option that the server can select at the time of connection to DoubleZero to have the FPGA drop 1RTT traffic for which the FPGA does not have keys. The drawback of this would be an increased connection spin-up latency, with a benefit of further reducing bad traffic reaching the server.
+
+The "Frame Swap" takes stream frames tagged by Wiredancer and replaces them with `RESET_STREAM`.
+
+----
+  
+<br>
+
+### 1. Encryption
+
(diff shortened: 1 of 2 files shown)


rfcs/rfcx-quic-changes-for-edge-filtering.md
```diff
@@ -0,0 +1,312 @@
...
+└──────────────┘   │     │      └────────────┘  | └────────────┘  │           │      │
+                   │     │                      |                 │           │      │
+                   │     │                      └–– Approve ──────│───────────┘      │
+                   │     └───────────────────────────────────────–┘                  │
+                   │                                                                 │
+                   │                                                                 │
+                   └─────────────────────────────────────────────────────────────––––┘
+```
+
+This initial architecture passes through any non-QUIC traffic, and any traffic for which the FPGA does not have the keys. A future improvement may add a join-time option for a server to drop 1RTT traffic for which the FPGA does not have keys. The drawback of this would be an increased connection spin-up latency, for a benefit of further reducing bad traffic reaching the server.
```
alexpyattaev: definition of join-time?

rfcs/rfcx-quic-changes-for-edge-filtering.md
```diff
@@ -0,0 +1,312 @@
...
+    
+----
+
+> 💡 Agave uses `MAX_STREAMS` as the primary method of rate limiting incoming transactions per client, and uses the `MAX_DATA` as a backstop. If Agave instead set the `MAX_DATA` to a huge value, the FPGA could always use a 4k FINAL Length parameter since the `MAX_DATA` parameter would be effectively unlimited (~563 trillion 4k transactions), and the client will not overrun it. 
+Firedancer imposes no limits and sets both `MAX_DATA` and `MAX_STREAMS` to 2^62, so would already work with this approach.
+
+----
+
+As a result of this discrepancy between the two validator software stacks, there are two proposed options for the `FINAL_SIZE`: If Agave is changed to match Firedancer with a 2^62 `MAX_DATA`, then the FPGA will always set `FINAL_LENGTH` to 4k. If Agave continues to use the `MAX_DATA` backstop, then the FPGA will make a best guess based on offset+len, and it is recommended (but not necessary) that Quinn is modified based on the recommendation in #1 above.
```
alexpyattaev: agave change is trivial

rfcs/rfcx-quic-changes-for-edge-filtering.md
```diff
@@ -0,0 +1,312 @@
...
+
+#### **Necessary Changes**
+
+**Connection IDs**: The server's CID must be 8 bytes. The server's CID must not rotate throughout the life of a connection. For middlebox parsing, CID must be a fixed length, because short header packets assume the recipient knows the Destination CID length. The FPGA needs to store keys in association with a particular connection, so the CID should not rotate.
+
+💡 *As long as the Server’s CID is a fixed length, the FPGA can filter inbound traffic. If we ever need to parse data sent towards the client, the client’s CID must be fixed too. Effort should be made to encourage clients to enforce these same rules*
+
+<br>
+
+**Packet Fragmentation**: Stream frames must be 1232 bytes, except for the last frame in a stream. If a frame is shorter than 1232 bytes and does not have the FIN flag, then the FPGA must replace it with a `RESET_STREAM` to prevent abuse of the connection.
```
alexpyattaev: this is way more than needed to parse tx header. why?

rfcs/rfcx-quic-changes-for-edge-filtering.md
```diff
@@ -0,0 +1,312 @@
...
+
+> 💡 This ensures the FPGA has the most information possible as early as possible, that transactions are broken up into a predictable pattern, and that a malicious sender does not break a transaction into tiny pieces to get around filtering. Currently Agave allows smaller frames, but only four total fragments. Since there are already rules about fragmentation, this adjustment of those rules allows the Edge Filtration to be more useful and efficient. This is already usually met by a normal sender.
+
+> 👉 There is an option to instead enforce some smaller (but reasonable) minimum size for the first frame in a stream. For example, requiring all the signatures, or signatures + header of a transaction to be in the first frame. However once some size constraint must be enforced, we might as well enforce something that will make both Validator and FPGA code paths more efficient (thus optimizing for processor time, rather than network bandwidth).
```
alexpyattaev: far more reasonable!

rfcs/rfcx-quic-changes-for-edge-filtering.md
```diff
@@ -0,0 +1,312 @@
...
+
+### Sharing Session Secrets
+Session secrets will be passed to the FPGA encrypted by an FPGA pubkey so that any other snoopers of network traffic cannot intercept them. If the FPGA's private key is compromised, then it can be rotated, and the validator software updated to match the new key. Since session secrets are ephemeral, previously captured secrets have no future value. Any validator or sender with concerns that a particular session may have been compromised needs only to disconnect and reconnect to establish new secrets.
+
+### FPGA Access to Transactions
+Some in the Solana community may be concerned that the DoubleZero FPGA will have access to transaction data as it passes through. The Solana Core Dev community has agreed that since DoubleZero is a trusted contributor to the Solana ecosystem, this is acceptable. A developer of Validator software would have similar access to transaction flow. Additionally, until recently the transactions were not encrypted in the first place, and the change to QUIC for TPU was for the purpose of flow control, not encryption. Any validator who does not wish to allow the DoubleZero FPGA this access can choose not opt into edge filtering.
```
alexpyattaev: A developer of Validator software has no access to TX flow, the operator does. Devs do not distribute binaries, only source.

alexpyattaev (commented):
Thanks for putting this together. Some initial feedback below:

rfcs/rfcx-quic-changes-for-edge-filtering.md
ben-dz: I'll update it to say "an option to be selected at the time of connection to DoubleZero"

rfcs/rfcx-quic-changes-for-edge-filtering.md
```diff
@@ -0,0 +1,312 @@
...
+
+<br>
+
+#### **Optional**
+
+**Coalescing:** Short header packets must not be coalesced, even after a long-header packet. If the UDP datagram contains a QUIC short header packet, then that must be the first and only thing in the packet.
+
+**Frame Ordering:** Stream Frames must be the first thing in a QUIC packet.
+
+**Single Stream Frame per Packet**: There must be only one Stream frame in a QUIC packet.
```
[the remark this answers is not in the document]
ben-dz: This was one that had been suggested (don't recall by who) back in September and everyone seemed on board with it. In a bare-metal embedded world or an FPGA world, the added processor overhead is workable, but perhaps running on top of Linux makes that concerning?

Were transactions in separate UDP packets before QUIC?

rfcs/rfcx-quic-changes-for-edge-filtering.md
ben-dz: Fair. I will just cut this. My point was more that there were other places someone could add code to do untoward things, but not really relevant.

rfcs/rfcx-quic-changes-for-edge-filtering.md
```diff
@@ -0,0 +1,312 @@
...
+
+- if Agave continues to use `MAX_DATA` as a backstop, it will be more reliable if the changes described in (1) above are made to Quinn. 
+
+- if Quinn receives a `RESET_STREAM` with a different `FINAL_LENGTH` then it has already determined, Quinn must not issue a `FNAL_SIZE_ERROR`. This is slightly different from the previous point, and addresses an edge case where the `RESET_STREAM` replaced a packet without the Fin but, but that packet has already been received by the server.
+
+**4k Transactions & Fragmentation:** Transactions which are fragmented across multiple packets may not be dropped until the last packet depending on the criteria causing the drop. There isn’t a workaround for this other than storing and forwarding, which is not practicable for the amount of traffic a single edge filtering node might be handling. The FPGA must issue the `RESET_STREAM` as soon as it knows that a drop is desired.
```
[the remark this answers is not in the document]
ben-dz: I'm not 100% on how we're going to handle it in the FPGA either, other than that we have to find a way to do so since fragmentation is inevitable. I have a few different ideas, but I think the handling system will partly depend on the fragmentation rules settled on. I'm coming around to a possible store-forward system being able to be performant enough if we place enough limits on fragmentation.

The point here is that the FPGA will issue the `RESET_STREAM` as soon as it knows the transaction is bad- which could be subbing for any frame in a multi-frame stream. So in making modifications to a receiving validator ahead of edge filtering support, software needs to account for three possibilities:
- FPGA subs reset_stream for first frame, but delivers the rest of the frames.
- FPGA subs reset_stream for a middle frame, having delivered some frames already, and delivering some additional one(s) after.
- FPGA subs reset_stream for last frame, having already delivered all preceding data in the stream.

The first and last are most likely scenarios.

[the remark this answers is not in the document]
alexpyattaev: delivering anything for a stream after reset is a protocol breaking change. hacking the server to ignore it is possible but not desirable. I'd prefer if FPGA zeroed out the payload or just dropped the last fragment. Granted it would not help with bandwidth much, but the server is welcome to blacklist the peer to save on bandwidth.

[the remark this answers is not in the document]
alexpyattaev: > A subsequent frame with a FIN bit would contain a smaller final size than the FINAL_SIZE included in the reset_stream.

yeah in this case its ok but then it will all get delivered to the application.... not ideal

> imo, I'd love to just require server-sides to never generate a final_size_error. Generating that error requires keeping state for closed streams, which is a waste anyway. 

maybe worth hacking it, yes,

rfcs/rfcx-quic-changes-for-edge-filtering.md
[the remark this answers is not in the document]
alexpyattaev: yes, it was 1 tx per packet. wasting iops is fine i guess, just not desirable for a niche feature. + requires hacking client side code. i'd try to avoid.

rfcs/rfcx-quic-changes-for-edge-filtering.md
ben-dz: Was the idea that with once 4k transactions exist there would be something like 16 allowable fragments (since ~1K transactions -> 4 fragments) to basically allow any normal fragmentation to happen, but prevent egregious cases?

For Quinn, in quinn-proto/src/connection/mod.rs, the populate_packet function fills in all other frames, then puts stream frame(s) in any leftover space. I'd suspect that as transactions get longer, it will lead to more fragmentation, since there will be housekeeping inserted into many packets containing bits of stream, shortening the space for stream data per packet.

Without sample wire data to look at, intuitively I see it being likely that streams will start with a small fragment in Quinn's implementation since a stream is likely to wind up tacked onto the end of a packet that contains housekeeping and the end of some other stream. Adjusting frame::Stream::SIZE_BOUND could be a quick hack for ensuring there's space for a larger fragment, but wouldn't let a whole transactions that is small fill in smaller spaces.

It seems like enforcing any kind of fragmentation rule that isn't just ruling out extreme edge cases is going to require more from txn senders that has previously been required.

alexpyattaev: > Was the idea that with once 4k transactions exist there would be something like 16 allowable fragments (since ~1K transactions -> 4 fragments) to basically allow any normal fragmentation to happen, but prevent egregious cases?

Something like that, yes. 

> For Quinn, in quinn-proto/src/connection/mod.rs, the populate_packet function fills in all other frames, then puts stream frame(s) in any leftover space. I'd suspect that as transactions get longer, it will lead to more fragmentation, since there will be housekeeping inserted into many packets containing bits of stream, shortening the space for stream data per packet.

any quic implementation is likely to work similarly

> It seems like enforcing any kind of fragmentation rule that isn't just ruling out extreme edge cases is going to require more from txn senders that has previously been required.

Yes, and that is a major concern when it comes to interop with other implementations.

ben-dz: Let me ruminate on this one a little longer. I may need to get further into implementation to see what's necessary vs nice-to-have.

Perhaps the reasonable ask is something like a 200 byte minimum size. That's roughly the minimum transaction size, In Quinn it could be implemented with frame::Stream::SIZE_BOUND being set to 225 instead of 25.

rfcs/rfcx-quic-changes-for-edge-filtering.md
```diff
@@ -0,0 +1,299 @@
...
+└──────────────┘   │     │      └────────────┘  | └────────────┘  │           │      │
+                   │     │                      |                 │           │      │
+                   │     │                      └–– Approve ──────│───────────┘      │
+                   │     └───────────────────────────────────────–┘                  │
+                   │                                                                 │
+                   │                                                                 │
+                   └─────────────────────────────────────────────────────────────––––┘
+```
+
+This initial architecture passes through any non-QUIC traffic, and any traffic for which the FPGA does not have the keys. A future improvement may add an option that the sefver can select at the time of connection to DoubleZero to have the FPGA drop 1RTT traffic for which the FPGA does not have keys. The drawback of this would be an increased connection spin-up latency, with a benefit of further reducing bad traffic reaching the server.
```
martinsander00: Typo I think?

> A future improvement may add an option that the sefver can select

serfver -> server

ben-dz: Whoops. Fixed.

ben-dz:
Closing this for now. Will reopen if we circle back to TPU path transaction filtering.

PR #2537, opened 2025-12-30
e2e: qa alldevices test: minimize latency between qa clients and devices

nikw9944:
## Summary of Changes
* Update the QA alldevices test to minimize latency between qa clients and devices

Process:
1. When the alldevices test starts, call GetLatency for each client
1. Create a list of devices for each host. 
    1. If there are multiple hosts with <50ms latency for that device, assign the device to the host with the fewest devices
    1. Otherwise, associate each device with the host with the best latency
1. Loop through n batches, where n is the largest number of devices associated with a host
    1. Look at each host
        1. If there are >= 2 hosts, and the current host has 0 devices left to test, remove the current host
        1. If there are 2 hosts, and the current host has 0 devices left to test, test the previously tested device again
     1. Test all hosts in current batch

Next steps:
1. Once we add more QA clients, we can further decrease the latency threshold.


snormore (approved):
:shipit:

PR #2540, opened 2026-01-01
feat: verification of side_z in AcceptLink

### LINKED ISSUE #2214: Verify that `link.side_z_pk == side_z_account.key` in `AcceptLink`

ANISH-SR:
## Summary of Changes
* Added an explicit check in process_accept_link to enforce link.side_z_pk == *side_z_account.key.
* The instruction now fails early with DoubleZeroError::InvalidAccountOwner if the Side-Z account does not match the link.
* Prevents accepting a link with a Side-Z device that differs from the one recorded on-chain.

## Testing Verification
* Existing tests pass.

Closes #2214

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/processors/link/accept.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/link/accept.rs
@@ -73,7 +73,6 @@ pub fn process_accept_link(
         return Err(DoubleZeroError::InvalidStatus.into());
     }
 
-    // Validate that the stored side_z_pk matches the provided side_z_account
     if link.side_z_pk != *side_z_account.key {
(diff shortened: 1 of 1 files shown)


smartcontract/programs/doublezero-serviceability/src/processors/link/accept.rs
```diff
@@ -75,6 +75,11 @@ pub fn process_accept_link(
         return Err(DoubleZeroError::InvalidStatus.into());
     }
 
+    // Validate that the stored side_z_pk matches the provided side_z_account
```
martinsander00: On the same note as my last comments in other issues, we can remove this comment too. Also make sure to sign the commit :)

Issue #2388, opened 2025-12-10
Phase 0: add first_free_index optimization to IpAllocator


vihu:
RFC11 specifies O(1) allocation optimization via `first_free_index: u32` field. On allocate, start scan at index; on deallocate, update if freed bit is before current index.

PR #2563, opened 2026-01-06
Resource Extension: add first_free_index optimization

elitegreg:
## Summary of Changes
* Adds a first_free_index member to the ID/IP Allocators to track the next free (u64) index in the allocation bitmap
* Implements a minimal optimization:
  * allocation skips ahead to the first free index
  * deallocation stores the deallocated from index as the first free index
  * If a index fills, just set next free to zero.
  * TODOs for improving in the future

## Testing Verification
* Existing tests pass

References #2388


juan-malbeclabs (approved):
LGTM

PR #2567, opened 2026-01-06
RFC 11: On chain activation for MulticastGroup entity

### LINKED ISSUE #2412: Phase 1: Modify `ActivateMulticastGroup` to allocate from ResourceExtension
Accept optional global ResourceExtension. Allocate multicast_ip from global Multicast bitmap.

vihu:
# Summary

Fixes #2412

Implements RFC-11 on chain activation for MulticastGroup entity - enabling on-chain IP allocation via ResourceExtension bitmaps.

- Add `multicast_slot: u32` field to MulticastGroup account for tracking bitmap slot index (enables deallocation)
- Modify `ActivateMulticastGroup` to optionally accept ResourceExtension account and allocate multicast_ip from bitmap
- Extend authorization to allow `foundation_allowlist` members (in addition to `activator_authority_pk`)
- Update SDK with `resource_extension_pubkey` parameter for `ActivateMulticastGroupCommand`

## Backward Compatibility
- When no ResourceExtension account is provided, uses the legacy path (provided `multicast_ip`, `multicast_slot = u32::MAX`)
- Deserialization defaults `multicast_slot` to `u32::MAX` for existing on-chain accounts

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/tests/test_helpers.rs
+++ b/smartcontract/programs/doublezero-serviceability/tests/test_helpers.rs
@@ -262,36 +262,6 @@ pub fn create_transaction_with_extra_accounts(
     )
 }
 
-/// Execute a transaction with optional extra accounts appended after payer and system_program.
-#[allow(dead_code)]
-pub async fn execute_transaction_with_extra_accounts(
-    banks_client: &mut BanksClient,
-    _recent_blockhash: solana_program::hash::Hash,
-    program_id: Pubkey,
-    instruction: DoubleZeroInstruction,
-    accounts: Vec<AccountMeta>,
-    payer: &Keypair,
-    extra_accounts: Vec<AccountMeta>,
-) {
-    print!("➡️  Transaction {instruction:?} (with extra accounts) ");
-
-    let recent_blockhash = banks_client
-        .get_latest_blockhash()
-        .await
-        .expect("Failed to get latest blockhash");
-    let mut transaction = create_transaction_with_extra_accounts(
-        program_id,
-        &instruction,
-        &accounts,
-        payer,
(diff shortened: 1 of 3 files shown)


elitegreg:
multicast_slot is not needed. The index is computed from the allocated address at time of deallocate. All of the multicast_slot code can go away.

vihu:
> multicast_slot is not needed. The index is computed from the allocated address at time of deallocate. All of the multicast_slot code can go away.

Roger, I'll address this shortly

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/activate.rs
```diff
@@ -36,6 +40,8 @@ pub fn process_activate_multicastgroup(
     let globalstate_account = next_account_info(accounts_iter)?;
     let payer_account = next_account_info(accounts_iter)?;
     let _system_program = next_account_info(accounts_iter)?;
+    // Optional: ResourceExtension account for on-chain IP allocation
```
juan-malbeclabs: Could you add this account before the payer to keep the rule that the last two accounts are always the payer and the system? In other cases, I use the account count to validate whether there is an extra account.

ie: https://github.com/malbeclabs/doublezero/blob/6f7e87598aaec04937a30679be4a222dbb5590bc/smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs#L85

vihu: Fixed in [85d5b98](https://github.com/malbeclabs/doublezero/pull/2567/commits/85d5b9827da71fc6f6be2542f5d749af41936c4e)

PR #2582, opened 2026-01-08
e2e: QA alldevices bugfixes

nikw9944:
## Summary of Changes
* Fix cases where the report shows success even when tests failed
* Ensure we don't have a stale value for doubleZeroIP; when a client is assigned the same host twice in a row, the 2nd batch should not assume it connected successfully in the 1st batch
* Don't connect more than 1 client to an exchange that has an allocate-addr client connected, since IBRLWithAllocatedIP mode doesn't provide intra-exchange connectivity


nikw9944:
Planning to add unit tests in a follow-up PR

PR #2627, opened 2026-01-14
fix: enforce zero pub/sub counts on DeactivateMulticastGroup

### LINKED ISSUE #2219: In `DeactivateMulticastGroup`, verify that both `multicastgroup.publisher_count` and `multicastgroup.subscriber_count` are empty

ANISH-SR:
## Summary of Changes
* DeactivateMulticastGroup now only succeeds when both publisher_count == 0 and subscriber_count == 0 (in addition to status == Deleting).
* If either count is non zero, the instruction fails and the account is not closed.
* Regression test added: New test verifies deactivation fails when publisher/subscriber counts are non-zero and passes when they’re zero.
* Updated Changelog.

## Testing Verification
* Existing rust + lint checks pass.

Closes #2219


juan-malbeclabs (approved):
LGTM

PR #2629, opened 2026-01-14
simultaneous tunnels RFC (#2628)

bgm-malbeclabs:
## Summary of Changes
RFC 

Closes https://github.com/malbeclabs/doublezero/issues/2628

## Testing Verification
N/A

### DIFF
--- a/rfcs/rfcxxx-simultaneous_tunnels.md
+++ b/rfcs/rfcxxx-simultaneous_tunnels.md
@@ -1,0 +1,90 @@
+# Simultaneous Tunnels 
+
+## Summary
+
+**Status: `Draft`**
+
+When the RFC has been implemented, a user will be able to have simultaneous tunnels. This will be a single unicast and a single multicast-capable tunnel. The previous work that has been done to enable multicast using PIM (protocol independent multicast) and RP (rendezvous point) will now be extended to enable multiple tunnel support.
+
+Note that this RFC does not include support for a user to become both a subscriber and a publisher for multicast feeds. That will be done in a subsequent RFC. The main reason for this bifurcation is that the changes in this RFC are largely internal to the protocol and network. In contrast, the multicast pub/sub changes are largely user-facing. More user feedback should be solicited, and more thought is needed about how the CLI should be enhanced, especially with other ongoing efforts like multi-tenancy and a push towards a single monorepo and CLI.
+
+## Motivation
+
+There are several desired use cases for DoubleZero that require the ability to transmit/receive both unicast traffic and multicast traffic. Currently, a user can have either a unicast or multicast tunnel but not both. Due to the implementation details of the IBRL mode of DoubleZero, which allows users to reuse their public IP addresses to communicate across the network, we're constrained in the ability to use a single tunnel interface to handle both unicast and multicast traffic. 
+
+## New Terminology
+
+N/A
+
+## Alternatives Considered
+
+1. **Only support one tunnel on a user's machine.** The ability to use a single tunnel for both unicast and multicast traffic would require the capability to support multicast signaling inside of the isolated routing table used for IBRL mode. There are a handful of unworkable ways to do this:
+   - Use a GRE overlay internal to DoubleZero, which would require creating N X copies of multicast traffic across the core of DoubleZero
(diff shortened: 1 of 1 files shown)


rfcs/rfcxxx-simultaneous_tunnels.md
```diff
@@ -0,0 +1,90 @@
...
+2. **Client**: Updates the CLI to provision a second tunnel by locating an available tunnel endpoint on the user's current device, or falling back to a nearby device within a 5ms latency threshold. The existing check preventing multiple tunnels will be removed.
+
+3. **Serviceability**: Introduces multicast tunnel slot tracking on the smart contract to control multicast user capacity per device, with a default of 0 slots for newly added devices.
+
+4. **Network**: Requires contributors to configure at least two public IPv4 addresses per device as tunnel endpoints, registered on the smart contract with the `user-tunnel-endpoint` tag. Separate max-user counts for unicast and multicast (default: 48 each) enable capacity planning.
+
+### Application Changes
+
+#### Activator
+* If the user is a multicast publisher, the activator will need to assign a DoubleZero IP from the next available global pool listed in the global config portion of the serviceability program. The range of addresses used for this is TBD.
```
thijsvanemmerik: range of addresses used for this is proposed: 147.51.126.0/23 according to rule 74

rfcs/rfcxxx-simultaneous_tunnels.md
```diff
@@ -0,0 +1,90 @@
...
+* On the smartcontract, these IP addresses should be registered as CYOA interfaces and tagged as `user-tunnel-endpoint`.   The `ip-net` flag should be used to allow contributors to assign their own address space.  
+* `user-tunnel-endpoint` IP addresses should be configured on Loopback interfaces to reduce the risk of an outage if a physical interface goes down.  However, if there is only a single CYOA interface, a `user-tunnel-endpoint` may be configured on a physical interface.
+* To support capacity planning, the smart-contract should support distinct max-user counts for unicast and multicast.  By default, both values should be set to 48.
+* Policing of user-tunnels, particularly important for multicast tunnels, will be addressed in a separate RFC.
+* Multicast publishers should be assigned an IP address from a global pool of address space defined in `global-config`.  It is proposed that this is set to 147.51.126.0/23, allowing 512 distinct multicast publishers.  If required to expand this block, there should be an option for including additional ranges.
```
thijsvanemmerik: "512 distinct multicast publishers" when there are no multicast subscribers right? They both seem to get assigned addresses from the same pool?

rfcs/rfcxxx-simultaneous_tunnels.md
```diff
@@ -0,0 +1,90 @@
...
+  1. Determine if the user has an existing tunnel to the device, and if so, attempt to find a second tunnel termination point in the interfaces list.
+  2. If there are no additional tunnel endpoints on the device, find the next lowest latency device within a 5ms latency threshold.
+  3. If no devices meet this criteria, return an error message to the user.
+* doublezerod will need to remove the check in place to prevent an additional tunnel from being provisioned. Most of the implementation work to support simultaneous tunnels was done prior.
+
+#### Controller
+* We don't believe there is any controller work that needs to occur for this feature work as this is not scoped for being a multicast publisher/subscriber over the same tunnel interface. 
+
+#### Serviceability
+* Contributors need the ability to add a second publicly reachable IP per device as a loopback interface and flag the interface as a tunnel endpoint. We believe this support is already in place.
```
thijsvanemmerik: should we have an extra flag to mark it specifically as multicast endpoint? I'm just thinking if a contributor creates a device it sets a public IP. Will that IP always be the one used for unicast tunnels? What if a contributor registers 3 CYOA interfaces with all `--user-tunnel-endpoint` set. Would the second one be used as multicast tunnel endpoint? 

Or can mutlicast and unicast tunnels be created by the client to each of the addresses, meaning interface 1 can serve unicast and multicast tunnels? The downside of this will be no clear distinction between types of tunnels to a specific interface, but the upside could be better load distribution accross interfaces and more efficient tunnel slots usage per DZD. Lets say in a region there are 60 unicast tunnels to a DZD and 0 multicast tunnels, then 12 tunnels need to connect to a less optimal DZD if we have configured 48 unicast and 48 multicast tunnel slots.

PR #2667, opened 2026-01-18
chore: Remove contributor modification from Device Update and Link Update

### LINKED ISSUE #2648: Remove contributor modification from Device Update and Link Update
The ability to change the contributor for a Device and a Link was introduced when the Contributor account was implemented. This was required at the time to migrate existing Devices, but it is no longer needed.

To avoid unintended changes and simplify the model, the option to modify the contributor must be removed from both update flows.

Scope:

- Remove contributor modification logic from Device Update
- Remove contributor modification logic from Link Update
- Ensure existing behavior remains unchanged for all other fields
- Update tests accordingly

Notes:

Contributor changes should not be allowed via update instructions going forward
Ownership and contributor relationships should remain immutable after creation

sranish:
## Summary of Changes
* Removed contributor changes from UpdateDevice and UpdateLink so contributor ownership is immutable after creation.
* Update flows simplified: all other fields behave exactly as before, contributor fields are ignored.
* contributor options removed from update commands and tests updated accordingly.

## Testing Verification
* all existing rust checks + lint checks pass.

Resolves #2648


ben-dz:
Sorry this sat unreviewed for so long. Thanks for putting it together, but I think the link side got addressed separately, and device side got moved to foundation-allowlist only, which is probably a useful contingency, so I'm going to close this.

PR #2737, opened 2026-01-28
client/cli: support multiple multicast groups per publisher/subscriber

snormore:
## Summary of Changes
  - Client `connect multicast` now accepts multiple group codes, joining each group
  - CLI `user subscribe` accepts multiple `--group` flags
  - Previously, connecting to a second group would error; now it joins any new groups
  - E2E tests updated for multi-group scenarios with control plane verification (PIM, MSDP, mroutes, BGP VRF)
  - E2E devnet fixes: race condition in misc network creation, retry transient docker inspect errors

  ## Testing Verification
  - Unit tests updated for multi-group support, including new `test_cli_user_subscribe_multiple_groups`
  - E2E publisher and subscriber tests extended to verify two multicast groups with control plane checks

### DIFF
--- a/e2e/ibrl_multicast_coexistence_test.go
+++ b/e2e/ibrl_multicast_coexistence_test.go
@@ -438,16 +438,23 @@ func verifyMulticastSubscriberPIMAdjacency(t *testing.T, log *slog.Logger, devic
 		pim, err := devnet.DeviceExecAristaCliJSON[*arista.ShowPIMNeighbors](t.Context(), device, arista.ShowPIMNeighborsCmd())
 		require.NoError(t, err, "error fetching pim neighbors from doublezero device")
 
-		neighbor, ok := pim.Neighbors[expectedLinkLocalAddr]
+		defaultVRF, ok := pim.VRFs["default"]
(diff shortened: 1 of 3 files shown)


client/doublezero/src/command/connect.rs
```diff
@@ -503,33 +520,52 @@ impl ProvisioningCliCommand {
...
+                };
+
+                // Wait for user to be activated before subscribing to additional groups
+                if mcast_group_pks.len() > 1 {
+                    self.poll_for_user_activated(client, &user_pk, spinner)?;
                 }
-            }?,
+
+                // Subscribe to remaining groups
+                for mcast_group_pk in mcast_group_pks.iter().skip(1) {
```
martinsander00: The existing user branch since it uses contains() does some filtering if the input has a duplicate group. But I think this branch doesn't do that

martinsander00: Ah I see you wrote a `test_connect_command_multicast_publisher_deduplicates_groups`, should we still allow duplicated input or throw error?

PR #2753, opened 2026-01-29
fix(smartcontract): reserve first IP of DzPrefixBlock for device

vihu:
## Summary of Changes

When a DzPrefixBlock resource is created, index 0 is automatically allocated/reserved, ensuring user allocations start from the second IP and the first ip is reserved.


elitegreg (approved):
With the way you implemented it here, it will effect `doublezero resource create --resource-type dz-prefix-block`, which is fine, and means you don't need to make any change to the migration

PR #2771, opened 2026-01-30
feat(cli): remove one tunnel restriction (#2725)

bgm-malbeclabs:
## Summary of Changes
This PR adds the ability from the CLI and the smart contract to add simultaneous tunnels: a single unicast and a single mulitcast tunnel (subscriber or publisher only). This also ensures that a second tunnel appears on a second device.

This also ensures that a mcast subscriber gets a device derived by tunnel_dst rather than dz_ip since subscribers aren't assigned a dz_ip.

This PR is split out from https://github.com/malbeclabs/doublezero/pull/2728 so that we can roll this out more safely and in two discrete steps (cli changes then activator,controller changes)

Closes https://github.com/malbeclabs/doublezero/issues/2725

## Testing Verification
* E2E tests were extended to include creating multiple tunnels for a single user
* The coexistence tests that were skipped are now enabled and pass
* This also adds tests to verify that `pubkey` is always the last deserialized field in both go and rust
* Claude helped extend the tests

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -19,27 +19,62 @@ fi
 
 # Check if any meaningful files are staged (excluding the changelog itself)
 has_changes=false
-changelog_updated=false
+changelog_staged=false
 
 for file in $STAGED_FILES; do
     if [[ "$file" == "CHANGELOG.md" ]]; then
-        changelog_updated=true
+        changelog_staged=true
     else
         has_changes=true
     fi
 done
 
-# If there are changes but changelog wasn't updated, fail
-if [ "$has_changes" = true ] && [ "$changelog_updated" = false ]; then
-    echo ""
-    echo "❌ CHANGELOG.md not updated"
-    echo ""
-    echo "Please update CHANGELOG.md with your changes."
-    echo ""
-    echo "To skip this check (use sparingly):"
-    echo "   git commit --no-verify"
-    echo ""
-    exit 1
+# If changelog is staged in this commit, we're good
+if [ "$changelog_staged" = true ]; then
+    exit 0
 fi
 
-exit 0
+# If no meaningful changes, we're good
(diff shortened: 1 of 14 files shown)


e2e/ibrl_multicast_coexistence_test.go
```diff
@@ -87,6 +78,331 @@ func TestE2E_IBRL_AllocatedAddr_Multicast_Publisher_Coexistence(t *testing.T) {
...
+		"deviceIP", actualMcastDevice.CYOANetworkIP,
+		"ibrlDeviceCode", ibrlDevice.Spec.Code,
+		"ibrlDeviceID", ibrlDevice.ID,
+		"ibrlDeviceIP", ibrlDevice.CYOANetworkIP)
+	waitForAgentConfigWithClient(t, log, dn, actualMcastDevice, client)
+	log.Info("--> Agent config pushed to multicast device")
+
+	// Give time for agent to apply the configuration
+	log.Info("==> Waiting for agent to apply configuration")
+	time.Sleep(10 * time.Second)
```
snormore: Can this be removed? We're waiting for tunnels below it anyway

.githooks/pre-commit
```diff
@@ -19,27 +19,62 @@ fi
...
+# If we get here, changelog hasn't been updated in this branch
+echo ""
+echo "❌ CHANGELOG.md not updated"
+echo ""
+echo "Please update CHANGELOG.md with your changes."
+echo ""
+echo "To skip this check (use sparingly):"
+echo "   git commit --no-verify"
+echo ""
+exit 1
```
snormore: Newline here 😅

e2e/ibrl_multicast_coexistence_test.go
bgm-malbeclabs: yep most probably

PR #2849, opened 2026-02-06
Multi-tenancy: vrf IDs resource created

elitegreg:
## Summary of Changes
* Adds resource for allocating VRF IDs

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/processors/globalconfig/set.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/globalconfig/set.rs
@@ -61,6 +61,7 @@ pub fn process_set_globalconfig(
     let multicastgroup_block_account = next_account_info(accounts_iter)?;
     let link_ids_account = next_account_info(accounts_iter)?;
     let segment_routing_ids_account = next_account_info(accounts_iter)?;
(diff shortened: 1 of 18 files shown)


smartcontract/cli/src/resource/allocate.rs
```diff
@@ -39,9 +39,10 @@ impl From<AllocateResourceCliCommand> for AllocateResourceCommand {
             | ResourceType::DzPrefixBlock => {
                 IdOrIp::Ip(x.parse::<NetworkV4>().expect("Failed to parse IP address"))
             }
-            ResourceType::TunnelIds | ResourceType::LinkIds | ResourceType::SegmentRoutingIds => {
```
juan-malbeclabs: I think the name would be VRF id. We are not referring to Segment Routing.



**Segment Routing (SR)**, instead, is a path steering mechanism. It does not isolate networks; it tells packets which path to follow across the network by encoding instructions (segments) inside the packet. Routers don’t choose the path hop-by-hop, the ingress router decides the full route.

**A VRF (Virtual Routing and Forwarding)** is a virtual routing table. It isolates traffic by creating multiple independent routing domains inside the same router — like having many logical routers running on one box. Two customers can use the same IP ranges without conflict because each packet is looked up in a different routing table.

PR #2856, opened 2026-02-07
smartcontract: move multicast group allowlists from AccessPass vecs to PDAs

martinsander00:
Resolves: https://github.com/malbeclabs/infra/issues/529

## Summary
- Replace `mgroup_pub_allowlist` and `mgroup_sub_allowlist` vecs inside AccessPass with dedicated `MGroupAllowlistEntry` PDAs, eliminating unbounded account growth and reallocation costs
- Add self-migration path: subscribe and create_subscribe processors automatically migrate existing vec entries to PDAs on use, removing them from the vec after PDA creation
- Allowlist add now creates a separate PDA account; allowlist remove closes the PDA and reclaims rent
- New `MGroupAllowlistEntry` account type (id=13) with `SEED_MGROUP_ALLOWLIST` seed, keyed by access pass, multicast group, and allowlist type (publisher/subscriber)

## Testing Verification
- Unit tests for MGroupAllowlistEntry serialization and validation
- Updated integration tests for allowlist add/remove with new PDA account
- Expanded multicast subscribe tests covering PDA-based allowlist checks and vec-to-PDA migration


juan-malbeclabs:
I think that although this improves AccessPass managment, it doesn’t solve the real problem: QA is leaving AccessPasses referencing multicast groups that it later deletes. We should first fix QA so it removes the multicast group from the AccessPasses before deleting it.

PR #2877, opened 2026-02-09
Multi-tenancy: Use VRF Id resource for allocation

elitegreg:
## Summary of Changes
* Use VRF ID resource for allocation


juan-malbeclabs (approved):
LGTM

PR #2879, opened 2026-02-09
Geolocation RFC

ben-dz:


rfcs/rfcX-geolocation-verification.md
```diff
@@ -0,0 +1,533 @@
...
+- Run a UDP listener (default port 8923) accepting signed Offset messages from DZDs
+- Pull configuration from the DZ Ledger and measure latency to target devices specified there
+- Generate composite Offset messages including references to DZD measurements attesting to the probe's location
+- Are registered onchain in the Telemetry Program
+
+### Offset
+A signed data structure representing the latency relationship between two entities (DZD↔Probe or Probe↔Target) and is sent to the Probe or Target. This is sent via UDP to the next link in the chain (From DZD->Probe and From Probe->Target)
+
+```rust
+struct LocationOffset {
```
nihar-dz: Is there a timestamp in this? More generally, how do we consider time in constructing references? We want to make sure an adversarial probe doesn't pick an old reference (when they were much closer to a DZD).

rfcs/rfcX-geolocation-verification.md
```diff
@@ -0,0 +1,533 @@
...
+- New Probe accounts coexist with existing Device accounts
+
+## Open Questions
+
+1. Should probes support IPv6? (Initially IPv4 only for simplicity)
+2. What's the optimal cache size for Offset storage? (Testing will determine, starting with 10k entries)
+3. Should probe metrics be posted onchain? (Yes for auditability; separate from target verification path)
+4. How to handle probe key rotation? (Manual for POC, automated in MVP)
+5. In the architecture diagram, the Probe has a line to the Target that says "Probe". Should this be TWAMP (requires configuration on the target), or ICMP, or a TCP syn/syn-ack on a port known to listen publicly? Or support all three options? For POC we can start with ICMP.
+6. How should we handle the latency between the probe and device changing over time? Should we always use the most recent measurement? Should we use the average since the last Signed Offset was sent? The avg/min/max of the previous epoch?
```
nihar-dz: I think this is the big open question — how we enforce something on staleness in constructing references. Admittedly a user, if given the timestamps of all measured offsets, can probably construct logic on their end to decide how to aggregate it. But it then effectively bypasses the reference data in the output, which would be a shame.

rfcs/rfcX-geolocation-verification.md
```diff
@@ -0,0 +1,533 @@
...
+### Offset
+A signed data structure representing the latency relationship between two entities (DZD↔Probe or Probe↔Target) and is sent to the Probe or Target. This is sent via UDP to the next link in the chain (From DZD->Probe and From Probe->Target)
+
+```rust
+struct LocationOffset {
+    signature: [u8; 64],          // Ed25519 signature
+    pubkey: [u8; 32],             // Signer's public key (DZD or Probe)
+    lat: f64,                     // Reference point latitude (WGS84)
+    lon: f64,                     // Reference point longitude (WGS84)
+    rtt_ns: u64,                  // Measured RTT in nanoseconds
```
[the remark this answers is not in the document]
ben-dz: Good point. We can always use minimums.

rfcs/rfcX-geolocation-verification.md
```diff
@@ -0,0 +1,533 @@
...
+**DZD-generated Offsets** contain no references (DZDs are roots of trust). <br> 
+**Probe-generated Offsets** include references to DZD Offsets, enabling targets to verify the entire measurement chain.
+
+> 💡 An enterprising user could use the existing link telemetry to confirm locations of DZDs relative to other DZDs. This is not covered by this RFC.
+
+### Location Offset
+The RTT to a target from the lat/lon in the Offset struct.
+
+### Child Probe
+A Probe assigned to a specific DZD for periodic latency measurement, defined onchain. DZDs only measure and send Offsets to their mated Probes.
```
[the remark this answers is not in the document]
ben-dz: whoops, we originally called it a "mated probe" but realized that it was likely to turn into a many-many relationship and needed a new name.

rfcs/rfcX-geolocation-verification.md
ben-dz: Adding slot.

PR #2896, opened 2026-02-10
Update user delete

juan-malbeclabs:
This pull request improves the user deletion process by ensuring that multicast group subscriptions are only attempted for groups that currently exist. The main change is to check for the existence of each multicast group before attempting to unsubscribe, which prevents errors when a user is associated with groups that may have been deleted.

Multicast group handling improvements:

* Added a call to `ListMulticastGroupCommand` to fetch the current list of multicast groups before processing user subscriptions.
* Updated the deletion logic to only execute `SubscribeMulticastGroupCommand` for groups that exist in the current multicast group list, preventing attempts to unsubscribe from non-existent groups. [[1]](diffhunk://#diff-11779c00a2224ce8e960532a9494334146983fe1f64b619779c6cf1d2c8c2a06R32-R35) [[2]](diffhunk://#diff-11779c00a2224ce8e960532a9494334146983fe1f64b619779c6cf1d2c8c2a06R45)

### DIFF
--- a/smartcontract/sdk/rs/src/commands/user/delete.rs
+++ b/smartcontract/sdk/rs/src/commands/user/delete.rs
@@ -226,7 +226,10 @@ mod tests {
             .in_sequence(&mut seq)
             .returning(move |_| {
                 let mut map = std::collections::HashMap::new();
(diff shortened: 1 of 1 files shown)


martinsander00 (approved):
Lint issues but lgtm

PR #2922, opened 2026-02-11
feat(serviceability): add TenantBillingConfig and epoch tracking to UpdatePaymentStatus

vihu:
# Summary

- Add TenantBillingConfig enum (FlatPerEpoch variant) to Tenant state for per-DZ-epoch flat-rate billing
- Extend UpdateTenant to accept an optional billing config
- Extend UpdatePaymentStatus with an optional last_deduction_dz_epoch field so the sentinel can atomically bump the epoch after a successful deduction
- Thread new fields through SDK and CLI layers, add integration tests, and add TENANT_README.md


juan-malbeclabs (approved):
LGTM

PR #3028, opened 2026-02-17
config agent: reduce full config check frequency from 5s to 1m and compare hashes instead

### LINKED ISSUE #3026: Reduce config agent resource consumption
Today, the doublezero config agent (controlplane/agent) downloads the full config to be applied from the controller and applies it in a configuration session. If Arista's ConfigAgent process detects any changes, the doublezero config agent applies them.  Checking for changes is a CPU-intensive operation for Arista ConfigAgent, and the largest device in mainnet-beta today has a configuration of over 200KB. This process is repeated every 5 seconds.

Changes are detected in a few cases:
1. The configuration has changed on the controller. For new users, we want to learn about this change immediately so we can apply it on the device and bring the user tunnel up
2. The configuration was changed manually by the contributor. This is generally done for troubleshooting purposes. Contributors know that they need to disable the agent if they want to avoid this

We can reduce CPU and network usage while staying optimized for case #1 by making the following changes to the doublezero config agent and controller:

- When the config agent starts, call GetConfig and apply it, and add it to a local in-memory config cache, along with its hash
- Every 5 seconds, call GetConfigHash (a new controller call) to get the current hash of the controller's config for the device
- If the hash has not changed, do nothing; if the hash has changed, call GetConfig and apply, and update the local cache
- Every 60 seconds, call GetConfig and apply, and update the cache

This will reduce the frequency at which we undo a contributors changes for scenario #2 above, and it will maintain the same 5-second update interval for scenario 1.

nikw9944:
## Summary
- Instead of applying config to the device every 5s, apply it only when the config hash changes, or when it has been 60s since it was last applied.

### Controller
- Refactor GetConfig by extracting config generation into reusable helper functions
- Add architecture documentation with sequence diagram

### Agent  
- Refactor main loop
- Implement a simple caching scheme for device config
- Instead of apply the full config from the controller every 5 seconds, only apply it if it has changed
- Also apply the full config after cache timeout (default 60s)

## Testing Verification
- Added unit tests for controlplane/agent/cmd/main.go
- No functionality has changed so e2e tests should run as-is


nikw9944:
#### BEFORE: Simple polling every 5 seconds
```
┌─────────┐                 ┌────────────┐                 ┌────────────┐                 ┌─────────┐
│  Agent  │                 │ Controller │                 │ Controller │                 │   EOS   │
│  main() │                 │ GetConfig()│                 │  Config    │                 │ Device  │
│         │                 │   (gRPC)   │                 │  Generator │                 │         │
└────┬────┘                 └─────┬──────┘                 └─────┬──────┘                 └────┬────┘
     │                            │                              │                              │
     │ Every 5s:                  │                              │                              │
     │                            │                              │                              │
     │ GetBgpNeighbors()          │                              │                              │
     ├─────────────────────────────────────────────────────────────────────────────────────────►│
     │◄─────────────────────────────────────────────────────────────────────────────────────────┤
     │ [peer IPs]                 │                              │                              │
     │                            │                              │                              │
     │ GetConfigFromServer()      │                              │                              │
     ├───────────────────────────►│                              │                              │
     │                            │ processConfigRequest()       │                              │
     │                            ├─────────────────────────────►│                              │
     │                            │                              │ generateConfig()             │
     │                            │                              │  • deduplicateTunnels()      │
     │                            │                              │  • renderConfig()            │
     │                            │                              │    (~50KB config text)       │
     │                            │◄─────────────────────────────┤                              │
     │                            │ [config string]              │                              │
     │◄───────────────────────────┤                              │                              │
     │ ConfigResponse             │                              │                              │
     │ {config: "..."}            │                              │                              │
     │                            │                              │                              │
     │ AddConfigToDevice(config)  │                              │                              │
     ├─────────────────────────────────────────────────────────────────────────────────────────►│
     │◄─────────────────────────────────────────────────────────────────────────────────────────┤
     │ [config applied]           │                              │                              │
     │                            │                              │                              │
     │ sleep(5s)                  │                              │                              │
     │ goto top                   │                              │                              │
     │                            │                              │                              │
```

#### AFTER: Hash-based polling (5s hash check, 5m full config fetch)
```
┌─────────┐                 ┌────────────┐                 ┌────────────┐                  ┌─────────┐
│  Agent  │                 │ Controller │                 │ Controller │                  │   EOS   │
│  main() │                 │GetConfigHash                 │  Config    │                  │ Device  │
│         │                 │ GetConfig()│                 │  Generator │                  │         │
└────┬────┘                 └─────┬──────┘                 └─────┬──────┘                  └────┬────┘
     │                            │                              │                              │
     │ Every 5s:                  │                              │                              │
     │                            │                              │                              │
     │ GetBgpNeighbors()          │                              │                              │
     ├─────────────────────────────────────────────────────────────────────────────────────────►│
     │◄─────────────────────────────────────────────────────────────────────────────────────────┤
     │ [peer IPs]                 │                              │                              │
     │                            │                              │                              │
     │ Decision: should fetch?    │                              │                              │
     │  • First run (no hash)?    │                              │                              │
     │  • 1m since last apply?    │                              │                              │
     │  • Hash changed?           │                              │                              │
     │                            │                              │                              │
     │ GetConfigHashFromServer()  │                              │                              │
     ├───────────────────────────►│                              │                              │
     │                            │ processConfigRequest()       │                              │
     │                            ├─────────────────────────────►│                              │
     │                            │                              │ generateConfig()             │
     │                            │                              │  • deduplicateTunnels()      │
     │                            │                              │  • renderConfig()            │
     │                            │                              │ SHA256(config)               │
     │                            │◄─────────────────────────────┤                              │
     │                            │ [hash only]                  │                              │
     │◄───────────────────────────┤                              │                              │
     │ ConfigHashResponse         │                              │                              │
     │ {hash: "abc123..."}        │                              │                              │
     │ (64 bytes)                 │                              │                              │
     │                            │                              │                              │
     │ Compare: hash != lastHash? │                              │                              │
     │                            │                              │                              │
     ├─── if YES (or first run or 5m timeout):                                                  │
     │                            │                              │                              │
     │    fetchConfigFromController()                            │                              │
     │    ├─► GetConfigFromServer()                              │                              │
     │    │   ──────────────────► │                              │                              │
     │    │                       │ processConfigRequest()       │                              │
     │    │                       ├─────────────────────────────►│                              │
     │    │                       │                              │ generateConfig()             │
     │    │                       │                              │  • deduplicateTunnels()      │
     │    │                       │                              │  • renderConfig()            │
     │    │                       │                              │    (entire config text)      │
     │    │                       │◄─────────────────────────────┤                              │
     │    │   ◄──────────────────│ [config string]               │                              │
     │    │   ConfigResponse      │                              │                              │
     │    │   {config: "..."}     │                              │                              │
     │    │                       │                              │                              │
     │    ├─► computeChecksum(config)                            │                              │
     │    │   [local SHA256]      │                              │                              │
     │    │                       │                              │                              │
     │    └─► return config+hash  │                              │                              │
     │                            │                              │                              │
     │    applyConfig()           │                              │                              │
     │    └─► AddConfigToDevice(config)                          │                              │
     │        ─────────────────────────────────────────────────────────────────────────────────►│
     │        ◄─────────────────────────────────────────────────────────────────────────────────┤
     │        [config applied]    │                              │                              │
     │                            │                              │                              │
     │    lastChecksum = hash     │                              │                              │
     │    lastApplyTime = now     │                              │                              │
     │                            │                              │                              │
     ├─── else: skip this cycle (hash unchanged, no work needed) |                              │
     │                            │                              │                              │
     │ sleep(5s)                  │                              │                              │
     │ goto top                   │                              │                              │
     │                            │                              │                              │
```

controlplane/controller/internal/controller/server.go
```diff
@@ -799,13 +732,94 @@ func (c *Controller) GetConfig(ctx context.Context, req *pb.ConfigRequest) (*pb.
...
+		getConfigPubkeyErrors.WithLabelValues(pubkey).Inc()
+		return "", nil, status.Errorf(codes.NotFound, "pubkey %s not found", pubkey)
+	}
+	if len(device.DevicePathologies) > 0 {
+		return "", nil, status.Errorf(codes.FailedPrecondition, "cannot render config for device %s: %v", pubkey, device.DevicePathologies)
+	}
+
+	// Find unknown BGP peers that need to be removed
+	peerFound := func(peer net.IP) bool {
+		for _, tun := range device.Tunnels {
```
packethog: The old code looped over `deviceForRender.Tunnels` which looks like it was a post-dedupe slice of tunnels. Should we still be doing that here?

https://github.com/malbeclabs/doublezero/pull/3028/changes#diff-f5e715ed1377ac408a4ecde2c60cbcb98327512c7d1496114ef143cb27e7afafL725

packethog (changes requested):

controlplane/controller/internal/controller/server.go
nikw9944: Using the raw list should not cause us to remove peers that should be kept, but I will change it to match the original.

controlplane/agent/cmd/agent/main_test.go
```diff
@@ -0,0 +1,282 @@
...
+			hash2 := computeChecksum(tt.input)
+
+			if hash1 != hash2 {
+				t.Errorf("hash should be consistent for same input")
+			}
+			if len(hash1) != 64 {
+				t.Errorf("SHA256 hash should be 64 hex characters, got %d", len(hash1))
+			}
+
+			if tt.name != "simple config" {
```
ben-dz: The expected hash for the "simple config" case (line 39) is wrong — the actual SHA256 of `"interface Tunnel500\n  description test"` is `a418fb48875fda25486c99b94b15adcb2128c7d1f7175008697a0b1dce3b2743`, not `0e5c3a6f84d0...`. This guard then silently skips verifying exactly that case, so the bad fixture never fails. Can you fix the fixture and remove the guard? As written the test documents an incorrect expected value while appearing to verify it.

controlplane/agent/cmd/agent/main_test.go
```diff
@@ -0,0 +1,282 @@
...
+				err := applyConfig(ctx, nil, tt.configText, tt.maxLockAge)
+				if err != nil {
+					t.Errorf("expected no error for empty config, got %v", err)
+				}
+			}
+		})
+	}
+}
+
+func TestCachingLogic(t *testing.T) {
```
ben-dz: This test duplicates the `shouldApply` if/else chain from `main()` verbatim inside the test body (lines 264-275), so it tests its own copy rather than the production code — if the logic in `main.go` regresses, this still passes. Suggest extracting the decision into a function, e.g.:

```go
func shouldApplyConfig(cachedHash, newHash string, lastApplied time.Time, timeout time.Duration, now time.Time) bool
```

called from `main()` and tested directly. Passing `now` explicitly also removes the need for the `cacheTime`/`currentTime` table fields.

controlplane/agent/cmd/agent/main_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+func TestApplyConfig(t *testing.T) {
+	tests := []struct {
+		name        string
+		configText  string
+		maxLockAge  int
+		expectError bool
+	}{
+		{
+			name:        "successful config application",
```
ben-dz: This "successful config application" case is never exercised — the loop body (line 198) only asserts when `tt.configText == ""`. Either drop the case or have `applyConfig` accept an interface so the EOS client can be mocked; as written it inflates apparent coverage.

controlplane/agent/cmd/agent/main.go
```diff
@@ -121,15 +122,52 @@ func main() {
...
+				log.Println("ERROR: fetchConfigFromController returned", err)
+				continue
+			}
+
+			// Only apply if config changed or timeout elapsed
+			shouldApply := false
+			if cachedConfigHash == "" {
+				// First run
+				shouldApply = true
+			} else if configHash != cachedConfigHash {
```
ben-dz: Heads-up on hash stability: this comparison only saves work if the controller renders byte-identical text for unchanged state. The agent side is deterministic (peers are sorted in `GetConfigFromServer`), but the controller renders from `stateCache` slices built by `GetProgramData`, which appends accounts in whatever order `getProgramAccounts` returns them — ordering isn't guaranteed stable across calls or RPC nodes. Only interfaces, VRFs, and multicast boundary lists are sorted (`server.go:184,422,553`); tunnels/users/peers are not. If ordering shifts between cache refreshes, the hash changes and the agent re-applies a semantically identical config. Worst case is today's behavior (not a correctness bug), but it can silently erode the CPU win. Worth either sorting the rendered collections controller-side, or watching the apply rate after rollout to confirm the optimization holds.

controlplane/agent/cmd/agent/main.go
```diff
@@ -20,52 +22,51 @@ import (
...
-	metricsEnable              = flag.Bool("metrics-enable", false, "Enable prometheus metrics")
-	metricsAddr                = flag.String("metrics-addr", ":8080", "Address to listen on for prometheus metrics")
+	localDevicePubkey           = flag.String("pubkey", "frtyt4WKYudUpqTsvJzwN6Bd4btYxrkaYNhBNAaUVGWn", "This device's public key on the doublezero network")
+	controllerAddress           = flag.String("controller", "18.116.166.35:7000", "The DoubleZero controller IP address and port to connect to")
+	device                      = flag.String("device", "127.0.0.1:9543", "IP Address and port of the Arist EOS API. Should always be the local switch at 127.0.0.1:9543.")
+	sleepIntervalInSeconds      = flag.Float64("sleep-interval-in-seconds", 5, "How long to sleep in between polls")
+	controllerTimeoutInSeconds  = flag.Float64("controller-timeout-in-seconds", 30, "How long to wait for a response from the controller before giving up")
+	configCacheTimeoutInSeconds = flag.Int("config-cache-timeout-in-seconds", 60, "Force full config fetch after this many seconds, even if hash unchanged")
```
ben-dz: Help text is stale from the dropped `GetConfigHash` design: the fetch still happens every 5s regardless — this timeout forces a full config **apply**, not a fetch. Suggest something like "Force config apply after this many seconds, even if hash unchanged". (Same staleness applies to the PR title/description, which still mention hash polling and the removed endpoint.)

controlplane/controller/README.md
```diff
@@ -2,6 +2,60 @@
...
+     │ Every 5s:                   │                              │                              │
+     │                             │                              │                              │
+     │ GetBgpNeighbors()           │                              │                              │
+     ├──────────────────────────────────────────────────────────────────────────────────────────►│
+     │◄──────────────────────────────────────────────────────────────────────────────────────────┤
+     │ [peer IPs]                  │                              │                              │
+     │                             │                              │                              │
+     │ GetConfigFromServer()       │                              │                              │
+     ├────────────────────────────►│                              │                              │
+     │                             │ processConfigRequest()       │                              │
```
ben-dz: `processConfigRequest()` and `generateConfig()` (line 37) don't exist — they were part of the removed controller refactor. The actual call path is `GetConfig()` (`internal/controller/server.go`) → `deduplicateTunnels()` → `renderConfig()` (`render.go`). Can you update the diagram to match the code as merged?

controlplane/agent/cmd/agent/main.go
nikw9944: This fine because the controller always renders identical config for any given state

CHANGELOG.md
```diff
@@ -281,6 +281,8 @@ All notable changes to this project will be documented in this file.
   - feat(smartcontract): add use_onchain_deallocation flag to MulticastGroup ([#2748](https://github.com/malbeclabs/doublezero/pull/2748))
 - CLI
   - Remove restriction for a single tunnel per user; now a user can have a unicast and multicast tunnel concurrently (but can only be a publisher _or_ a subscriber) ([2728](https://github.com/malbeclabs/doublezero/pull/2728))
+- Device agents
+  - Reduce config agent network and CPU usage by checking config checksums every 5 seconds, and reducing full config check frquency to 1m
```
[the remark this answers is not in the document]
nikw9944: Fixed, thanks.

controlplane/agent/cmd/agent/main_test.go
nikw9944: Fixed the expected hash and removed the guard. The test now verifies the correct SHA256 value.

controlplane/agent/cmd/agent/main_test.go
nikw9944: Extracted `shouldApplyConfig(cachedHash, newHash, lastApplied, timeout, now)` as a standalone function in main.go. Tests now call it directly instead of duplicating the logic.

controlplane/agent/cmd/agent/main_test.go
nikw9944: Removed the case. Testing actual config application needs an EOS client mock, which is out of scope here.

controlplane/agent/cmd/agent/main.go
nikw9944: Updated to "Force config apply after this many seconds, even if hash unchanged".

controlplane/controller/README.md
nikw9944: Updated the diagram to show the actual call chain: `GetConfig()` → `deduplicateTunnels()` → `renderConfig()`.

ben-dz (approved):

packethog (approved):

PR #3034, opened 2026-02-18
client: add onchain reconciler to daemon

snormore:
## Summary of Changes
- Add an onchain reconciler to the daemon (`doublezerod`) that polls onchain state and automatically provisions/tears down tunnels, replacing the CLI-driven provisioning model
- Move status enrichment (device, metro, tenant, latency) from the CLI to the daemon's `GET /v2/status` endpoint, eliminating 4 redundant RPC calls per status invocation
- Add `enable`/`disable` CLI subcommands and daemon endpoints to control the reconciler, with persistent state across daemon restarts
- Refactor CLI `connect` to simply enable the reconciler and wait for the daemon to provision, rather than performing provisioning directly
- Replace file-based DB (`doublezerod.json`) with onchain-driven state; add automatic migration from legacy state files

## Diff Breakdown
| Category     | Files | Lines (+/-) | Net    |
|--------------|-------|-------------|--------|
| Core logic   |    16 | +1879 / -1644 | +235 |
| Scaffolding  |     8 | +152 / -28    | +124 |
| Tests        |    12 | +2563 / -1011 | +1552 |
| Fixtures     |    14 | +18 / -131    | -113  |
| Docs         |     4 | +206 / -12    | +194  |
| **Total**    |    54 | +4818 / -2826 | +1992 |

~1900 lines of core logic changes (net +235 after removing legacy DB/provisioning code), supported by ~2500 lines of new tests and ~200 lines of docs/RFC.

<details>
<summary>Key files (click to expand)</summary>

- [`client/doublezerod/internal/manager/manager.go`](https://github.com/malbeclabs/doublezero/pull/3034/files#diff-27424874932194e8d91a18571b7c6a3bf1f1bfa057986e442db5324e6fced881) — onchain reconciler loop, tunnel provisioning from onchain state, enable/disable control, tunnel-src resolution
- [`client/doublezero/src/command/status.rs`](https://github.com/malbeclabs/doublezero/pull/3034/files#diff-4f704cf90f1dcb068f7142fc69c3b334575f5606872d549c25298747c6e130cd) — simplified to read enriched fields from daemon v2/status instead of making SDK RPC calls
- [`client/doublezero/src/command/connect.rs`](https://github.com/malbeclabs/doublezero/pull/3034/files#diff-0e50f0119aa96b168465129d4b6f763c1125ae34defc1d682d9858bf9aac8184) — refactored to enable reconciler and poll for daemon-driven provisioning instead of provisioning directly
[...]

PR #3066, opened 2026-02-20
client: handle existing IBRL tunnel on reconnect

martinsander00:
## Summary
- `doublezero connect ibrl` previously failed silently when the daemon already had a unicast service in memory (e.g. recovered from the state file after a daemon restart or reboot), printing `❌ Error provisioning service: unicast service already provisioned` but then claiming `✅ User Provisioned` and leaving the user with a broken tunnel
- The connect command now checks daemon status before provisioning: if a unicast service is already active (BGP Session Up), it exits successfully without re-provisioning; if the service exists but is not active (e.g. Network Unreachable, BGP Session Down), it returns a clear error directing the user to run `doublezero disconnect ibrl` first
- `doublezero disconnect` now cleans up stale daemon services even when the onchain user account has already been deleted before disconnect runs (e.g. in a ban/unban flow where the account is deleted as the unban mechanism). Previously, the daemon's tunnel service was silently left behind, causing a subsequent `connect ibrl` to fail with "already provisioned" and caused e2e `TestE2E_UserBan/user-ban-ibrl` to fail
- Two unit tests cover both connect cases: already-active tunnel skips provision, stale tunnel returns an actionable error

## Testing Verification
- `test_connect_ibrl_already_active_skips_provision`: verifies provisioning is not called when BGP is up (mockall panics on unexpected `provisioning()` call)
- `test_connect_ibrl_stale_tunnel_returns_error`: verifies the error message includes the current status and the disconnect command

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 - CLI
   - Remove redundant `connect ibrl` unit tests that were duplicates of hybrid-device equivalents
   - `doublezero connect ibrl` now checks the daemon's tunnel status before attempting to provision. If an IBRL tunnel is already active (BGP Session Up), the command exits successfully without re-provisioning. If a tunnel exists but is not active (e.g. stale after a daemon restart), the command returns a clear error directing the user to run `doublezero disconnect ibrl` first.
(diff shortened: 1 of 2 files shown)


martinsander00:
Handled by https://github.com/malbeclabs/doublezero/pull/3034

PR #3071, opened 2026-02-22
client: fix multicast subscriber tunnel src resolution for NAT environments

snormore:
## Summary
- Always resolve tunnel source from the routing table for multicast connections, fixing subscribers behind NAT (e.g. AWS) where the public IP isn't bound to a local interface
- Extends the approach from #2497
- Remove unused `has_allocated_dz_ip()` method and its tests

## Diff Breakdown
| Core logic |     2 | +8 / -60   |  -52 |
| Tests      |     1 | +90 / -9   |  +81 |
| Config     |     2 | +4 / -0    |   +4 |

Core change removes the conditional and always resolves tunnel src for multicast; also removes dead code.

<details>

- [`client/doublezero/src/command/connect.rs`](https://github.com/malbeclabs/doublezero/pull/3071/files#diff-0e50f0119aa96b168465129d4b6f763c1125ae34defc1d682d9858bf9aac8184) — unconditionally resolve tunnel src in `execute_multicast`; add dedicated NAT test; update existing multicast tests to expect resolve_route
- [`smartcontract/.../state/user.rs`](https://github.com/malbeclabs/doublezero/pull/3071/files#diff-0e24a7e1473ca2f2de965d23e72a3b5212b2d7d3255e6042659604eb08ee19f6) — remove unused `has_allocated_dz_ip()` and its tests

</details>

## Testing Verification
- New `test_connect_command_multicast_resolves_tunnel_src_for_nat` verifies resolved local IP is used as tunnel_src instead of public IP
- All existing multicast tests updated to expect resolve_route calls


client/doublezero/src/command/connect.rs
```diff
@@ -293,15 +293,14 @@ impl ProvisioningCliCommand {
             })
             .collect::<Result<Vec<_>, _>>()?;
 
-        let tunnel_src = if user.has_allocated_dz_ip() {
```
snormore: Note this returns false if `dz_ip == Ipv4Addr::UNSPECIFIED` so tunnel src isn't getting resolved for new multicast users:
```rust
pub fn has_allocated_dz_ip(&self) -> bool {
    self.dz_ip != Ipv4Addr::UNSPECIFIED && self.dz_ip != self.client_ip
}
```

packethog (approved):
nit: Can you update the language to say this is subscribers only? Publishing behind a NAT works fine.

PR #3120, opened 2026-02-27
geolocation: add GeoProbe state and CRUD instructions

nikw9944:
## Summary of Changes
- Add `GeoProbe` account type to the Geolocation Program with `CreateGeoProbe`, `UpdateGeoProbe`, and `DeleteGeoProbe` instructions (foundation-gated via Serviceability CPI)
- Create validates exchange is activated, IP is publicly routable, code length ≤32 bytes; delete enforces `reference_count == 0`
- Add comprehensive IP validation that rejects all non-publicly-routable addresses (RFC 1918, loopback, multicast, link-local, reserved ranges, etc.)
- Add GeoProbe PDA derivation (`["doublezero", "probe", code.as_bytes()]`) and extend serializer with `try_acc_close` for account deletion
- Supports rfcs/rfc16-geolocation-verification.md

## Diff Breakdown
| Category     |  Lines  |
|--------------|-------|
| Prod code   |    ~500 |
| Test code |  ~1000 |

## Testing Verification
- Unit tests and integration tests

### DIFF
--- a/smartcontract/programs/CLAUDE.md
+++ b/smartcontract/programs/CLAUDE.md
@@ -45,3 +45,6 @@
 ### Program Upgrades
 
 1. **Use standard interfaces**: Use `solana-loader-v3-interface` to parse `UpgradeableLoaderState` rather than implementing your own parser. The interface crate provides well-tested, maintained implementations.
+
+### Pull Requests
+- Make sure `make rust-lint` and `make rust-test` both pass before posting pull requests
--- a/smartcontract/programs/doublezero-geolocation/src/state/geo_probe.rs
+++ b/smartcontract/programs/doublezero-geolocation/src/state/geo_probe.rs
@@ -1,7 +1,4 @@
-use crate::{
-    error::{GeolocationError, Validate},
-    state::accounttype::AccountType,
-};
+use crate::state::accounttype::AccountType;
 use borsh::{BorshDeserialize, BorshSerialize};
 use solana_program::{account_info::AccountInfo, msg, program_error::ProgramError, pubkey::Pubkey};
 use std::{fmt, net::Ipv4Addr};
@@ -85,20 +82,6 @@ impl TryFrom<&AccountInfo<'_>> for GeoProbe {
     }
 }
 
-impl Validate for GeoProbe {
-    fn validate(&self) -> Result<(), GeolocationError> {
-        if self.account_type != AccountType::GeoProbe {
-            return Err(GeolocationError::InvalidAccountType);
-        }
-
-        // Note: Code length and parent devices count are validated at instruction time
-        // and enforced by instruction constraints, so we don't need to re-validate here.
-        // These conditions should never occur in a properly deserialized account.
-
-        Ok(())
-    }
-}
-
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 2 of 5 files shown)


smartcontract/programs/doublezero-geolocation/src/state/geo_probe.rs
```diff
@@ -0,0 +1,221 @@
...
+            return Err(GeolocationError::InvalidCodeLength);
+        }
+        if self.parent_devices.len() > MAX_PARENT_DEVICES {
+            msg!(
+                "Too many parent devices: {} (max {})",
+                self.parent_devices.len(),
+                MAX_PARENT_DEVICES
+            );
+            return Err(GeolocationError::MaxParentDevicesReached);
+        }
```
karl-dz: No instructions add parent devices yet. But I'm wondering if this would ever hit, too? I'm guessing when an instruction that introduces adding parent devices, there would be validation of whether this violates MAX_PARENT_DEVICES, so this condition would never be true.

Curious about a scenario where this would be true.

smartcontract/programs/doublezero-geolocation/src/state/geo_probe.rs
```diff
@@ -0,0 +1,221 @@
...
+        Ok(())
+    }
+}
```
karl-dz: Now that I've written the two comments above, I'm wondering why this trait even needs to be implemented with any of the accounts you have in this smart contract. What does calling `validate` protect against exactly?

smartcontract/programs/doublezero-geolocation/tests/geo_probe_test.rs
```diff
@@ -0,0 +1,320 @@
...
+
+    assert_eq!(probe.account_type, AccountType::GeoProbe);
+    assert_eq!(probe.owner, payer.pubkey());
+    assert_eq!(probe.exchange_pk, exchange_pubkey);
+    assert_eq!(probe.public_ip, Ipv4Addr::new(8, 8, 8, 8));
+    assert_eq!(probe.location_offset_port, 4242);
+    assert_eq!(probe.metrics_publisher_pk, args.metrics_publisher_pk);
+    assert_eq!(probe.reference_count, 0);
+    assert_eq!(probe.code, code);
+    assert_eq!(probe.parent_devices.len(), 0);
```
karl-dz: FYI, if GeoProbe derives `PartialEq` then it would be better to `assert_eq!(probe, expected_probe)` in case you add more fields to your geo probe account in the future. With the way this is written, you run the risk of missing asserting that the new fields will be equal what you expect

smartcontract/programs/doublezero-geolocation/tests/geo_probe_test.rs
```diff
@@ -0,0 +1,320 @@
...
+
+    let tx = Transaction::new_signed_with_payer(
+        &[ix],
+        Some(&payer.pubkey()),
+        &[&payer],
+        *recent_blockhash.read().await,
+    );
+
+    let result = banks_client.process_transaction(tx).await;
+    assert!(result.is_err());
```
karl-dz: Does not seem necessary since you unwrap the error later on. Should the claude markdown be informed to not perform this `is_err` check?

smartcontract/programs/CLAUDE.md
```diff
@@ -28,6 +28,18 @@
 
 2. **Use BorshDeserializeIncremental**: For instruction arguments that may gain new optional fields over time, use `BorshDeserializeIncremental` or derive `BorshDeserialize`.
 
+### Testing
+
+1. **Assert specific errors**: Tests should assert specific error types (e.g., `ProgramError::Custom(17)`), not just `.is_err()`. This catches regressions where the instruction fails for the wrong reason.
+
+2. **Don't test framework functionality**: Avoid writing tests that only exercise SDK/framework behavior (e.g., testing that `Pubkey::find_program_address` is deterministic or produces different outputs for different inputs). Focus tests on your program's logic.
```
karl-dz: I guess when the unit tests were generated in this PR, claude forgot about this

PR #3148, opened 2026-03-02
smartcontract/cli: improve tenant get output with table/JSON formatting

juan-malbeclabs:
## Summary of Changes
- Added `--json` flag to `tenant get` command for machine-readable output
- Replaced flat key-value output with formatted tables using `tabled` for human-readable display
- Tenant config and administrators are now shown as separate tables
- Administrators list is only shown when non-empty
- Expanded test coverage for all combinations of `--json` and admin presence

## Diff Breakdown
| Core logic |     1 | +148 / -14  | +134 |

Single-file change adding structured output (table + JSON) to the `tenant get` CLI command.

<details>

- `smartcontract/cli/src/tenant/get.rs` — rewrote output logic to use `tabled` tables and `serde_json` for `--json` mode; added `ConfigDisplay` and `AdminDisplay` structs; expanded tests

</details>

## Testing Verification
- Unit tests cover table output (with and without administrators) and JSON output (with and without administrators)
- All four combinations of `--json` flag × admin presence are exercised

### DIFF
--- a/smartcontract/cli/src/tenant/get.rs
+++ b/smartcontract/cli/src/tenant/get.rs
@@ -17,25 +17,17 @@ pub struct GetTenantCliCommand {
 
 #[derive(Tabled, Serialize)]
 struct ConfigDisplay {
-    #[tabled(rename = "account")]
     pub account: String,
-    #[tabled(rename = "code")]
     pub code: String,
-    #[tabled(rename = "vrf_id")]
     pub vrf_id: u16,
-    #[tabled(rename = "metro_routing")]
     pub metro_routing: bool,
-    #[tabled(rename = "route_liveness")]
     pub route_liveness: bool,
-    #[tabled(rename = "reference_count")]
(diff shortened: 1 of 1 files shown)


smartcontract/cli/src/tenant/get.rs
```diff
@@ -1,29 +1,87 @@
...
+    pub route_liveness: bool,
+    #[tabled(rename = "reference_count")]
+    pub reference_count: u32,
+    #[tabled(rename = "owner")]
+    pub owner: String,
+}
+
+#[derive(Tabled, Serialize)]
+struct AdminDisplay {
+    #[tabled(rename = "administrator")]
```
vihu: Is the `rename` necessary if it's the same as the field (above as well), just wondering whether it would work as is without the `#[tabled(rename=...)]` (assuming field is the name we want to keep in the table)

PR #3150, opened 2026-03-02
tools/twamp - Add signed sender/reflector for signed TWAMP messages

ben-dz:
## Summary

- Add Ed25519-signed TWAMP sender and reflector to support inbound probing flow per [RFC16](rfcs/rfc16-geolocation-verification.md) Phase 1.5
- Signed probes (108 bytes) carry sender pubkey + signature; signed replies (204 bytes) echo the probe and add reflector pubkey + signature, enabling mutual authentication
- Reflector enforces a pubkey allowlist (atomically updatable) and rate-limits signature verification per sender (minimum 55s between verifications for the same key)
- Linux-only implementation using raw syscalls (epoll, SO_TIMESTAMPNS, SO_BINDTODEVICE) with compile-time fallback stubs for non-Linux platforms

**Code:** ~500 lines
**Test:** ~650 lines

## Testing Verification

- Unit tests for packet serialization/deserialization round-trips, signature verification, and invalid packet handling
- Reflector tests cover authorization rejection, rate-limited signature verification, concurrent authorized key updates, and oversized/undersized packet handling
- Sender tests cover probe/reply round-trip, reflector pubkey mismatch rejection, and tampered reply detection


tools/twamp/pkg/light/signed_reflector_linux.go
```diff
@@ -0,0 +1,202 @@
...
+			if _, ok := r.authorizedKeys.Load(probe.SenderPubkey); !ok {
+				continue
+			}
+
+			// Rate-limit signature verifications per pubkey.
+			now := time.Now()
+			if interval := SignedReflectorVerifyInterval; interval > 0 {
+				if last, ok := r.lastVerify.Load(probe.SenderPubkey); ok {
+					if now.Sub(last.(time.Time)) < interval {
+						continue
```
snormore: So will this just ignore any probes that exceed the rate limit? From the sender's side it'll look like a loss?

tools/twamp/pkg/light/signed_packet.go
```diff
@@ -0,0 +1,152 @@
...
+const (
+	SignedProbePacketSize = 108
+	SignedReplyPacketSize = 204
+
+	signedProbePayloadSize = 44  // bytes 0-43: fields signed by sender
+	signedReplyPayloadSize = 140 // bytes 0-139: fields signed by reflector
+)
+
+// SignedProbePacket is sent from Target to Probe in the inbound probing flow.
+type SignedProbePacket struct {
```
snormore: Want to move all of this into `pkg/signed` instead of `pkg/light`?

Issue #3190, opened 2026-03-06
geolocation: Update inbound probing to include time delta since last inbound


ben-dz:
TEEs can't actually do time measurements. Therefore we need to provide a customer running in TEE the time delta between each TWAMP ping that the geoprobe reflector receives.

## Required Changes To signed reflector & ReplyPacket
- Signed TWAMP reflector does not need to verify signature of inbound message, just needs to verify that it's an authorized pubkey.
- Signed TWAMP reflector needs to allow at least two messages back to back, and then force a DOS-prevention wait.
- Signed TWAMP reply needs to include SinceLastRxSec And SinceLastRxFrac, with the time between the arrival of the previous SignedTwamp, and the arrival of the SignedTWAMP this reply is replying to.

## Required Changes to geolocation-target-sender
- Probes will now be considered as "pairs" Send probe 0.
- Send a follow up probe 1 as soon as target-sender receives reply 0, even before verifying reply 0.
- Receive reply 1. Reply 0 and Reply 1 both verify, print the SinceLastRx as "Probe Measured RTT"
- Also print the lower of the RTTs as measured normally as "Target Measured RTT"

PR #3194, opened 2026-03-06
Geolocation: add location and timing fields to signed TWAMP replies for paired probing

ben-dz:
Resolves: #3190

## Summary of Changes
- Extend the signed TWAMP reply packet with five new fields (`MeasurementSlot`, `Lat`, `Lng`, `SinceLastRxNs`, `RttNs`) and embed DZD `LocationOffset` blobs, enabling inbound geolocation probing for TEE-based targets that cannot measure time
- Implement paired probing in the target-sender: two probes sent back-to-back let the reflector compute the inter-arrival delta (`SinceLastRxNs`) as a proxy RTT
- Add IP-binding for probe pairs on the reflector — both probes in a pair must originate from the same source IP, preventing cross-sender interference
- Add pair-based rate limiting (2 probes per window per sender pubkey) to bound reply throughput
- Add `ParseOffsetInfo` to extract location/timing data from Borsh-encoded `LocationOffset` blobs without importing the full geoprobe package
- Re-add probe signature verification logging in the target-sender for defense-in-depth audit trails
- Update RFC16 to document the paired-probing inbound measurement flow (steps 5–9)

Related: [RFC16 — Geolocation Verification](rfcs/rfc16-geolocation-verification.md)

## Diff Breakdown
| Core logic   |     4 | +319 / -146 | +173  |
| Tests        |     3 | +321 / -57  | +264  |
| Docs         |     1 | +36 / -19   | +17   |
| Config/build |     1 | +1 / -1     |  0    |

~60% of the diff is test coverage; core logic is concentrated in 4 files.

<details>

- [`controlplane/telemetry/cmd/geoprobe-target-sender/main.go`](https://github.com/malbeclabs/doublezero/pull/3194/files#diff-7e892eb9746bff2f3cd77676edb134f1b774e2a03b4f378391d1c979645d7c00) — Rewrite probe loop to send paired probes; add probe signature verification logging; add `--format` flag for text/JSON output with new location and timing fields
- [`tools/twamp/pkg/signed/reflector_linux.go`](https://github.com/malbeclabs/doublezero/pull/3194/files#diff-4a412449845c51f90365163b49587cf84a488e8306ba4ea5a40ec9a909d09a6e) — Add `pairSourceIP` IP-binding, pair-based rate limiting, `SinceLastRxNs` inter-arrival timing, and `LocationOffset` embedding in replies
- [`tools/twamp/pkg/signed/packet.go`](https://github.com/malbeclabs/doublezero/pull/3194/files#diff-b1af8e52a71e03573feab54cce8909004955cac6b82ad5b4817c664200e39eab) — Extend `ReplyPacket` with five new signed fields; add `ParseOffsetInfo` for Borsh-encoded offset extraction
- [`tools/twamp/pkg/signed/reflector_test.go`](https://github.com/malbeclabs/doublezero/pull/3194/files#diff-d9ee93efb6507c6e587de69ab67e484b6a2cb050fc8f354b3f0c53fb178d84a4) — Add tests for IP-binding rejection, corrupted second-probe signature rejection, and pair-based rate limiting
- [`tools/twamp/pkg/signed/packet_test.go`](https://github.com/malbeclabs/doublezero/pull/3194/files#diff-f5c0b99fd1494ddc94d5f31f2eeecaeea1a7a5871dd84ab6d47dcf91af3d62d1) — Add tests for new reply fields, `ParseOffsetInfo`, byte layout, and tamper detection
- [`controlplane/telemetry/cmd/geoprobe-target-sender/main_test.go`](https://github.com/malbeclabs/doublezero/pull/3194/files#diff-9222af122670a29430d7d9907e6f81f4be6fe6a3fa6b38a0d510ad9689960f6f) — Update test fixtures for new paired-probe output fields
- [`rfcs/rfc16-geolocation-verification.md`](https://github.com/malbeclabs/doublezero/pull/3194/files#diff-437d757532ab5ce5c2d80f57ccd5ffa0020207d136e1f76c668bd4696c9516f3) — Document inbound paired-probing flow and updated reply packet format

</details>

## Testing Verification
- All `tools/twamp/pkg/signed/...` tests pass (38 tests), including new tests for IP-binding rejection, second-probe signature rejection, pair-based rate limiting with `SinceLastRxNs` assertions, and `ParseOffsetInfo` field-skip validation
- All `controlplane/telemetry/cmd/geoprobe-target-sender/...` tests pass with updated paired-probe output format
- `geoprobe-agent` binary builds cleanly
- Manual end-to-end testing with deployed geoprobe-agent and target-sender binaries

### DIFF
--- a/tools/twamp/pkg/signed/reflector_linux.go
+++ b/tools/twamp/pkg/signed/reflector_linux.go
@@ -213,7 +213,10 @@ func (r *LinuxReflector) Run(ctx context.Context) error {
 			// capturing `now` — its latency only affects reply 1, which doesn't
 			// impact the measurement (TEE targets can't measure time). Both
 			// probes in a pair must come from the same source IP.
-			fromAddr := from.(*unix.SockaddrInet4)
+			fromAddr, ok := from.(*unix.SockaddrInet4)
+			if !ok {
+				continue
+			}
 			if state.pairCount == 0 {
 				state.pairSourceIP = fromAddr.Addr
 			} else {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ All notable changes to this project will be documented in this file.
 - SDK
   - Add read-only Go SDK for `doublezero-geolocation` program with state deserialization, PDA derivation, and RPC client for querying geoprobe configuration
 - Telemetry
-  - Embed LocationOffsets from parent DZDs in signed TWAMP replies so inbound probes carry geolocation context, and make signed TWAMP replies more like LocationOffsets to copule with a new double-probe system for inbound probing.
+  - Embed LocationOffsets from parent DZDs in signed TWAMP replies so inbound probes carry geolocation context, and make signed TWAMP replies more like LocationOffsets to couple with a new double-probe system for inbound probing.
 - Client
   - Increase default route liveness probe interval (TxMin/RxMin) from 300ms to 1s and raise MaxTxCeil from 1s to 3s to preserve backoff headroom
 - Smartcontract


tools/twamp/pkg/signed/reflector_linux.go
```diff
@@ -182,24 +192,72 @@ func (r *LinuxReflector) Run(ctx context.Context) error {
...
 
-			if !probe.Verify() {
-				continue
+			// Pair authentication: the measurement loop runs from Probe 0 Rx to
+			// Probe 1 Rx, so we skip verification on probe 0 to keep the reply
+			// fast and the inter-arrival gap accurate. Probe 1 is verified after
+			// capturing `now` — its latency only affects reply 1, which doesn't
+			// impact the measurement (TEE targets can't measure time). Both
+			// probes in a pair must come from the same source IP.
+			fromAddr := from.(*unix.SockaddrInet4)
```
snormore: This cast should never really fail here I think but just in case did you want to do something like this?
```go
fromAddr, ok := from.(*unix.SockaddrInet4)
if !ok {
    continue
}
```

tools/twamp/pkg/signed/reflector_linux.go
```diff
@@ -182,24 +192,72 @@ func (r *LinuxReflector) Run(ctx context.Context) error {
...
+			if state.pairCount == 0 {
+				state.pairSourceIP = fromAddr.Addr
+			} else {
+				if fromAddr.Addr != state.pairSourceIP {
+					continue
+				}
+				// Verify the probe signature. Comment out these two lines to
+				// rely solely on IP-binding for pair authentication.
+				// if !probe.Verify() {
```
snormore: Uncomment this?

ben-dz: It's there so that I can add it back easily (and I wanted to remember where I'd stick it if I changed my mind), but I don't think it's necessary. The origin of the follow up probe has to have the same IP as the original probe. So the only realistic spoof probe would be the hypervisor around the TEE target. But since we include the original probe message in the outbound message, so the target-sender code in the TEE can see that it doesn't match its own signature.

The ed25519 verify is not quick, but by putting it there (and the reason I it as a hint about where to put it), it won't affect the measurement for _this_ target. However it could introduce jitter and higher RTT times to other probes happening at the same time, so it's better not to have it unless we need it.

PR #3206, opened 2026-03-09
[New Permission 3/5] smartcontract: enforce Permission-based authorization in existing instructions

### LINKED ISSUE #3183: serviceability: Permission account model for scalable, granular access control
## Background

DoubleZero's current authorization model relies on three allowlists stored in GlobalState:

- `foundation_allowlist` — full administrative access (locations, devices, links, tenants, contributors, access passes, etc.)
- `qa_allowlist` — QA-specific access
- `activator_authority_pk` / `sentinel_authority_pk` / `health_oracle_pk` — single-key authority slots for automated roles

This model was a practical starting point, but it does not scale as DoubleZero's contributor and operator ecosystem grows:

- **No granularity.** A key is either in the foundation allowlist (full admin) or it isn't. There is no way to grant a key the right to manage access passes but not devices, or devices but not tenants.
- **No auditability.** There is no on-chain record of _who_ has _what_ access, when it was granted, or by whom.
- **No revocation path.** Removing a key from an allowlist is an all-or-nothing operation. There is no suspend/resume cycle.
- **Does not scale.** The allowlist is a vector stored in a single GlobalState account. It is bounded and expensive to update as the number of operators grows.
- **Single points of failure.** The `activator_authority_pk` and `sentinel_authority_pk` slots are single keys; rotating them requires a program-wide update.

## Proposed Solution: Permission Accounts

Introduce a new on-chain account type — **Permission** — that acts as the authorization credential for a given pubkey.

### Account structure

```
Permission {
    account_type: AccountType::Permission,
    owner:        Pubkey,         // who created / manages this account
    bump_seed:    u8,
    status:       PermissionStatus,  // Activated | Suspended
    user_payer:   Pubkey,            // the key being authorized
    permissions:  u128,              // bitmask of granted flags
}
```

The PDA is derived from `(program_id, user_payer)`, so it is deterministic and verifiable by any instruction without an index or lookup.

### Permission flags (bitmask)

| Bit | Name | Legacy equivalent | Scope |
|-----|------|-------------------|-------|
| 0 | `foundation` | foundation_allowlist | Full administrative access |
| 1 | `permission-admin` | foundation_allowlist | Create/update/delete Permission accounts |
| 2 | `infra-admin` | foundation_allowlist | Locations, exchanges |
| 3 | `network-admin` | foundation_allowlist + activator | Devices, interfaces, links |
| 4 | `tenant-admin` | foundation_allowlist + sentinel | Tenants |
| 5 | `multicast-admin` | foundation_allowlist + activator + sentinel | Multicast groups |
| 6 | `reservation` | reservation_authority_pk | IP/tunnel allocation |
| 7 | `activator` | activator_authority_pk | Approve/reject entities |
| 8 | `sentinel` | sentinel_authority_pk | Route control |
| 9 | `user-admin` | foundation_allowlist + activator | User lifecycle |
| 10 | `access-pass-admin` | foundation_allowlist + sentinel | Access pass management |
| 11 | `health-oracle` | health_oracle_pk | Device/link health reporting |
| 12 | `qa` | qa_allowlist | QA/testing access |
| 13 | `globalstate-admin` | foundation_allowlist | Feature flags, allowlists, authority keys |
| 14 | `contributor-admin` | foundation_allowlist | Contributor management |

### Authorization flow

Each instruction that currently checks allowlists is migrated to call a shared `authorize(program_id, accounts_iter, payer, globalstate, required_flags)` function:

1. If a Permission PDA account is present in the transaction's remaining accounts, validate its PDA derivation, check its status is `Activated`, and verify the bitmask satisfies `required_flags`.
2. If no Permission account is present, fall back to the existing GlobalState allowlist/authority-key checks — **unless** `RequirePermissionAccounts` feature flag is set, in which case the legacy path is rejected.

This design is backward-compatible by default: existing keys that are in the allowlists continue to work without any migration on their part.

### Feature flag: `RequirePermissionAccounts`

A new `FeatureFlag::RequirePermissionAccounts` (bit 1 in GlobalState `feature_flags`) controls the enforcement mode:

- **OFF (default):** Both paths are accepted. Existing allowlist members work without a Permission account.
- **ON:** The legacy path is blocked. All instructions require a Permission account. Exception: foundation members retain access to `permission-admin` operations to prevent lockout.

The SDK's `execute_transaction` automatically detects whether the payer has a Permission account on-chain and appends it to the transaction when found — the caller does not need to manage this explicitly.

### Step 1 — Introduce the data model ✅
- Add `Permission` account type, `permission_flags` constants, and `PermissionStatus`.
- Add `authorize()` helper in `authorize.rs` with OR-semantics bitmask check.
- Migrate the instructions that use the most granular roles first: `SetAccessPass`, `CloseAccessPass`, `DeleteUser`, `BanUser`, `RequestBan`, `CreatePermission`, `UpdatePermission`, `SuspendPermission`, `ResumePermission`, `DeletePermission`.

### Step 2 — CLI tooling ✅
- `permission set --user-payer <PUBKEY> --add <flag>... --remove <flag>...` (upsert, incremental delta)
- `permission get --user-payer <PUBKEY>` (shows human-readable flag names)
- `permission list` (shows human-readable flag names)
- `permission suspend --user-payer <PUBKEY>`
- `permission resume --user-payer <PUBKEY>`
- `permission delete --user-payer <PUBKEY>`

### Step 3 — SDK transparent injection ✅
The Rust SDK's `execute_transaction` checks on-chain whether the payer has a Permission account and automatically appends it as a trailing read-only account. No SDK call-site changes are required.

### Step 4 — Migrate remaining instructions (next)
Migrate processors that still check `foundation_allowlist` directly (device, contributor, exchange, location, link, tenant, multicast, globalstate) to go through `authorize()`.

### Step 5 — Enable `RequirePermissionAccounts` on testnet
After Step 4, enable the flag on testnet. Operators must have a Permission account to perform any operation. Run the two-phase integration test (`start-test-permissions.sh`) as the acceptance gate.

### Step 6 — Deprecate GlobalState allowlists
- Remove the write path for `foundation_allowlist`, `qa_allowlist`.
- Replace `activator_authority_pk`, `sentinel_authority_pk`, `health_oracle_pk` with Permission accounts for those roles.
- Enable `RequirePermissionAccounts` on mainnet-beta.

### Step 7 — Remove legacy fields from GlobalState
Once no traffic relies on the legacy path, remove the allowlist fields from `GlobalState` and simplify `authorize()` to the permission-account path only.

juan-malbeclabs:
> **Permission rollout — stacked PR series.** Review/merge order:
> 1. [#3206 — enforce Permission-based authorization in existing instructions](https://github.com/malbeclabs/doublezero/pull/3206)
> 2. [#3942 — define topology/resource/index permission flags (3/4)](https://github.com/malbeclabs/doublezero/pull/3942) — stacked on #3206
> 3. [#3943 — enforce topology/resource/index permission flags (4/4)](https://github.com/malbeclabs/doublezero/pull/3943) — stacked on #3942
>
> Retarget each PR's base to `main` as its upstream merges.
> **👉 You are here: #3206 (base of the follow-up stack).**

## Summary of Changes
- Wire `authorize()` into the privileged instruction processors that gate on it: `accesspass/{close,set}` (`ACCESS_PASS_ADMIN`) and `user/{delete,requestban}` (`USER_ADMIN`). Each appends the caller's Permission PDA as an optional trailing account, read last by `authorize()`. The user owner can still delete their own account without a Permission account.
- Add `execute_authorized_transaction` (and its `_quiet` variant) to the Rust SDK. They share the existing `execute_transaction_inner` builder, so compute-budget, preflight, and error-reporting behavior are identical to `execute_transaction`; the only difference is the optional trailing Permission PDA, which is resolved at most once per client (retried on transient RPC errors, then memoized).
- Switch the affected SDK commands (`accesspass/{close,set}`, `user/{delete,requestban}`, `tenant/delete`, and the `permission/*` CRUD commands) to `execute_authorized_transaction` so the Permission account is included transparently.
- Drop the retired activator authority from the `USER_ADMIN`/`NETWORK_ADMIN`/`MULTICAST_ADMIN` legacy fallbacks. `ACCESS_PASS_ADMIN` legacy authority (foundation + sentinel + feed) is applied uniformly across its instructions and documented in `authorize.rs`.

## Testing Verification
- New DZClient-level tests assert the assembled instruction list carries the protocol-max compute-budget/heap-frame requests and that the Permission account is appended after `payer`+`system`, with and without an on-chain permission account.
- `authorize` unit tests cover the legacy fallbacks (including the activator now being denied for the admin roles) and the Permission-account path (PDA, ownership, discriminator, status, and flag checks).
- `cargo test -p doublezero_sdk` and `cargo test -p doublezero-serviceability` pass.

Related RFC/series: the Permission account and `authorize()` mechanism land in the earlier PRs of this series; this PR enforces them in existing instructions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -568,7 +568,7 @@ All notable changes to this project will be documented in this file.
 - Onchain Programs
   - Serviceability: add `Permission` account with `CreatePermission`, `UpdatePermission`, `DeletePermission`, `SuspendPermission`, and `ResumePermission` instructions for managing per-keypair permission bitmasks onchain
 - SDK
-  - Split `execute_transaction` into `execute_transaction` (no auth) and `execute_authorized_transaction` (injects Permission PDA) to avoid breaking processors that use `accounts.len()` for optional-account detection
+  - Add `execute_authorized_transaction` (and its `_quiet` variant) alongside `execute_transaction`. The authorized variants append the payer's Permission PDA (read-only) as the trailing account when it exists on-chain, so `authorize()` can find it. All variants share the same builder, so the protocol-max compute-budget/heap-frame requests, preflight, and error-reporting behavior are identical to `execute_transaction`; the only difference is the optional trailing Permission account. The Permission PDA lookup is retried on transient RPC errors and memoized per client.
 - CLI
   - Add `permission get`, `permission list`, and `permission set` commands with table and JSON output; `permission set` supports incremental `--add` / `--remove` flags and creates or updates the account as needed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -565,6 +565,12 @@ All notable changes to this project will be documented in this file.
   - Add onchain parent DZD discovery to geoprobe-agent: periodically queries the Geolocation program for this probe's parent devices and resolves their metrics publisher keys from Serviceability, replacing the need for static `--parent-dzd` CLI flags. Static parents from CLI are merged with onchain parents, with onchain taking precedence for duplicate keys.
   - Optimize inbound probe-measured RTT accuracy: pre-sign both TWAMP probes before network I/O so probe 1 fires immediately after reply 0 with no signing delay, measure Tx-to-Rx interval (reply 0 Tx → probe 1 Rx) instead of Rx-to-Rx to exclude processing overhead on both sides, use kernel `SO_TIMESTAMPNS` receive timestamps on the reflector, and add a 15ms busy-poll window on the sender to avoid scheduler wakeup latency
   - Optimize outbound probe RTT accuracy: send a staggered warmup probe on a separate socket 2ms before the measurement probe to wake the reflector's thread, then take the min RTT of both
+- Onchain Programs
+  - Serviceability: add `Permission` account with `CreatePermission`, `UpdatePermission`, `DeletePermission`, `SuspendPermission`, and `ResumePermission` instructions for managing per-keypair permission bitmasks onchain
+- SDK
+  - Split `execute_transaction` into `execute_transaction` (no auth) and `execute_authorized_transaction` (injects Permission PDA) to avoid breaking processors that use `accounts.len()` for optional-account detection
+- CLI
+  - Add `permission get`, `permission list`, and `permission set` commands with table and JSON output; `permission set` supports incremental `--add` / `--remove` flags and creates or updates the account as needed
 
 ## [v0.11.0](https://github.com/malbeclabs/doublezero/compare/client/v0.10.0...client/v0.11.0) - 2026-03-12
 
--- a/smartcontract/programs/doublezero-serviceability/src/authorize.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/authorize.rs
@@ -31,11 +31,11 @@ use solana_program::{
 ///   SENTINEL          → sentinel_authority_pk
 ///   HEALTH_ORACLE     → health_oracle_pk
 ///   FEED_AUTHORITY     → feed_authority_pk
-///   USER_ADMIN        → foundation_allowlist OR activator_authority_pk
+///   USER_ADMIN        → foundation_allowlist
 ///   ACCESS_PASS_ADMIN → foundation_allowlist OR sentinel_authority_pk OR feed_authority_pk
-///   NETWORK_ADMIN     → foundation_allowlist OR activator_authority_pk
+///   NETWORK_ADMIN     → foundation_allowlist
 ///   TENANT_ADMIN      → foundation_allowlist OR sentinel_authority_pk
-///   MULTICAST_ADMIN   → foundation_allowlist OR activator_authority_pk OR sentinel_authority_pk
+///   MULTICAST_ADMIN   → foundation_allowlist OR sentinel_authority_pk
 ///   PERMISSION_ADMIN  → foundation_allowlist (also allowed even when RequirePermissionAccounts is set)
 ///   INFRA_ADMIN       → foundation_allowlist
 ///   GLOBALSTATE_ADMIN → foundation_allowlist
@@ -57,11 +57,13 @@ where
             if permission_account.key != &expected_pda {
                 return Err(ProgramError::InvalidArgument);
             }
-            if permission_account.data_is_empty() {
-                return Err(DoubleZeroError::NotAllowed.into());
-            }
+            // Verify program ownership immediately after the PDA-address check, before
+            // inspecting the account's contents.
             if permission_account.owner != program_id {
                 return Err(ProgramError::InvalidAccountData);
+            }
+            if permission_account.data_is_empty() {
+                return Err(DoubleZeroError::NotAllowed.into());
             }
             let permission = Permission::try_from(permission_account)?;
             if permission.status != PermissionStatus::Activated {
@@ -117,14 +119,17 @@ fn check_legacy_any(payer: &Pubkey, globalstate: &GlobalState, any_of: u128) -> 
     if any_of & permission_flags::FEED_AUTHORITY != 0 && globalstate.feed_authority_pk == *payer {
         return true;
     }
-    // USER_ADMIN in legacy = foundation or activator (historical user management authorities).
+    // USER_ADMIN in legacy = foundation only. (The activator authority has been
+    // retired from the system, so it no longer grants user-management rights.)
     if any_of & permission_flags::USER_ADMIN != 0
-        && (globalstate.foundation_allowlist.contains(payer)
-            || globalstate.activator_authority_pk == *payer)
+        && globalstate.foundation_allowlist.contains(payer)
     {
         return true;
     }
-    // ACCESS_PASS_ADMIN in legacy = foundation, sentinel, or feed authority.
+    // ACCESS_PASS_ADMIN in legacy = foundation, sentinel, or feed authority. This
+    // mirrors the historical accesspass/set authority and is applied uniformly to
+    // all ACCESS_PASS_ADMIN instructions (so accesspass/close, previously
+    // foundation+feed only, now also accepts the sentinel authority).
     if any_of & permission_flags::ACCESS_PASS_ADMIN != 0
         && (globalstate.foundation_allowlist.contains(payer)
             || globalstate.sentinel_authority_pk == *payer
@@ -132,10 +137,10 @@ fn check_legacy_any(payer: &Pubkey, globalstate: &GlobalState, any_of: u128) -> 
     {
         return true;
     }
-    // NETWORK_ADMIN in legacy = foundation or activator.
+    // NETWORK_ADMIN in legacy = foundation only. (The activator authority has
+    // been retired from the system.)
     if any_of & permission_flags::NETWORK_ADMIN != 0
-        && (globalstate.foundation_allowlist.contains(payer)
-            || globalstate.activator_authority_pk == *payer)
+        && globalstate.foundation_allowlist.contains(payer)
     {
         return true;
     }
@@ -146,10 +151,10 @@ fn check_legacy_any(payer: &Pubkey, globalstate: &GlobalState, any_of: u128) -> 
     {
         return true;
     }
-    // MULTICAST_ADMIN in legacy = foundation, activator, or sentinel.
+    // MULTICAST_ADMIN in legacy = foundation or sentinel. (The activator authority
+    // has been retired from the system.)
     if any_of & permission_flags::MULTICAST_ADMIN != 0
         && (globalstate.foundation_allowlist.contains(payer)
-            || globalstate.activator_authority_pk == *payer
             || globalstate.sentinel_authority_pk == *payer)
     {
         return true;
@@ -347,11 +352,12 @@ mod tests {
     }
 
     #[test]
-    fn test_legacy_user_admin_via_activator() {
+    fn test_legacy_user_admin_activator_denied() {
+        // The activator authority has been retired and no longer grants USER_ADMIN.
         let program_id = Pubkey::new_unique();
         let payer = Pubkey::new_unique();
         let gs = gs_with_activator(&payer);
-        assert!(authorize_legacy(&program_id, &payer, &gs, permission_flags::USER_ADMIN).is_ok());
+        assert!(authorize_legacy(&program_id, &payer, &gs, permission_flags::USER_ADMIN).is_err());
     }
 
     #[test]
@@ -429,12 +435,13 @@ mod tests {
     }
 
     #[test]
-    fn test_legacy_network_admin_via_activator() {
+    fn test_legacy_network_admin_activator_denied() {
+        // The activator authority has been retired and no longer grants NETWORK_ADMIN.
         let program_id = Pubkey::new_unique();
         let payer = Pubkey::new_unique();
         let gs = gs_with_activator(&payer);
         assert!(
-            authorize_legacy(&program_id, &payer, &gs, permission_flags::NETWORK_ADMIN).is_ok()
+            authorize_legacy(&program_id, &payer, &gs, permission_flags::NETWORK_ADMIN).is_err()
         );
     }
 
@@ -485,12 +492,13 @@ mod tests {
     }
 
     #[test]
-    fn test_legacy_multicast_admin_via_activator() {
+    fn test_legacy_multicast_admin_activator_denied() {
+        // The activator authority has been retired and no longer grants MULTICAST_ADMIN.
         let program_id = Pubkey::new_unique();
         let payer = Pubkey::new_unique();
         let gs = gs_with_activator(&payer);
         assert!(
-            authorize_legacy(&program_id, &payer, &gs, permission_flags::MULTICAST_ADMIN).is_ok()
+            authorize_legacy(&program_id, &payer, &gs, permission_flags::MULTICAST_ADMIN).is_err()
         );
     }
 
--- a/smartcontract/programs/doublezero-serviceability/src/authorize.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/authorize.rs
@@ -36,7 +36,9 @@ use solana_program::{
 ///   NETWORK_ADMIN     → foundation_allowlist
 ///   TENANT_ADMIN      → foundation_allowlist OR sentinel_authority_pk
 ///   MULTICAST_ADMIN   → foundation_allowlist OR sentinel_authority_pk
-///   PERMISSION_ADMIN  → foundation_allowlist (also allowed even when RequirePermissionAccounts is set)
+///   PERMISSION_ADMIN  → foundation_allowlist (always allowed for foundation, even under
+///                       RequirePermissionAccounts or when the payer's own Permission account
+///                       is missing/suspended/under-privileged — see foundation_permission_recovery)
 ///   INFRA_ADMIN       → foundation_allowlist
 ///   GLOBALSTATE_ADMIN → foundation_allowlist
 ///   CONTRIBUTOR_ADMIN → foundation_allowlist
@@ -62,14 +64,23 @@ where
             if permission_account.owner != program_id {
                 return Err(ProgramError::InvalidAccountData);
             }
-            if permission_account.data_is_empty() {
-                return Err(DoubleZeroError::NotAllowed.into());
-            }
-            let permission = Permission::try_from(permission_account)?;
-            if permission.status != PermissionStatus::Activated {
-                return Err(DoubleZeroError::NotAllowed.into());
-            }
-            if permission.permissions & any_of_flags == 0 {
+            // Whether the supplied Permission account itself grants one of the flags.
+            let granted = !permission_account.data_is_empty()
+                && Permission::try_from(permission_account)
+                    .map(|p| {
+                        p.status == PermissionStatus::Activated && p.permissions & any_of_flags != 0
+                    })
+                    .unwrap_or(false);
+            if !granted {
+                // The SDK auto-appends the payer's Permission PDA whenever it exists
+                // on-chain. Without the recovery below, a foundation member whose own
+                // Permission account is suspended, under-privileged, or uninitialized
+                // would be routed through this branch and denied the very PERMISSION_ADMIN
+                // instruction needed to repair it — re-introducing the lockout the
+                // None-branch fallback exists to prevent.
+                if foundation_permission_recovery(globalstate, payer_key, any_of_flags) {
+                    return Ok(());
+                }
                 return Err(DoubleZeroError::NotAllowed.into());
             }
         }
@@ -81,9 +92,7 @@ where
             ) {
                 // Even in strict mode, foundation members can manage permissions to
                 // prevent being locked out of the permission system.
-                if any_of_flags & permission_flags::PERMISSION_ADMIN != 0
-                    && globalstate.foundation_allowlist.contains(payer_key)
-                {
+                if foundation_permission_recovery(globalstate, payer_key, any_of_flags) {
                     return Ok(());
                 }
                 return Err(DoubleZeroError::NotAllowed.into());
@@ -94,6 +103,19 @@ where
         }
     }
     Ok(())
+}
+
+/// Foundation lockout recovery: a foundation member may always exercise
+/// `PERMISSION_ADMIN`, even in strict mode or when their own Permission account is
+/// missing, suspended, or lacks the flag. This guarantees the permission system can
+/// never lock foundation out of managing permissions.
+fn foundation_permission_recovery(
+    globalstate: &GlobalState,
+    payer_key: &Pubkey,
+    any_of_flags: u128,
+) -> bool {
+    any_of_flags & permission_flags::PERMISSION_ADMIN != 0
+        && globalstate.foundation_allowlist.contains(payer_key)
 }
 
 /// Returns true if `payer` satisfies at least one of the requested flags using legacy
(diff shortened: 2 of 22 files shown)


elitegreg (changes requested):

smartcontract/programs/doublezero-serviceability/tests/create_subscribe_user_test.rs
elitegreg: This whole file should have probably not been touched. All that is changed here is that it got rid of the onchain allocation tests.

elitegreg (changes requested):

smartcontract/programs/doublezero-serviceability/tests/create_subscribe_user_test.rs
juan-malbeclabs: I've updated the code and refreshed the PR. Please review it again.

smartcontract/sdk/rs/src/commands/user/closeaccount.rs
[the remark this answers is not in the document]
juan-malbeclabs: I've updated the code and refreshed the PR. Please review it again.

smartcontract/sdk/rs/src/commands/user/ban.rs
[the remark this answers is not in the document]
juan-malbeclabs: I've updated the code and refreshed the PR. Please review it again.

smartcontract/sdk/rs/src/commands/user/create_subscribe.rs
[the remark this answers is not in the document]
juan-malbeclabs: I've updated the code and refreshed the PR. Please review it again.

smartcontract/sdk/rs/src/client.rs
```diff
@@ -482,6 +483,66 @@ impl DZClient {
...
+            if self.client.get_account(&permission_pda).is_ok() {
+                trailing.push(AccountMeta::new_readonly(permission_pda, false));
+            }
+        }
+
+        let mut transaction = Transaction::new_with_payer(
+            &[Instruction::new_with_bytes(
+                self.program_id,
+                &data,
+                [accounts, trailing].concat(),
```
ben-dz: **C1 (Critical): compute-budget + heap-frame instructions dropped for the entire SDK.** `execute_transaction` now routes through `build_and_send`, which builds a transaction with only the main instruction. `execute_transaction_inner` always prepended `set_compute_unit_limit(1_400_000)` and `request_heap_frame(256 KiB)` (client.rs:213-217); the module comment (client.rs:60-65) says *every* transaction requires the protocol-max budget. Every non-quiet `execute_transaction` caller (device/link/user create, multicastgroup, DeleteUser/RequestBanUser) loses this and will fail with ComputeBudgetExceeded / heap allocation panic when exceeding default 200k CU / 32 KiB. Prepend the two ComputeBudgetInstructions here, or thread `with_permission` through `execute_transaction_inner` and delete `build_and_send`.

smartcontract/sdk/rs/src/client.rs
```diff
@@ -482,6 +483,66 @@ impl DZClient {
...
+
+        if let Some(TransactionError::InstructionError(_index, InstructionError::Custom(number))) =
+            result.value.err
+        {
+            return Err(eyre!(DoubleZeroError::from(number)));
+        } else if let Some(err) = result.value.err {
+            return Err(eyre!(err));
+        }
+
+        self.client
```
ben-dz: **H1 (High): preflight + error-reporting semantics silently changed.** `build_and_send` uses `simulate_transaction` then `send_and_confirm_transaction` with default config (preflight ON) vs the old `skip_preflight: true`, and loses the structured `SimulationError`/`SimulationTransactionError` extraction and the landed-tx log recovery in `execute_transaction_inner`. This is an unannounced semantic change to the whole SDK. Apply one preflight/error strategy uniformly.

smartcontract/sdk/rs/src/client.rs
```diff
@@ -583,6 +644,29 @@ impl DoubleZeroClient for DZClient {
...
+        accounts: Vec<AccountMeta>,
+    ) -> eyre::Result<Signature> {
+        let mut accounts = accounts;
+        if let Some(payer) = self.payer.as_ref() {
+            let (permission_pda, _) = get_permission_pda(&self.program_id, &payer.pubkey());
+            if self.client.get_account(&permission_pda).is_ok() {
+                accounts.push(AccountMeta::new_readonly(permission_pda, false));
+            }
+        }
+        self.execute_transaction_inner(instruction, accounts, true)
```
ben-dz: **H2 (High): `execute_authorized_transaction_quiet` appends the Permission account in the wrong position and is dead code.** It pushes `permission` onto `accounts` before `execute_transaction_inner` appends `[payer, system]`, yielding `[..., permission, payer, system]`. But `build_and_send` (the loud variant) yields `[..., payer, system, permission]`, and processors read the Permission account as the trailing account after payer+system (authorize.rs:53-73). If ever wired up, the processor would read `payer` where it expects the Permission PDA and fail `expected_pda` (authorize.rs:57). Verified zero callers. Delete it, or unify onto the shared builder with a quiet flag + test.

smartcontract/sdk/rs/src/client.rs
```diff
@@ -482,6 +483,66 @@ impl DZClient {
...
+        );
+
+        let blockhash = self.client.get_latest_blockhash().map_err(|e| eyre!(e))?;
+        transaction.sign(&[&payer], blockhash);
+
+        debug!("Simulating transaction: {transaction:?}");
+
+        let result = self.client.simulate_transaction(&transaction)?;
+        if result.value.err.is_some() {
+            eprintln!("Program Logs:");
```
ben-dz: **H3 (High): loud failure path replaces a race-tolerant path for activator-driven instructions.** `DeleteUser`/`RequestBanUser` now use `execute_authorized_transaction`, which eprintln!s program logs and returns a hard, non-structured error on simulation failure. The deliberate quiet path (`SimulationError`/`SimulationTransactionError`, consumed by `execute_transaction_inner`'s quiet branch) that suppresses expected race-failure noise is not used here, and the quiet authorized variant that would address it (H2) is unused/mis-ordered. Provide a correct quiet authorized variant before automation drives these.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/close.rs
```diff
@@ -70,13 +74,15 @@ pub fn process_close_access_pass(
...
-    {
-        return Err(DoubleZeroError::NotAllowed.into());
-    }
+    authorize(
+        program_id,
+        accounts_iter,
+        payer_account.key,
+        &globalstate,
+        permission_flags::ACCESS_PASS_ADMIN,
+    )?;
```
ben-dz, line 85: **M1 (Medium): legacy authorization broadened — sentinel newly authorized.** Old check was `foundation_allowlist OR feed_authority_pk`; the new ACCESS_PASS_ADMIN legacy mapping (authorize.rs:128-134) is `foundation OR sentinel OR feed`, so the sentinel authority can now close access passes on the legacy path where it couldn't before. Confirm intent and document, or use a narrower flag.

smartcontract/programs/doublezero-serviceability/src/processors/user/requestban.rs
```diff
@@ -100,9 +101,13 @@ pub fn process_request_ban_user(
...
     let globalstate = GlobalState::try_from(globalstate_account)?;
-    if !globalstate.foundation_allowlist.contains(payer_account.key) {
-        return Err(DoubleZeroError::NotAllowed.into());
-    }
+    authorize(
+        program_id,
+        accounts_iter,
+        payer_account.key,
+        &globalstate,
+        permission_flags::USER_ADMIN,
```
ben-dz, line 109: **M1 (Medium): legacy authorization broadened — activator newly authorized.** RequestBan's old check was `foundation_allowlist` only; the new USER_ADMIN legacy mapping (authorize.rs:121-126) is `foundation OR activator`. Same broadening applies to user/delete.rs:134-142 (non-owner delete). Plausibly intended (activator manages users), but confirm and document these authorization-surface expansions.

smartcontract/sdk/rs/src/client.rs
```diff
@@ -482,6 +483,66 @@ impl DZClient {
...
+            .ok_or_eyre("No default signer found, run \"doublezero keygen\" to create a new one")?;
+        let data = instruction.pack();
+
+        let mut trailing = vec![
+            AccountMeta::new(payer.pubkey(), true),
+            AccountMeta::new(program::id(), false),
+        ];
+        if with_permission {
+            let (permission_pda, _) = get_permission_pda(&self.program_id, &payer.pubkey());
+            if self.client.get_account(&permission_pda).is_ok() {
```
ben-dz: **M3 (Medium): extra un-retried `get_account` RPC per authorized send (N+1).** This existence probe runs on every authorized transaction and skips the client's `rpc_retry_builder`. In `tenant/delete.rs` (fans out per user to DeleteUser + SetAccessPass) it adds one round-trip per iteration on top of the new double-simulation, and is a guaranteed wasted RPC for admins without a Permission PDA. Resolve/cache the permission AccountMeta once.

smartcontract/sdk/rs/src/doublezeroclient.rs
```diff
@@ -51,6 +51,23 @@ pub trait DoubleZeroClient {
         accounts: Vec<AccountMeta>,
     ) -> eyre::Result<Signature>;
 
+    /// Like `execute_transaction` but appends the payer's Permission PDA
+    /// (read-only) when it exists on-chain, so `authorize()` can find it.
+    /// Use this for instructions whose processor calls `authorize()`.
+    fn execute_authorized_transaction(
+        &self,
+        instruction: DoubleZeroInstruction,
+        accounts: Vec<AccountMeta>,
```
ben-dz, line 60: **L3 (Low): trait doc oversells consistency.** The doc presents `execute_authorized_transaction` and `execute_authorized_transaction_quiet` as parallel, but their implementations diverge (see H2) and the quiet one is dead. Update or remove.

CHANGELOG.md
```diff
@@ -565,6 +565,12 @@ All notable changes to this project will be documented in this file.
...
   - Optimize inbound probe-measured RTT accuracy: pre-sign both TWAMP probes before network I/O so probe 1 fires immediately after reply 0 with no signing delay, measure Tx-to-Rx interval (reply 0 Tx → probe 1 Rx) instead of Rx-to-Rx to exclude processing overhead on both sides, use kernel `SO_TIMESTAMPNS` receive timestamps on the reflector, and add a 15ms busy-poll window on the sender to avoid scheduler wakeup latency
   - Optimize outbound probe RTT accuracy: send a staggered warmup probe on a separate socket 2ms before the measurement probe to wake the reflector's thread, then take the min RTT of both
+- Onchain Programs
+  - Serviceability: add `Permission` account with `CreatePermission`, `UpdatePermission`, `DeletePermission`, `SuspendPermission`, and `ResumePermission` instructions for managing per-keypair permission bitmasks onchain
+- SDK
+  - Split `execute_transaction` into `execute_transaction` (no auth) and `execute_authorized_transaction` (injects Permission PDA) to avoid breaking processors that use `accounts.len()` for optional-account detection
```
ben-dz: **Changelog (Low): inaccurate.** The SDK entry says the split exists 'to avoid breaking processors that use accounts.len() for optional-account detection,' but `build_and_send` also silently changed compute-budget, preflight, and error behavior for all `execute_transaction` callers. Revert that incidental change (preferred) or document it here.

ben-dz (dismissed):
The on-chain `authorize()` wiring across the 4 migrated processors (accesspass close/set, user delete/requestban) is sound and not bypassable. The blocking issue is in the Rust SDK: `execute_transaction` was rerouted through a new `build_and_send` helper that silently drops the protocol-max compute-budget and heap-frame instructions for *every* transaction in the SDK (not just the permission ones), and also changes preflight (skip_preflight → preflight-on) and error-reporting behavior. A second `execute_authorized_transaction_quiet` is dead code that appends the Permission account in the wrong position. None of the new SDK behavior is covered by tests. Also flagging legacy-fallback authorization expansions (sentinel can now close access passes; activator can now ban/delete users) to confirm intent. C1 must be fixed before merge.

- M2 (Medium): No test exercises the assembled transaction. SDK command tests only swap expect_execute_transaction → expect_execute_authorized_transaction on the automock, which short-circuits build_and_send. Nothing verifies compute-budget presence, the conditional permission append, the get_account probe, or the trailing position of the Permission account — the exact gap hiding C1 and H2. Add a DZClient-level test asserting the final Vec<Instruction>/Vec<AccountMeta> with and without an on-chain permission account.
- L2 (Low): Partial migration — only 4 instructions are migrated to authorize(); user/create, user/create_subscribe, user/ban, user/closeaccount, multicastgroup/subscribe, and device/link/tenant CRUD still use inline allowlist checks. Not a regression (verified all accounts.len()-sensitive unmigrated processors still use the non-appending execute_transaction, so no account-count collision). Keep the authorize.rs legacy-fallback table in sync with the unmigrated processors as migration continues, and align the PR summary with the diff (it lists more 'wired' instructions than the diff actually touches).

- `smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set.rs:70` — medium: **M4 (Medium): fragile positional optional-account detection.** `accounts.len() >= 7` infers tenant presence; it is correct alongside the appended Permission account only because tenant adds 2 accounts and permission adds 1 (no-tenant+perm=6 <7; tenant+perm=8 >=7). Any future trailing-layout change silently misclassifies an account as a tenant. Encode optional-account presence in instruction args (as user/delete does), and/or validate the candidate Permission account's AccountType discriminator inside authorize(); add tests for 6- and 8-account shapes.
- `smartcontract/programs/doublezero-serviceability/src/authorize.rs:63` — low: **L1 (Low): check ordering.** `data_is_empty` (line 60) is checked before `owner == program_id` (line 63). Safe because the PDA-address equality at line 57 already blocks account substitution. Optional hardening: assert `owner == program_id` immediately after the PDA check.

smartcontract/sdk/rs/src/client.rs
juan-malbeclabs: Fixed in 4660f087. Removed `build_and_send`; `execute_transaction` routes back through `execute_transaction_inner`, which prepends `set_compute_unit_limit(MAX)` + `request_heap_frame(MAX)` for every transaction (see the new `assemble_instructions`). New test `prepends_compute_budget_instructions` locks this in.

smartcontract/sdk/rs/src/client.rs
juan-malbeclabs: Fixed in 4660f087 by the same unification — every send now goes through `execute_transaction_inner`, so `skip_preflight: true` plus the structured `SimulationError`/`SimulationTransactionError` extraction and landed-tx log recovery are applied uniformly. The `simulate_transaction` + preflight-on detour is gone.

smartcontract/sdk/rs/src/client.rs
juan-malbeclabs: Fixed in 4660f087. `execute_authorized_transaction_quiet` now shares `execute_transaction_inner` with `with_permission=true`, so the Permission account is appended as the trailing account (after payer+system), matching the loud variant and `authorize()`s read order. Covered by `permission_account_is_appended_after_payer_and_system`.

smartcontract/sdk/rs/src/client.rs
juan-malbeclabs: Resolved in 4660f087. Both authorized variants now share the builder, so a correct quiet authorized variant exists. DeleteUser/RequestBanUser use `execute_authorized_transaction`, which restores the exact loud+structured behavior they had on main (they used `execute_transaction` there), so there is no race-tolerance regression versus main. The activator that drove these has since been retired.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/close.rs
juan-malbeclabs: Addressed in 4660f087. `ACCESS_PASS_ADMIN` keeps foundation+sentinel+feed (the historical `accesspass/set` authority) and is now applied uniformly to close; documented in `authorize.rs` as a deliberate expansion (close was previously foundation+feed).

smartcontract/programs/doublezero-serviceability/src/processors/user/requestban.rs
juan-malbeclabs: Fixed in 4660f087. The activator authority has been retired from the system, so it is removed from the USER_ADMIN, NETWORK_ADMIN, and MULTICAST_ADMIN legacy fallbacks. USER_ADMIN legacy is now foundation-only, matching the original delete/requestban authority. Tests updated to assert the activator is denied.

smartcontract/sdk/rs/src/client.rs
juan-malbeclabs: Fixed in 4660f087. The Permission PDA probe now goes through `rpc_retry_builder` and is memoized per client (`permission_account_cache`), so it resolves at most once instead of one un-retried `get_account` per authorized send.

smartcontract/sdk/rs/src/doublezeroclient.rs
juan-malbeclabs: Fixed in 4660f087. The two authorized variants now share `execute_transaction_inner`, so they are genuinely parallel and the quiet variant is no longer dead/divergent — the doc is now accurate.

CHANGELOG.md
juan-malbeclabs: Fixed in 4660f087. Rewrote the SDK entry: the authorized variants differ from `execute_transaction` only by the optional trailing Permission account; compute-budget, preflight, and error-reporting behavior are identical.

juan-malbeclabs:
## Review feedback addressed (4660f087)

Thanks for the thorough review. Inline replies cover C1, H1, H2, H3, M1, M3, L3, and the changelog note. Summary findings:

- **M2** — Added DZClient-level tests asserting the assembled instruction list: compute-budget instructions are present, and the Permission account is appended as the trailing account both with and without an on-chain permission account (`assemble_instructions_tests` in `client.rs`). The build/send path was extracted into a pure `assemble_instructions` so it is testable without RPC.
- **M4** — `authorize()` already rejects a misclassified account (it validates the PDA address, program ownership, and the `AccountType::Permission` discriminator), so the trailing Permission account can never be read as a tenant. Documented the exact `accounts.len()` arithmetic invariant in `accesspass/set.rs`. Encoding tenant presence in instruction args is a worthwhile follow-up but is an ABI change, so left out of this PR.
- **L1** — Reordered the checks in `authorize()` so program ownership is verified immediately after the PDA-address check, before inspecting account data.
- **L2** — The activator broadening is removed (see M1); the activator has been retired from the system. PR description updated to match the actual (rebased) diff.

smartcontract/sdk/rs/src/commands/user/delete.rs
```diff
@@ -120,7 +120,7 @@ impl DeleteUserCommand {
 
         accounts.push(AccountMeta::new(user.owner, false));
 
-        client.execute_transaction(
+        client.execute_authorized_transaction(
```
vihu, line 123: This makes `USER_ADMIN` Permission-based delete/request-ban fail for multicast users that still have roles.

`DeleteUserCommand` and `RequestBanUserCommand` now authorize the final `DeleteUser` / `RequestBanUser` instruction, but they first clear multicast roles through `UpdateMulticastGroupRolesCommand`. That command still sends a plain `execute_transaction`, and the on-chain role-update processor only allows the access-pass payer or a foundation member.

So a non-foundation, non-owner payer with a valid `USER_ADMIN` Permission account cannot complete delete/request-ban for a multicast user with publisher/subscriber refs: pre-cleanup fails, and skipping cleanup also fails because the target processors reject non-empty `publishers` / `subscribers` with `ReferenceCountNotZero`.

This is related to the earlier “partial migration” note, but it is an end-to-end caller-flow regression: the migrated `USER_ADMIN` operations still depend on an unmigrated prerequisite transaction. Maybe make removal-only role cleanup usable by the same authority that can delete/request-ban the user, or move the cleanup into the `USER_ADMIN`-authorized processors.

smartcontract/programs/doublezero-serviceability/src/authorize.rs
```diff
@@ -117,25 +119,28 @@ fn check_legacy_any(payer: &Pubkey, globalstate: &GlobalState, any_of: u128) ->
     if any_of & permission_flags::FEED_AUTHORITY != 0 && globalstate.feed_authority_pk == *payer {
         return true;
     }
-    // USER_ADMIN in legacy = foundation or activator (historical user management authorities).
+    // USER_ADMIN in legacy = foundation only. (The activator authority has been
```
vihu, line 144: The foundation `PERMISSION_ADMIN` recovery path can be bypassed by SDK auto-injection when the payer’s own Permission PDA exists but is bad.

`authorize()` only reaches the foundation lockout fallback in the `None` branch. But the SDK now auto-appends the payer Permission PDA whenever it exists. If a foundation admin’s own Permission account is suspended or missing `PERMISSION_ADMIN`, Permission CRUD commands take this `Some(permission_account)` branch and return `NotAllowed` before the foundation fallback is considered.

That means a foundation key can be unable to repair/resume its own Permission account via the normal SDK/CLI path, despite the explicit “prevent being locked out” fallback.

Maybe for `PERMISSION_ADMIN`, allow foundation fallback even when the supplied payer Permission account is unusable, or provide an explicit SDK/CLI recovery path that intentionally omits the payer Permission PDA.

elitegreg:
**accesspass/close.rs — sentinel can now close *any* access pass**

Switching `close` to the `ACCESS_PASS_ADMIN` legacy set (`foundation || sentinel || feed`) broadens the close authority to include the sentinel authority, which the PR description notes is intentional. One thing I'd like to confirm: the existing own-only guard a bit further down only restricts the *feed* authority:

```rust
// Feed authority can only close access passes they own
if globalstate.feed_authority_pk == *payer_account.key
    && accesspass.owner != *payer_account.key
{
    return Err(DoubleZeroError::NotAllowed.into());
}
```

Since sentinel now passes the same authorization check but isn't covered by this guard, sentinel can close *any* access pass, not just ones it owns. Is that the intended scope for sentinel, or should the own-only restriction extend to it as well?

smartcontract/sdk/rs/src/commands/user/delete.rs
juan-malbeclabs: Good catch — this was a real end-to-end gap. `DeleteUserCommand`/`RequestBanUserCommand` authorize the final instruction with `USER_ADMIN`, but the role cleanup ran through `UpdateMulticastGroupRoles`, whose processor only accepted the access-pass owner or a foundation member, so a `USER_ADMIN`-only payer was blocked on the prerequisite.

Fixed in 14027857: the role-update processor now accepts `USER_ADMIN` for the **removal-only** case (`!publisher && !subscriber`) — i.e. the same authority that can delete/request-ban the user can now strip its roles as a prerequisite. Adding roles still requires the access-pass owner or foundation. On the SDK side `UpdateMulticastGroupRolesCommand` goes through `execute_authorized_transaction` so the payer's Permission account is appended when present (and is harmlessly ignored for owner/foundation callers). Added integration coverage for both the allowed removal and the rejected add (`test_unsubscribe_user_admin_permission_allowed` / `test_subscribe_user_admin_permission_rejected`).

smartcontract/programs/doublezero-serviceability/src/authorize.rs
juan-malbeclabs: Agreed, and confirmed the bypass: since the SDK auto-appends the payer's Permission PDA whenever it exists, a foundation member with a suspended / under-privileged / uninitialized own Permission account was routed through the `Some` branch and denied the very `PERMISSION_ADMIN` instruction needed to repair it — the `None`-branch fallback never ran.

Fixed in 20dcbed1: extracted `foundation_permission_recovery()` and applied it in the `Some` branch too. When the supplied Permission account doesn't grant the flag, a foundation member requesting `PERMISSION_ADMIN` is still allowed. The PDA-address and program-ownership checks remain hard errors (those indicate a malformed request, not a lockout), and recovery is scoped to `PERMISSION_ADMIN` only. Added tests for the suspended and missing-flag recovery cases, the non-foundation denial, and that recovery doesn't leak to other flags (e.g. `USER_ADMIN`).

juan-malbeclabs:
@elitegreg Good catch, and yes, the broadening to sentinel is intentional.

To answer the question directly: the own-only guard should not extend to sentinel. It's a feed-authority-specific carve-out by design, and the same pattern holds in set.rs (the update path guards only on feed_authority_pk as well). The intent is two tiers inside ACCESS_PASS_ADMIN:

- foundation + sentinel → operator-tier admins, unrestricted scope
- feed → delegated authority, limited to access passes it owns

The reason own-only makes sense for feed but not for sentinel is ownership semantics: owner is set to the payer at creation (set.rs:197), and feed authority owns the passes it creates. Sentinel is a monitoring/enforcement authority — it doesn't create passes for itself, so it would essentially never be the owner. Applying the own-only guard to it would mean it could close almost nothing, which defeats the purpose of granting it ACCESS_PASS_ADMIN.

So the real trust decision here is whether sentinel should hold ACCESS_PASS_ADMIN at all (which is the intended scope) rather than scoping it down with the feed guard. Leaving sentinel unrestricted is the consistent and intended behavior. No change planned here.

elitegreg (approved):

PR #3266, opened 2026-03-15
activator/client: handle device full when tunnel IDs exhausted

martinsander00:
## Summary of Changes
- Cap tunnel ID allocation per device (500–627) so the activator gracefully rejects users when a device is full instead of allocating invalid tunnel IDs
- Update the CLI to stop retrying on rejection and surface the rejection reason as an error, so operators see a clear "device full" message instead of hanging indefinitely

## Diff Breakdown
| Core logic   |     5 | +37 / -16   | +21  |
| Tests        |     1 | +48 / -7    | +41  |

Majority of new lines are tests for the bounded allocator; core logic is compact.

<details>

- `activator/src/idallocator.rs` — add `with_max` constructor and `Option<u16>` return from `next_available`; 3 new test cases for bounded allocation
- `client/doublezero/src/command/connect.rs` — stop polling on `Rejected` status; return rejection reasons as an error instead of silently succeeding
- `activator/src/process/user.rs` — handle `None` from tunnel ID allocation by rejecting the user with a descriptive onchain log message
- `activator/src/states/devicestate.rs` — cap tunnel IDs at 627 via `IDAllocator::with_max`
- `activator/src/process/iface_mgr.rs` — unwrap `Option` with descriptive panic for segment routing IDs
- `activator/src/process/link.rs` — unwrap `Option` with descriptive panic for link IDs

</details>

## Testing Verification
- All 13 `IDAllocator` unit tests pass, including 3 new tests covering bounded allocation (upper bound respected, gap filling, exhaustion)
- CLI rejection path returns `Err` with reason detail, ensuring `doublezero connect` exits non-zero on device-full rejection

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 - Activator
-  - Cap tunnel ID allocation per device (500–627) and gracefully reject users when a device is full instead of allocating invalid tunnel IDs
+  - Cap tunnel ID allocation per device based on `max_users` and gracefully reject users when a device is full instead of allocating invalid tunnel IDs
 - Client
   - Surface rejection reason as a CLI error on `doublezero connect` instead of hanging indefinitely when the activator rejects a user (e.g., device full)
   - Demote passive-mode liveness session-down log messages from Info to Debug to reduce log noise when no dataplane action is taken


activator/src/states/devicestate.rs
```diff
@@ -50,7 +50,7 @@ impl DeviceState {
                 .iter()
                 .map(|b| IPBlockAllocator::new((*b).into()))
                 .collect(),
-            tunnel_ids: IDAllocator::new(500, vec![]),
+            tunnel_ids: IDAllocator::with_max(500, 627, vec![]),
```
packethog: I don't think we can set an arbitrary upper bound here. This can be changed at anytime by the contributor. If they move it up past 128, the activator would never assign a tunnel id.

Issue #2924, opened 2026-02-11
New geoProbe targets should have their first probe sent immediately


ben-dz:
When on-chain discovery of geoProbe targets is implemented, we should ensure that a first probe is sent immediately, in addition to the probe being added to the list of probes to be sent on regular intervals.

This will alleviate a long delay before a new target receives its first LocationOffset.

PR #3327, opened 2026-03-20
geoprobe: probe newly discovered targets immediately

ben-dz:
Resolves: #2924

## Summary of Changes
- Newly discovered onchain targets are probed immediately upon discovery instead of waiting up to 5 minutes for the next measurement ticker
- Extracted `sendCompositeOffsets` from `runMeasurementCycle` so the offset composition/signing/sending logic can be reused for immediate probes
- Added `Pinger.MeasureOne` to measure a single target using the same warmup+measurement probe pattern as `MeasureAll`

## Diff Breakdown
| Core logic   |     2 | +87 / -6    | +81  |

All net-new logic; no tests, config, or docs changes.

<details>

- `controlplane/telemetry/internal/geoprobe/pinger.go` — new `MeasureOne` method for single-target measurement
- `controlplane/telemetry/cmd/geoprobe-agent/main.go` — immediate probe on target discovery; extracted `sendCompositeOffsets` helper

</details>

## Testing Verification
- Existing geoprobe unit tests pass (including pinger concurrency tests)
- Build verified in dev container

### DIFF
--- a/controlplane/telemetry/internal/geoprobe/pinger_test.go
+++ b/controlplane/telemetry/internal/geoprobe/pinger_test.go
@@ -646,3 +646,67 @@ func TestPinger_DualProbe_BothFail(t *testing.T) {
 	assert.NotContains(t, results, addr,
 		"should not have result when both probes fail")
 }
+
+func TestPinger_MeasureOne_UnknownProbe(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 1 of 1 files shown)


ben-dz:
Good thought- if it turns into a problem, we can spin off a temporary thread to do the new first-time probes.

PR #3337, opened 2026-03-20
geolocation: add target_update_count to GeoProbe for lightweight target change detection

### LINKED ISSUE #3304: geolocation: add target_update_count to GeoProbe for lightweight change detection
The geoprobe-agent polls all GeolocationUser accounts every 60s to discover targets assigned to it. Each account can be up to ~284 KB, making this the most expensive RPC call in the agent's loop. To avoid unnecessary scans, add a target_update_count field to the onchain GeoProbe account. The add_target and remove_target instructions already touch the probe for reference_count, so incrementing a second counter is trivial. Since the agent already fetches its own GeoProbe account every 60s for parent discovery, it can check whether the counter has changed before deciding to do the full user scan. The field should use BorshDeserializeIncremental and be appended at the end of the struct, so existing accounts deserialize with a default of 0 — no migration or breaking change required.

Changes that don't go through add_target/remove_target (e.g., payment status updates, user suspension) don't need to increment the counter. A forced full refresh every ~5 minutes catches those cases, which is acceptable given the lower urgency of payment enforcement convergence compared to target routing changes.

ben-dz:
Closes #3304

## Summary of Changes
- Add `target_update_count: u32` field to the onchain `GeoProbe` account, incremented by `add_target` and `remove_target` processors. Non-breaking: uses `BorshDeserializeIncremental` so existing accounts default to 0.
- Extend the Go SDK to deserialize the new field with backward compatibility (EOF → 0).
- Wire up the geoprobe-agent so parent discovery publishes the counter via a shared `atomic.Uint32`, and target discovery skips the expensive `GetGeolocationUsers` RPC scan when the counter hasn't changed. A forced full refresh every 5th tick (~5 min at 60s intervals) catches non-counter changes like payment status updates.

## Diff Breakdown
| Core logic   |     5 | +109 / -53  | +56   |
| Scaffolding  |     8 | +11 / -0    | +11   |
| Tests        |     6 | +278 / -25  | +253  |

~80% of the diff is tests; core logic is compact.

<details>

- `controlplane/telemetry/internal/geoprobe/target_discovery.go` — skip-scan logic: check shared `target_update_count` atomic, force full refresh every 5th tick, nil-return for skipped scans
- `smartcontract/programs/doublezero-geolocation/src/state/geo_probe.rs` — add `target_update_count` field with `BorshDeserializeIncremental`, backward-compat deserialization, updated Display/tests
- `controlplane/telemetry/internal/geoprobe/onchain_discovery.go` — parent discovery stores `probe.TargetUpdateCount` into shared atomic after each poll
- `controlplane/telemetry/cmd/geoprobe-agent/main.go` — create shared `atomic.Uint32`, pass to both parent and target discovery configs
- `sdk/geolocation/go/state.go` — add `TargetUpdateCount uint32` to Go `GeoProbe` struct, backward-compat deserialization

</details>

## Testing Verification
- Rust unit tests pass (39 geolocation lib tests), including new backward-compat test that truncates serialized data to simulate pre-`target_update_count` accounts
- Integration test assertions added: `target_update_count` increments after `add_target`/`remove_target`, unchanged after `update_payment_status`
- Go SDK round-trip and backward-compat tests pass
- 4 new geoprobe-agent target discovery tests: counter unchanged → scan skipped, counter changed → full scan, forced full refresh ignores counter, nil counter → always scans
- New parent discovery test verifies `target_update_count` is stored in the shared atomic
- Full workspace `cargo build`, `cargo clippy`, `rust-fmt`, `go test`, `golangci-lint` all pass

### DIFF
--- a/controlplane/telemetry/cmd/geoprobe-agent/main.go
+++ b/controlplane/telemetry/cmd/geoprobe-agent/main.go
@@ -36,7 +36,6 @@ const (
 	defaultEvictionInterval      = 30 * time.Minute
 	defaultVerifyInterval        = 29 * time.Second
 	discoveryInterval            = 60 * time.Second
-	parentDiscoveryInterval      = 60 * time.Second
 )
 
 var (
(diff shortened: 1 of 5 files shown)


nikw9944:
We should probably explicitly handle (and log) when the counter wraps.

PR #3378, opened 2026-03-25
geoprobe: create sockets on-demand instead of holding them open per target

ben-dz:
## Summary of Changes
- Replace persistent per-target sender pairs with on-demand socket creation: sockets are created at probe time and closed immediately after, eliminating the long-lived ephemeral-port consumption that could exhaust the source-port range
- `Pinger` now tracks only target addresses; `probeTarget()` creates and closes a sender pair each measurement cycle via the extracted `createSenderPair()` helper
- Tighten `retryOnBindError` now that socket creation is on the probe hot path: drop the wasted backoff after the final attempt and lower the base delay 50ms→20ms, cutting worst-case bind-retry latency from ~350ms to ~60ms per socket (this helper is shared with the publisher's `AddProbe`, which benefits too)
- Warn when sender-pair creation nears the worst-case retry budget (~100ms, just under the ~120ms/pair ceiling), surfacing sustained bind contention without flagging every retry
- Consolidate duplicated probe logic from `probeWorker`/`MeasureOne` into `probeTarget()`, and inject sender creation through a `senderFactory` for testing

## Diff Breakdown
| Category   | Files | Lines (+/-) | Net |
|------------|-------|-------------|-----|
| Core logic |     2 | +133 / -160 | -27 |
| Tests      |     2 | +103 / -63  | +40 |
| **Total**  |     4 | +236 / -223 | +13 |

Net reduction in core logic — removing persistent socket management outweighs the new on-demand lifecycle and instrumentation.

<details>

- [`controlplane/telemetry/internal/geoprobe/pinger.go`](https://github.com/malbeclabs/doublezero/pull/3378/files#diff-1b07aba6b82794f8591f10188a953966d4db60895d6480153bd5b4625750b8da) — on-demand socket lifecycle: `targets` map replaces `senders`; new `createSenderPair()`/`probeTarget()`; `senderFactory` injection; slow-creation WARN
- [`controlplane/telemetry/internal/geoprobe/pinger_test.go`](https://github.com/malbeclabs/doublezero/pull/3378/files#diff-9d4e81fe02e35b543010c3d773233311b087a01b71548840ec777864b35f5054) — switch tests to `senderFactory` injection (`mockFactory`); assertions moved from `senders` to `targets`
- [`controlplane/telemetry/internal/geoprobe/retry_test.go`](https://github.com/malbeclabs/doublezero/pull/3378/files#diff-023db9299a9c916153e34b7faa418a784cac133dd2360bdc02c91c0f1a0542ca) — add `TestRetryOnBindError_NoBackoffAfterFinalAttempt` pinning the no-trailing-sleep behavior
- [`controlplane/telemetry/internal/geoprobe/retry.go`](https://github.com/malbeclabs/doublezero/pull/3378/files#diff-c08e40de7d6876c3672d879649c5965dda661d2e9a357040901d1243d11a2ae5) — skip backoff after the final attempt; clarify doc comment

</details>

## Testing Verification
- Existing dual-probe scenarios preserved (min RTT, one-fails-uses-other, both-fail, `MeasureOne`, concurrent add/remove)
- Added `TestRetryOnBindError_NoBackoffAfterFinalAttempt`; the prior tests only asserted call count, so the wasted final-attempt sleep went uncaught
- Full `geoprobe` package passes under `-race`, including the 1000- and 2000-probe large-scale tests that create real sockets and exercise the bind-retry path under contention

### DIFF
--- a/controlplane/telemetry/internal/geoprobe/pinger.go
+++ b/controlplane/telemetry/internal/geoprobe/pinger.go
@@ -20,6 +20,8 @@ const (
 	senderRetryMin      = 50 * time.Millisecond
 )
 
+type senderFactory func(ctx context.Context, log *slog.Logger, iface string, local, remote *net.UDPAddr) (twamplight.Sender, error)
+
 type PingerConfig struct {
 	Logger              *slog.Logger
 	ProbeTimeout        time.Duration
@@ -31,21 +33,21 @@ type PingerConfig struct {
 type Pinger struct {
 	log       *slog.Logger
 	cfg       *PingerConfig
-	senders   map[string]*senderEntry
-	sendersMu sync.Mutex
+	targets   map[string]ProbeAddress
+	targetsMu sync.Mutex
+	newSender senderFactory
 }
 
-type senderEntry struct {
-	addr         ProbeAddress
-	sender       twamplight.Sender
-	warmupSender twamplight.Sender
+func defaultSenderFactory(ctx context.Context, log *slog.Logger, iface string, local, remote *net.UDPAddr) (twamplight.Sender, error) {
+	return newSenderWithRetry(ctx, log, iface, local, remote)
 }
 
 func NewPinger(cfg *PingerConfig) *Pinger {
 	return &Pinger{
-		log:     cfg.Logger,
-		cfg:     cfg,
-		senders: make(map[string]*senderEntry),
+		log:       cfg.Logger,
+		cfg:       cfg,
+		targets:   make(map[string]ProbeAddress),
+		newSender: defaultSenderFactory,
 	}
 }
 
@@ -55,13 +57,12 @@ func newSenderWithRetry(ctx context.Context, log *slog.Logger, iface string, loc
 	})
 }
 
-func (p *Pinger) AddProbe(ctx context.Context, addr ProbeAddress) error {
-	p.sendersMu.Lock()
-	defer p.sendersMu.Unlock()
+func (p *Pinger) AddProbe(_ context.Context, addr ProbeAddress) error {
+	p.targetsMu.Lock()
+	defer p.targetsMu.Unlock()
 
 	key := addr.String()
-
-	if _, exists := p.senders[key]; exists {
+	if _, exists := p.targets[key]; exists {
 		p.log.Debug("Probe already exists", "probe", key)
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


controlplane/telemetry/internal/geoprobe/pinger.go
```diff
@@ -70,120 +71,110 @@ func (p *Pinger) AddProbe(ctx context.Context, addr ProbeAddress) error {
...
+	if err != nil {
+		p.log.Warn("Failed to create senders", "probe", addr.String(), "error", err)
+		return 0, false
+	}
+	defer sender.Close()
+	defer warmup.Close()
 
-	key := addr.String()
+	probeCtx, cancel := context.WithTimeout(ctx, p.cfg.ProbeTimeout)
+	defer cancel()
```
nikw9944, line 141: The dual-probe rationale comment from the old probeWorker was dropped during extraction ("Send a warmup probe first to wake the reflector's thread... Both run on separate sockets... We take the min RTT"). probeTarget() is now the sole home of this non-obvious logic — note especially that the var named `warmup` probes immediately while `sender` is the delayed measurement. Re-add the comment here to preserve the 'why'.

controlplane/telemetry/internal/geoprobe/pinger.go
```diff
@@ -31,21 +33,21 @@ type PingerConfig struct {
...
+	targets   map[string]ProbeAddress
+	targetsMu sync.Mutex
+	newSender senderFactory
 }
 
-type senderEntry struct {
-	addr         ProbeAddress
-	sender       twamplight.Sender
-	warmupSender twamplight.Sender
+func defaultSenderFactory(ctx context.Context, log *slog.Logger, iface string, local, remote *net.UDPAddr) (twamplight.Sender, error) {
```
nikw9944: `defaultSenderFactory` has the identical signature to `newSenderWithRetry` and only forwards its args. NewPinger could assign `newSender: newSenderWithRetry` directly and drop this wrapper.

controlplane/telemetry/internal/geoprobe/pinger_test.go
```diff
@@ -28,6 +29,17 @@ func (m *mockSender) LocalAddr() *net.UDPAddr {
 	return &net.UDPAddr{IP: net.IPv4zero, Port: 0}
 }
 
+func mockFactory(sender, warmup *mockSender) senderFactory {
```
nikw9944, line 38: mockFactory distinguishes measurement vs. warmup by an unsynchronized `call++` and an odd/even rule. Fine today because every test using it registers a single target (createSenderPair runs serially, exactly two calls, sender-then-warmup). But a future multi-target test would race on `call` under -race and mis-pair senders if the call order changes. Add a comment noting the single-target assumption, or guard `call` with a mutex.

controlplane/telemetry/internal/geoprobe/pinger.go
```diff
@@ -70,120 +71,110 @@ func (p *Pinger) AddProbe(ctx context.Context, addr ProbeAddress) error {
...
+	delete(p.targets, key)
+	p.log.Info("Removed probe", "probe", key)
+	return nil
+}
+
+func (p *Pinger) createSenderPair(ctx context.Context, addr ProbeAddress) (sender, warmup twamplight.Sender, err error) {
+	resolvedAddr := &net.UDPAddr{IP: net.ParseIP(addr.Host), Port: int(addr.TWAMPPort)}
+	iface := p.cfg.ManagementNamespace
+
+	sender, err = p.newSender(ctx, p.log, iface, &net.UDPAddr{IP: net.IPv4zero, Port: 0}, resolvedAddr)
```
nikw9944, line 109: Socket creation moved from AddProbe (once per target, off hot path) to probeTarget (twice per target every cycle), and the two newSender calls in createSenderPair run sequentially. With retryOnBindError backing off up to ~350ms under bind contention, this adds per-probe latency. Acceptable trade-off for the port-exhaustion goal; if cycle latency becomes a concern, the two sockets could be created concurrently.

nikw9944 (commented):
Clean, faithful refactor: the Pinger no longer holds two long-lived UDP sockets per target and instead creates/closes a sender pair per measurement in the consolidated `probeTarget()`. Verified on-branch: `go test -race` passes with no data races, and `go vet`/`go build` are clean. Since these are UDP sockets, Close() frees the ephemeral port immediately, so the design goal (bound concurrent port usage to ~2×numWorkers) holds. No critical or high issues — findings are non-blocking suggestions around a dropped explanatory comment, a redundant factory wrapper, test-mock fragility, an untested error path, and the latency cost of moving socket creation onto the hot path.

- Low: the new socket-creation failure paths in createSenderPair (first-sender error; warmup error → first sender closed) and probeTarget (Warn + return 0,false) are untested — the injected test factory never returns an error. The previous design surfaced creation failures synchronously via AddProbe's return value (which was covered); that signal moved to probe time and lost its test. A mockFactory variant returning an error would cover it cheaply.

controlplane/telemetry/internal/geoprobe/pinger.go
ben-dz: Restored in `probeTarget`, with the naming caveat you flagged spelled out:

```go
// Send a warmup probe first to wake the reflector's thread, then send the
// measurement probe after a short delay. Both run on separate sockets so
// neither blocks the other; we take the min RTT. Note the naming: warmup
// fires immediately, while sender is the delayed measurement probe.
```

controlplane/telemetry/internal/geoprobe/pinger.go
ben-dz: Done — removed `defaultSenderFactory`; `NewPinger` now assigns `newSender: newSenderWithRetry` directly. Identical signatures, so the wrapper was pure forwarding.

controlplane/telemetry/internal/geoprobe/pinger_test.go
ben-dz: Went with the comment rather than a mutex. A mutex would make `call++` race-free, but it wouldn't fix the hazard you're pointing at: under concurrent `createSenderPair` calls the interleaving is still nondeterministic, so sender/warmup would mis-pair regardless — it'd just silence `-race` while leaving the logic broken. The comment documents the single-target constraint and keeps `-race` as a tripwire for whoever adds the first multi-target test.

controlplane/telemetry/internal/geoprobe/pinger.go
ben-dz: Kept the two sockets sequential — agreed the trade-off is acceptable. But tracing the latency into `retryOnBindError` turned up two fixes worth making now that this runs every cycle instead of once per target in `AddProbe`:

1. It backed off *after* the final attempt before giving up, so the worst case was `50+100+200ms` but the trailing 200ms was dead time with no retry behind it (and the "Bind failed, retrying" log on the last attempt was misleading). It now only sleeps *between* attempts.
2. Lowered the base backoff `50ms→20ms` — the transient `EINVAL` it guards clears quickly.

Worst-case per-socket backoff drops `~350ms→~60ms` (pair: `~700ms→~120ms`). Added `TestRetryOnBindError_NoBackoffAfterFinalAttempt` to lock in the no-trailing-sleep behavior — the existing test only asserted the call count, which is why this slipped through.

Also added a WARN in `createSenderPair` when pair creation crosses ~100ms, i.e. just below the worst-case retry budget (~120ms/pair), so we can see whether we're hitting bad bind contention often rather than flagging every retry. Note the `retryOnBindError` change is shared with `publisher.go`'s `AddProbe`, which is strictly better off too.

nikw9944 (approved):

Issue #3343, opened 2026-03-23
AdminGroupBits ResourceExtension


ben-malbeclabs:
Add persistent bitmap to track allocated admin-group bits. Bits are never reused after deletion.

**Milestone:** RFC-18: Onchain Foundation
**RFC:** https://github.com/malbeclabs/doublezero/pull/3288

---

**Files affected:**
- `smartcontract/programs/doublezero-serviceability/src/state/resource_extension.rs`
- `smartcontract/programs/doublezero-serviceability/src/processors/resource/`

**Size:** Medium

**Tests:**
- Bitmap exhaustion returns explicit error
- Allocation always returns lowest available bit
- Deallocated bits are not reused

**Blockers:** None

ben-malbeclabs:
Implemented in #3497 (RFC-18 PR 1).

ben-malbeclabs:
Implemented in #3497 (merged).

PR #3389, opened 2026-03-27
smartcontract: add AdminGroupBits ResourceExtension (#3343)

Rhovian:
## Summary of Changes
This PR adds `AdminGroupBits` as a new `ResourceExtension` type, including PDA derivation, seed wiring, range definition (0..32), and focused allocator/resource-extension tests for lowest-available allocation and exhaustion.

Assumptions for follow-up work:
- This PR does not change global bootstrap or SDK/CLI `SetGlobalConfig` account lists, so `AdminGroupBits` is expected to be created explicitly by the future RFC-18 `LinkColorInfo` flow. 
- The “bits are never reused after deletion” policy is also deferred to that consumer logic by not calling `deallocate_id()` when a color is removed, rather than changing the generic `IdAllocator` behavior.

## Testing Verification
Screenshot 2026-03-26 at 6 19 43 PM

cc @ben-malbeclabs

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/id_allocator.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/id_allocator.rs
@@ -231,6 +231,38 @@ mod tests {
     }
 
     #[test]
+    fn test_admin_group_bits_exhaustion() {
+        let mut aligned_data = AlignedBitmap([0u8; 8]);
+        let mut allocator = IdAllocator::new((0, 32)).unwrap();
+
+        for expected in 0..32 {
+            let id = allocator.allocate(&mut aligned_data.0);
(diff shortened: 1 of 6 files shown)


ben-dz:
Hi @Rhovian — thanks for this, and apologies that it sat without a review. This functionality ended up landing in #3497 as part of the broader RFC-18 topology/flex-algo work. Your design matched where we ended up. Main difference is the bit range grew from 0..32 to 1..127 as the flex-algo design evolved. Since it's fully covered on main and this branch now conflicts, I'm going to close this PR. Appreciated the clean implementation and the allocator tests. Sorry that this fell by the wayside. 

We've been working to be better about reviewing external contributions, and have better systems now to shepherd external work through CI and get it merged.

PR #3446, opened 2026-04-03
feat(cli): handle non-user owned disconnects (#3443)

bgm-malbeclabs:
## Summary of Changes
This PR adds in a graceful error message when a user tries to disconnect when they don't own the connection

## Testing Verification
* Added unit tests to verify CLI behavior

Closes https://github.com/malbeclabs/doublezero/issues/3443

### DIFF
--- a/client/doublezero/src/command/disconnect.rs
+++ b/client/doublezero/src/command/disconnect.rs
@@ -17,7 +17,7 @@ use doublezero_cli::{
 
 use doublezero_sdk::{
     commands::user::{delete::DeleteUserCommand, get::GetUserCommand, list::ListUserCommand},
-    UserType,
+    GetGlobalStateCommand, UserType,
 };
 
 #[allow(clippy::upper_case_acronyms)]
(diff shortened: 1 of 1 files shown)


client/doublezero/src/command/disconnect.rs
```diff
@@ -79,44 +116,33 @@ impl DecommissioningCliCommand {
                 None => {}
             }
 
+            // Skip oracle-owned users — only the oracle can delete them.
+            // The oracle's orphan cleanup will remove the DZ Ledger user
+            // automatically once the onchain seat becomes inactive.
+            if user.owner != client.get_payer() {
+                spinner.println(format!(
+                    "⚠️  User {pubkey} is managed by an external service (owner: {}). \
```
snormore: If the user is owned by the shred oracle can we tell them in the message here that they should use `doublezero-solana` to `withdraw` if they want to disconnect. Otherwise if owned by somebody user show this message

PR #3470, opened 2026-04-06
geoprobe: track minimum measurements with best/second-best cache

### LINKED ISSUE #3438: Geolocation: Target and Target-Sender should use minimum measurement
Both `geoprobe-target-sender` and `geoprobe-target` programs should behave more like how we expect clients to use the service. The minimum measurement over an allowed lookback is the relevant measurement for location, since jitter can cause some measurements to read high, but the measurement won't be artificially low.

These should now have a CLI specified TTL (default 1h if not specified) for measurements. Drawing inspiration from how the `geoprobe-agent` keeps track of its second-best parent offset, let's have the targets do the same. They always have a "best" and "second best". If the "best" TTL expires, it's replaced by the second best.

If the `--verbose` flag is present, the targets log every measurement received like it currently does, but also notes whether the best/second best have been updated. 
If the `--verbose` flag is not present, then the target only outputs when the best changes, even if it is still receiving new measurements.

ben-dz:
Resolves #3438

## Summary of Changes
- Add a generic `MinCache` that tracks the lowest-RTT measurement over a rolling TTL window using a best/backup promotion pattern
- Integrate `MinCache` into both `geoprobe-target` (inbound offset tracking) and `geoprobe-target-sender` (outbound probe tracking) to suppress redundant output and surface only new-best or promotion events
- Add `--max-measurement-age` / `--max-offset-age` CLI flags to control the TTL window

## Diff Breakdown
| Core logic   |     3 | +255 / -29  | +226 |
| Tests        |     1 | +161 / -0   | +161 |

~60% core logic, ~40% tests

<details>

- `controlplane/telemetry/internal/geoprobe/mincache.go` — new generic MinCache with best/backup expiry, promotion, and thread-safe accessors
- `controlplane/telemetry/internal/geoprobe/mincache_test.go` — unit tests covering all cache transitions (best, backup, promotion, expiry, staleness)
- `controlplane/telemetry/cmd/geoprobe-target-sender/main.go` — integrate MinCache into probe pair loop; only print on new-best/promoted; add `--max-measurement-age` flag and JSON/text output for cache state
- `controlplane/telemetry/cmd/geoprobe-target/main.go` — integrate MinCache into offset handler; suppress output unless best changes; add `--max-offset-age` flag

</details>

## Testing Verification
- Unit tests for all `MinCache` state transitions: first measurement, lower-RTT replacement, backup insertion, expiry-based promotion, both-expired, stale backup replacement, equal RTT, no-change case
- Tests use short TTLs (20–200ms) with sleeps to exercise real expiry behavior

### DIFF
--- a/controlplane/telemetry/cmd/geoprobe-target-sender/main.go
+++ b/controlplane/telemetry/cmd/geoprobe-target-sender/main.go
@@ -203,17 +203,16 @@ func probePair(ctx context.Context, log *slog.Logger, sender signed.Sender, seq 
 		reply1SigValid:      reply1SigValid,
 	}
 
-	prevBestRtt, hadPrevBest := cache.BestRttNs()
-	cacheResult := cache.Update(m)
+	info := cache.Update(m)
 
-	if *verbose || cacheResult.Changed() {
+	if *verbose || info.Changed() {
 		logPairedResult(log, seq, probeMeasuredRttNs, targetMeasuredRtt,
 			reply0ProbeSigValid, reply0SigValid, reply1ProbeSigValid, reply1SigValid,
-			result.Reply1, cacheResult, prevBestRtt, hadPrevBest)
+			result.Reply1, info)
 	}
 }
 
-func logPairedResult(log *slog.Logger, seq uint32, probeMeasuredRttNs uint64, targetMeasuredRtt time.Duration, reply0ProbeSigValid, reply0SigValid, reply1ProbeSigValid, reply1SigValid bool, reply *signed.ReplyPacket, cacheResult geoprobe.UpdateResult, prevBestRtt uint64, hadPrevBest bool) {
+func logPairedResult(log *slog.Logger, seq uint32, probeMeasuredRttNs uint64, targetMeasuredRtt time.Duration, reply0ProbeSigValid, reply0SigValid, reply1ProbeSigValid, reply1SigValid bool, reply *signed.ReplyPacket, info geoprobe.UpdateInfo) {
 	authorityPK := solana.PublicKeyFromBytes(reply.AuthorityPubkey[:])
 	geoprobePK := solana.PublicKeyFromBytes(reply.GeoprobePubkey[:])
 	offsets := parseOffsets(reply.Offsets)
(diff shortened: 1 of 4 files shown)


nikw9944:
### Code review

Found 3 issues:

1. **`UpdatePromoted` semantic is misleading when new measurement becomes backup after promotion.** When `best` expires and `backup` is promoted, if the incoming measurement has higher RTT than the promoted backup, it gets stored as the new backup but `Update()` returns `UpdatePromoted`. The callers then print "Backup promoted to best (previous best: X)" — while the promotion did happen, the `UpdatePromoted` result conflates "a promotion occurred" with "the incoming value became backup", making it unclear what actually changed. Consider separating the promotion event from the backup-update event, or documenting that `UpdatePromoted` means "promotion happened during this call" regardless of where the new value landed.

https://github.com/malbeclabs/doublezero/blob/7e0dc2f9ea98ca5f477144aa88323543e70f3faa/controlplane/telemetry/internal/geoprobe/mincache.go#L104-L123

2. **`prevBestRtt` snapshot is stale during promotion, showing misleading "previous best" values.** Both callers capture `prevBestRtt` via `BestRttNs()` before calling `Update()`. When `best` is expired but `backup` is alive, `BestRttNs()` returns the backup's RTT via read-through. `Update()` then promotes that backup to best. The result: "previous best" displays the same value as the new best, which is semantically wrong — it should show the expired best's RTT, but that information is already lost by the time `BestRttNs()` is called.

https://github.com/malbeclabs/doublezero/blob/7e0dc2f9ea98ca5f477144aa88323543e70f3faa/controlplane/telemetry/cmd/geoprobe-target-sender/main.go#L205-L212
https://github.com/malbeclabs/doublezero/blob/7e0dc2f9ea98ca5f477144aa88323543e70f3faa/controlplane/telemetry/cmd/geoprobe-target/main.go#L352-L355

3. **`MinCacheMap` has no eviction of dead per-key caches.** `MinCacheMap.Get()` creates a `MinCache` per key and never removes it. In `geoprobe-target`, keys are geoprobe pubkeys. If probes deregister over time, the map grows without bound. The same file already implements TTL-based cleanup in `rateLimiter` (lines 209-227) — a similar pattern could be applied here.

https://github.com/malbeclabs/doublezero/blob/7e0dc2f9ea98ca5f477144aa88323543e70f3faa/controlplane/telemetry/internal/geoprobe/mincache.go#L162-L194

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #3503, opened 2026-04-09
device-health-oracle: add controller_success device activation criterion

nikw9944:
Resolves: #3493

## Summary of Changes

- Add `controller_success` device activation criterion that queries ClickHouse to verify devices have called the controller at least once per minute over the burn-in period before advancing health
- Introduce criteria-based evaluation pattern with stage-aware health progression (`Pending → ReadyForLinks → ReadyForUsers`) to support the many criteria planned in [RFC-12](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc12-network-provisioning.md)
- Skip onchain health writes when the value is already at the desired state
- ClickHouse connection is optional via `CLICKHOUSE_ADDR` env var — backward compatible when not set

## Diff Breakdown
| Core logic   |     4 | +332 / -4   | +328 |
| Scaffolding  |     3 | +72 / -2    |  +70 |
| Tests        |     3 | +437 / -0   | +437 |
| Docs         |     2 | +4 / -2     |   +2 |

~330 lines of core logic implementing the criteria pattern and ClickHouse integration, well-covered by ~440 lines of tests.

<details>

- [`criteria.go`](https://github.com/malbeclabs/doublezero/pull/3503/files#diff-4fd40f530c225ddc7f12656d12ff3edacdf21e045027f1c1444c6e3a3bfca64c) — DeviceCriterion/LinkCriterion interfaces, BurnInTimes context helpers, and stage-aware evaluators
- [`clickhouse.go`](https://github.com/malbeclabs/doublezero/pull/3503/files#diff-4e66562452cebb316aec46805cd7345d34a82fbc3d17cec0b2509bc7895f5ec6) — ClickHouse client with ControllerCallCoverage query and database name validation
- [`controller_success.go`](https://github.com/malbeclabs/doublezero/pull/3503/files#diff-79edd4570aac4cdcdfb6d26e555134aeca31259c4984f3cfd04b6bf445bcd525) — ControllerSuccessCriterion querying ClickHouse for per-minute call coverage
- [`worker.go`](https://github.com/malbeclabs/doublezero/pull/3503/files#diff-cb4d126aa894fbc1f5b6dec64090e91f14bee1fa90b19535ada50dc4afd9ca02) — Evaluator integration, GetBlockTime burn-in resolution, skip-update optimization

</details>

## Testing Verification

- Evaluator enforces stage progression — devices at `Pending` advance to `ReadyForLinks` (not directly to `ReadyForUsers`), and failing criteria block advancement
- ClickHouse client tested via mocked `driver.Conn`, verifying query construction, database name quoting, and error propagation
- Burn-in slot count correctly selects provisioning (200K) vs drained (5K) based on `status.IsDrained()`
- Zero-length burn-in window (new environment) short-circuits to pass without querying ClickHouse

### DIFF
--- a/controlplane/device-health-oracle/cmd/device-health-oracle/main.go
+++ b/controlplane/device-health-oracle/cmd/device-health-oracle/main.go
@@ -129,6 +129,43 @@ func main() {
 	serviceabilityExecutor := serviceability.NewExecutor(log, rpcClient, &signer, networkConfig.ServiceabilityProgramID)
 	telemetryClient := telemetry.New(log, rpcClient, nil, networkConfig.TelemetryProgramID)
 
+	// Initialize ClickHouse-dependent criteria.
+	var deviceCriteria []worker.DeviceCriterion
+	if chAddr := os.Getenv("CLICKHOUSE_ADDR"); chAddr != "" {
+		chDB := os.Getenv("CLICKHOUSE_DB")
+		if chDB == "" {
+			chDB = *env
+		}
+		chUser := os.Getenv("CLICKHOUSE_USER")
+		if chUser == "" {
+			chUser = "default"
+		}
+		chPass := os.Getenv("CLICKHOUSE_PASS")
+		chTLSDisabled := os.Getenv("CLICKHOUSE_TLS_DISABLED") == "true"
+
+		chClient, err := worker.NewClickHouseClient(chAddr, chDB, chUser, chPass, chTLSDisabled)
+		if err != nil {
+			log.Warn("ClickHouse connection failed, continuing without controller_success criterion", "addr", chAddr, "error", err)
+		} else {
+			defer chClient.Close()
+			log.Info("ClickHouse enabled", "addr", chAddr, "db", chDB, "user", chUser, "tls", !chTLSDisabled)
+			controllerSuccess := worker.NewControllerSuccessCriterion(chClient, log)
+			deviceCriteria = append(deviceCriteria, controllerSuccess)
+		}
+	} else {
+		log.Error("ClickHouse disabled (CLICKHOUSE_ADDR not set), no controller_success criterion")
+	}
+
+	deviceEvaluator := &worker.DeviceHealthEvaluator{
+		ReadyForLinksCriteria: deviceCriteria,
+		ReadyForUsersCriteria: nil,
+		Log:                   log,
+	}
+	linkEvaluator := &worker.LinkHealthEvaluator{
+		ReadyForServiceCriteria: nil,
+		Log:                     log,
+	}
+
 	worker.MetricBuildInfo.WithLabelValues(version, commit, date).Set(1)
 	go func() {
 		listener, err := net.Listen("tcp", *metricsAddr)
@@ -155,6 +192,8 @@ func main() {
 		Env:                     *env,
 		ProvisioningSlotCount:   *provisioningSlotCount,
 		DrainedSlotCount:        *drainedSlotCount,
+		DeviceEvaluator:         deviceEvaluator,
+		LinkEvaluator:           linkEvaluator,
 	})
 	if err != nil {
 		log.Error("Failed to create worker", "error", err)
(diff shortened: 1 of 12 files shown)


controlplane/device-health-oracle/cmd/device-health-oracle/main.go
```diff
@@ -129,6 +130,48 @@ func main() {
...
+				chClient,
+				*provisioningSlotCount,
+				*drainedSlotCount,
+				*slotDurationMs,
+				log,
+			)
+			deviceCriteria = append(deviceCriteria, controllerSuccess)
+		}
+	} else {
+		log.Info("ClickHouse disabled (CLICKHOUSE_ADDR not set), no controller_success criterion")
```
nikw9944: This should log an error, since device-health-oracle will basically be broken when clickhouse is not available.

controlplane/device-health-oracle/internal/worker/worker.go
```diff
@@ -74,6 +74,26 @@ func (w *Worker) tick(ctx context.Context) {
 		"drainedSlotCount", w.cfg.DrainedSlotCount,
 		"drainedSlot", drainedSlot)
 
+	// Resolve burn-in boundary slots to wall-clock times for criteria evaluation.
+	burnIn := BurnInTimes{Now: time.Now()}
+	if provisioningSlot > 0 {
```
martinsander00, line 85: this would never happen in mainnet or tn but should we have a branch with:
When `provisioningSlot == 0`, set `burnIn.ProvisioningStart = burnIn.Now`?

Same for `DrainedStart` when `drainedSlot == 0`.  

You test this in the `TestControllerSuccessCriterion_ZeroBurnIn_Passes` test but the prod code doesn't have that same behavior

controlplane/device-health-oracle/cmd/device-health-oracle/main.go
```diff
@@ -129,6 +129,43 @@ func main() {
...
+		if err != nil {
+			log.Warn("ClickHouse connection failed, continuing without controller_success criterion", "addr", chAddr, "error", err)
+		} else {
+			defer chClient.Close()
+			log.Info("ClickHouse enabled", "addr", chAddr, "db", chDB, "user", chUser, "tls", !chTLSDisabled)
+			controllerSuccess := worker.NewControllerSuccessCriterion(chClient, log)
+			deviceCriteria = append(deviceCriteria, controllerSuccess)
+		}
+	} else {
+		log.Error("ClickHouse disabled (CLICKHOUSE_ADDR not set), no controller_success criterion")
```
martinsander00, line 156: controller uses `Info` when CH envs are not set, geoprobe just returns nil so maybe this should either be Info or not log?

controlplane/device-health-oracle/cmd/device-health-oracle/main.go
```diff
@@ -129,6 +129,43 @@ func main() {
 	serviceabilityExecutor := serviceability.NewExecutor(log, rpcClient, &signer, networkConfig.ServiceabilityProgramID)
 	telemetryClient := telemetry.New(log, rpcClient, nil, networkConfig.TelemetryProgramID)
 
+	// Initialize ClickHouse-dependent criteria.
+	var deviceCriteria []worker.DeviceCriterion
+	if chAddr := os.Getenv("CLICKHOUSE_ADDR"); chAddr != "" {
+		chDB := os.Getenv("CLICKHOUSE_DB")
+		if chDB == "" {
+			chDB = *env
```
martinsander00: other components default to `*default`, not `*env`

controlplane/device-health-oracle/internal/worker/criteria.go
```diff
@@ -0,0 +1,142 @@
...
+	return serviceability.DeviceHealthReadyForUsers
+}
+
+func (e *DeviceHealthEvaluator) checkAll(ctx context.Context, device serviceability.Device, criteria []DeviceCriterion) bool {
+	devicePubkey := device.PubKey[:]
+	for _, c := range criteria {
+		passed, reason := c.Check(ctx, device)
+		if !passed {
+			e.Log.Info("Device criterion not met",
+				"device", devicePubkey,
```
martinsander00, line 102: `"device", devicePubkey.String(),`

right?

controlplane/device-health-oracle/cmd/device-health-oracle/main.go
nikw9944: My thought is that it should be an actionable error because device-health-oracle requires telemetry and clickhouse data to do its job, so if it's not configured for clickhouse it's broken. And I think we should change the controller to log an error too. We should probably have it log the "ClickHouse disabled" error on every tick too, not just on startup. What do you think?

martinsander00 (approved):

PR #3507, opened 2026-04-09
sentinel: improve multicast publisher find/create commands

snormore:
## Summary of Changes
- Fix owner bug: multicast publisher users were being created with the sentinel payer as owner instead of the validator's IBRL user owner, causing incorrect PDA derivation for access passes and user accounts
- Add multi-value \`--client\` filter to both \`find\` and \`create\` commands (OR semantics, repeatable flag)
- Add \`--ip\` filter to \`create\` command for targeting specific validator IPs; when no candidates match, show per-IP diagnosis explaining the exact skip reason (no IBRL user, already a publisher, not in validator metadata, client/stake filter mismatch)
- Add "NEAREST DEVICE" column to both \`find\` and \`create\` plan output showing the closest DZ device with available multicast publisher capacity (geographic Haversine distance from the validator's current device)
- Re-evaluate target device per-candidate during \`create\` execution so slots filled earlier in the run are reflected in subsequent device selection; warn-and-skip if no device has capacity
- Consistent table formatting: \`create\` plan now uses \`print_table\` (same markdown-style table as \`find\`)
- Strengthen e2e test by using a dedicated sentinel keypair (distinct from manager) and asserting the created multicast publisher owner matches the IBRL user's owner

## Diff Breakdown
| Core logic   |     5 | +364 / -28   | +336 |
| Scaffolding  |     3 | +106 / -43   |  +63 |
| Tests        |     1 | +37  / -2    |  +35 |

Primarily new logic: nearest-device selection, device info fetching, and the owner bug fix account for the bulk of additions.

<details>

- `crates/sentinel/src/nearest_device.rs` — new module: Haversine distance, device capacity check, `find_nearest_device_for_multicast`
- `crates/sentinel/src/dz_ledger_reader.rs` — `DzDeviceInfo` struct and `fetch_device_infos` (batch fetches Device + Location accounts)
- `controlplane/doublezero-admin/src/cli/sentinel.rs` — CLI wiring for all new flags, NEAREST DEVICE column, dynamic device re-evaluation loop, consistent table formatting
- `crates/sentinel/src/multicast_create.rs` — `CandidateFilters` gets `clients: Vec<String>` and `ips: Vec<Ipv4Addr>`; `find_candidates` accepts `device_infos`
- `crates/sentinel/src/dz_ledger_writer.rs` — owner bug fix: explicit `owner` param used for PDA derivation, `user_payer`, and `owner` field (was using `payer`)
- `crates/sentinel/src/multicast_find.rs` — `FindFilters.clients: Vec<String>` replaces `client: Option<String>`
- `e2e/sentinel_multicast_publisher_test.go` — dedicated sentinel keypair setup; asserts `u.Owner == dn.Manager.Pubkey`

</details>

## Testing Verification
- Unit tests in `multicast_create.rs`, `multicast_find.rs`, `nearest_device.rs`, and `dz_ledger_writer.rs` cover filter logic, capacity checks, Haversine distance, and instruction building with correct owner
- E2E test `TestE2E_SentinelMulticastPublisherCreatesPublishers` updated to use a separate sentinel keypair so the owner assertion would have caught the bug before this fix


martinsander00:
Will approve. Just curious on why we use haversine distance instead of lowest latency device?

martinsander00 (approved):

PR #3508, opened 2026-04-09
activator: register legacy tunnel endpoints on reload to prevent duplicate underlay pairs

ben-dz:
## Summary of Changes
- Fix duplicate tunnel underlay pairs caused by activator restarts not tracking legacy users' effective tunnel endpoints
- Users with `tunnel_endpoint=0.0.0.0` (pre-UTE feature) use `device.public_ip` in the controller, but the activator's reload path skipped registering that endpoint as in-use — allowing a second tunnel for the same client to be assigned the same IP

## Diff Breakdown
| Core logic   |     1 | +6 / -1     |  +5  |

Minimal targeted fix in the startup reservation path.

<details>

- `activator/src/processor.rs` — register `device.public_ip` as in-use for legacy users during `reserve_user_allocations`

</details>

## Testing Verification
- All 76 existing activator unit tests pass
- The `DeviceState` endpoint tracking tests cover the underlying allocation/dedup logic; this fix ensures the reload path feeds it correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,29 +8,21 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
-- Telemetry
-  - Device telemetry agent now posts `agent_version` and `agent_commit` in the `DeviceLatencySamplesHeader` when initializing new sample accounts, enabling version attribution of onchain telemetry data
-- Smartcontract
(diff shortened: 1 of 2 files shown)


juan-malbeclabs (approved):
LGTM

PR #3511, opened 2026-04-09
bugfix(activator): prevent duplicate IP allocation (#3510)

bgm-malbeclabs:
## Summary of Changes

We had an [issue](https://malbeclabs.slack.com/archives/C0804H5R7JP/p1775760432506799) where two users had the same device `public_ip` which was causing the controller to repeatedly error. 

 1. Activator starts, loads existing users
 2. User 509 (5AGV...) has tunnel_endpoint = 0.0.0.0 → has_tunnel_endpoint() is  false → not registered in tunnel_endpoints_in_use
 3. User 573 (3QNM...) arrives as Pending with client_ip = 64.130.32.201
 4. get_available_tunnel_endpoint(64.130.32.201) checks the map → empty for this client_ip
 5. Device has no UserTunnelEndpoint interfaces → falls back to device.public_ip
 6. Returns public_ip (same endpoint user 509 is already using) → duplicate tunnel pair

Fix is to register the effective tunnel endpoint rather than race. 

## Testing Verification
* Added regression test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
-- Telemetry
-  - Device telemetry agent now posts `agent_version` and `agent_commit` in the `DeviceLatencySamplesHeader` when initializing new sample accounts, enabling version attribution of onchain telemetry data
+- Activator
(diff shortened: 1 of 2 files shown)


bgm-malbeclabs:
duplicate

PR #3513, opened 2026-04-10
smartcontract,controlplane: add flex-algo link/tenant CLI extensions and
  migrate command

ben-malbeclabs:
RFC-18 flex-algo · PR 3 of 5 · see `rfcs/rfc-0018-flex-algo.md`
Depends on: #3512 (PR 2)
Series: #3497 · #3512 · #3513 · #3514 · #3515

## Summary of Changes
- Extends `doublezero link get/list/update` to display topology assignments (`link_topologies`) and drain status; adds `--link-topology` (comma-separated topology names, `default` to clear all) and `--unicast-drained` flags to `link update`; adds `--topology` filter to `link list`
- Extends `doublezero tenant get/list/update` to display included topologies (`include_topologies`) and adds `--include-topologies` (comma-separated topology names, `default` to clear) to `tenant update`
- Adds `doublezero-admin migrate flex-algo [--dry-run]` command that backfills link topology assignments and VPNv4 loopback flex-algo node segments across all existing devices and links

## Diff Breakdown
| Category    | Files | Lines (+/-) | Net  |
|-------------|-------|-------------|------|
| Core logic  |    11 | +471 / -57  | +414 |
| Scaffolding |     7 | +16 / -5    |  +11 |

Mostly core logic — the migrate command alone is 180 lines of backfill and dry-run orchestration.

<details>
<summary>Key files (click to expand)</summary>

- `controlplane/doublezero-admin/src/cli/migrate.rs` — new: `migrate flex-algo` command; validates UNICAST-DEFAULT PDA exists, backfills link topologies for all links, backfills flex-algo node segments for all VPNv4 loopback interfaces on all devices; supports `--dry-run`
- `smartcontract/cli/src/link/list.rs` — adds `--topology <name>` filter and displays `link_topologies`/`unicast_drained` columns, resolving topology pubkeys to human-readable names
- `smartcontract/cli/src/link/get.rs` — displays topology assignments and drain status; fetches topology map to resolve pubkeys to names
- `smartcontract/cli/src/link/update.rs` — adds `--link-topology` (accepts comma-separated list of topology names; use `default` to clear) and `--unicast-drained` flags
- `smartcontract/cli/src/tenant/list.rs` — displays `include_topologies` column
- `smartcontract/cli/src/tenant/get.rs` — displays included topology names
- `smartcontract/cli/src/tenant/update.rs` — adds `--include-topologies` (accepts comma-separated list; use `default` to clear)

</details>

## Testing Verification
[...]

PR #3497, opened 2026-04-08
smartcontract: add TopologyInfo account, flex-algo state, and topology processors

ben-malbeclabs:
RFC-18 flex-algo · PR 1 of 5 · see `rfcs/rfc-0018-flex-algo.md`
Series: #3497 · #3512 · #3513 · #3514 · #3515

## Summary of Changes
- Adds `TopologyInfo` onchain account (RFC-18 flex-algo) with `admin_group_bit`, `flex_algo_number`, and `TopologyConstraint`; creates `AdminGroupBits` resource extension singleton for bit allocation
- Adds four new foundation-only instructions (107–110): `CreateTopology`, `DeleteTopology`, `ClearTopology`, `BackfillTopology`
- Extends `Link` with `link_topologies` (Vec<Pubkey>) and `link_flags` (LINK_FLAG_UNICAST_DRAINED); extends `Tenant` with `include_topologies`; adds `flex_algo_node_segments` to Interface V2
- `ActivateLink` now requires the unicast-default topology account and auto-tags the link into it
- `UpdateLink` gains foundation-gated `link_topologies` update and contributor-gated `unicast_drained` flag

## Diff Breakdown
| Category    | Files | Lines (+/-)   | Net   |
|-------------|-------|---------------|-------|
| Core logic  |    22 | +1,306 / -38  | +1,268|
| Scaffolding |    49 | +133 / -3     |  +130 |
| Tests       |    27 | +4,592 / -74  | +4,518|

The scaffolding is mechanical struct field additions (`link_topologies: vec![]`, `include_topologies: vec![]`, `flex_algo_node_segments: vec![]`) across CLI, SDK, activator, and client to keep the workspace compiling. Core logic is concentrated in the new topology processors, state types, and link/activate updates.

<details>
<summary>Key files (click to expand)</summary>

- `smartcontract/programs/doublezero-serviceability/src/state/interface.rs` — updates Interface V2 (discriminant 1) to include `flex_algo_node_segments: Vec<FlexAlgoNodeSegment>`; V1 accounts (pre-CYOA format) are left as-is; pre-RFC-18 V2 accounts on mainnet are upgraded in-place by `MigrateDeviceInterfaces` before this deserialization path is used
- `smartcontract/programs/doublezero-serviceability/src/state/topology.rs` — new `TopologyInfo` account: `admin_group_bit` (u8), `flex_algo_number` (128+bit), `TopologyConstraint` (IncludeAny/Exclude)
[...]

Issue #3355, opened 2026-03-23
cEOS full acceptance criteria for flex-algo


ben-malbeclabs:
Full end-to-end acceptance test suite using cEOS testcontainers. Scaffolding can be written during the Controller — Topology Config milestone with assertions filled in as pieces land.

**Milestone:** RFC-18: cEOS E2E Tests
**RFC:** https://github.com/malbeclabs/doublezero/pull/3288

---

**Files affected:**
- `e2e/` (new flex-algo test file)

**Size:** Large

**Test scenarios:**
- Topology creation pushes correct device config including `UNICAST-DRAINED group 1` alias and `exclude 1` in all `include-any` definitions
- Admin-group applied to tagged interface; removed on clear
- Link tagging exclude list overrides onchain assignment
- Flex-algo topology: `show isis flex-algo` shows only tagged links in algo 128
- Colored tunnel RIB: `show tunnel rib system-colored-tunnel-rib brief` shows (endpoint, color 1)
- VPN unicast path selection via color 1 tunnel
- Per-tenant topology — single topology stamped correctly
- Per-tenant topology — multi-topology stamped; lower-metric color selected
- Per-tenant topology — fallback on node-segment withdrawal
- Community stamping per-device and exclude respected
- Multicast RPF unaffected (algo 0, all links)
- Unicast drain: link excluded from all constrained topologies, remains in algo-0, multicast unaffected
- New topology creation → node-segment lines appear on all devices (backfill path)
- Full revert on `enabled: false`

**Blockers:** All previous milestones complete

ben-malbeclabs:
**Status update (2026-04-10):** Branch `bc/rfc18-flex-algo-e2e` exists with 4 test functions (`TestE2E_FlexAlgo_TopologyCreate`, `TestE2E_FlexAlgo_ControllerConfig`, `TestE2E_FlexAlgo_LinkTopologyAssignment`, `TestE2E_FlexAlgo_InterfaceNodeSegment`), all currently `t.Skip`ped pending two blockers:
1. RFC-18 CLI commands (`doublezero topology create` etc.) from PRs #3512/#3513 not yet merged
2. cEOS 4.33.1F does not populate `system-colored-tunnel-rib` from flex-algo paths — full traffic-path assertions are not possible on cEOS; test scope will be limited to onchain state and controller config verification.

ben-malbeclabs:
**Status update (2026-05-12):**

[...]

PR #3515, opened 2026-04-10
sdk/go,controller: add TopologyInfo deserialization and flex-algo controller support

ben-malbeclabs:
RFC-18 flex-algo · PR 5 of 5 · see `rfcs/rfc-0018-flex-algo.md`
Depends on: #3497 (PR 1) — does not require PRs 2–4; can merge after PR 1
Series: #3497 · #3512 · #3513 · #3514 · #3515

## Summary of Changes
- Adds `TopologyInfo` account type to the Go SDK (`state.go`, `deserialize.go`, `client.go`), including `TopologyType`, `TopologyConstraint`, `LinkFlagUnicastDrained`, `LinkTopologies`/`LinkFlags` on `Link`, and `IncludeTopologies` on `Tenant`
- Auto-loads `/etc/doublezero-controller/features.yaml` at startup if present (silently skips if absent); when `flex_algo.enabled: true`, populates topology data into the state cache, resolves tenant color communities via `resolveTenantColors`, and emits IS-IS flex-algo node segment and link configuration into the Arista EOS template
- Template emits correct flex-algo definitions for both `include-any` constraint (`administrative-group include any N exclude 0`) and `exclude` constraint (`administrative-group exclude N,0`) topologies; gated per-link via `link_tagging.exclude` and per-tenant via `community_stamping`
- Updates `e2e/compatibility_test.go` to set the mandatory upgrade boundary for RFC-18 interface/link operations to `before: "0.18.0"`

## Diff Breakdown
| Category    | Files | Lines (+/-)  | Net   |
|-------------|-------|--------------|-------|
| Core logic  |     6 | +368 / -6    | +362  |
| Scaffolding |     2 | +22 / -0     |  +22  |
| Tests       |     4 | +482 / -2    | +480  |
| Fixtures    |     2 | +415 / -0    | +415  |
| Docs        |     1 | +4 / -0      |   +4  |

Most of the weight is tests and fixtures; core logic is the controller cache/template and Go SDK deserialization.

> **Note on diff size:** The full diff vs `main` is large because this branch is stacked on PR 1 (#3497), which has not yet merged. The table above reflects only PR 5's contribution. After PR 1 merges and this branch is rebased, the diff shrinks to the ~15 files shown here.

<details>

- [`controlplane/controller/internal/controller/server.go`](https://github.com/malbeclabs/doublezero/pull/3515/files#diff-f5e715ed1377ac408a4ecde2c60cbcb98327512c7d1496114ef143cb27e7afaf) — populates `Topologies` map in state cache; resolves `LinkTopologies` pubkeys to topology names and `UnicastDrained` from `LinkFlags`; computes `TenantTopologyColors`; `resolveTenantColors` falls back to `unicast-default` when tenant has no explicit topology assignments
- [`controlplane/controller/internal/controller/templates/tunnel.tmpl`](https://github.com/malbeclabs/doublezero/pull/3515/files#diff-d67a980326aca35fdabdd5445fbbab4716f1ae412af4540770d5a9ee116b04d9) — IS-IS flex-algo node segment config, link admin-group tagging, BGP `next-hop resolution ribs` and `set extcommunity` stamping, and full rollback (`no` commands) when disabled; handles both `include-any` and `exclude` constraint topologies
- [`controlplane/controller/internal/controller/features_config.go`](https://github.com/malbeclabs/doublezero/pull/3515/files#diff-24f09aab331b4d309662e5ee6550fd38fa173676ae868db381d9af23cf645f1d) — `FeaturesConfig` struct with `flex_algo.enabled`, `link_tagging.exclude`, and `community_stamping` controls; auto-loaded from `/etc/doublezero-controller/features.yaml`
- [`controlplane/controller/internal/controller/models.go`](https://github.com/malbeclabs/doublezero/pull/3515/files#diff-bdc4e34592a7435dcde7dab78ffd735dcb9847447f0113bd9f3afc8d3d01f864) — adds `LinkTopologies`, `UnicastDrained`, `FlexAlgoNodeSegments`, `TenantTopologyColors` to controller model types; `TopologyModel` and `FlexAlgoEnabled()` for template rendering
- [`smartcontract/sdk/go/serviceability/state.go`](https://github.com/malbeclabs/doublezero/pull/3515/files#diff-057f627e5f2629b35f1976fba6eaa1b3c507c511b0f47d2f80a3a1d0c684c4a1) — adds `TopologyType`, `TopologyConstraint`, `TopologyInfo`, `LinkFlagUnicastDrained`; extends `Link` with `LinkTopologies`/`LinkFlags` and `Tenant` with `IncludeTopologies`
- [`smartcontract/sdk/go/serviceability/deserialize.go`](https://github.com/malbeclabs/doublezero/pull/3515/files#diff-b71f90b1b3e00de9bd820920e63f5615edb9b7d110c2b39f38895847509a7844) — adds `DeserializeTopologyInfo`; extends `DeserializeLink` with `LinkTopologies`/`LinkFlags` and `DeserializeTenant` with `IncludeTopologies`

</details>

## Testing Verification
- `TestGetConfig_FlexAlgo` (new): render tests for flex-algo disabled, enabled with both constraint types (`include-any` and `exclude`), and link excluded from tagging — all pass
- `Test_resolveTenantColors` (new): 5 table-driven cases covering fallback to `unicast-default`, missing topology, single/multiple known pubkeys, and unknown pubkey — all pass
- `TestE2E_IBRL`, `TestE2E_IBRL_WithAllocatedAddr`, `TestE2E_Multicast` — all pass; flex-algo config is suppressed in e2e (no `features.yaml` present)
- `exclude` constraint flex-algo syntax validated on physical lab switches (chi-dn-dzd5–dzd8)

### DIFF
--- a/controlplane/controller/internal/controller/features_config.go
+++ b/controlplane/controller/internal/controller/features_config.go
@@ -7,7 +7,8 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-// FeaturesConfig is loaded from a YAML file at controller startup.
+// FeaturesConfig is optionally loaded from /etc/doublezero-controller/features.yaml at
+// controller startup. If the file is absent the controller runs with all features disabled.
 // It gates flex-algo topology config, link tagging, and BGP color community stamping.
 type FeaturesConfig struct {
 	Features struct {
--- a/controlplane/controller/internal/controller/features_config.go
+++ b/controlplane/controller/internal/controller/features_config.go
@@ -1,0 +1,68 @@
+package controller
+
+import (
+	"io"
+	"slices"
+
+	"gopkg.in/yaml.v3"
+)
+
+// FeaturesConfig is loaded from a YAML file at controller startup.
+// It gates flex-algo topology config, link tagging, and BGP color community stamping.
+type FeaturesConfig struct {
+	Features struct {
+		FlexAlgo FlexAlgoConfig `yaml:"flex_algo"`
+	} `yaml:"features"`
+}
+
+// FlexAlgoConfig controls IS-IS Flex-Algo and BGP color extended community behaviour.
+type FlexAlgoConfig struct {
+	Enabled           bool                    `yaml:"enabled"`
+	LinkTagging       LinkTaggingConfig       `yaml:"link_tagging"`
+	CommunityStamping CommunityStampingConfig `yaml:"community_stamping"`
+}
+
+// LinkTaggingConfig controls which links receive IS-IS TE admin-group attributes.
+type LinkTaggingConfig struct {
+	Exclude struct {
+		Links []string `yaml:"links"` // link pubkeys to skip
+	} `yaml:"exclude"`
+}
+
+// IsExcluded returns true if the given link pubkey is in the exclude list.
+func (c *LinkTaggingConfig) IsExcluded(linkPubKey string) bool {
+	return slices.Contains(c.Exclude.Links, linkPubKey)
+}
+
+// CommunityStampingConfig controls BGP color extended community stamping per tenant/device.
+// A device is stamped if All is true, OR its pubkey is in Devices, OR the tenant's pubkey
+// is in Tenants — unless the device pubkey is in Exclude.Devices (overrides all).
+type CommunityStampingConfig struct {
+	All     bool     `yaml:"all"`
+	Tenants []string `yaml:"tenants"` // tenant pubkeys
+	Devices []string `yaml:"devices"` // device pubkeys
+	Exclude struct {
+		Devices []string `yaml:"devices"`
+	} `yaml:"exclude"`
+}
+
+// ShouldStamp returns true if BGP color communities should be stamped for the
+// given (tenantPubKey, devicePubKey) pair.
+func (c *CommunityStampingConfig) ShouldStamp(tenantPubKey, devicePubKey string) bool {
+	if slices.Contains(c.Exclude.Devices, devicePubKey) {
+		return false
+	}
+	if c.All {
+		return true
+	}
+	return slices.Contains(c.Tenants, tenantPubKey) || slices.Contains(c.Devices, devicePubKey)
+}
+
+// LoadFeaturesConfig parses a features YAML config from the given reader.
+func LoadFeaturesConfig(r io.Reader) (*FeaturesConfig, error) {
+	var cfg FeaturesConfig
+	if err := yaml.NewDecoder(r).Decode(&cfg); err != nil {
+		return nil, err
+	}
+	return &cfg, nil
+}
--- a/controlplane/controller/internal/controller/templates/tunnel.tmpl
+++ b/controlplane/controller/internal/controller/templates/tunnel.tmpl
@@ -107,6 +107,15 @@ interface {{ .Name }}
    {{- end }}
    {{- if and .IsVpnv4Loopback .NodeSegmentIdx}}
    node-segment ipv4 index {{ .NodeSegmentIdx }}
+   {{- if $.FlexAlgoEnabled }}
+   {{- range .FlexAlgoNodeSegments }}
+   node-segment ipv4 index {{ .NodeSegmentIdx }} flex-algo {{ $.Strings.ToUpper .TopologyName }}
+   {{- end }}
+   {{- else }}
+   {{- range .FlexAlgoNodeSegments }}
+   no node-segment ipv4 index {{ .NodeSegmentIdx }} flex-algo {{ $.Strings.ToUpper .TopologyName }}
+   {{- end }}
+   {{- end }}
    {{- end }}
    {{- if and .Ip.IsValid .IsLoopback }}
    isis enable 1
@@ -123,6 +132,21 @@ interface {{ .Name }}
    {{- end }}
    isis hello padding
    isis network point-to-point
+   {{- if and .Ip.IsValid .IsPhysical .Metric .IsLink (not .IsSubInterfaceParent) (not .IsCYOA) (not .IsDIA) }}
+   {{- if $.FlexAlgoEnabled }}
+   traffic-engineering
+   {{- if and .LinkTopologies (not ($.Config.Features.FlexAlgo.LinkTagging.IsExcluded .PubKey)) }}
+   traffic-engineering administrative-group {{ $.Strings.Join " " ($.Strings.ToUpperEach .LinkTopologies) }}{{ if .UnicastDrained }} UNICAST-DRAINED{{ end }}
+   {{- else if .UnicastDrained }}
+   traffic-engineering administrative-group UNICAST-DRAINED
+   {{- else }}
+   no traffic-engineering administrative-group
+   {{- end }}
+   {{- else if $.Config }}
+   no traffic-engineering administrative-group
+   no traffic-engineering
+   {{- end }}
+   {{- end }}
    {{- end }}
 !
 {{- end }}
@@ -247,6 +271,11 @@ router bgp 65342
       {{- range .UnknownBgpPeers }}
       no neighbor {{ . }}
       {{- end }}
+      {{- if $.FlexAlgoEnabled }}
+      next-hop resolution ribs tunnel-rib colored system-colored-tunnel-rib tunnel-rib system-tunnel-rib
+      {{- else if $.Config }}
+      no next-hop resolution ribs
+      {{- end }}
    !
 {{- range $vrfId := .UnicastVrfs }}
    vrf vrf{{ $vrfId }}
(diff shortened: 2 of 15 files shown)


controlplane/controller/internal/controller/features_config.go
nikw9944, line 1: The controller already has a config file in /etc/doublezero-controller/config.env.  It would be nice to add these config items there so we don't have a 3rd way of configuring the controller (command line flags, env file, yaml file). But yaml is nice because it lets us structure the config better. If we do use yaml, for consistency we should probably move the items from config.env into the yaml file.

ben-malbeclabs: @nikw9944 can you look at the latest commit please?  

The binary now auto-loads /etc/doublezero-controller/features.yaml at startup if present, silently skipping if absent.  The --features-config flag is gone. I think we should just have a separate PR to migrate config.env into the YAML.

nikw9944 (commented):
## Code Review — PR #3515: sdk/go,controller: add TopologyInfo deserialization and flex-algo controller support

Related: `rfcs/rfc-0018-flex-algo.md`

Tests pass, lint is clean. Overall this is solid work — the feature-gate pattern is well-designed, the cleanup/rollback path (disabled flex-algo emitting `no` commands) is thoughtful, and the test coverage on render paths is good.

---

### High

**1. CHANGELOG is stale after the latest commit**
`CHANGELOG.md:14` — Still says `Add --features-config flag accepting a YAML file` but the latest commit (`7cb9ba85`) removed the `--features-config` flag in favor of auto-loading `/etc/doublezero-controller/features.yaml`. The CHANGELOG should reflect the current behavior (auto-load, no flag).

**2. CHANGELOG references nonexistent `ListTopologies` client method**
`CHANGELOG.md:16` — Says "Go serviceability SDK adds … `ListTopologies` client method" but no `ListTopologies` function exists anywhere in the codebase. The SDK adds the `TopologyInfo` type, deserialization, and the dispatch case in `GetProgramData` — but there is no standalone `ListTopologies` method. This should be corrected to avoid confusing readers.

---

### Medium

**3. `LinkFlags` bit 0 is a magic number**
`server.go:415` — `link.LinkFlags&0x01 != 0` for `UnicastDrained`. A named constant (e.g., `const LinkFlagUnicastDrained = 0x01`) in the SDK's `state.go` would be more self-documenting and avoid magic bitmask values scattered across consumers.

**4. No direct unit test for `resolveTenantColors`**
`server.go:780-798` — The `resolveTenantColors` function has meaningful branching logic (empty vs. non-empty `includeTopologies`, fallback to `unicast-default`) but is only tested indirectly through render tests that pre-compute `TenantTopologyColors`. A focused unit test would catch regressions in the fallback behavior (e.g., what happens when the topology map has no `unicast-default` entry, or when `includeTopologies` contains a pubkey not in the map).

**5. Redundant condition in template**
`tunnel.tmpl:135` — The inner condition `{{- if and .Ip.IsValid .IsPhysical .Metric .IsLink (not .IsSubInterfaceParent) (not .IsCYOA) (not .IsDIA) }}` is identical to the outer condition at line 123. The inner block is always entered when the outer block is entered, so this is a no-op guard. Consider removing the redundant inner check for clarity, or add a comment explaining if it's intentional (e.g., defensive against future template restructuring).

---

### Low

**6. `resolveTenantColors` default fallback uses string matching**
`server.go:791` — `strings.EqualFold(topo.Name, "unicast-default")` — the name `"unicast-default"` is a well-known topology name used in multiple places. Consider extracting it as a named constant (e.g., `const DefaultTopologyName = "unicast-default"`) to avoid string literal duplication.

**7. `exclude` constraint topologies silently produce no flex-algo definition**
`tunnel.tmpl:349` — The `{{- if eq .ConstraintStr "include-any" }}` guard means topologies with `exclude` constraint skip the `flex-algo N NAME` definition block entirely. This is probably correct behavior, but a brief comment in the template explaining why `exclude` topologies don't get a flex-algo definition would help future readers.

controlplane/controller/internal/controller/templates/tunnel.tmpl
```diff
@@ -123,6 +132,21 @@ interface {{ .Name }}
    {{- end }}
    isis hello padding
    isis network point-to-point
+   {{- if and .Ip.IsValid .IsPhysical .Metric .IsLink (not .IsSubInterfaceParent) (not .IsCYOA) (not .IsDIA) }}
+   {{- if $.FlexAlgoEnabled }}
+   traffic-engineering
+   {{- if and .LinkTopologies (not ($.Config.Features.FlexAlgo.LinkTagging.IsExcluded .PubKey)) }}
```
nikw9944, line 137: Would be nice to have a test that proves this exclusion feature is working.

controlplane/controller/internal/controller/templates/tunnel.tmpl
```diff
@@ -247,6 +271,11 @@ router bgp 65342
       {{- range .UnknownBgpPeers }}
       no neighbor {{ . }}
       {{- end }}
+      {{- if $.FlexAlgoEnabled }}
+      next-hop resolution ribs tunnel-rib colored system-colored-tunnel-rib tunnel-rib system-tunnel-rib
```
nikw9944, line 273: Should this also be gated by the device exclusion config? Or is the idea that as long as the physical interfaces are not configured for flex-algo, this config here doesn't really do anything?  Same question applies to the flex-algo command below.

controlplane/controller/internal/controller/templates/tunnel.tmpl
```diff
@@ -284,12 +313,51 @@ router isis 1
...
+      {{- else }}
+      {{- range $.AllTopologies }}
+      no flex-algo {{ $.Strings.ToUpper .Name }} level-2
+      {{- end }}
+      {{- end }}
+   {{- if $.FlexAlgoEnabled }}
+   traffic-engineering
+      no shutdown
+      is-type level-2
+   {{- else if $.Config }}
```
nikw9944, line 327: Do we need to specify `if $.Config` here?

nikw9944 (changes requested):

nikw9944:
In my opinion as a followup you should add an e2e test showing that the correct config is generated in the following scenarios:
- flex algo disabled
- flex algo enabled
- flex algo enabled but device is excluded

controlplane/controller/internal/controller/templates/tunnel.tmpl
ben-malbeclabs: Addressed in the latest commit — `TestGetConfig_FlexAlgo/flex_algo_enabled_excluded_link_emits_no_administrative_group` sets the link pubkey in `LinkTagging.Exclude.Links` and asserts the interface gets `no traffic-engineering administrative-group` rather than the tagging line.

controlplane/controller/internal/controller/templates/tunnel.tmpl
ben-malbeclabs: No gating needed here. `LinkTagging.Exclude` is link-level — it controls which links advertise admin-group memberships in IS-IS. The `next-hop resolution ribs` line and the `router traffic-engineering` block below are device-level: all flex-algo-enabled devices need them regardless of whether their individual links are tagged, because they're about BGP next-hop resolution on the receiving side and understanding the topology as a participant. A device with excluded links still needs to resolve colored routes and hold the flex-algo definitions to act as a destination.

controlplane/controller/internal/controller/templates/tunnel.tmpl
ben-malbeclabs: Yes — it distinguishes two distinct states: `Config == nil` (features file not loaded, no change to TE config) vs `Config != nil && Enabled == false` (features file present but disabled — explicitly roll back TE config). Without the `$.Config` guard, a device with no features file would get `no traffic-engineering` emitted unconditionally, which could interfere with any manually-configured TE.

ben-malbeclabs (commented):
Thanks for the thorough review. Addressed in the latest commit:

**1 & 2 — CHANGELOG:** Both entries corrected — auto-load description replaces the flag reference, and `GetProgramData` dispatch case replaces the nonexistent `ListTopologies` method.

**3 — `LinkFlags` magic number:** Added `LinkFlagUnicastDrained` constant to `sdk/go/serviceability/state.go` and updated `server.go` to use it.

**4 — Unit test for `resolveTenantColors`:** Added `Test_resolveTenantColors` with 5 table-driven cases covering: empty `include_topologies` falling back to `unicast-default`, empty with no `unicast-default` (returns empty string), single known pubkey, multiple known pubkeys in slice order, and unknown pubkey silently skipped.

**5 — Redundant template condition:** Removed the duplicate inner guard.

**6 — `"unicast-default"` string literal:** Extracted as `const defaultTopologyName`.

**7 — `exclude` constraint topologies:** `exclude` is a valid user-creatable topology type — operators can create one via `doublezero topology create --constraint exclude`, at which point the activator allocates a flex-algo number and color for it. Without a corresponding flex-algo definition pushed to the device, the color community would be stamped on BGP routes but the device would have no algorithm to resolve them against — a silent misconfiguration. Added the `else if eq .ConstraintStr "exclude"` branch and validated the syntax in the physical lab. `TestGetConfig_FlexAlgo/flex_algo_enabled` now verifies both constraint types render correctly.

ben-malbeclabs:
The latest commit also adds `TestGetConfig_FlexAlgo` with 3 render test cases covering the scenarios you mentioned: flex-algo disabled (asserts `no router traffic-engineering` and `no traffic-engineering administrative-group`), flex-algo enabled (asserts the full TE block including both `include-any` and `exclude` constraint definitions), and flex-algo enabled with a link in the exclude list (asserts `no traffic-engineering administrative-group` on that interface). These are config render tests rather than full cEOS container tests — the latter is tracked in #3355.

nikw9944 (approved):

PR #3522, opened 2026-04-11
edge: add Top-of-Book parser framework

armcconnell:
## Summary
- Introduces the `edge` package with a pluggable `Parser` interface, a parser registry, and the first implementation: a decoder for the DoubleZero Top-of-Book feed v0.1.0 wire format.
- Parser is stateful: it learns instrument definitions from refdata messages and uses them to enrich marketdata (quote/trade) messages with symbol, price/qty exponents, and derived floats. Marketdata arriving before its definition is held in a bounded per-instrument buffer (most-recent-wins, capped at 1000 instruments) and flushed the moment the definition lands, so a cold-start reader still produces meaningful output at the first refdata cycle.
- Wire decoder is a hand-written positional reader over the fixed little-endian format using \`encoding/binary\` and a small sticky-error reader helper. No third-party deps, no code generation.
- Emits generic \`Record\` values; intentionally decoupled from output sinks, which land in the next PR.

## Part 1 of 4 — edge feed parser stack

1. **this PR** — edge: add Top-of-Book parser framework ← start here
2. edge: add JSON/CSV/Unix-socket output sinks
3. edge: add feed runner, manager, HTTP API, and CLI
4. edge: add integration tests, synthetic publisher, and devnet guide

Each PR compiles and tests standalone. Feature is off-by-default until PR 3 lands; early PRs are dormant code.

## Testing Verification
- Unit tests cover wire decode for each message type, bounded-buffer overwrite semantics, cap-reached drops, and refdata-triggered flush.
- Also exercised against real testnet Top-of-Book traffic on a QA node — processed 1M+ records end-to-end without buffer growth or parse errors.

## Size note
~1000 lines of production code, above doublezero's ~500-line guideline. The stack already splits the feature into 4 PRs; the parser itself doesn't subdivide further without creating non-functional intermediate states. Bulk of the diff is the wire decoder (~340 LOC, straight-line positional reads) and the parser state machine (~600 LOC), both of which benefit from being reviewed together.

### DIFF
--- a/client/doublezerod/internal/edge/parser.go
+++ b/client/doublezerod/internal/edge/parser.go
@@ -1,0 +1,63 @@
+package edge
+
+import "time"
+
+// Record represents a single decoded feed message ready for output.
+type Record struct {
+	Type           string         `json:"type"`
+	Timestamp      time.Time      `json:"ts"`
+	ChannelID      uint8          `json:"channel_id"`
+	SequenceNumber uint64         `json:"seq"`
+	InstrumentID   uint32         `json:"instrument_id,omitempty"`
+	Symbol         string         `json:"symbol,omitempty"`
+	Fields         map[string]any `json:"fields,omitempty"`
+}
+
+// Parser decodes raw UDP datagrams into structured records.
+// Implementations are stateful: they maintain instrument definitions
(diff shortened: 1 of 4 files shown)


armcconnell:
Closing — moving edge feed parser to a standalone application outside of doublezerod.

PR #3523, opened 2026-04-11
edge: add JSON/CSV/Unix-socket output sinks

armcconnell:
## Part 2 of 4 — edge feed parser stack

1. #3522 — edge: add Top-of-Book parser framework ← **start here**
2. **this PR** — edge: add JSON/CSV/Unix-socket output sinks
3. edge: add feed runner, manager, HTTP API, and CLI (coming)
4. edge: add integration tests, synthetic publisher, and devnet guide (coming)

Draft until #3522 lands. Reviewers welcome to read for context; I'll mark ready-for-review after rebasing onto main.

## Summary
- Introduces the \`OutputSink\` interface and three implementations: JSON-lines file, CSV file (with header row auto-inferred from the first record's shape), and Unix-socket broadcast with drop-on-slow-consumer semantics so a stuck reader can't backpressure the parser.
- \`NewSink\` factory keyed on format + path; paths beginning with \`unix://\` route to the socket sink, others to file sinks.
- CSV sink pivots the parser's flat \`Fields\` map onto a stable column set so heterogeneous record types (quotes, trades, instrument definitions, heartbeats) line up in one sheet.
- The parser still knows nothing about sinks — these are pure consumers of \`[]Record\` produced by the parser from #3522.

## Testing Verification
- Unit tests cover each sink: CSV header inference + column-pivot, JSON line format, socket broadcast to multiple connected clients, and slow-consumer drop behavior.

PR #3524, opened 2026-04-11
edge: add feed runner, manager, HTTP API, and CLI

armcconnell:
## Part 3 of 4 — edge feed parser stack

1. #3522 — edge: add Top-of-Book parser framework ← **start here**
2. #3523 — edge: add JSON/CSV/Unix-socket output sinks
3. **this PR** — edge: add feed runner, manager, HTTP API, and CLI
4. edge: add integration tests, synthetic publisher, and devnet guide (coming)

Draft until the prior two land. This is the PR where the edge feature **turns on** — after it merges, operators can call \`doublezero edge enable\` and a feed will actually flow.

## Summary
- \`Runner\` owns one feed: two multicast listener goroutines (marketdata + refdata on separate UDP ports), a parser (#3522), a sink (#3523), and atomic counters. A periodic goroutine logs a liveness summary every 30s so operators can see activity without polling the API.
- \`Manager\` owns many runners keyed by code. \`Enable\` / \`Disable\` / \`Close\` lifecycle is owned by the manager, not by HTTP request context. This is deliberate: the runner context is explicitly rooted in the manager's lifetime, not the caller's, so an HTTP handler returning does not tear down the feed.
- HTTP endpoints \`ServeEnable\` / \`ServeDisable\` / \`ServeStatus\` exposed on the existing doublezerod Unix socket API.
- Rust CLI adds \`doublezero edge enable/disable/status\` subcommands with \`--marketdata-port\` / \`--refdata-port\` flags. Status supports \`--json\` for machine-readable output.
- \`runtime/run.go\` wires the manager against the existing serviceability reader (to resolve group codes to multicast IPs) and the existing user manager (for onchain subscription checks), so edge feeds respect the allowlist automatically.

## Testing Verification
- Manager unit tests cover port validation (required, must-differ), subscription enforcement, and lifecycle transitions.
- Validated against real testnet Top-of-Book traffic on a QA node: deployed a patched doublezerod, ran \`edge enable\` against a real multicast group with an onchain subscription, and confirmed both sockets bound, IGMP membership established, and records flowed end-to-end at sustained throughput (>1M records over the test window) with \`buffered=0\` after the first refdata cycle.

## Size note
~625 production lines across Go + Rust, slightly above the ~500-line target. The unit is cohesive (runner lifecycle + HTTP surface + CLI binding) and splitting further would leave intermediate states where the feature compiles but can't be invoked end-to-end.

### DIFF
--- a/client/doublezero/src/cli/command.rs
+++ b/client/doublezero/src/cli/command.rs
@@ -2,10 +2,10 @@ use super::multicast::MulticastCliCommand;
 use crate::{
     cli::{
         accesspass::AccessPassCliCommand, config::ConfigCliCommand,
-        contributor::ContributorCliCommand, device::DeviceCliCommand, exchange::ExchangeCliCommand,
-        globalconfig::GlobalConfigCliCommand, link::LinkCliCommand, location::LocationCliCommand,
-        permission::PermissionCliCommand, resource::ResourceCliCommand, tenant::TenantCliCommand,
-        user::UserCliCommand,
+        contributor::ContributorCliCommand, device::DeviceCliCommand, edge::EdgeCliCommand,
+        exchange::ExchangeCliCommand, globalconfig::GlobalConfigCliCommand, link::LinkCliCommand,
+        location::LocationCliCommand, permission::PermissionCliCommand,
+        resource::ResourceCliCommand, tenant::TenantCliCommand, user::UserCliCommand,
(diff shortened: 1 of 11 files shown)


armcconnell:
Closing — moving edge feed parser to a standalone application outside of doublezerod.

PR #3526, opened 2026-04-13
telemetry: move geoprobe-agent metrics into geoprobe package

### LINKED ISSUE #3432: geoprobe: consolidate geoprobe-agent metrics into package-level metrics
## Summary

The standalone `geoprobe-agent` binary (`controlplane/telemetry/cmd/geoprobe-agent/main.go`) instruments Prometheus metrics directly in its `main.go`, using package-level vars from `controlplane/telemetry/internal/metrics/geolocation_metrics.go` with the `doublezero_device_geoprobe_agent_*` prefix.

Issue #2942 added a new `Metrics` struct at the geoprobe package level (`controlplane/telemetry/internal/geoprobe/metrics.go`) with `source` and `device_pubkey` constant labels, using the `doublezero_geoprobe_*` prefix. This struct is used by the telemetry agent's embedded geoprobe coordinator and discovery.

The standalone `geoprobe-agent` should be migrated to use the same package-level `Metrics` struct (with `source=geoprobe-agent`), so that:

1. Both agents emit metrics with a consistent naming scheme (`doublezero_geoprobe_*`)
2. Both agents include `source` and `device_pubkey` labels for disambiguation
3. The old `doublezero_device_geoprobe_agent_*` metrics in `geolocation_metrics.go` can be removed

## Tasks

- [ ] Update `geoprobe-agent` `main.go` to create a `geoprobe.NewMetrics(geoprobe.SourceGeoProbeAgent, devicePubkey)` instance
- [ ] Replace all references to `metrics.GeoProbe*` vars with the new `Metrics` struct fields
- [ ] Add any metrics from `geolocation_metrics.go` that are not yet represented in the `Metrics` struct (e.g., `offsets_received`, `offsets_rejected`, `parents_discovered`, `build_info`)
- [ ] Remove the old `geolocation_metrics.go` file once migration is complete
- [ ] Update or add tests

## Context

- Related: #2941, #2942

nikw9944:
Fixes #3432

## Summary
- Move geoprobe metrics from the shared `internal/metrics` package into a `Metrics` struct in `internal/geoprobe`, replacing package-level `promauto` globals with explicit dependency injection via `NewMetrics()`
- Add `source` and `device_pubkey` constant labels to all geoprobe metrics, enabling per-device filtering in dashboards
- Shorten metric names from `doublezero_device_geoprobe_agent_*` to `doublezero_geoprobe_*`
- Delete the now-unused `geolocation_metrics.go` from the shared metrics package
- Moving geoprobe metrics from package-level globals into a struct with source and device_pubkey constant labels lets multiple binaries emit the same metric names without collision, so dashboards can filter by source and device without needing separate metric prefixes per consumer.

## Lines of Code

| File | Lines | Status |
|------|-------|--------|
| `geoprobe/metrics.go` | 190 | New |
| `geoprobe/metrics_test.go` | 144 | New (test) |
| `geoprobe-agent/main.go` | ~30 | Modified |
| `metrics/geolocation_metrics.go` | 139 | Deleted |
| **Net new (excl. tests)** | **~51** | |

## Testing Verification
- Unit tests verify all 12 collectors are registered, constant labels are applied correctly, and duplicate registration panics as expected
- All existing `offsetCache` tests in `geoprobe-agent` continue to pass unchanged
- `go vet` clean across the telemetry module


ben-dz (changes requested):
The new constant label is named device_pubkey, but the geoprobe-agent runs on geoprobe servers, not on DZDs/devices. The value passed is geoProbePubkey.String() — that's a geoprobe pubkey, not a device pubkey. This will cause confusion in PromQL queries and dashboards. Should probably be geoprobe_pubkey or just pubkey.

NewMetrics() is called unconditionally at https://github.com/malbeclabs/doublezero/blob/9831f4c0b019c0d9d9f2d0c17d8943fbaa46201d/controlplane/telemetry/cmd/geoprobe-agent/main.go#L340, before the if *metricsEnable check. All 12 collectors get registered to prometheus.DefaultRegisterer even if no HTTP server is started to serve them. Not a regression (promauto did the same), but might want to fix it.

ben-dz (approved):

PR #3480, opened 2026-04-07
geolocation: add result destination for LocationOffsets

ben-dz:
Resolves: #3468

## Summary of Changes
- Rust SDK `SetResultDestinationCommand` that builds the onchain transaction with user PDA + probe account metas
- CLI `user set-result-destination --destination <host:port>` (or `--clear`) with client-side validation of IP routability and domain format before transaction submission
- Display `result_destination` in `user get` output (shows `"none"` when unset)
- Go SDK `ResultDestination string` field on `GeolocationUser` with backwards-compatible deserialization (old accounts default to `""`)
- Subcommand registration in the geolocation CLI binary

Stacked on #3501 which adds the onchain `SetResultDestination` instruction and `result_destination` field.

## Diff Breakdown
| Category    | Files | Lines (+/-)  | Net   |
|-------------|-------|--------------|-------|
| Core logic  |     3 | +425 / -8    | +417  |
| Scaffolding |     6 | +13 / -0     | +13   |
| Tests       |     1 | +49 / -2     | +47   |

Core logic is the CLI command (290 lines including validation + tests), Rust SDK command (118 lines), and Go SDK deserialization. Scaffolding is trait wiring and subcommand registration.

<details>
<summary>Key files (click to expand)</summary>

- `smartcontract/cli/src/geolocation/user/set_result_destination.rs` — CLI command with clap args, client-side `host:port` validation (IP routability + domain format per RFC 1035), mock tests for set/clear/invalid inputs
- `smartcontract/sdk/rs/src/geolocation/geolocation_user/set_result_destination.rs` — Rust SDK command: validates code, derives user PDA, passes probe account metas to `execute_transaction`
- `sdk/geolocation/go/state.go` — `ResultDestination string` field, serialize/deserialize with try-decode-default for backwards compat
- `sdk/geolocation/go/state_test.go` — round-trip and backwards-compat tests (truncated data defaults to empty)
- `smartcontract/cli/src/geoclicommand.rs` — `set_result_destination` trait method + impl

</details>

## Testing Verification
- CLI mock tests: set destination, clear, missing destination flag, 9 invalid destination cases (private IPs, bad ports, malformed domains)
- Go SDK: round-trip test with non-empty destination, empty-targets test, backwards-compat test (truncated data)
- All tests pass: 289 CLI, 165 SDK, 82 geolocation, Go SDK green

Issue #3468, opened 2026-04-06
geolocation: add result destination to GeolocationUser (onchain + CLI + SDK)


ben-dz:
Parent: #3430

## Summary

Add `result_destination_ip` and `result_destination_port` fields to the `GeolocationUser` onchain account, with a new instruction to set/clear them, CLI support, and Go SDK deserialization.

## Onchain changes

**`GeolocationUser` struct** (`smartcontract/programs/doublezero-geolocation/src/state/geolocation_user.rs`):
- Switch from `BorshDeserialize` to `BorshDeserializeIncremental`
- Append two fields after `targets`:
  - `result_destination_ip: Ipv4Addr` — `#[incremental(default = Ipv4Addr::UNSPECIFIED)]` (4 bytes)
  - `result_destination_port: u16` — `#[incremental(default = 0)]` (2 bytes)
- `0.0.0.0:0` means "not set" (current behavior preserved)

**New instruction: `SetResultDestination`**
- Args: `ip: Ipv4Addr`, `port: u16`
- Accounts: `user_account` (writable), `payer_account` (signer, must be owner)
- Validates IP is public (reuse `validate_public_ip`) when setting; allows `0.0.0.0:0` to clear
- Bumps `target_update_count` on all referenced probes so agents re-discover the change

**Backwards compatibility:**
- Follows the same pattern as `GeoProbe.target_update_count` (`BorshDeserializeIncremental` + `#[incremental(default)]`)
- Existing accounts without these bytes deserialize with defaults
- Add a backwards-compat test (serialize without new fields, verify deserialization defaults)

## CLI changes

**`smartcontract/cli/src/geolocation/user/`**:
- New command: `dz-cli geolocation user set-result-destination --user <code> --ip <addr> --port <port>`
- To clear: `dz-cli geolocation user set-result-destination --user <code> --clear`
- Display result destination in `dz-cli geolocation user get` output

## Go SDK changes

**`sdk/geolocation/go/state.go`**:
- Add `ResultDestinationIP [4]byte` and `ResultDestinationPort uint16` to `GeolocationUser` struct
- Deserialize with "try decode, default on error" pattern (same as `GeoProbe.TargetUpdateCount`)
- Add backwards-compat test (truncated data defaults to zero values)

## Testing

- Rust unit tests: serialization roundtrip with new fields, backwards-compat without new fields
- Rust integration tests: `SetResultDestination` success, unauthorized signer, invalid IP, clear
[...]

PR #3534, opened 2026-04-15
geoprobe: route LocationOffsets to result destination

### LINKED ISSUE #3469: geolocation: route LocationOffsets to result destination in geoprobe-agent
Parent: #3430
Depends on: #3468

## Summary

Update the geoprobe-agent to check a user's result destination when sending LocationOffsets, and route offsets there instead of the target IP when set.

## Target discovery changes

**`controlplane/telemetry/internal/geoprobe/target_discovery.go`**:
- `TargetUpdate` and `ICMPTargetUpdate` need to carry the result destination alongside each target
- When extracting targets from `GeolocationUser`, read the user's `result_destination_ip`/`result_destination_port`
- If set, override the "send-to" address for all of that user's outbound targets
- The `ProbeAddress` struct (or a wrapper) needs to distinguish the **measure address** (ICMP ping / TWAMP target) from the **delivery address** (where to send the LocationOffset)

## Offset delivery changes

**`controlplane/telemetry/cmd/geoprobe-agent/main.go`** (`sendCompositeOffsets`):
- Currently builds `targetAddr` from `addr.Host:addr.Port` (the measurement target)
- When a delivery address is present, use that instead
- ICMP targets with no result destination: log a warning and skip sending (nothing is listening)

## Testing

- Unit tests for target discovery: user with result destination set → delivery address overrides target address
- Unit tests for target discovery: user without result destination → delivery address equals target address
- Unit tests: ICMP target without result destination → warning logged, offset not sent
- E2E test (if feasible): ICMP target with result destination receives offsets at the specified endpoint

ben-dz:
## Summary

- Route outbound LocationOffsets to the user's `ResultDestination` when set on `GeolocationUser`, instead of sending to the measurement target address
- Skip ICMP targets that have no result destination (nothing is listening at the measurement IP)
- Add DNS resolution with caching (5min TTL) for domain-based result destinations
- Validate delivery addresses against public-IP scope checks to prevent DNS rebinding/SSRF attacks
- Validate `ResultDestination` format at discovery time to surface errors early

## Details

The `ResultDestination` field was added to `GeolocationUser` in #3480/#3501. This PR implements the geoprobe-agent side: reading the field during target discovery, carrying it through as `DeliveryAddrs` on `TargetUpdate`/`ICMPTargetUpdate`, and routing offsets to the specified address in `sendCompositeOffsets`.

Key design decisions:
- Delivery addresses are carried as a separate `map[ProbeAddress]string` rather than modifying `ProbeAddress` itself, to preserve Pinger map key semantics
- `DNSCache.Resolve()` validates resolved IPs against `ValidateScope()` to prevent DNS rebinding
- Change detection for delivery addresses triggers target updates even when the target list is unchanged

Depends on: #3468

## Testing Verification

- Unit tests for outbound target with result destination → delivery override populated
- Unit tests for user without result destination → no delivery overrides
- Unit tests for ICMP target with result destination → delivery override in ICMP update
- Unit tests for mixed users (with/without result destination)
- Unit tests for domain-based result destination passthrough
- Unit tests for delivery address change detection → triggers update
- Unit tests for invalid result destination format → ignored with warning
- DNS cache: IP passthrough, domain resolution and caching, TTL expiry, concurrent access
- DNS cache security: private IP rejection, DNS rebinding to private IP rejected

Fixes #3469

### DIFF
--- a/controlplane/telemetry/cmd/geoprobe-agent/main.go
+++ b/controlplane/telemetry/cmd/geoprobe-agent/main.go
@@ -624,6 +624,7 @@ func main() {
 			getCurrentSlot:     getCurrentSlot,
 			signedReflector:    signedReflector,
 			metrics:            m,
+			dnsCache:           geoprobe.NewDNSCache(5 * time.Minute),
 			targetUpdateCh:     targetUpdateCh,
 			icmpTargetUpdateCh: icmpTargetUpdateCh,
 			inboundKeyCh:       inboundKeyCh,
@@ -739,9 +740,12 @@ type measurementLoop struct {
 	getCurrentSlot  func(ctx context.Context) (uint64, error)
 	signedReflector signed.Reflector
 	metrics         *geoprobe.Metrics
+	dnsCache        *geoprobe.DNSCache
 
-	targets     []geoprobe.ProbeAddress
-	icmpTargets []geoprobe.ProbeAddress
+	targets           []geoprobe.ProbeAddress
+	icmpTargets       []geoprobe.ProbeAddress
+	deliveryAddrs     map[geoprobe.ProbeAddress]string
+	icmpDeliveryAddrs map[geoprobe.ProbeAddress]string
 
 	targetUpdateCh     <-chan geoprobe.TargetUpdate
 	icmpTargetUpdateCh <-chan geoprobe.ICMPTargetUpdate
@@ -815,10 +819,11 @@ func (ml *measurementLoop) run() error {
 				func(addr geoprobe.ProbeAddress) (uint64, bool) { return ml.pinger.MeasureOne(ml.ctx, addr) },
 			)
 			ml.targets = newTargets
+			ml.deliveryAddrs = update.DeliveryAddrs
 			ml.metrics.TargetsDiscovered.Set(float64(len(ml.targets)))
 			ml.log.Info("Updated targets from discovery", "totalTargets", len(ml.targets))
 			if len(rttData) > 0 {
-				sendCompositeOffsets(ml.ctx, ml.log, rttData, ml.cache, ml.signer, ml.senderConn, ml.getCurrentSlot, ml.metrics)
+				sendCompositeOffsets(ml.ctx, ml.log, rttData, ml.cache, ml.signer, ml.senderConn, ml.getCurrentSlot, ml.metrics, ml.deliveryAddrs, nil, ml.dnsCache)
 			}
 
 		case icmpUpdate := <-ml.icmpTargetUpdateCh:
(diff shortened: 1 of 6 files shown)


nikw9944 (commented):
## Code Review: geoprobe: route LocationOffsets to result destination

**RFC Reference:** [RFC-16: Geolocation Verification](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc16-geolocation-verification.md) — specifically the `SetResultDestination` instruction and GeolocationUser.`result_destination` field.

### Overall Assessment

This is a well-structured PR that implements the geoprobe-agent side of result destination routing. The design decisions are sound — carrying delivery addresses as a separate map rather than modifying `ProbeAddress` is the right call, and the DNS caching with SSRF protection is a good security measure. The test coverage is comprehensive. A few items worth discussing:

### Issues

**1. ICMP target skip logs at `Warn` level (main.go:963-965)**

```go
log.Warn("ICMP target has no result destination, skipping offset delivery",
    "target", addr.Host)
```

Once ICMP targets exist in production, this will fire on *every* measurement cycle for any ICMP target without a result destination. That's potentially every 5 seconds per target. This should likely be `Debug` level, since the absence of a result destination on an ICMP target is an expected (non-exceptional) configuration — it just means the user hasn't set up a listener yet. If you want visibility for the initial rollout, at minimum consider rate-limiting or logging only on first occurrence per target.

**2. `sendCompositeOffsets` signature is getting long (main.go:934)**

The function now takes 11 parameters. This is starting to strain readability — especially the three new trailing parameters (`deliveryAddrs`, `icmpTargets`, `dnsCache`). Consider grouping the new delivery-related params into a small struct (e.g., `deliveryContext`) in a follow-up. Not blocking.

**3. Merged delivery map in `runCycle()` could shadow on key collision (main.go:903-916)**

```go
mergedDelivery := make(map[geoprobe.ProbeAddress]string, len(ml.deliveryAddrs)+len(ml.icmpDeliveryAddrs))
for k, v := range ml.deliveryAddrs {
    mergedDelivery[k] = v
}
for k, v := range ml.icmpDeliveryAddrs {
    mergedDelivery[k] = v
}
```

If a `ProbeAddress` somehow appears in both `deliveryAddrs` and `icmpDeliveryAddrs`, the ICMP entry wins silently. In practice this shouldn't happen because outbound targets use `Validate()` (requiring TWAMPPort) while ICMP targets use `ValidateICMP()` (TWAMPPort=0), so the `ProbeAddress` keys should be disjoint. But the code doesn't enforce or assert this. A comment noting this invariant would help future readers.

**4. `discover()` returns a single flat `deliveryAddrs` map, then `discoverAndSend()` re-splits it (target_discovery.go:102-115)**

The `discover()` function merges all delivery addresses into one map, then `discoverAndSend()` immediately splits them back into `outboundDelivery` and `icmpDelivery` by looking up against an `outboundSet`. This round-trip is slightly wasteful. Consider returning two maps from `discover()` directly — it would align the return values with how they're consumed and remove the splitting logic. Low priority.

### Positives

- **Security**: DNS rebinding protection via `ValidateScope()` on resolved IPs is exactly right. Validating at resolution time (not just at discovery time) prevents time-of-check/time-of-use issues on the DNS→IP boundary.
- **Validation at discovery time**: Checking `ResultDestination` format (host:port) during `discover()` rather than on every send cycle is a good pattern — log once, don't spam.
- **Change detection**: The `deliveryAddrsEqual()` check triggers updates even when only the delivery address changes (not the target list). This is important because `SetResultDestination` bumps `target_update_count` on probes, and without this check the agent would see new targets but the delivery override wouldn't propagate.
- **E2E test**: The `TestE2E_GeoprobeResultDestination` test is thorough — it sets up two separate containers (measurement target vs result destination) and verifies the offset arrives at the result destination. This is the right way to test this feature.
- **ICMP fix in existing test**: The `TestE2E_GeoprobeIcmpTargets` fix adding `setGeolocationUserResultDestination` is a good catch — ICMP targets have no listener, so without a result destination the offsets have nowhere to go.
- **Test coverage**: Unit tests cover all the important permutations — outbound with/without result dest, ICMP with result dest, mixed users, domain names, change detection, and invalid format handling.

nikw9944 (approved):

PR #3543, opened 2026-04-16
qa: poll USDC balance for finalization in multicast settlement test

packethog:
## Summary of Changes
* Poll the USDC balance with `require.Eventually` in `validate_balance_after_pay` and `validate_balance_after_withdraw` instead of a single read, with a 30s timeout and 5s interval.
* Fixes a race between `FeedSeatPay`/`FeedSeatWithdraw` (which return after tx **confirmation**) and `GetUSDCBalance` at `CommitmentFinalized` — mainnet finalization lags confirmation by ~13s, so a one-shot finalized read can still see the pre-debit balance.
* PR #3542 (commitment bump alone) was insufficient: the failure reproduced immediately at [run 24534512863](https://github.com/malbeclabs/infra/actions/runs/24534512863/job/71732161416), with before/after balance identical because the finalized-slot view predated the pay tx.
* 5s poll interval chosen to avoid RPC rate limiting.

## Testing Verification
* `go build -tags=qa ./e2e/...` and `go vet -tags=qa ./e2e/...` pass.
* Run failure signature reproduced: `balanceBeforePay=1008431000`, `balanceAfterPay=1008431000`, debit=0, polled one ~115ms after `FeedSeatPay` returned. With polling, the assertion has up to 30s for the finalized slot to advance past the pay tx.

### DIFF
--- a/e2e/internal/qa/client_settlement.go
+++ b/e2e/internal/qa/client_settlement.go
@@ -197,7 +197,7 @@ func (c *Client) GetUSDCBalance(ctx context.Context) (uint64, error) {
 	}
 
 	solanaClient := rpc.New(c.SolanaRPCURL)
-	result, err := solanaClient.GetTokenAccountBalance(ctx, ata, rpc.CommitmentFinalized)
+	result, err := solanaClient.GetTokenAccountBalance(ctx, ata, rpc.CommitmentConfirmed)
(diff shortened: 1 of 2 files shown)


karl-dz (approved):
We should investigate this closer once we get QA working

PR #3554, opened 2026-04-21
cli: verify MulticastPublisherBlock in resource verify

elitegreg:
## Summary of Changes

- Extend `doublezero resource verify` to cover the `MulticastPublisherBlock` resource extension, the global IP block from which multicast publisher users allocate their `dz_ip`. Previously only 7 of 10 `ResourceType` variants were verified; publishers' `dz_ip` allocations were invisible to the tool.
- Check in both directions: every publisher user's in-range `dz_ip` must be allocated in the extension, and every allocation must have a matching user. Discrepancies flow through the existing `insert_usage` / `check_discrepancies` helpers and are auto-fixable via `--fix`.
- Ignore legacy publishers whose `dz_ip` was allocated before this extension existed and falls outside the block's range (e.g. prior to `MulticastPublisherBlock`). Without this guard, those would be reported as `UsedButNotAllocated` and `--fix` would then fail because `allocate_specific` rejects out-of-range IPs.
- Add `MulticastPublisherBlock` to the "Resources checked" summary output.

## Diff Breakdown

| Core logic   |     1 | +67 / -2    |  +65 |
| Tests        |     1 | +238 / -2   | +236 |
| **Total**    |     2 | +307 / -4   | +303 |

Mostly tests and a small, pattern-following addition to the verify CLI. Core logic is the new `verify_multicast_publisher_block` function plus imports/wiring.

<details>

- [`smartcontract/cli/src/resource/verify.rs`](https://github.com/malbeclabs/doublezero/pull/new/gm/resource-verify-mcast-pub/files#diff-cb0415493763d7495f42a3f46a2bc2479c1ba3800a3e42dfc46b7eace2db5584) — new `verify_multicast_publisher_block` function, output + counter wiring, plus 2 new tests (happy path, out-of-range ignore) and updates to 4 existing tests to register the new extension.

</details>

## Testing Verification

- New unit test `test_verify_multicast_publisher_block_happy_path` — extension has `148.51.120.5` allocated, one publisher user holds it, expect no discrepancies.
- New unit test `test_verify_multicast_publisher_ignores_out_of_range_dz_ip` — covers two edge cases: a publisher with a legacy `dz_ip` outside the block's range, and a non-publisher `UserType::Multicast` user with an in-range `dz_ip`; both must be ignored.
- All 7 `resource::verify` tests pass; full `doublezero_cli` suite (296 tests) passes.

### DIFF
--- a/smartcontract/cli/src/resource/verify.rs
+++ b/smartcontract/cli/src/resource/verify.rs
@@ -14,8 +14,8 @@ use doublezero_serviceability::{
         interface::{InterfaceType, LoopbackType},
         link::Link,
         multicastgroup::MulticastGroup,
-        resource_extension::ResourceExtensionOwned,
-        user::User,
+        resource_extension::{Allocator, ResourceExtensionOwned},
+        user::{User, UserType},
     },
 };
 use solana_sdk::pubkey::Pubkey;
@@ -63,6 +63,7 @@ pub struct VerifyResourceResult {
     pub segment_routing_ids_checked: usize,
     pub link_ids_checked: usize,
     pub multicast_group_block_checked: usize,
+    pub multicast_publisher_block_checked: usize,
 }
 
 impl VerifyResourceResult {
@@ -114,6 +115,11 @@ impl VerifyResourceCliCommand {
             out,
             "  MulticastGroupBlock: {}",
             result.multicast_group_block_checked
+        )?;
+        writeln!(
+            out,
+            "  MulticastPublisherBlock: {}",
+            result.multicast_publisher_block_checked
         )?;
         writeln!(out)?;
 
@@ -551,6 +557,9 @@ fn verify_resources<C: CliCommand>(client: &C) -> eyre::Result<VerifyResourceRes
         &mut result,
     );
 
+    // Verify MulticastPublisherBlock
+    verify_multicast_publisher_block(&program_id, &users, &resource_extensions, &mut result);
+
     Ok(result)
 }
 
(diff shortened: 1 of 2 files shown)


smartcontract/cli/src/resource/verify.rs
```diff
@@ -928,6 +937,61 @@ fn verify_multicast_group_block(
...
+            .discrepancies
+            .push(ResourceDiscrepancy::ExtensionNotFound { resource_type });
+        return;
+    };
+
+    // Pull the base network so we can ignore legacy dz_ips that pre-date this
+    // extension and fall outside the block's range.
+    let base_net = match &extension.allocator {
+        Allocator::Ip(ip_alloc) => ip_alloc.base_net,
+        Allocator::Id(_) => return,
```
juan-malbeclabs, line 960: If `MulticastPublisherBlock` ever has an Id allocator (e.g. a misconfiguration), this silently exits: the counter stays at zero, no discrepancy is recorded, and the operator has no signal that something is wrong. Since the allocator type is load-bearing for which users are in scope here, a mismatch should probably be loud, either an `ExtensionNotFound` discrepancy or a new variant. Not a blocker, but worth a follow-up.

smartcontract/cli/src/resource/verify.rs
```diff
@@ -928,6 +937,61 @@ fn verify_multicast_group_block(
...
+    };
+
+    let allocated: HashSet<IdOrIp> = extension.iter_allocated().into_iter().collect();
```
juan-malbeclabs, line 963: The `.into_iter()` is redundant  `.collect()` calls `IntoIterator` implicitly. Same pattern appears in every other verifier in this file, so this could be a follow-up cleanup, but: `extension.iter_allocated().collect()` is enough.

juan-malbeclabs (approved):
Clean implementation, follows the existing verifier pattern well. Two small comments inline.

smartcontract/cli/src/resource/verify.rs
elitegreg: This is enforced in code, there is a match statement that maps MulticastPublisherBlock to an IpAllocator

PR #3558, opened 2026-04-21
sdk/geolocation: add Go write SDK for add/remove target operations

### LINKED ISSUE #3557: Geolocation: Complete Go Write SDK
Combine the requests in issue #3296 and #3011, into a Geolocation Go Write SDK to match the Rust equivalents. Issue #3011 was worked on and has a draft branch that can be used or not based on what makes sense given the evolution of the code since then. The Read half of the SDK already exists and complete enough.

The Go Write SDK does not need to include everything. It can skip Foundation commands, such as probe creation and management, marking users at paid etc. That will continue to be CLI, or possible rust automations in the future. There's no need for Go code to do that. This should just be the SDK a user would need to automate adding/removing targets from their (already existing) user account, and possible changing their offset destination, or other behaviors that a user might want to automate.

ben-dz:
Resolves: #3557

## Summary of Changes
- New Go write SDK for the geolocation program focused on target management: instruction builders for `AddTarget` and `RemoveTarget`, plus a transaction executor.
- Instruction builders follow the telemetry write SDK pattern: `*InstructionConfig` struct + `Validate()` + `Build*Instruction()` returning a self-contained `solana.Instruction`, using `near/borsh-go` to match the Rust program's Borsh layout.
- User lifecycle (`CreateGeolocationUser`, `UpdateGeolocationUser`, `DeleteGeolocationUser`) and `SetResultDestination` are deferred to a follow-on PR (#3567) to keep this reviewable — my immediate consumer only needs add/remove targets.

## Diff Breakdown
| Category   | Files | Lines (+/-) |  Net |
| Core logic |     5 | +475 / -0   | +475 |
| Tests      |     4 | +389 / -0   | +389 |

Pure-additive new package; ~45% of the diff is tests.

## Testing Verification
- Borsh serialization tests pin discriminator bytes, field ordering, and IP/port/target-pk layout against the Rust program's expected layout for Add/Remove.
- Private-IP validation covers all reserved ranges (10/8, 172.16/12, 192.168/16, 127/8, 0.0.0.0) for outbound target types.
- Account-meta ordering verified against the Rust SDK command builders.
- Executor error paths covered: missing signer, missing program ID.

### DIFF
--- a/sdk/geolocation/go/add_target.go
+++ b/sdk/geolocation/go/add_target.go
@@ -29,7 +29,7 @@ func (c *AddTargetInstructionConfig) Validate() error {
 
 	switch c.TargetType {
 	case GeoLocationTargetTypeOutbound, GeoLocationTargetTypeOutboundIcmp:
-		if err := validateNotPrivateIP(c.IPAddress); err != nil {
+		if err := validatePublicIP(c.IPAddress); err != nil {
 			return err
 		}
 	case GeoLocationTargetTypeInbound:
@@ -40,31 +40,6 @@ func (c *AddTargetInstructionConfig) Validate() error {
 		return fmt.Errorf("unknown target type: %d", c.TargetType)
 	}
 
-	return nil
-}
-
-// validateNotPrivateIP checks that an IPv4 address is not in a private/reserved range.
-func validateNotPrivateIP(ip [4]uint8) error {
-	// 10.0.0.0/8
-	if ip[0] == 10 {
-		return fmt.Errorf("IP address %d.%d.%d.%d is in private range 10.0.0.0/8", ip[0], ip[1], ip[2], ip[3])
-	}
-	// 172.16.0.0/12
-	if ip[0] == 172 && ip[1] >= 16 && ip[1] <= 31 {
-		return fmt.Errorf("IP address %d.%d.%d.%d is in private range 172.16.0.0/12", ip[0], ip[1], ip[2], ip[3])
-	}
-	// 192.168.0.0/16
-	if ip[0] == 192 && ip[1] == 168 {
-		return fmt.Errorf("IP address %d.%d.%d.%d is in private range 192.168.0.0/16", ip[0], ip[1], ip[2], ip[3])
-	}
-	// 127.0.0.0/8
-	if ip[0] == 127 {
-		return fmt.Errorf("IP address %d.%d.%d.%d is in loopback range 127.0.0.0/8", ip[0], ip[1], ip[2], ip[3])
-	}
-	// 0.0.0.0
-	if ip[0] == 0 && ip[1] == 0 && ip[2] == 0 && ip[3] == 0 {
-		return fmt.Errorf("IP address 0.0.0.0 is not allowed")
-	}
 	return nil
 }
 
--- a/sdk/geolocation/go/add_target.go
+++ b/sdk/geolocation/go/add_target.go
@@ -1,0 +1,117 @@
+package geolocation
+
+import (
+	"fmt"
+
+	"github.com/gagliardetto/solana-go"
+	"github.com/near/borsh-go"
+)
+
+type AddTargetInstructionConfig struct {
+	Code               string
+	ProbePK            solana.PublicKey
+	TargetType         GeoLocationTargetType
+	IPAddress          [4]uint8
+	LocationOffsetPort uint16
+	TargetPK           solana.PublicKey
+}
+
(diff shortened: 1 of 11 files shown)


sdk/geolocation/go/add_target.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	default:
+		return fmt.Errorf("unknown target type: %d", c.TargetType)
+	}
+
+	return nil
+}
+
+// validateNotPrivateIP checks that an IPv4 address is not in a private/reserved range.
+func validateNotPrivateIP(ip [4]uint8) error {
```
vihu: I don't know of a good place to put this so just adding it here: this verification is too loose compared to what happens in `doublezero-geolocation/src/validation.rs`  which also rejects CGNAT, link-local etc. I'd recommend basically porting all of that logic here as well (ideally 1:1 if possible).

sdk/geolocation/go/remove_target.go
```diff
@@ -0,0 +1,88 @@
...
+	if c.ServiceabilityGlobalState.IsZero() {
+		return fmt.Errorf("serviceability global state public key is required")
+	}
+	if c.TargetType > GeoLocationTargetTypeOutboundIcmp {
+		return fmt.Errorf("unknown target type: %d", c.TargetType)
+	}
+	return nil
+}
+
+func BuildRemoveTargetInstruction(
```
vihu, line 44: This function doesn’t validate inbound `TargetPK`, but inbound removals are matched by `target_pk` only on-chain. In `process_remove_targe`t, the program ignores` ip_address` for Inbound and looks up the target using `target_pk` ([smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/remove_target.rs:79](https://github.com/malbeclabs/doublezero/blob/855a28ca5390a16cd27b57454ddcf1f226f9703f/smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/remove_target.rs#L79)).

Since inbound adds already reject a default target_pk ([smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/add_target.rs:81](https://github.com/malbeclabs/doublezero/blob/855a28ca5390a16cd27b57454ddcf1f226f9703f/smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/add_target.rs#L81)), the SDK currently allows callers to build an inbound remove instruction that can never match anything and will always fail with `TargetNotFound`. Can we add the same non-zero `TargetPK` validation here for `GeoLocationTargetTypeInbound`?

sdk/geolocation/go/executor.go
```diff
@@ -0,0 +1,176 @@
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(250 * time.Millisecond):
+		}
+	}
+	return errors.New("signature not found after wait")
+}
+
+func (e *executor) waitForTransactionFinalized(ctx context.Context, sig solana.Signature) (*solanarpc.GetTransactionResult, error) {
```
vihu, line 150: So we should either:
- document that the caller needs to provide some deadline/timeout
- or optionally have some sort of default `WithFinalizationTimeout` (60s-120s seems like a good middle ground)

sdk/geolocation/go/executor_test.go
```diff
@@ -0,0 +1,78 @@
+package geolocation_test
```
vihu, line 1: After doing a review pass with claude:

Minimum additions worth having in this test suite:

- Happy path with a mock `ExecutorRPCClient`: blockhash → send → visible → finalized.
- `GetLatestBlockhash` returns error.
- `SendTransactionWithOpts` returns error.
- Finalization status transitions: `Processed` → `Confirmed` → `Finalized`.
- `waitForSignatureVisible` times out (Value[0] stays nil).
- Context cancellation mid-wait.
- `SkipPreflight=true` passthrough is observed by the mock.

The `mockRPCClient` pattern in `main_test.go:41-54` already exists for `RPCClient`; extending it (or adding a sibling mock) for `ExecutorRPCClient` is straightforward.

sdk/geolocation/go/add_target.go
ben-dz: Good point. New validation.go covers all those cases.

sdk/geolocation/go/executor_test.go
ben-dz: Probably worth having. But is the explanation for the bloat in PR size.

vihu (approved):
LGTM

PR #3564, opened 2026-04-22
test(client/doublezero): return different min, max, avg from make_latency

axaysagathiya:
## Summary of Changes
Updated `make_latency` function in tests to accept and return distinct latency values for minimum, average, and maximum latency fields.

Closes #3495 

## Testing Verification
`cd client/doublezero && cargo test dzd_latency`

### DIFF
--- a/client/doublezero/src/dzd_latency.rs
+++ b/client/doublezero/src/dzd_latency.rs
@@ -414,9 +414,9 @@ mod tests {
         devices.insert(pk3, dev3);
 
         let latencies = vec![
-            make_latency(&pk1.to_string(), 12000000, 13000000, 14000000, true),
-            make_latency(&pk2.to_string(), 9000000, 10000000, 11000000, true),
-            make_latency(&pk3.to_string(), 15000000, 16000000, 17000000, true),
+            make_latency(&pk1.to_string(), 12_000_000, 5_000_000, 14_000_000, true),
+            make_latency(&pk2.to_string(), 9_000_000, 20_000_000, 11_000_000, true),
+            make_latency(&pk3.to_string(), 15_000_000, 15_000_000, 17_000_000, true),
         ];
 
         let mut controller = MockServiceController::new();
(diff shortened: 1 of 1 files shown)


axaysagathiya:
Is the current approach acceptable, or do we want to modify it?

We can accept the `min_latency_ns` as a function argument.
Then, we can calculate 
the `avg_latency_ns` as `min_latency_ns` plus 1 millisecond, and 
the `max_latency_ns` as `min_latency_ns` plus 2 milliseconds.

martinsander00:
Thanks for picking this up! Two suggestions to make the change worth it:

**1.** Convert the two existing "min vs avg" tests to use `make_latency`, they're natural callers for the new signature:
- `test_retrieve_latencies_sorts_by_min_not_avg` (lines 1013–1032)
- `test_best_latency_current_device_seeded_by_min_not_avg` (lines 1066–1085)

**2.** Make one existing test adversarial. Right now every call site passes monotonic triples (`10M, 11M, 12M`), so swapping `min_latency_ns` → `avg_latency_ns` in the ranking code wouldn't fail anything. In `test_best_latency_selects_lowest`, change the values so min and avg disagree on the winner (e.g., pk1 min=12M/avg=5M, pk2 min=9M/avg=20M, pk3 min=15M/avg=15M), same assertion, but now it actually catches a wrong-field bug.

martinsander00:
two things:

1. can you use the same number notation we use in the rest of the codebase? Rn looks like you are using two different notations (1_000_000 vs 1000000)
2. seems like due to some cross repo access policies our workflows don't run on forks. I'll chat with the team to see what we can do about it

axaysagathiya:
> seems like due to some cross repo access policies our workflows don't run on forks. I'll chat with the team to see what we can do about it

hello @martinsander00 
Any update on this? Since workflows don’t run on forks, would granting contributor access help (so I can push a branch), or is there another approach you’d prefer?

martinsander00:
We will move some parts of the codebase around to fix this issue. Apologies for the delay. Hopefully we will do this work in the next few weeks.

martinsander00:
@axaysagathiya commit are not yet signed btw. You need to use a GPG key and see a "Verified" label next to the commit hash

ben-dz (approved):

PR #3566, opened 2026-04-22
smartcontract: detect orphaned ResourceExtension accounts in resource verify

elitegreg:
## Summary
- Extend \`doublezero resource verify\` to detect \`ResourceExtension\` accounts on-chain whose PDA does not correspond to any currently-expected resource type (global singleton or per-device extension for a live device/prefix). These accumulate when devices are deleted or a device's \`dz_prefixes\` list shrinks.
- With \`--fix\`, propose closing each orphan in the existing y/N confirmation flow and close via a new \`CloseResourceByPubkeyCommand\`.
- Global singletons \`VrfIds\` and \`AdminGroupBits\` — not currently verified against usage — are still recognized as legitimate, not flagged as orphans.

## Design notes
- The on-chain \`process_closeaccount_resource_extension\` only needs the resource account + owner + globalstate, so for orphans (whose \`ResourceType\` can no longer be derived from live state) we bypass \`GetResourceCommand\` and pass pubkey/owner directly. Existing \`CloseResourceCommand\` now delegates to \`CloseResourceByPubkeyCommand\`.
- Expected-PDA set is built from loaded state: 8 globals + \`TunnelIds(device, 0)\` and \`DzPrefixBlock(device, i)\` for \`i in 0..device.dz_prefixes.len()\` per device. Anything else in \`resource_extensions\` is an orphan.
- Closing requires foundation allowlist membership — failures surface with the same \`FAILED: ...\` shape used elsewhere in the fix flow.

## Testing Verification
- 4 new unit tests in \`verify.rs\`:
  - Orphan detection when a device is deleted (dangling \`TunnelIds\`).
  - Orphan detection when \`dz_prefixes\` shrinks (stale \`DzPrefixBlock(device, i>=len)\`).
  - \`VrfIds\` / \`AdminGroupBits\` globals present but not verified — confirmed not flagged.
  - Report output contains the new "Orphaned resource extensions" section with the orphan PDA and \`associated_with\` pubkey.

### DIFF
--- a/smartcontract/cli/src/resource/verify.rs
+++ b/smartcontract/cli/src/resource/verify.rs
@@ -2,8 +2,8 @@ use crate::doublezerocommand::CliCommand;
 use clap::Args;
 use doublezero_program_common::types::NetworkV4;
 use doublezero_sdk::commands::resource::{
-    allocate::AllocateResourceCommand, create::CreateResourceCommand,
-    deallocate::DeallocateResourceCommand,
+    allocate::AllocateResourceCommand, closeaccount::CloseResourceByPubkeyCommand,
+    create::CreateResourceCommand, deallocate::DeallocateResourceCommand,
 };
 use doublezero_serviceability::{
     pda::get_resource_extension_pda,
@@ -49,6 +49,16 @@ pub enum ResourceDiscrepancy {
         first_account_type: String,
         second_account_pubkey: Pubkey,
         second_account_type: String,
+    },
+    /// ResourceExtension account exists onchain but does not correspond to any
+    /// currently-expected PDA (global singleton or per-device extension for a
+    /// live device/prefix). Typically caused by device deletion or a shrunk
+    /// dz_prefixes list.
+    OrphanedExtension {
+        pubkey: Pubkey,
+        associated_with: Pubkey,
+        owner: Pubkey,
+        allocator_kind: &'static str,
     },
 }
 
@@ -134,6 +144,7 @@ impl VerifyResourceCliCommand {
             let mut used_not_allocated: Vec<&ResourceDiscrepancy> = Vec::new();
             let mut extensions_not_found: Vec<&ResourceDiscrepancy> = Vec::new();
             let mut duplicate_usages: Vec<&ResourceDiscrepancy> = Vec::new();
+            let mut orphaned_extensions: Vec<&ResourceDiscrepancy> = Vec::new();
 
             for d in &result.discrepancies {
                 match d {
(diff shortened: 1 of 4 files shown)


smartcontract/sdk/rs/src/commands/resource/closeaccount.rs
```diff
@@ -15,21 +15,40 @@ pub struct CloseResourceCommand {
 
 impl CloseResourceCommand {
     pub fn execute(&self, client: &dyn DoubleZeroClient) -> eyre::Result<Signature> {
-        let (globalstate_pubkey, _globalstate) = GetGlobalStateCommand
```
martinsander00, line 18: why do we delete this? can't we keep both so that we log "Globalstate not initialized" before "device not found"?

smartcontract/cli/src/resource/verify.rs
```diff
@@ -560,9 +641,74 @@ fn verify_resources<C: CliCommand>(client: &C) -> eyre::Result<VerifyResourceRes
...
+    program_id: &Pubkey,
+    devices: &HashMap<Pubkey, Device>,
+    resource_extensions: &HashMap<Pubkey, ResourceExtensionOwned>,
+    result: &mut VerifyResourceResult,
+) {
+    let mut expected: HashSet<Pubkey> = HashSet::new();
+
+    // Global singletons. VrfIds and AdminGroupBits aren't verified against
+    // usage above but must still be treated as legitimate, not orphans.
+    for resource_type in [
```
martinsander00, line 664: maybe it's for another pr but would it be worth it to do something like this in [doublezero/smartcontract/programs/doublezero-serviceability/src/resource.rs](https://github.com/malbeclabs/doublezero/blob/main/smartcontract/programs/doublezero-serviceability/src/resource.rs):

```
#[derive(BorshSerialize, BorshDeserialize, Clone, Copy, Debug, PartialEq, strum::EnumIter)]
  pub enum GlobalResource {                                                                                                           
      DeviceTunnelBlock,
      UserTunnelBlock,                                                                                                                
      MulticastGroupBlock,                                  
      MulticastPublisherBlock,                                                                                                        
      LinkIds,                                              
      SegmentRoutingIds,
      VrfIds,                                                                                                                         
      AdminGroupBits,
  }                                                                                                                                   
                                                            
  pub enum ResourceType {
      Global(GlobalResource),
      DzPrefixBlock(Pubkey, usize),
      TunnelIds(Pubkey, usize),                                                                                                       
  }
```

so then we can verify the type doing something like:
```
for g in GlobalResource::iter() {                         
      let (pda, _, _) = get_resource_extension_pda(program_id, ResourceType::Global(g));
      expected.insert(pda);                                                                                                           
  }
```

elitegreg: I'm pretty sure that refactor would break compatibility at this point. It's a good idea, but I don't see the value in this situation.

smartcontract/sdk/rs/src/commands/resource/closeaccount.rs
elitegreg: You are misreading the diff. This isn't removed, it's just shifted down into the next command where it is used.

martinsander00 (approved):

PR #3567, opened 2026-04-22
sdk/geolocation: add Go write SDK for user lifecycle and result destination

ben-dz:
Stacked on #3558. Once that merges, GitHub retargets this PR to `main` automatically.

## Summary of Changes
- Adds the remaining Go write SDK surface deferred out of #3558: instruction builders for `CreateGeolocationUser`, `UpdateGeolocationUser`, `DeleteGeolocationUser`, and `SetResultDestination`.
- Adds a high-level `executor.SetResultDestination(ctx, code, destination, opts)` helper that fetches the user, derives the exact set of probe accounts from `user.Targets` via the exported `DeriveUniqueProbePKs`, and submits the instruction — so Go consumers don't have to re-implement the dedupe logic the Rust CLI does today.
- Extends `ExecutorRPCClient` with `GetAccountInfo` so the executor can read onchain state before writing.

## Diff Breakdown
| Category   | Files | Lines (+/-) |  Net |
| Core logic |     5 | +287 / -2   | +285 |
| Tests      |     7 | +712 / -3   | +709 |

Pure-additive; ~70% of the diff is tests.

## Testing Verification
- Borsh serialization tests pin discriminator bytes, field ordering, and `Option<Pubkey>` encoding against the Rust program's expected layout for each new instruction.
- Account-meta ordering verified against the Rust SDK command builders.
- `SetResultDestination` helper error paths covered: missing signer, missing program ID, account not found, owner mismatch, empty code; dedupe behavior covered via pure-function tests of `DeriveUniqueProbePKs` (empty, all-unique, duplicates-preserving-first-occurrence).

### DIFF
--- a/sdk/geolocation/go/add_target.go
+++ b/sdk/geolocation/go/add_target.go
@@ -29,7 +29,7 @@ func (c *AddTargetInstructionConfig) Validate() error {
 
 	switch c.TargetType {
 	case GeoLocationTargetTypeOutbound, GeoLocationTargetTypeOutboundIcmp:
-		if err := validateNotPrivateIP(c.IPAddress); err != nil {
+		if err := validatePublicIP(c.IPAddress); err != nil {
 			return err
 		}
 	case GeoLocationTargetTypeInbound:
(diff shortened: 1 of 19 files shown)


ben-dz:
Closing since we still don't have a use yet. Will reopen and rebase if it becomes needed.

PR #3577, opened 2026-04-23
feat(cli): allow unsubscribe without disconnect

### LINKED ISSUE #3575: Support disconnecting from individual multicast groups w/o restarting the tunnel
A user may publish or subscribe to multiple multicast groups. 
The cli support incrementally adding groups via the cli, but does not support having the ability to disconnect individual groups.

bgm-malbeclabs:
## Summary of Changes
This PR adds the ability for a user to unsubscribe or no longer publish without the need for a disconnect. This is only for connections managed by the user, ie not the oracle. There's a warning if a user tries to unpublish or unsubscribe from their only remaining group.

Closes https://github.com/malbeclabs/doublezero/issues/3575

## Testing Verification
* Updated e2e test
* Resolved a flaky single socket issue
* rust rests

### DIFF
--- a/client/doublezero/src/command/multicast.rs
+++ b/client/doublezero/src/command/multicast.rs
@@ -60,11 +60,11 @@ impl MulticastUnsubscribeCliCommand {
     pub async fn execute(self, client: &dyn CliCommand) -> eyre::Result<()> {
         let controller = ServiceControllerImpl::new(None);
         let client_ip = crate::command::helpers::resolve_client_ip(&controller).await?;
-        self.execute_inner(client, client_ip).await
+        self.execute_inner(client, client_ip)
     }
 
     /// Testable core: takes an already-resolved client_ip.
-    async fn execute_inner(self, client: &dyn CliCommand, client_ip: Ipv4Addr) -> eyre::Result<()> {
+    fn execute_inner(self, client: &dyn CliCommand, client_ip: Ipv4Addr) -> eyre::Result<()> {
         let spinner = init_command(2);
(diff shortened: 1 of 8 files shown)


client/doublezero/src/command/multicast.rs
```diff
@@ -0,0 +1,827 @@
...
+
+impl MulticastUnsubscribeCliCommand {
+    pub async fn execute(self, client: &dyn CliCommand) -> eyre::Result<()> {
+        let controller = ServiceControllerImpl::new(None);
+        let client_ip = crate::command::helpers::resolve_client_ip(&controller).await?;
+        self.execute_inner(client, client_ip).await
+    }
+
+    /// Testable core: takes an already-resolved client_ip.
+    async fn execute_inner(self, client: &dyn CliCommand, client_ip: Ipv4Addr) -> eyre::Result<()> {
```
ben-dz: Does this need to be async?

client/doublezero/src/command/multicast.rs
```diff
@@ -0,0 +1,827 @@
...
+                continue;
+            }
+            let carry_pub = user.publishers.contains(&group_pk);
+            client.update_multicastgroup_roles(UpdateMulticastGroupRolesCommand {
+                user_pk,
+                group_pk,
+                client_ip,
+                publisher: carry_pub,
+                subscriber: false,
+            })?;
```
ben-dz: I'd make an argument for log-and-continue here, since any that came before would have worked, why not try the ones after too?

ben-dz (approved):

PR #3578, opened 2026-04-24
controller: write config agent version info to ClickHouse

nikw9944:
## Summary of Changes
- Add a `controller_agent_versions` ClickHouse table using `ReplacingMergeTree(updated_at)` keyed by `device_pubkey` to track the latest agent and controller version per device
- The controller writes on every GetConfig poll; ClickHouse merges rows down to one per device. Lake queries with `FINAL` for deduplicated results
- Keep `controller_grpc_getconfig_success` lean (`timestamp` + `device_pubkey` only) — separates event tracking from version state
- Companion: malbeclabs/lake#534 adds `controller_agent_versions` to Lake's external remote table proxies
- Allows malbeclabs/lake to display device config agent version info
- Schema is still simple so not adding goose migrations at this point

## Diff Breakdown
| Category     | Files | Lines (+/-)   | Net  |
|--------------|-------|---------------|------|
| Core logic   |     2 | +98 / -12     |  +86 |
| Scaffolding  |     1 | +1 / -1       |   +0 |
| Docs         |     1 | +21 / -2      |  +19 |
| **Total**    |     4 | +120 / -15    | +105 |

Small feature — mostly core logic (ClickHouse writer + flush), with minimal scaffolding.

## Testing Verification
- `go build ./controlplane/controller/...` — builds clean
- `go test ./controlplane/controller/...` — all existing tests pass
- `go vet ./controlplane/controller/...` — no issues

### DIFF
--- a/controlplane/controller/README.md
+++ b/controlplane/controller/README.md
@@ -26,11 +26,14 @@ CREATE USER controller_devnet IDENTIFIED BY 'your_password';
 
 -- Grant permissions
 GRANT SELECT, INSERT, CREATE TABLE, SHOW COLUMNS ON devnet.controller_grpc_getconfig_success TO controller_devnet;
+GRANT SELECT, INSERT, CREATE TABLE, SHOW COLUMNS ON devnet.controller_agent_versions TO controller_devnet;
 ```
 
-The controller will automatically create the `controller_grpc_getconfig_success` table on startup and batch-insert GetConfig events every 10 seconds.
+The controller automatically creates both tables on startup and batch-inserts every 10 seconds.
 
-#### Table Schema
+#### Table Schemas
+
+**GetConfig events** — one row per successful GetConfig poll:
 
 ```sql
 CREATE TABLE devnet.controller_grpc_getconfig_success (
@@ -39,4 +42,20 @@ CREATE TABLE devnet.controller_grpc_getconfig_success (
 ) ENGINE = MergeTree
 PARTITION BY toYYYYMM(timestamp)
 ORDER BY (timestamp, device_pubkey)
+```
+
+**Agent versions** — latest agent and controller version per device. Uses `ReplacingMergeTree` so ClickHouse merges rows down to one per device; query with `FINAL` for deduplicated results:
+
+```sql
+CREATE TABLE devnet.controller_agent_versions (
+    device_pubkey LowCardinality(String),
+    updated_at DateTime64(3),
+    agent_version LowCardinality(String) DEFAULT '',
+    agent_commit LowCardinality(String) DEFAULT '',
+    agent_date LowCardinality(String) DEFAULT '',
+    controller_version LowCardinality(String) DEFAULT '',
+    controller_commit LowCardinality(String) DEFAULT '',
+    controller_date LowCardinality(String) DEFAULT ''
+) ENGINE = ReplacingMergeTree(updated_at)
+ORDER BY device_pubkey
 ```
--- a/controlplane/controller/README.md
+++ b/controlplane/controller/README.md
@@ -109,6 +109,8 @@ GRANT SELECT, INSERT, CREATE TABLE, SHOW COLUMNS ON devnet.controller_agent_vers
 
 The controller automatically creates both tables on startup and batch-inserts every 10 seconds.
 
+Apply the grants **before** deploying a controller version that adds a table: table creation degrades per-table, so a failing `CREATE TABLE` (e.g. a missing grant) disables writes to that table until restart while the other table keeps working. Failures are logged at `ERROR`.
+
 #### Table Schemas
 
 **GetConfig events** — one row per successful GetConfig poll:
@@ -137,3 +139,8 @@ CREATE TABLE devnet.controller_agent_versions (
 ) ENGINE = ReplacingMergeTree(updated_at)
 ORDER BY device_pubkey
 ```
+
+Caveats for consumers of this table:
+
+- Rows are **last-reported values, agent-asserted over an unauthenticated endpoint** with latest-write-wins semantics — treat them as informational, not verified fleet state.
+- Blank version reports are skipped, so a device whose agent never reports version fields gets no row (indistinguishable from never having polled), and a downgrade to a non-reporting agent leaves the old row in place. Check `updated_at` freshness or join against `controller_grpc_getconfig_success` for liveness.
--- a/controlplane/controller/README.md
+++ b/controlplane/controller/README.md
@@ -104,11 +104,14 @@ CREATE USER controller_devnet IDENTIFIED BY 'your_password';
 
 -- Grant permissions
 GRANT SELECT, INSERT, CREATE TABLE, SHOW COLUMNS ON devnet.controller_grpc_getconfig_success TO controller_devnet;
+GRANT SELECT, INSERT, CREATE TABLE, SHOW COLUMNS ON devnet.controller_agent_versions TO controller_devnet;
 ```
 
-The controller will automatically create the `controller_grpc_getconfig_success` table on startup and batch-insert GetConfig events every 10 seconds.
+The controller automatically creates both tables on startup and batch-inserts every 10 seconds.
 
-#### Table Schema
+#### Table Schemas
+
+**GetConfig events** — one row per successful GetConfig poll:
 
 ```sql
 CREATE TABLE devnet.controller_grpc_getconfig_success (
@@ -117,4 +120,20 @@ CREATE TABLE devnet.controller_grpc_getconfig_success (
 ) ENGINE = MergeTree
 PARTITION BY toYYYYMM(timestamp)
 ORDER BY (timestamp, device_pubkey)
+```
+
+**Agent versions** — latest agent and controller version per device. Uses `ReplacingMergeTree` so ClickHouse merges rows down to one per device; query with `FINAL` for deduplicated results:
+
+```sql
+CREATE TABLE devnet.controller_agent_versions (
+    device_pubkey LowCardinality(String),
+    updated_at DateTime64(3),
+    agent_version LowCardinality(String) DEFAULT '',
+    agent_commit LowCardinality(String) DEFAULT '',
+    agent_date LowCardinality(String) DEFAULT '',
+    controller_version LowCardinality(String) DEFAULT '',
+    controller_commit LowCardinality(String) DEFAULT '',
+    controller_date LowCardinality(String) DEFAULT ''
+) ENGINE = ReplacingMergeTree(updated_at)
+ORDER BY device_pubkey
 ```
--- a/controlplane/controller/internal/controller/clickhouse.go
+++ b/controlplane/controller/internal/controller/clickhouse.go
@@ -17,13 +17,30 @@ type getConfigEvent struct {
 	DevicePubkey string
 }
 
+type versionEvent struct {
+	DevicePubkey      string
+	UpdatedAt         time.Time
+	AgentVersion      string
+	AgentCommit       string
+	AgentDate         string
+	ControllerVersion string
+	ControllerCommit  string
+	ControllerDate    string
+}
+
 type ClickhouseWriter struct {
 	conn              clickhouse.Conn
 	log               *slog.Logger
 	db                string
 	mu                sync.Mutex
 	events            []getConfigEvent
+	versions          []versionEvent
 	consecutiveErrors int
+
+	// Controller build info, stamped on every version event.
+	controllerVersion string
+	controllerCommit  string
+	controllerDate    string
 }
 
 // consecutiveErrorThreshold is the number of consecutive flush failures
@@ -53,7 +70,7 @@ func buildClickhouseOptions(addr, db, user, pass string, disableTLS bool) *click
 	return opts
 }
 
-func NewClickhouseWriter(log *slog.Logger, addr, db, user, pass string, disableTLS bool) (*ClickhouseWriter, error) {
+func NewClickhouseWriter(log *slog.Logger, addr, db, user, pass string, disableTLS bool, controllerVersion, controllerCommit, controllerDate string) (*ClickhouseWriter, error) {
 	chOpts := buildClickhouseOptions(addr, db, user, pass, disableTLS)
 	conn, err := clickhouse.Open(chOpts)
 	if err != nil {
@@ -63,13 +80,16 @@ func NewClickhouseWriter(log *slog.Logger, addr, db, user, pass string, disableT
 		return nil, fmt.Errorf("error pinging clickhouse: %w", err)
 	}
 	return &ClickhouseWriter{
-		conn: conn,
-		log:  log,
-		db:   db,
+		conn:              conn,
+		log:               log,
+		db:                db,
+		controllerVersion: controllerVersion,
+		controllerCommit:  controllerCommit,
+		controllerDate:    controllerDate,
 	}, nil
 }
 
-func (cw *ClickhouseWriter) CreateTable(ctx context.Context) error {
+func (cw *ClickhouseWriter) CreateTables(ctx context.Context) error {
 	err := cw.conn.Exec(ctx, fmt.Sprintf(`
 		CREATE TABLE IF NOT EXISTS "%s".controller_grpc_getconfig_success (
 			timestamp DateTime64(3),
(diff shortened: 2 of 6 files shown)


controlplane/controller/internal/controller/clickhouse.go
```diff
@@ -139,6 +193,31 @@ func (cw *ClickhouseWriter) flush(ctx context.Context) {
...
+	if err := batch.Send(); err != nil {
+		_ = batch.Close()
+		cw.recordFlushError("error sending clickhouse versions batch", err)
+		return
+	}
+	if err := batch.Close(); err != nil {
+		cw.recordFlushError("error closing clickhouse versions batch", err)
+		return
+	}
+	cw.log.Debug("flushed version events to clickhouse", "count", len(versions))
```
ben-dz: `flushVersions` never resets `consecutiveErrors = 0` on success, unlike `flushEvents` (line 192). Both paths share the counter that drives WARN→ERROR escalation. Common case is masked because `flushEvents` runs first and resets, but if the events flush errors early (e.g. PrepareBatch failure) while versions succeed, the counter stays elevated though writes are flowing. Reads as a copy-paste omission. Fix: reset here on success, or reset once in `flush()` after both sub-flushes succeed.

controlplane/controller/internal/controller/server.go
```diff
@@ -824,6 +824,13 @@ func (c *Controller) GetConfig(ctx context.Context, req *pb.ConfigRequest) (*pb.
 			Timestamp:    reqStart,
 			DevicePubkey: req.GetPubkey(),
 		})
+		c.clickhouse.RecordVersion(versionEvent{
```
ben-dz: Write amplification: `RecordVersion` runs on every GetConfig poll, but agent version changes very rarely, so this inserts a near-static ReplacingMergeTree row per device per poll. Intentional per the PR design (merges down, reads use FINAL) and matches the sibling getconfig_success pattern, but the churn-to-value ratio is high here. Optional: keep a per-device last-seen and only RecordVersion on change. Separately (L2): this runs even when all three agent fields are empty, so a blank report from an old/restarting agent overwrites the last good version row until the next non-empty poll — consider skipping when all agent fields are empty.

ben-dz (commented):
Small, additive, backward-compatible feature; build/vet/tests pass on the branch. No critical or high issues. One Medium worth addressing before merge: (M1) `flushVersions` does not reset the shared `consecutiveErrors` counter on success like `flushEvents` does, a copy-paste asymmetry that skews WARN→ERROR log escalation. Plus two Low design notes: write amplification (a near-static ReplacingMergeTree row per device per poll) and empty-field overwrite of a good version row. No injection, no secret leakage; schema change is additive.

controlplane/controller/internal/controller/clickhouse.go
nikw9944: Fixed in da0fbe8e — took the second option: `flushEvents`/`flushVersions` now return success, and `flush()` resets `consecutiveErrors` once after all sub-flushes succeed. This also avoids the inverse masking (versions succeeding resetting the counter while events flushes keep failing). Added no-op guard when both buffers are empty so an idle tick doesn't reset the counter.

controlplane/controller/internal/controller/server.go
nikw9944: The L2 part is fixed in da0fbe8e: `RecordVersion` now skips events where all three agent fields are empty, so a blank report can't overwrite the last good row (covered by a new unit test).

On the write amplification: keeping write-per-poll deliberately. Only-on-change dedup has two downsides here: a row lost to a failed flush (batches are dropped on error) would never be rewritten until the next version change, and `updated_at` would stop reflecting freshness — with periodic writes a consumer can tell the version info is current. The volume is one small row per device per ~10s poll, which ReplacingMergeTree merges away cheaply.

controlplane/controller/internal/controller/clickhouse.go
```diff
@@ -79,8 +99,26 @@ func (cw *ClickhouseWriter) CreateTable(ctx context.Context) error {
...
+			agent_commit LowCardinality(String) DEFAULT '',
+			agent_date LowCardinality(String) DEFAULT '',
+			controller_version LowCardinality(String) DEFAULT '',
+			controller_commit LowCardinality(String) DEFAULT '',
+			controller_date LowCardinality(String) DEFAULT ''
+		) ENGINE = ReplacingMergeTree(updated_at)
+		ORDER BY device_pubkey
+	`, cw.db))
+	if err != nil {
+		return fmt.Errorf("error creating agent_versions table: %w", err)
```
ben-dz: M1: `CreateTables` is all-or-nothing — if this second CREATE fails, `Controller.Run` sets `c.clickhouse = nil` (server.go:741-743) and the pre-existing `controller_grpc_getconfig_success` writes are silently lost too, surfaced only as a single WARN. ClickHouse grants are per-table (the README diff adds the new GRANT), so an existing deployment upgraded before applying the grant hits exactly this. Consider degrading per-table (per-stream enabled flags), or at minimum log at ERROR and call out the deploy-ordering requirement (grant first, then deploy).

controlplane/controller/internal/controller/clickhouse.go
```diff
@@ -106,20 +159,34 @@ func (cw *ClickhouseWriter) Run(ctx context.Context) {
...
+	}
+	ok := true
+	if len(events) > 0 {
+		ok = cw.flushEvents(ctx, events) && ok
+	}
+	if len(versions) > 0 {
+		ok = cw.flushVersions(ctx, versions) && ok
+	}
+	if ok {
+		cw.consecutiveErrors = 0
```
ben-dz, line 231: M2: `consecutiveErrors` is read/reset here and read-modify-written in `recordFlushError`/`logFlushError` without holding `cw.mu`. The single-flusher assumption doesn't hold: with `tls-listen-port` set, main.go clones the options slice (main.go:278) and both Controller instances share this writer (main.go:264), each starting `go c.clickhouse.Run(ctx)` (server.go:746) — two concurrent ticker loops, a genuine data race on the counter. The race predates the PR, but this reset is a new unlocked write site and the PR restructures the counter logic, so it's the natural place to fix: guard with `cw.mu`, use an atomic, or prevent double `Run`.

controlplane/controller/internal/controller/server.go
```diff
@@ -1040,6 +1040,13 @@ func (c *Controller) GetConfig(ctx context.Context, req *pb.ConfigRequest) (*pb.
...
```
ben-dz: M3: this persists unauthenticated, agent-asserted data with latest-write-wins semantics. `GetConfig` has no caller auth (no ClientAuth/mTLS, no auth interceptor; the only gate is the pubkey existing in the ledger cache, and device pubkeys are public onchain). A forged request can silently replace a device's true version row under `FINAL`, and the blank-skip prevents the genuine agent from resetting it. The endpoint being unauthenticated is pre-existing, but the persistent replace-semantics sink is new — worth documenting the trust model wherever lake presents this table as fleet state.

controlplane/controller/internal/controller/clickhouse.go
```diff
@@ -106,20 +159,34 @@ func (cw *ClickhouseWriter) Run(ctx context.Context) {
...
+		return
+	}
+	ok := true
+	if len(events) > 0 {
+		ok = cw.flushEvents(ctx, events) && ok
+	}
+	if len(versions) > 0 {
+		ok = cw.flushVersions(ctx, versions) && ok
+	}
+	if ok {
```
ben-dz, line 230: L1: with two sub-flushes per tick, a tick where both fail increments `consecutiveErrors` twice (one `recordFlushError` per sub-flush), so the WARN→ERROR threshold of 3 is crossed on the second bad tick, not the third — and the comment on `consecutiveErrorThreshold` still says "consecutive flush failures". Either increment at most once per `flush()` (symmetric with this once-per-flush reset) or update the comment.

controlplane/controller/internal/controller/clickhouse_test.go
```diff
@@ -67,6 +67,34 @@ func TestClickhouseWriterLogEscalation(t *testing.T) {
 	}
 }
 
+func TestRecordVersion(t *testing.T) {
```
ben-dz: L2: the record side is covered, but the new flush semantics aren't — reset-once-only-when-all-sub-flushes-succeed, the no-op-tick guard not resetting the counter, and `flushVersions` itself (INSERT column list vs `batch.Append` argument order) have zero coverage. `clickhouse.Conn` is an interface, so a fake conn/batch makes this cheap.

controlplane/controller/internal/controller/clickhouse.go
```diff
@@ -90,6 +128,21 @@ func (cw *ClickhouseWriter) Record(event getConfigEvent) {
 	cw.mu.Unlock()
 }
 
+func (cw *ClickhouseWriter) RecordVersion(event versionEvent) {
```
ben-dz: L3: `agent_version/commit/date` come verbatim from the request with no length cap, feeding `LowCardinality(String)` columns — a hostile caller can send large or high-cardinality values, bloating the in-memory buffer (bounded only by the 10s flush window) and degrading dictionary efficiency. Cheap hardening: truncate the three fields to a sane length (e.g. 128 bytes) here.

controlplane/controller/README.md
```diff
@@ -118,3 +121,19 @@ CREATE TABLE devnet.controller_grpc_getconfig_success (
 PARTITION BY toYYYYMM(timestamp)
 ORDER BY (timestamp, device_pubkey)
 ```
+
+**Agent versions** — latest agent and controller version per device. Uses `ReplacingMergeTree` so ClickHouse merges rows down to one per device; query with `FINAL` for deduplicated results:
```
ben-dz, line 127: L4: worth one sentence here noting the rows are last-reported values — a device whose agent never populates version fields gets no row at all (indistinguishable from "never polled"), and a downgrade to a non-reporting agent keeps the stale row indefinitely, so consumers should check `updated_at` freshness or join against `controller_grpc_getconfig_success`.

ben-dz (commented):
Small, additive, well-scoped feature; build/vet/tests (including -race) pass on the branch, and the earlier review-round fixes (once-per-flush counter reset, blank-report skip) are correctly implemented. No critical or high findings. Three mediums worth addressing: (M1) CreateTables is all-or-nothing, so a missing per-table GRANT on upgrade fails the new table's CREATE and silently disables the pre-existing getconfig pipeline; (M2) consecutiveErrors is accessed outside the mutex while TLS+plaintext dual mode runs two flush loops on one shared writer — a real data race this PR touches; (M3) the new table persists unauthenticated agent-asserted version strings with latest-write-wins replace semantics. Plus lows: stale counter-threshold comment / double-increment per tick, no flush-level test coverage, no length cap on agent strings, blank-skip staleness ambiguity worth a README sentence, optional in-batch dedup, silently-overwritten versionEvent.Controller* fields, nine positional constructor params, and partial-batch-send-after-Append-error idiom doubled.

- Checked and clean: row values are parameterized via batch.Append (only operator-configured cw.db is interpolated into identifiers, pre-existing pattern); CLICKHOUSE_PASS is never logged; buffers are drained before each flush attempt so memory is bounded while ClickHouse is down; the shutdown path flushes with a fresh context.Background(); updated_at = reqStart is the right choice for a replace-by-latest table; the schema change is additive.

controlplane/controller/internal/controller/clickhouse.go
```diff
@@ -129,14 +196,40 @@ func (cw *ClickhouseWriter) flush(ctx context.Context) {
...
 		cw.recordFlushError("error closing clickhouse batch", err)
-		return
+		return false
 	}
-	cw.consecutiveErrors = 0
 	cw.log.Debug("flushed getconfig events to clickhouse", "count", len(events))
+	return true
+}
+
+func (cw *ClickhouseWriter) flushVersions(ctx context.Context, versions []versionEvent) bool {
```
ben-dz, line 268: L5 (optional): with ~5s agent polls and a 10s flush this inserts ~2 redundant rows per device per tick; a small map-based dedup keeping only the newest event per `device_pubkey` at the top of `flushVersions` halves volume while preserving the `updated_at` freshness and failed-flush-self-heal properties argued for in the review thread.

controlplane/controller/internal/controller/clickhouse.go
```diff
@@ -17,13 +17,30 @@ type getConfigEvent struct {
 	DevicePubkey string
 }
 
+type versionEvent struct {
+	DevicePubkey      string
+	UpdatedAt         time.Time
+	AgentVersion      string
+	AgentCommit       string
+	AgentDate         string
+	ControllerVersion string
```
ben-dz, line 32: L6: these `Controller*` fields are caller-settable but unconditionally overwritten in `RecordVersion` (lines 138-140). Either narrow `RecordVersion`'s input to the device/agent fields or comment the struct so a future caller doesn't expect their values to survive.

controlplane/controller/internal/controller/clickhouse.go
```diff
@@ -53,7 +70,7 @@ func buildClickhouseOptions(addr, db, user, pass string, disableTLS bool) *click
 	return opts
 }
 
-func NewClickhouseWriter(log *slog.Logger, addr, db, user, pass string, disableTLS bool) (*ClickhouseWriter, error) {
+func NewClickhouseWriter(log *slog.Logger, addr, db, user, pass string, disableTLS bool, controllerVersion, controllerCommit, controllerDate string) (*ClickhouseWriter, error) {
```
ben-dz: L7: nine positional parameters, three of them an easy-to-transpose string triple. A small `BuildInfo{Version, Commit, Date string}` struct passed once would help. Not blocking.

controlplane/controller/internal/controller/clickhouse.go
```diff
@@ -129,14 +196,40 @@ func (cw *ClickhouseWriter) flush(ctx context.Context) {
...
+func (cw *ClickhouseWriter) flushVersions(ctx context.Context, versions []versionEvent) bool {
+	batch, err := cw.conn.PrepareBatch(ctx, fmt.Sprintf(
+		`INSERT INTO "%s".controller_agent_versions (device_pubkey, updated_at, agent_version, agent_commit, agent_date, controller_version, controller_commit, controller_date)`, cw.db,
+	))
+	if err != nil {
+		cw.recordFlushError("error preparing clickhouse versions batch", err)
+		return false
+	}
+	for _, v := range versions {
+		if err := batch.Append(v.DevicePubkey, v.UpdatedAt, v.AgentVersion, v.AgentCommit, v.AgentDate, v.ControllerVersion, v.ControllerCommit, v.ControllerDate); err != nil {
```
ben-dz, line 287: L8: an `Append` error here would be a systematic type/programming error affecting every row, yet the partial batch is still sent with only a WARN (and no `consecutiveErrors` increment, since it uses `logFlushError`). Returning `false` on the first Append error would be more honest. Copied from the pre-existing events path, so low priority, but the PR doubled the pattern.

controlplane/controller/internal/controller/clickhouse.go
nikw9944: Fixed in dc07a2e2 — `CreateTables` now degrades per-table (failed CREATE logged at ERROR, only that table's writes disabled; error returned only when both fail). README notes the grant-before-deploy ordering.

controlplane/controller/internal/controller/clickhouse.go
nikw9944: Fixed in dc07a2e2 — all `consecutiveErrors` access (and the new per-table flags) is now under `cw.mu`.

controlplane/controller/internal/controller/server.go
nikw9944: Documented in dc07a2e2 — README now flags rows as agent-asserted over an unauthenticated endpoint with latest-write-wins semantics: informational, not verified fleet state. Will mirror the caveat on the lake side.

controlplane/controller/internal/controller/clickhouse.go
nikw9944: Fixed in dc07a2e2 — the counter now advances at most once per `flush()` pass, so the threshold counts failed ticks and the comment stays accurate.

controlplane/controller/internal/controller/clickhouse_test.go
nikw9944: Added in dc07a2e2 — fake `driver.Conn`/`driver.Batch` tests now cover reset-on-success, no-op ticks not resetting, once-per-tick increment + WARN→ERROR escalation, append-error abort, per-device dedup, per-table degrade, and the versions column order.

controlplane/controller/internal/controller/clickhouse.go
nikw9944: Fixed in dc07a2e2 — the three agent fields are truncated to 128 bytes in `RecordVersion`.

controlplane/controller/README.md
nikw9944: Added in dc07a2e2 — README caveats now cover last-reported semantics, no row for never-reporting agents, stale rows after downgrades, and checking `updated_at` / joining `controller_grpc_getconfig_success`.

controlplane/controller/internal/controller/clickhouse.go
nikw9944: Done in dc07a2e2 — `flushVersions` keeps only the newest event per device before batching.

controlplane/controller/internal/controller/clickhouse.go
nikw9944: Fixed in dc07a2e2 — `RecordVersion` now takes a narrowed `agentVersionReport`; the controller fields exist only on the internal `versionEvent`.

controlplane/controller/internal/controller/clickhouse.go
nikw9944: Done in dc07a2e2 — the constructor now takes a `Build{Version, Commit, Date}` struct (named `Build` since `BuildInfo` is taken by the Prometheus gauge).

controlplane/controller/internal/controller/clickhouse.go
nikw9944: Fixed in dc07a2e2 for both paths — the first `Append` error closes the batch and returns `false` instead of sending a partial batch.

ben-dz (approved):
Follow-up commit fully addresses the prior review: CreateTables degrades per-table (missing-grant upgrades no longer kill the existing getconfig pipeline; grant-before-deploy documented), all consecutiveErrors/flag access is under the mutex with the dual-listener sharing invariant documented, the README states the table is agent-asserted over an unauthenticated endpoint plus consumer staleness caveats, the counter advances once per flush tick, agent fields are truncated to 128 bytes, flushVersions dedupes to the newest row per device, RecordVersion takes a narrowed report type, the constructor takes a Build struct, and Append errors abort the batch instead of sending a partial one. New fake-Conn/Batch tests substantively pin all of this (escalation timing, column order vs appended values, dedup by UpdatedAt, per-table degrade). Verified independently: build, vet, and go test -race pass at the new head. Remaining nits are trivial (byte-level truncation can split a UTF-8 rune; a transient CREATE failure disables a stream until restart, which the README documents).

- Trivial, non-blocking: truncateAgentField slices at a byte boundary and can split a multi-byte UTF-8 rune at 128 bytes; harmless for storage, fine to leave.
- Trivial, non-blocking: a transient CREATE failure at startup disables that table's writes until restart — documented in the README, so accepted as a tradeoff.

PR #3583, opened 2026-04-24
sentinel: set tunnel endpoint on multicast publisher create

snormore:
## Summary of Changes
- Sentinel now picks a concrete `tunnel_endpoint` when creating a multicast publisher instead of sending `Ipv4Addr::UNSPECIFIED`. It prefers a `user_tunnel_endpoint` interface IP on the user's device and falls back to the device's `public_ip`, excluding any IP already used by another user at the same `client_ip` — mirroring the CLI behavior at `client/doublezero/src/command/connect.rs:310`.
- Adds a pure-function `tunnel_endpoint` module (`select_tunnel_endpoint`, `in_use_tunnel_endpoints`, `select_tunnel_endpoint_for_user`) with unit tests.
- Extends `MulticastDzLedgerClient` with `fetch_all_device_endpoints` so `poll_cycle` can resolve endpoints for each candidate; the RPC impl derives them from activated Device accounts.
- `DzUser` now carries `tunnel_endpoint` and `DzDeviceInfo` carries `public_ip` + `user_tunnel_endpoints`; `build_create_multicast_publisher_instructions` takes an explicit `tunnel_endpoint` argument.
- Threads the same selection into the admin CLI's `create-validator-multicast-publishers` command.

## Diff Breakdown
| Category     | Files | Lines (+/-)     | Net   |
|--------------|-------|-----------------|-------|
| Core logic   |     3 | +393 / -15      |  +378 |
| Scaffolding  |     5 | +32  / -1       |  +31  |
| **Total**    |     8 | +425 / -16      |  +409 |

Overwhelmingly additive: new selection module + wiring, plus small field additions to existing data types.

<details>

- `crates/sentinel/src/tunnel_endpoint.rs` — new module: `select_tunnel_endpoint`, `in_use_tunnel_endpoints`, `select_tunnel_endpoint_for_user`, and unit tests
- `crates/sentinel/src/multicast_publisher.rs` — adds `fetch_all_device_endpoints` to the trait, implements it in `RpcMulticastDzLedgerClient`, wires selection into `poll_cycle`, updates mocks and adds a test asserting the multicast publisher skips an IBRL-occupied UTE
- `controlplane/doublezero-admin/src/cli/sentinel.rs` — computes the exclude list from `all_users` for each candidate and passes the selected endpoint to `build_create_multicast_publisher_instructions`
- `crates/sentinel/src/dz_ledger_reader.rs` — reads `user.tunnel_endpoint` into `DzUser`; populates `public_ip` + `user_tunnel_endpoints` on `DzDeviceInfo` from device interfaces
- `crates/sentinel/src/dz_ledger_writer.rs` — `build_create_multicast_publisher_instructions` takes `tunnel_endpoint: Ipv4Addr` instead of hardcoding `UNSPECIFIED`

</details>

## Testing Verification
- Added `tunnel_endpoint_is_selected_excluding_ibrl_endpoint` in `multicast_publisher.rs`: device exposes two UTE IPs, the IBRL user occupies the first, and the asserted `create_multicast_publisher` call receives the second.
- Added unit tests covering `select_tunnel_endpoint` across all branches (UTE preferred, UTE excluded, public_ip fallback, everything excluded, no endpoints at all) and `select_tunnel_endpoint_for_user` (exclude-list derivation, unknown device).
- `cargo test -p doublezero-sentinel -p doublezero-admin` — 56 + 8 tests pass.

### DIFF
--- a/crates/sentinel/src/tunnel_endpoint.rs
+++ b/crates/sentinel/src/tunnel_endpoint.rs
@@ -1,0 +1,204 @@
+use std::net::Ipv4Addr;
+
+use solana_sdk::pubkey::Pubkey;
+
+/// Candidate tunnel endpoints for a device.
+#[derive(Debug, Clone)]
+pub struct DeviceEndpoints {
+    pub public_ip: Ipv4Addr,
+    /// IPs from device interfaces where `user_tunnel_endpoint == true`.
+    pub user_tunnel_endpoints: Vec<Ipv4Addr>,
+}
+
+impl Default for DeviceEndpoints {
+    fn default() -> Self {
+        Self {
+            public_ip: Ipv4Addr::UNSPECIFIED,
+            user_tunnel_endpoints: Vec::new(),
+        }
+    }
+}
+
+/// Select a tunnel endpoint for a user, preferring a `user_tunnel_endpoint`
+/// interface IP and falling back to the device's `public_ip`. Any IP in
+/// `exclude_ips` is skipped — these are the endpoints already used by other
+/// user accounts at the same `client_ip`.
+///
+/// Returns `Ipv4Addr::UNSPECIFIED` when no endpoint is available; the activator
+/// will then reject the create.
+pub fn select_tunnel_endpoint(
+    public_ip: Ipv4Addr,
+    user_tunnel_endpoints: &[Ipv4Addr],
+    exclude_ips: &[Ipv4Addr],
+) -> Ipv4Addr {
+    for ep in user_tunnel_endpoints {
+        if !exclude_ips.contains(ep) {
+            return *ep;
+        }
+    }
+
+    if public_ip != Ipv4Addr::UNSPECIFIED && !exclude_ips.contains(&public_ip) {
+        return public_ip;
+    }
+
+    Ipv4Addr::UNSPECIFIED
+}
+
+/// Collect the set of tunnel endpoints already in use by other user accounts
+/// at the same `client_ip`.
+pub fn in_use_tunnel_endpoints<'a, I>(users: I, client_ip: Ipv4Addr) -> Vec<Ipv4Addr>
+where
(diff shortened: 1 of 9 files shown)


crates/sentinel/src/tunnel_endpoint.rs
```diff
@@ -0,0 +1,204 @@
...
+
+/// Collect the set of tunnel endpoints already in use by other user accounts
+/// at the same `client_ip`.
+pub fn in_use_tunnel_endpoints<'a, I>(users: I, client_ip: Ipv4Addr) -> Vec<Ipv4Addr>
+where
+    I: IntoIterator<Item = &'a crate::dz_ledger_reader::DzUser>,
+{
+    users
+        .into_iter()
+        .filter(|u| u.client_ip == client_ip && u.tunnel_endpoint != Ipv4Addr::UNSPECIFIED)
```
packethog: Are we going to create a problem with this if an existing user has an unspecified address? If the user has an existing connection onchain with an unspecified tunnel endpoint, they're implicitly using the device public ip, but we would never add that as a in use endpoint here.

packethog (changes requested):

crates/sentinel/src/tunnel_endpoint.rs
snormore: Good catch, fixed in 18e5242d. `in_use_tunnel_endpoints` (and the equivalent helper in the admin CLI) now resolve UNSPECIFIED users to their device's `public_ip` using the device-endpoint map, mirroring the logic in `client/doublezero/src/command/connect.rs:939`. Added unit tests for the legacy case and an integration test `legacy_unspecified_tunnel_endpoint_excludes_device_public_ip` in the poll_cycle path.

packethog (approved):

elitegreg (approved):

PR #3588, opened 2026-04-24
cli: add tunnel_endpoint to user list output

juan-malbeclabs:
## Summary of Changes
- Added `tunnel_endpoint` field to `UserDisplay` in `doublezero user list`, exposing it in both table and JSON (`--json`) output
- Updated e2e fixtures to include the new column

## Diff Breakdown
| Category   | Files | Lines (+/-) | Net |
|------------|-------|-------------|-----|
| Core logic |     1 | +4 / -2     |  +2 |
| Tests      |     3 | +7 / -7     |   0 |
| **Total**  |     4 | +11 / -9    |  +2 |

Minimal change: one new field surfaced in the CLI, test fixtures updated to match.

<details>

- `smartcontract/cli/src/user/list.rs` -- adds `tunnel_endpoint: Ipv4Addr` to `UserDisplay` and maps it from `user.tunnel_endpoint`

</details>

## Testing Verification
- Unit tests in `user::list` updated and passing (table and JSON output assertions cover the new column)
- e2e fixtures for IBRL and IBRLWithAllocatedIP updated to include `tunnel_endpoint` column (`0.0.0.0` for users with no explicit endpoint set)

### DIFF
--- a/smartcontract/cli/src/user/list.rs
+++ b/smartcontract/cli/src/user/list.rs
@@ -113,6 +113,7 @@ pub struct UserDisplay {
     pub accesspass: String,
     pub tunnel_id: u16,
     pub tunnel_net: NetworkV4,
+    pub tunnel_endpoint: Ipv4Addr,
     pub status: UserStatus,
     pub bgp_status: BGPStatus,
     /// Raw BGP RTT in nanoseconds (kept on the struct for JSON output).
@@ -383,6 +384,7 @@ impl ListUserCliCommand {
                     },
                     tunnel_id: user.tunnel_id,
                     tunnel_net: user.tunnel_net,
+                    tunnel_endpoint: user.tunnel_endpoint,
                     status: user.status,
                     bgp_status: user.bgp_status,
                     bgp_rtt_ns: user.bgp_rtt_ns,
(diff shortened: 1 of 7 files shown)


smartcontract/cli/src/user/list.rs
```diff
@@ -113,6 +113,7 @@ pub struct UserDisplay {
     pub accesspass: String,
     pub tunnel_id: u16,
     pub tunnel_net: NetworkV4,
+    pub tunnel_endpoint: Ipv4Addr,
```
ben-dz, line 116: Informational, no action needed: an unset tunnel endpoint renders as `0.0.0.0` rather than a friendlier placeholder (e.g. `-`). This is consistent with `user get` and with how other IP columns in this table render unset values, so it's intentional and fine.

ben-dz (approved):
Adds a `tunnel_endpoint` column to `doublezero user list` (table + JSON), sourced directly from the `User` account's `tunnel_endpoint: Ipv4Addr`. The change is minimal, type-correct, and consistent with `user get` (which also renders an unset endpoint as `0.0.0.0`). Struct field ordering matches the updated fixtures and JSON assertion. Unit tests (`user::list`) pass 10/10. e2e fixtures with data rows were updated; the one stale fixture (`doublezero_user_list_user_removed.txt`) has no data rows after removal and `DiffCLITable` compares data rows only, so it is harmless. No correctness issues found.

PR #3592, opened 2026-04-27
geolocation: refactor write paths to in-place targets cursor

### LINKED ISSUE #3591: geolocation: replace per-call Vec<GeolocationTarget> deserialize with in-place byte cursor
## Background

`GeolocationUser` packs all targets into a single `Vec<GeolocationTarget>` field. Every `add_target` / `remove_target` call borsh-deserializes the full Vec onto the BPF heap, and the `solana-program` 2.x default bump allocator hardcodes that heap to 32 KiB.

Empirical measurement (see [tests/add_target_cu_benchmark.rs](smartcontract/programs/doublezero-geolocation/tests/add_target_cu_benchmark.rs)):

| N targets | account bytes | CU | result |
|---:|---:|---:|---|
| 200 | 14,306 | 65,989 | ok |
| 250 | 17,856 | — | OOM at deserialize |

So the de-facto cap today is ~200 targets per user, not the 4096 declared by `MAX_TARGETS`. CU is nowhere near binding (~287 CU per existing target).

## Why the easier fix isn't enough

The obvious workaround — override the global allocator to use 256 KiB and pair with `ComputeBudgetInstruction::request_heap_frame(256 * 1024)` — was prototyped and benchmarked. It lifts the ceiling from ~200 to ~1750 targets (CU at the new ceiling: ~509 K of 1.4 M budget). That's an 8-9× win, but it's still well below the existing 4096 cap, and it's a hard breaking change for every caller (any tx that forgets `request_heap_frame` faults instantly with an access violation, even at N=0). Coordinated CLI/SDK/test updates required.

In short: the allocator workaround is a half-measure, not a fix. We'd ship breaking-change plumbing across every call site to land 1750 — fewer targets than the cap we already declare.

## A multi-probe-Vec target struct doesn't help either

Restructuring `GeolocationTarget` to hold `Vec<Pubkey>` of probes (one struct per logical target instead of one per target+probe pair) reduces the entry count by a factor of K, but each entry grows by 32 bytes per probe. With Vec materialization on the heap, you save a small constant (the per-entry header amortized across K probes) but the dominant `K × 32 × N` term still scales heap with total combos. We'd hit OOM almost as fast.

The heap-saving must come from not materializing the targets list at all.

## Proposal

Stop materializing the targets list on the heap. Read and write targets in place, by byte offset, against the account buffer.

Concretely:

- A `TargetsCursor<'a>` accessor over `&'a [u8]` that exposes `len()`, `get(i)`, `iter() -> impl Iterator<Item = GeolocationTarget>`, `find_by(...)`. Per-element borsh deserialize is fine — one `GeolocationTarget` (71 bytes) is trivial.
- Rewrite `add_target` / `remove_target` to use the cursor: dup-scan via the iterator, append by reallocating +71 bytes and writing into the new tail, swap-remove by copying 71 bytes and reallocating -71 bytes. No `Vec<GeolocationTarget>` ever exists.
- Pin layout invariants with `const_assert!(size_of::<GeolocationTarget>() == 71)` and named offset/stride constants in the cursor module.

## Effects

- Per-call heap demand becomes O(1) regardless of N. Heap stops binding.
- CU per call drops — no full borsh round-trip — though dup-scan stays O(N).
- No CLI change required. No `request_heap_frame` needed.
- Wire format unchanged. No migration of existing accounts.

## Estimated cap after this change

Binding constraint shifts from heap to CU. Today's 287 CU/existing-target breaks down roughly as `~70 deserialize + ~70 reserialize + ~150 dup-scan/housekeeping`; the cursor eliminates the first two halves. At ~140-150 CU/entry, a 1.4 M budget allows roughly **8K-12K target+probe combos** per user, with headroom for `resize_account_if_needed` CPI overhead. Hard ceiling (account size, 10 MiB) at ~147K combos. The actual number wants a fresh benchmark once implemented.

A defensible new `MAX_TARGETS` lands in the 4K-8K range — verified by measurement, not declared on faith. The current 4096 cap would be reachable with margin for the first time.

## Forward compatibility for `GeolocationTarget`

The struct currently uses plain `BorshDeserialize`. Any field addition requires a deliberate migration regardless of whether we're on the Vec or the cursor.

**The "switch to BorshDeserializeIncremental later" plan was load-bearing but doesn't actually work.** Two stacked reasons:

1. The BDI macro generates only a `TryFrom<&[u8]>` impl, not `BorshDeserialize` ([borsh-incremental 0.1.2 lib.rs:142-150](https://docs.rs/borsh-incremental/0.1.2)). Swapping the derive on `GeolocationTarget` immediately breaks `targets: Vec<GeolocationTarget>` on the parent — Vec's deserializer requires `T: BorshDeserialize`. Won't compile.
2. Even with both derives held simultaneously, `Vec<T>::deserialize` calls the plain `BorshDeserialize` path. The "no bytes consumed → use default" check that makes BDI work only fires at the genuine end of the outer reader. Inside packed Vec elements, bytes from the next element look like more data, and a new optional field consumes them. Silent misalignment, not a graceful degrade.

So the realistic plan is:
1. **Keep `GeolocationTarget` sealed.** Grow via new `GeoLocationTargetType` variants, side-car PDAs, or new account types — not by appending fields. Same discipline as serviceability's stable wire types.
2. **If a real shape change becomes unavoidable, use versioned parent + one-shot migration.** Add `target_format_version: u8` to `GeolocationUser` (as a trailing field via existing BDI grace). Cursor dispatches stride by version. Ship a `MigrateTargets` instruction that rewrites the account in v1 layout. No mixed-stride account ever exists. For very large accounts the migration may need chunking to fit CU.

The cursor doesn't introduce these constraints — they exist today. It just makes the load-bearing assumption (`STRIDE = 71`) a `const_assert!` instead of a folk belief, which catches a future change at compile time rather than as runtime corruption.

## Out of scope

- Replacing the per-user-Vec architecture with one-PDA-per-target. Different tradeoff (per-target rate limits, billing-driven caps); evaluate separately if we want to push beyond ~10K combos.
- Changing `MAX_TARGETS`. Set it once we have the post-cursor benchmark.

ben-dz:
Resolves: #3591

## Summary of Changes
- Replace `GeolocationUser::try_from` with a new `GeolocationUserView` parser that records the targets section's byte offset and **skips** past the packed `Vec<GeolocationTarget>` instead of materializing it. Per-call heap usage is now bounded by the user's `code` and `result_destination` lengths, independent of `N`.
- Add a `TargetsCursor` for reads and free byte-level helpers (`append_target_bytes`, `swap_remove_target_bytes`) for in-place mutations against the account buffer. Each one stack-decodes a single 71-byte `GeolocationTarget` at a time.
- Refactor every on-chain write path (`add_target`, `remove_target`, `delete`, `update`, `update_payment_status`, `set_result_destination`) to use the new view + cursor instead of materializing the full user. `set_result_destination` rewrites its probe-validation as a streaming cursor scan with a small bounded `unique_from_targets` set, removing the previous `HashSet<Pubkey>` heap risk.
- Add a `stride_matches_wire_format` test plus an interleaved append/swap-remove oracle test that drives the byte helpers against a `Vec<GeolocationTarget>` mirror — catches off-by-one errors in the byte arithmetic.
- Add a CU benchmark (`tests/add_target_cu_benchmark.rs`) that sweeps `add_target` from N=0 up through N=4095. Pre-cursor this OOM'd at N≈250; post-cursor the full `MAX_TARGETS=4096` cap is now reachable, with N=4095 consuming ~913 K of the 1.4 M CU budget. Linear at ~221 CU/existing-target.

**Out of scope** (deliberate, separate PRs):
- `MAX_TARGETS` is intentionally untouched at 4096. Raising it should be a separate change with its own benchmark sweep above the current cap.
- No SDK / CLI / Go changes — wire format and instruction discriminators are unchanged.

## Diff Breakdown
| Core logic | 8 | +551 / -88 | +463 |
| Scaffolding (state/mod.rs) | 1 | +2 / -0 | +2 |
| Tests (inline + benchmark) | 3 | +690 / -0 | +690 |

Roughly 60% of the diff is tests (inline cursor/view tests + the benchmark). Production code change is ~460 net lines, and the processor refactors are net-small — most of the volume lives in the two new self-contained state modules.

<details>

- `smartcontract/programs/doublezero-geolocation/src/state/targets_cursor.rs` (+502/-0) — new module: `STRIDE` const + `stride_matches_wire_format` canary, `TargetsCursor<'a>` (`get`, `iter`, `find_match`), and the `append_target_bytes` / `swap_remove_target_bytes` byte-level mutators with an interleaved oracle test.
- `smartcontract/programs/doublezero-geolocation/src/state/geolocation_user_view.rs` (+384/-0) — new module: `GeolocationUserView` parses metadata and skips the targets section, then exposes `cursor`, `append_target`, `swap_remove_target`, `write_prefix`, `write_result_destination` helpers that handle the resize + memmove bookkeeping in place.
- `smartcontract/programs/doublezero-geolocation/tests/add_target_cu_benchmark.rs` (+229/-0) — `#[ignore]` benchmark sweeping `add_target` CU vs. existing-target count up to MAX_TARGETS-1.
- `smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/set_result_destination.rs` (+65/-30) — replaces the `HashSet<Pubkey>` of unique probes with a streaming cursor scan + bounded sorted `Vec<Pubkey>` (capped at `provided.len() + 1`) that preserves the original error semantics for count-mismatch vs. unrelated-probe cases.
- `smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/add_target.rs` (+22/-18) — switches to View + cursor for the dup-scan and `view.append_target` for the write.
- `smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/remove_target.rs` (+20/-21) — switches to View + cursor for the index lookup and `view.swap_remove_target` for the delete.
- `smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/update_payment_status.rs` (+9/-7) — reads via View, writes via `view.write_prefix`.
- `smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/delete.rs` (+5/-5) — `view.targets_count != 0` check replaces `user.targets.is_empty()`.

</details>

## Testing Verification
- 67 lib unit tests pass, including 30 new ones across the cursor and view modules: `stride_matches_wire_format`, cursor `iter` / `get` / `find_match` correctness, header round-trip across N=0..1000, byte-helper edge cases (last-index swap-remove, empty trailing, out-of-range, truncated buffer), and an interleaved append/swap-remove oracle test that diffs against a `Vec<GeolocationTarget>` mirror.
- 29 existing integration tests pass unmodified — wire format and instruction discriminators are unchanged.
- `cargo clippy --all-targets -- -Dclippy::all -Dwarnings` clean.
- `cargo +nightly fmt --check` clean.
- Empirical: `add_target` benchmark at N=4095 lands at 913,095 CU and 290,851 account bytes (was OOM at N≈250 pre-cursor). Numbers posted to issue #3591.

### DIFF
--- a/smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/add_target.rs
+++ b/smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/add_target.rs
@@ -3,9 +3,8 @@ use crate::{
     serializer::try_acc_write,
     state::{
         geo_probe::GeoProbe,
-        geolocation_user::{
-            GeoLocationTargetType, GeolocationTarget, GeolocationUser, MAX_TARGETS,
-        },
+        geolocation_user::{GeoLocationTargetType, GeolocationTarget, MAX_TARGETS},
+        geolocation_user_view::GeolocationUserView,
     },
     validation::validate_public_ip,
 };
@@ -62,16 +61,16 @@ pub fn process_add_target(
         return Err(ProgramError::InvalidAccountData);
     }
 
-    let mut user = GeolocationUser::try_from(user_account)?;
+    let mut view = GeolocationUserView::try_from_account(user_account)?;
 
-    if user.owner != *payer_account.key {
+    if view.owner != *payer_account.key {
         msg!("Signer is not the account owner");
         return Err(GeolocationError::Unauthorized.into());
     }
 
     let mut probe = GeoProbe::try_from(probe_account)?;
 
-    if user.targets.len() >= MAX_TARGETS {
+    if (view.targets_count as usize) >= MAX_TARGETS {
         msg!("Cannot add target: already at maximum of {}", MAX_TARGETS);
         return Err(GeolocationError::MaxTargetsReached.into());
     }
@@ -90,29 +89,34 @@ pub fn process_add_target(
         }
     }
 
-    let duplicate = user.targets.iter().any(|t| {
-        t.target_type == args.target_type
-            && t.geoprobe_pk == geoprobe_pk
-            && t.ip_address == args.ip_address
-            && t.target_pk == args.target_pk
-    });
-    if duplicate {
-        msg!("Target already exists");
-        return Err(GeolocationError::TargetAlreadyExists.into());
+    // Duplicate check: scan the targets section without materializing it.
+    {
+        let data = user_account.try_borrow_data()?;
+        let cursor = view.cursor(&data)?;
+        let duplicate = cursor.find_match(|t| {
+            t.target_type == args.target_type
+                && t.geoprobe_pk == geoprobe_pk
+                && t.ip_address == args.ip_address
+                && t.target_pk == args.target_pk
+        })?;
+        if duplicate.is_some() {
+            msg!("Target already exists");
+            return Err(GeolocationError::TargetAlreadyExists.into());
+        }
     }
 
-    user.targets.push(GeolocationTarget {
+    let new_target = GeolocationTarget {
         target_type: args.target_type,
         ip_address: args.ip_address,
         location_offset_port: args.location_offset_port,
         target_pk: args.target_pk,
         geoprobe_pk,
-    });
+    };
+    view.append_target(user_account, payer_account, accounts, &new_target)?;
 
     probe.reference_count = probe.reference_count.saturating_add(1);
     probe.target_update_count = probe.target_update_count.wrapping_add(1);
 
-    try_acc_write(&user, user_account, payer_account, accounts)?;
     try_acc_write(&probe, probe_account, payer_account, accounts)?;
 
     Ok(())
--- a/smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/delete.rs
+++ b/smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/delete.rs
@@ -1,6 +1,6 @@
 use crate::{
     error::GeolocationError, processors::check_foundation_allowlist, serializer::try_acc_close,
-    state::geolocation_user::GeolocationUser,
+    state::geolocation_user_view::GeolocationUserView,
 };
 use solana_program::{
     account_info::{next_account_info, AccountInfo},
@@ -35,9 +35,9 @@ pub fn process_delete_geolocation_user(
         return Err(ProgramError::InvalidAccountData);
     }
 
-    let user = GeolocationUser::try_from(user_account)?;
+    let view = GeolocationUserView::try_from_account(user_account)?;
 
-    if user.owner != *payer_account.key {
+    if view.owner != *payer_account.key {
         check_foundation_allowlist(
             program_config_account,
             serviceability_globalstate_account,
@@ -46,10 +46,10 @@ pub fn process_delete_geolocation_user(
         )?;
     }
 
-    if !user.targets.is_empty() {
+    if view.targets_count != 0 {
         msg!(
             "Cannot delete GeolocationUser with {} remaining targets",
-            user.targets.len()
+            view.targets_count
         );
         return Err(GeolocationError::TargetsNotEmpty.into());
     }
--- a/smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/remove_target.rs
+++ b/smartcontract/programs/doublezero-geolocation/src/processors/geolocation_user/remove_target.rs
@@ -4,7 +4,8 @@ use crate::{
     serializer::try_acc_write,
     state::{
         geo_probe::GeoProbe,
-        geolocation_user::{GeoLocationTargetType, GeolocationUser},
+        geolocation_user::GeoLocationTargetType,
+        geolocation_user_view::GeolocationUserView,
     },
 };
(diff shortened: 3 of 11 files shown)


juan-malbeclabs (approved):
Approach is solid. 

One request before merge: add an integration test that verifies `result_destination` survives an `add_target` + `remove_target` cycle. The current tests in `geolocation_user_test.rs` only call `set_result_destination` with a static target in place, nothing exercises the shift-forward/shift-back sequence through `banks_client`. That sequence is the core invariant of this PR and should have a CI signal.

Approving with that ask.

ben-dz:
> Approach is solid.
> 
> One request before merge: add an integration test that verifies `result_destination` survives an `add_target` + `remove_target` cycle. The current tests in `geolocation_user_test.rs` only call `set_result_destination` with a static target in place, nothing exercises the shift-forward/shift-back sequence through `banks_client`. That sequence is the core invariant of this PR and should have a CI signal.
> 
> Approving with that ask.

Good thought. Test added.

PR #3594, opened 2026-04-27
client/doublezerod/latency: fix flaky TestSingleSocketPing — race + stray-reply misattribution

ben-dz:
## Summary of Changes
Two distinct bugs in `SingleSocketPing` were combining to make `TestSingleSocketPing_Localhost` flaky in CI:

- **Data race on `rttEntry.sent`.** The sender goroutine wrote `states[i].rtts[seq].sent = time.Now()` while the reader goroutine read it inside `receiveTime.Sub(...)` to compute RTT. The kernel socket roundtrip provides logical happens-before but Go's race detector can't see across socket I/O, so this surfaced as an actual race report. Fixed with a per-state `sync.Mutex` covering the cross-goroutine access; the reader's update block is now one critical section with `totalReceived.Add(1)` outside the lock.
- **Stray echo replies were misattributed to pending slots.** The reader filtered incoming ICMP replies by peer IP and `seq` only, not by `echo.ID`. Any ambient echo reply on the host (e.g. responses to other processes' pings to localhost) with a sequence in `[0, pingCount)` was matched to one of our slots — whose `sent` was zero, so `receiveTime.Sub(zero)` saturated to `MaxDuration` and poisoned `min`/`avg`/`max` in `buildResults`. This is what produced the `got 61562 <= 20520 <= 9223372036854775807` assertion seen in CI. Fixed by matching on `echo.ID` (we set it to the target index when sending; raw ICMP sockets under `NET_RAW` preserve it), with a defensive peer-IP sanity check and `IsZero` guard on `sent` for belt-and-suspenders correctness. The `ipToIndices` map is no longer needed.

`buildResults` switches to index-based iteration so it doesn't copy `probeState` (which now contains a `sync.Mutex`) and trip govet. No locking is needed there — it runs only after `readerDone` closes, which establishes happens-before with the reader.

Originally surfaced as the flaky `go-test` failure in PR #3592 CI; isolated and fixed here so it can land independently.

## Diff Breakdown
| Core logic | 1 | +47 / -19 | +28 |

Single-file fix, no test changes — existing tests now pass cleanly under \`-race\`.

<details>

- \`client/doublezerod/internal/latency/singlesocket.go\` — adds \`sync.Mutex\` to \`probeState\`, locks around \`.sent\` accesses, switches the reader to match incoming replies by \`echo.ID\` (with peer-IP sanity check), drops the now-redundant \`ipToIndices\` map, and refactors \`buildResults\` to iterate by index.

</details>

## Testing Verification
- \`go test -race -count=20 -run TestSingleSocketPing ./client/doublezerod/internal/latency/...\` inside \`dz-go-test\` (which mirrors CI's \`NET_RAW\`/\`NET_ADMIN\` capabilities) — clean across 20 iterations, no race output, no assertion failures.
- \`go test -race ./client/doublezerod/internal/latency/...\` — full latency package, single pass — \`ok ... 56.161s\`.
- \`golangci-lint run ./client/doublezerod/internal/latency/...\` — \`0 issues\`.
- \`go vet ./client/doublezerod/internal/latency/...\` — clean.

### DIFF
--- a/client/doublezerod/internal/latency/singlesocket.go
+++ b/client/doublezerod/internal/latency/singlesocket.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"log/slog"
 	"net"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ben-dz:
Fixed in #3577

PR #3596, opened 2026-04-27
geolocation/sdk: bump compute budget for target-scan instructions

### LINKED ISSUE #3595: Geolocation: Increase CU Budget for Instructions that need it
AddTarget, RemoveTaret, and SetResultDestination all do O(n) scans through the target list, which can result in exhausting the compute budget with the default 200k. Need to set higher compute unit limit for those commands.

ben-dz:
Resolves: #3595

## Summary of Changes
- Rust and Go geolocation SDKs now prepend a `SetComputeUnitLimit(1_400_000)` instruction to `AddTarget`, `RemoveTarget`, and (Rust only) `SetResultDestination` transactions. These handlers do an O(n) scan over every target onchain; the default 200K CU per-tx limit exhausted around 743 targets, far below the program's `MAX_TARGETS=4096` ceiling. Consumers no longer need to think about compute budgets up to that bound.
- Rust SDK uses an exhaustive `match` over `GeolocationInstruction`, so adding a new variant forces an explicit CU-budget decision rather than silently falling through to the default.
- Go SDK renames `Build{Add,Remove}TargetInstruction` → `Build{Add,Remove}TargetInstructions` and changes the return type from `solana.Instruction` to `[]solana.Instruction`. The rename surfaces the breaking type change as a compile error rather than a silent slice/single mismatch; callers feed the returned slice into `executor.ExecuteTransactions`.
- The Go SDK currently has no `SetResultDestination` builder, so that handler is only bumped on the Rust side.

## Diff Breakdown
| Core logic |     4 | +182 / -22  | +160 |
| Tests      |     4 | +129 / -68  |  +61 |

Mostly SDK plumbing with matching test updates; no scaffolding/config/docs.

<details>

- `smartcontract/sdk/rs/src/geolocation/client.rs` — adds `TARGET_SCAN_COMPUTE_UNIT_LIMIT`, exhaustive-match `compute_unit_limit_for(...)` helper, and prepends the compute-budget instruction inside `GeoClient::execute_transaction`. Bucket-coverage unit tests for every variant.
- `sdk/geolocation/go/add_target.go` — `BuildAddTargetInstructions` returns `[]solana.Instruction` with a compute-budget prefix.
- `sdk/geolocation/go/remove_target.go` — same change for `BuildRemoveTargetInstructions`.
- `sdk/geolocation/go/constants.go` — adds `TargetMutationComputeUnitLimit`.
- `sdk/geolocation/go/{add_target,remove_target,instructions,executor}_test.go` — updated for new names/return type, plus regression tests asserting the prefix is present and requests the right limit.

</details>

## Testing Verification
- Reproduced the original failure mode (`exceeded CUs meter at BPF instruction` at ~743 targets) on software-devnet in CH7 prior to the fix; the stress run will now scale to `MAX_TARGETS=4096`.
- New unit tests in both SDKs assert per-variant bucket assignment and that `AddTarget`/`RemoveTarget` (and `SetResultDestination` in Rust) carry the 1.4M-CU prefix; the Rust tests use exhaustive variant lists so a new instruction can't silently slip through.
- Verified `add_target_cu_benchmark.rs` budget assumption (1.4M CU sufficient up to N=4095) is unchanged; this PR only flips the SDK to actually request that budget.

### DIFF
--- a/sdk/geolocation/go/add_target_test.go
+++ b/sdk/geolocation/go/add_target_test.go
@@ -9,7 +9,7 @@ import (
 )
 
 // addTargetMainIx returns the AddTarget instruction from the slice (index 1, after
-// the SetComputeUnitLimit prefix). assertHasComputeBudget verifies the prefix.
+// the SetComputeUnitLimit prefix). assertComputeBudgetPrefix verifies the prefix.
 func addTargetMainIx(t *testing.T, ixs []solana.Instruction) solana.Instruction {
 	t.Helper()
 	require.Len(t, ixs, 2, "expected [SetComputeUnitLimit, AddTarget]")
--- a/sdk/geolocation/go/add_target_test.go
+++ b/sdk/geolocation/go/add_target_test.go
@@ -8,14 +8,38 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestBuildAddTargetInstruction_Outbound(t *testing.T) {
+// addTargetMainIx returns the AddTarget instruction from the slice (index 1, after
+// the SetComputeUnitLimit prefix). assertHasComputeBudget verifies the prefix.
+func addTargetMainIx(t *testing.T, ixs []solana.Instruction) solana.Instruction {
+	t.Helper()
+	require.Len(t, ixs, 2, "expected [SetComputeUnitLimit, AddTarget]")
+	assertComputeBudgetPrefix(t, ixs[0])
+	return ixs[1]
+}
+
+// assertComputeBudgetPrefix checks that ix is a SetComputeUnitLimit instruction
+// requesting TargetMutationComputeUnitLimit. The discriminator for SetComputeUnitLimit
+// in the ComputeBudget program is 2 (RequestUnits=0, RequestHeapFrame=1,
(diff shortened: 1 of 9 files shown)


sdk/geolocation/go/add_target_test.go
```diff
@@ -8,14 +8,38 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestBuildAddTargetInstruction_Outbound(t *testing.T) {
+// addTargetMainIx returns the AddTarget instruction from the slice (index 1, after
+// the SetComputeUnitLimit prefix). assertHasComputeBudget verifies the prefix.
```
nikw9944: assertHasComputeBudget -> assertComputeBudgetPrefix ?

sdk/geolocation/go/executor_test.go
nikw9944: Rename to TestExecuteTransactions_NoPrivateKey, adding an `s` before the `_` to make it plural like ExecuteTransactions?  And maybe get rid of the old ExecuteTransaction() in executor.go since I don't see it being used anywhere?

nikw9944 (approved):

PR #2863, opened 2026-02-07
serviceability: add per-device unicast and multicast user limits

snormore:
## Summary of Changes
- Add separate limits for unicast and multicast users per device (`max_unicast_users`, `max_multicast_users`)
- Track user counts by type (`unicast_users_count`, `multicast_users_count`) 
- Enforce limits onchain in `CreateUser` and `CreateSubscribeUser` instructions
- Add client-side pre-flight capacity check before user creation
- New error codes: `MaxUnicastUsersExceeded` (73), `MaxMulticastUsersExceeded` (74)
- CLI: display new fields in `device get/list`, add `--max-unicast-users` and `--max-multicast-users` flags to `device update`
- Zero = unlimited semantics for backward compatibility
- [RFC-13: Per-Device Unicast and Multicast User Limits](https://github.com/malbeclabs/doublezero/blob/snor/separate-unicast-multicast-limits/rfcs/rfc13-per-device-unicast-multicast-user-limits.md)
- Resolves https://github.com/malbeclabs/infra/issues/537

## Testing Verification
- Rust unit tests for limit enforcement, counter updates, and zero=unlimited behavior
- E2E test (`TestE2E_UserLimits`) validates unicast and multicast limit enforcement

PR #3600, opened 2026-04-28
serviceability: allow count > max in Device::validate

ben-dz:
## Summary

- Removes the four `count <= max` invariants from `Device::validate` (`max_users`, `max_unicast_users`, `max_multicast_subscribers`, `max_multicast_publishers`) so operators can lower a per-device cap below the live count.
- Admission-time gates in `processors/user/create_core.rs` continue to reject new users when the cap is hit, so reducing the cap simply prevents further growth and the live count drains through natural churn.
- Restores the pre-#2259 behavior for `max_users`, and extends the same operator-friendly behavior to the per-type caps added in #2863.

## Why

Today `doublezero device update --max-multicast-subscribers N` (or any of the other caps) fails with `MaxMulticastSubscribersExceeded` whenever the current count exceeds N — making the cap a hard floor at the live count. There's no graceful way to shrink a cap. The structural invariant in `validate()` duplicates the admission check at user creation time; only the admission check meaningfully gates behavior, so removing the structural one is safe.

Example on testnet:

```
$ doublezero --env testnet device update --pubkey 6E1fuqbDBG5ejhYEGKHNkWG5mSTczjy4R77XCKEdUtpb --max-multicast-subscribers 1
Program log: Max multicast subscribers exceeded: multicast_subscribers_count = 2, max_multicast_subscribers = 1
Error: Max multicast subscribers exceeded
```

## Testing Verification

- Updated the two unit tests that asserted the old behavior (`test_state_device_validate_error_max_users_exceeded`, `test_max_users_exceeded`) to instead assert that `validate()` succeeds when `count > max` for all four caps.
- Ran integration tests in `device_test.rs` (11 pass) and `user_tests.rs` (6 pass) to confirm nothing else relied on the structural invariant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ All notable changes to this project will be documented in this file.
   - Fix BGP status submitter to collect socket stats and tunnel interfaces from all tenant VRF namespaces (`ns-vrf<N>`), not only `ns-vrf1`; users whose tenant has a non-default `VrfId` were previously always reporting "tunnel not found" and had their onchain BGP status left stale
 - CLI
   - Add `--narrow` flag to `doublezero user list` that hides `location`, `cyoa_type`, `accesspass`, and `tunnel_net`, abbreviates `user_type`, and summarizes `groups` as one publisher entry plus one subscriber entry with independent `+N` overflow counts; default output is unchanged
+- Smartcontract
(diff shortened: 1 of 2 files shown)


ben-malbeclabs (approved):
LGTM

PR #3634, opened 2026-05-01
client: add configurable doublezerod socket path

juan-malbeclabs:
## Summary of Changes
- Cherry-picks commit `1c28a1f` from @linuskendall that adds a `--sock-file` global CLI flag to the `doublezero` CLI
- Adds a `--sock-file` global flag (aliases: `--socket`, `--socket-path`) to the `doublezero` CLI to override the default Unix socket path used to communicate with `doublezerod`
- The socket path is resolved via a priority chain: explicit per-call argument -> global CLI override (`--sock-file`) -> compile-time default (`/var/run/doublezerod/doublezerod.sock`)
- Fixes `service_controller_check()` and `service_controller_can_open()` which previously always checked the hardcoded path, ignoring any configured socket

## Testing Verification
- New unit test `test_service_controller_uses_explicit_socket_path_for_checks` verifies both check methods return `true` when given an explicit valid path

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
+- Client
+  - Add `--sock-file` global flag (aliases: `--socket`, `--socket-path`) to the `doublezero` CLI to override the default Unix socket path used to communicate with `doublezerod` (`/var/run/doublezerod/doublezerod.sock`)
 - Controller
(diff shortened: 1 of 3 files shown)


nikw9944 (approved):
Looks good, but I found some of the names to be confusing. I think something like this would help: 

SOCKET_PATH_OVERRIDE -> GLOBAL_SOCKET_PATH

set_default_socket_path -> set_global_socket_path

Issue #3607, opened 2026-04-29
activator removal: tracker


elitegreg:
Tracker for the DoubleZero activator removal effort.

**Master plan:** `activator-removal-plan.md` (in greg's working tree; will be added to the repo if useful).

## Purpose

Retire `doublezero-activator`. Onchain allocation (RFC-11) supersedes it. The deployed activator is frozen — no new builds, no new releases — but existing instances keep running until the supported CLI floor rises high enough that the activator-invoked instructions can be removed from the program.

## Phasing overview

| Phase | What ships | Repo | Gate |
|------:|---|---|---|
| 1 | Stop deploying then stop building/releasing the activator | infra, doublezero | none |
| 2 | E2E runs onchain-allocation only; activator container removed from tests | doublezero | none |
| 3 | Delete `activator/` crate | doublezero | phases 1, 2 merged |
| 4 | Treat onchain-allocation as always-on; remove flag-gated branches; deprecate the flag | doublezero | phase 3 merged |
| 5 | Status enum cleanup (rename unused variants `*Deprecated`) | doublezero | phase 4 merged |
| 6 | In-repo docs and comments scrub | doublezero | phase 5 merged |
| 7a | Remove `User` activate/reject/closeaccount instructions | doublezero | **gate ≥ 0.12.0** |
| 7b | Remove device/link/multicastgroup activate/reject/closeaccount instructions | doublezero | **gate ≥ 0.14.1** |
| 8 | Tear down monitoring, shut down EC2 hosts in all environments, delete Terraform module | infra | all of 1–7 merged + product OK to lose ability to support older CLIs |

### Phase 1 — freeze deployment, then build and release

- [x] PR 1.1 — infra: stop deploying activator (keep instances running) — malbeclabs/infra#1175
- [x] PR 1.2 — doublezero: drop activator release pipeline — #3608

### Phase 2 — e2e migrates to onchain-allocation only

- [x] PR 2.1 — delete legacy BackwardCompatibility e2e shard — #3609
- [x] PR 2.2 — floor BackwardCompatibility shard at CLI 0.12.0 / 0.14.1 — #3610
- [x] PR 2.3 — convert remaining e2e shards to onchain-only; drop activator container — #3611

### Phase 3 — delete the activator crate

- [x] PR 3.1 — delete `activator/` crate from workspace — #3612

### Phase 4 — onchain-allocation always-on; deprecate the flag

[...]

PR #3648, opened 2026-05-01
smartcontract,controller: RFC-18 devnet fixes — node segment assignment and community stamping

ben-malbeclabs:
## Summary

- Atomic flex-algo node segment allocation on `CreateDeviceInterface`: Vpnv4 loopbacks under onchain allocation now allocate a `FlexAlgoNodeSegment` for each existing topology in the same transaction. New/reprovisioned devices no longer need a separate `AssignTopologyNodeSegments` step. The CLI auto-discovers topologies and passes them.
- New `UpdateDeviceInterface(update_topologies=true)` reconciliation path: deallocate removed topologies, allocate added ones, preserve unchanged. Authorized for the contributor owner or the foundation allowlist.
- `DeleteDeviceInterface` now also deallocates the loopback's flex-algo node segments under onchain deallocation.
- Controller: stamp the default `UNICAST-DEFAULT` topology color for users whose access pass has no tenant.
- Mechanical rename `BackfillTopology` → `AssignTopologyNodeSegments` across the program, CLI, and Rust SDK (instruction discriminant 110 unchanged).

## Context

Two issues discovered during the RFC-18 devnet deployment:

**1. Missing flex-algo node segments on new devices.** The activator used to call `BackfillTopology` after activating a VPNv4 loopback, assigning `FlexAlgoNodeSegment` entries for each existing topology. The activator was removed (#3607), and this responsibility was never ported. New or reprovisioned devices (e.g. dzd4 via the daily provisioning test) were missing node segments, causing the controller to log `flex_algo.enabled=true but VPNv4 loopback has no flex_algo_node_segments`.

The fix extends the onchain `CreateDeviceInterface` instruction to accept optional topology PDA accounts and allocate one `FlexAlgoNodeSegment` per topology atomically. `UpdateDeviceInterface` gains an `update_topologies` mode for ongoing reconciliation when topologies are added/removed after device creation, and `DeleteDeviceInterface` deallocates the segments alongside the base loopback SR ID. Topology accounts are validated by program-owner and by the first byte (`AccountType::Topology`) — a cheap check that prevents passing arbitrary program-owned accounts.

**2. Community stamping skipping tenantless users.** `resolveTenantColors` was only called inside the tenant lookup branch. All 985 unicast users on mainnet and 474/535 on testnet have no tenant, so they were silently getting no color. The fix adds an else branch that resolves the default UNICAST-DEFAULT color for tenantless users.

Related RFC: `rfcs/rfc-0018-flex-algo.md`

## Testing Verification

- `cargo test -p doublezero-serviceability` — full suite passes, including new `test_update_topologies_*` cases (adds/removes/clear/full-swap/no-op/rejects-non-vpnv4/rejects-duplicates) and `test_update_topologies_allowed_for_contributor` which verifies a non-foundation contributor-owner key can drive `update_topologies`
- `cargo test -p doublezero_cli device::interface::get` — `flex_algo_node_segments` now render as `topology_name:sr_id` rows; `list_topology` is fetched once and cached per get
- Verified mainnet has 0 users with tenants, testnet has 40/535 — confirming the community stamping gap


elitegreg (commented):
This solution pretty much requires that you never have more than about single digit or so number of topologies. As written today, it could be up to 25, but that assumes there is room in the packet, which is why I'm saying 8 or so. Is this a reasonable limitation? It seems to me like it is not as we may have several topologies per tenant?

ben-malbeclabs:
> This solution pretty much requires that you never have more than about single digit or so number of topologies. As written today, it could be up to 25, but that assumes there is room in the packet, which is why I'm saying 8 or so. Is this a reasonable limitation? It seems to me like it is not as we may have several topologies per tenant?

It would be a decent limit. I can't imagine us having that many.

I think other technologies become more interesting for us without have 10s of Flex-Algo topologies.

Worst case, if we did need more than 8, or even more than 25, is that difficult to achieve?

smartcontract/programs/doublezero-serviceability/src/processors/device/interface/create.rs
```diff
@@ -215,6 +229,20 @@ pub fn process_create_device_interface(
...
                 node_segment_idx = allocate_id(segment_routing_ids_ext)?;
+
+                // Allocate a flex-algo node segment for each topology
+                for topo_account in &topology_accounts {
+                    assert_eq!(
+                        topo_account.owner, program_id,
+                        "Invalid Topology Account Owner"
+                    );
+                    assert!(!topo_account.data_is_empty(), "Topology account is empty");
+                    let topo_segment_idx = allocate_id(segment_routing_ids_ext)?;
```
elitegreg, line 246: allocated node segment ids should be deallocated in the interface/remove instruction.

elitegreg (changes requested):

elitegreg:
> > This solution pretty much requires that you never have more than about single digit or so number of topologies. As written today, it could be up to 25, but that assumes there is room in the packet, which is why I'm saying 8 or so. Is this a reasonable limitation? It seems to me like it is not as we may have several topologies per tenant?
> 
> It would be a decent limit. I can't imagine us having that many.
> 
> I think other technologies become more interesting for us without have 10s of Flex-Algo topologies.
> 
> Worst case, if we did need more than 8, or even more than 25, is that difficult to achieve?

Difficult to achieve? Probably what we would need to do is come up with a more flexible implementation and create new version of the create interface command and retire the old one (this latter case is a big deal, with the activator instructions being the first instructions we are actually retiring). I think I would feel better if we had an instruction for adding a flex algo to an interface, and the having the interface create sdk (client side) call create interface, followed by iteratively adding the flex algo (each would allocate a segment routing id). My preference would be to do it this way now and not bork up create interface in a way that may only be temporary.

smartcontract/cli/src/device/interface/get.rs
```diff
@@ -48,6 +51,31 @@ impl GetDeviceInterfaceCliCommand {
             .find(|i| i.name.to_lowercase() == self.name.to_lowercase())
             .ok_or_else(|| eyre::eyre!("Interface '{}' not found", self.name))?;
 
+        // Resolve flex-algo topology PDAs to names. If the lookup fails (e.g. the
+        // topology was deleted), fall back to a truncated pubkey.
+        let topology_name_for = |pk: &solana_sdk::pubkey::Pubkey| -> Option<String> {
```
elitegreg: This should create a HashMap of pk->String so list_topology doesn't have to run for each topology.

smartcontract/programs/doublezero-serviceability/src/processors/device/interface/create.rs
elitegreg, line 246: should also assert that this can deserialize as a topology, otherwise, they could pass any program owned, non-empty account. A full deserialize isn't necessary. I believe we have a method for getting the account type from the first byte.

smartcontract/programs/doublezero-serviceability/src/processors/device/interface/update.rs
```diff
@@ -226,6 +253,69 @@ pub fn process_update_device_interface(
         iface.node_segment_idx = node_segment_idx;
     }
 
+    // Reconcile flex-algo node segments against the desired topology set.
+    // Existing entries whose topology is still in the set keep their SR ID; entries
+    // for removed topologies have their SR ID deallocated; new topologies get a
+    // freshly allocated SR ID.
+    if value.update_topologies {
+        if !globalstate.foundation_allowlist.contains(payer_account.key) {
```
elitegreg: Allow contributor key as well.

smartcontract/programs/doublezero-serviceability/src/processors/device/interface/update.rs
```diff
@@ -226,6 +253,69 @@ pub fn process_update_device_interface(
...
+        );
+
+        let mut desired: HashSet<Pubkey> = HashSet::new();
+        for topo_account in &topology_accounts {
+            assert_eq!(
+                topo_account.owner, program_id,
+                "Invalid Topology Account Owner"
+            );
+            assert!(!topo_account.data_is_empty(), "Topology account is empty");
+            if !desired.insert(*topo_account.key) {
```
elitegreg, line 300: validate that the first byte of the account matches the topology account type.

smartcontract/programs/doublezero-serviceability/src/processors/device/interface/update.rs
```diff
@@ -226,6 +253,69 @@ pub fn process_update_device_interface(
...
+                return Err(DoubleZeroError::InvalidArgument.into());
+            }
+        }
+
+        let mut kept: Vec<FlexAlgoNodeSegment> =
+            Vec::with_capacity(iface.flex_algo_node_segments.len());
```
elitegreg: shouldn't the capacity be set based on the larger of desired and flex_algo_node_segments?

elitegreg (approved):

elitegreg:
> > > This solution pretty much requires that you never have more than about single digit or so number of topologies. As written today, it could be up to 25, but that assumes there is room in the packet, which is why I'm saying 8 or so. Is this a reasonable limitation? It seems to me like it is not as we may have several topologies per tenant?
> > 
> > 
> > It would be a decent limit. I can't imagine us having that many.
> > I think other technologies become more interesting for us without have 10s of Flex-Algo topologies.
> > Worst case, if we did need more than 8, or even more than 25, is that difficult to achieve?
> 
> Difficult to achieve? Probably what we would need to do is come up with a more flexible implementation and create new version of the create interface command and retire the old one (this latter case is a big deal, with the activator instructions being the first instructions we are actually retiring). I think I would feel better if we had an instruction for adding a flex algo to an interface, and the having the interface create sdk (client side) call create interface, followed by iteratively adding the flex algo (each would allocate a segment routing id). My preference would be to do it this way now and not bork up create interface in a way that may only be temporary.

I have decided Ben's way of implementing this is fine for now. If we hit a limit for account arguments, we can add another instruction to add more topologies.

smartcontract/programs/doublezero-serviceability/src/processors/device/interface/create.rs
```diff
@@ -97,6 +104,12 @@ pub fn process_create_device_interface(
         None
     };
 
+    // Read topology PDA accounts (optional, for Vpnv4 loopback flex-algo assignment)
+    let mut topology_accounts = Vec::new();
+    for _ in 0..value.topology_count {
+        topology_accounts.push(next_account_info(accounts_iter)?);
+    }
```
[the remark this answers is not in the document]
elitegreg: Not needed. onchain allocation is always on now.

smartcontract/programs/doublezero-serviceability/src/processors/device/interface/create.rs
```diff
@@ -215,6 +229,26 @@ pub fn process_create_device_interface(
...
+                        topo_account.owner, program_id,
+                        "Invalid Topology Account Owner"
+                    );
+                    assert!(!topo_account.data_is_empty(), "Topology account is empty");
+                    let topo_type = AccountType::from(topo_account.try_borrow_data()?[0]);
+                    assert_eq!(
+                        topo_type,
+                        AccountType::Topology,
+                        "Invalid Topology Account Type"
+                    );
```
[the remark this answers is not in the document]
elitegreg: consistent with the rest of the program

smartcontract/programs/doublezero-serviceability/src/processors/device/interface/update.rs
```diff
@@ -226,6 +253,77 @@ pub fn process_update_device_interface(
...
+            let topo_type = AccountType::from(topo_account.try_borrow_data()?[0]);
+            assert_eq!(
+                topo_type,
+                AccountType::Topology,
+                "Invalid Topology Account Type"
+            );
+            if !desired.insert(*topo_account.key) {
+                return Err(DoubleZeroError::InvalidArgument.into());
+            }
+        }
```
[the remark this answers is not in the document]
elitegreg: consistent with the rest of the program

smartcontract/cli/src/device/interface/get.rs
```diff
@@ -48,6 +51,32 @@ impl GetDeviceInterfaceCliCommand {
...
 
+        // Resolve flex-algo topology PDAs to names. Cache the topology map up
+        // front so we don't run `list_topology` once per segment; on lookup
+        // miss (e.g. the topology was deleted), fall back to a truncated pubkey.
+        let flex_algo_node_segments = if interface.flex_algo_node_segments.is_empty() {
+            String::new()
+        } else {
+            let topology_map = client
+                .list_topology(ListTopologyCommand)
+                .unwrap_or_default();
```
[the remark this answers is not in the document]
elitegreg: not needed

smartcontract/programs/doublezero-serviceability/src/processors/device/interface/create.rs
```diff
@@ -215,6 +229,26 @@ pub fn process_create_device_interface(
...
                 node_segment_idx = allocate_id(segment_routing_ids_ext)?;
+
+                // Allocate a flex-algo node segment for each topology
+                for topo_account in &topology_accounts {
+                    assert_eq!(
+                        topo_account.owner, program_id,
+                        "Invalid Topology Account Owner"
+                    );
+                    assert!(!topo_account.data_is_empty(), "Topology account is empty");
+                    let topo_type = AccountType::from(topo_account.try_borrow_data()?[0]);
```
bgm-malbeclabs, line 240: will these panic if they are not asserted?

bgm-malbeclabs (approved):

smartcontract/programs/doublezero-serviceability/src/processors/device/interface/create.rs
elitegreg: Yes if the assertions fail, the instruction will terminate which is intended and consistent with the other instructions in the program.

elitegreg: We tend to assert on logic/programmer errors and Error on runtime errors.

smartcontract/programs/doublezero-serviceability/src/processors/device/interface/create.rs
```diff
@@ -215,6 +229,26 @@ pub fn process_create_device_interface(
...
+                        topo_type,
+                        AccountType::Topology,
+                        "Invalid Topology Account Type"
+                    );
+                    let topo_segment_idx = allocate_id(segment_routing_ids_ext)?;
+                    flex_algo_node_segments.push(FlexAlgoNodeSegment {
+                        topology: *topo_account.key,
+                        node_segment_idx: topo_segment_idx,
+                    });
+                }
```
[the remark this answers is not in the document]
elitegreg: Will fix in another PR

smartcontract/sdk/rs/src/commands/device/interface/create.rs
```diff
@@ -62,6 +66,16 @@ impl CreateDeviceInterfaceCommand {
...
             accounts.push(AccountMeta::new(device_tunnel_block_ext, false));
             accounts.push(AccountMeta::new(segment_routing_ids_ext, false));
+
+            // For Vpnv4 loopbacks, append topology PDAs so the onchain program
+            // can allocate FlexAlgoNodeSegment entries atomically.
+            if self.loopback_type == LoopbackType::Vpnv4 {
+                for name in &self.topology_names {
+                    let (topology_pda, _) = get_topology_pda(&client.get_program_id(), name);
+                    accounts.push(AccountMeta::new_readonly(topology_pda, false));
+                    topology_count += 1;
```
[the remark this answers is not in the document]
elitegreg: Will fix in another PR

smartcontract/cli/src/device/interface/update.rs
```diff
@@ -163,6 +171,16 @@ impl UpdateDeviceInterfaceCliCommand {
             .transpose()
             .map_err(|e| eyre::eyre!("Invalid status: {}", e))?;
 
+        // clap's `num_args = 0..` produces Some(vec!["", ...]) for `--topologies ""`,
+        // so strip empty/whitespace-only entries.
+        let topology_names = self.topologies.as_ref().map(|names| {
+            names
+                .iter()
+                .filter(|n| !n.trim().is_empty())
+                .map(|n| n.trim().to_string())
```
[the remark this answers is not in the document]
elitegreg: Will fix in another PR

smartcontract/programs/doublezero-serviceability/src/processors/device/interface/create.rs
bgm-malbeclabs: ok cool

smartcontract/sdk/rs/src/commands/device/interface/update.rs
```diff
@@ -50,18 +54,29 @@ impl UpdateDeviceInterfaceCommand {
...
         }
 
+        let mut topology_count: u8 = 0;
+        if update_topologies {
+            for name in self.topology_names.as_ref().unwrap() {
+                let (topology_pda, _) = get_topology_pda(&client.get_program_id(), name);
+                accounts.push(AccountMeta::new_readonly(topology_pda, false));
+                topology_count += 1;
+            }
+        }
```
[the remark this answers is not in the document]
elitegreg: Will fix in another PR

PR #3651, opened 2026-05-02
sdk/serviceability: make enum String() methods defensive against unknown values

### LINKED ISSUE #3650: SDK: make Go serviceability enum String() methods defensive against unknown values
## Summary

The Go SDK's enum `String()` methods in `smartcontract/sdk/go/serviceability/state.go` use the `[...]string{...}[i]` indexing pattern, which panics on out-of-range values. When an on-chain layout change misaligns deserialization, the SDK panics deep inside `String()` instead of surfacing a clear "unknown enum value N" log line. This makes incidents from SDK/on-chain version drift much harder to diagnose than they need to be.

## Background incident

On 2026-05-01, lake-indexer's `RefreshServiceability` panicked for ~22h on devnet because the indexer was on an older SDK that didn't know about RFC-18 flex-algo's `Interface` V3 layout. After the activator ran `MigrateDeviceInterfaces`, the indexer started reading multi-interface devices at the wrong offset and `InterfaceType.String()` panicked with `runtime error: index out of range [116] with length 3`. With a defensive stringer, the symptom would have been a single `InterfaceType: unknown(116)` log line and the root cause would have been obvious in minutes, not 22 hours.

Lake-side immediate fix: malbeclabs/lake#556. Lake-side regression-test follow-up: malbeclabs/lake#557.

## Proposed change

Make every `String()` method that maps an enum to a finite string array defensive:

```go
func (i InterfaceType) String() string {
    s := [...]string{"invalid", "loopback", "physical"}
    if int(i) >= len(s) {
        return fmt.Sprintf("unknown(%d)", i)
    }
    return s[i]
}
```

A sweep for the `[...]string{...}[i]` pattern in `state.go` should turn up the full set. Likely candidates from a quick read:

- `InterfaceStatus`
- `InterfaceType`
- `LoopbackType`
- `InterfaceCYOA`
- `LinkStatus`, `LinkLinkType`, `LinkDesiredStatus`
- `DeviceStatus`, `UserStatus`, `ContributorStatus`
- `RoutingMode`
- `BGPStatus`

Worth a sweep of other SDK packages (telemetry, etc.) for the same pattern.

## Why this matters

The deserializer is intentionally lenient — `ByteReader` returns zero values past EOF rather than panicking — but a single panic-prone stringer downstream defeats the lenient design at the first call site that prints or logs the value. Defensive stringers preserve the lenient deserialization model end-to-end.

ben-dz:
## Summary

* Add bounds checks to 10 array-indexed `String()` methods in the Go serviceability SDK so out-of-range enum values return `"unknown(N)"` instead of panicking
* Prevents SDK crashes when onchain layout changes (e.g., RFC-18 flex-algo migration) cause deserialization offset drift — turns a panic into a diagnosable log line
* Fixes #3650

Affected methods: `DeviceDeviceType`, `InterfaceStatus`, `InterfaceType`, `LoopbackType`, `InterfaceCYOA`, `InterfaceDIA`, `RoutingMode`, `TenantPaymentStatus`, `UserUserType`, `CyoaType`.

## Testing Verification

* Added `TestEnumStringDefensive` with 20 subtests covering all 10 enum types — validates both max valid boundary values and out-of-range (255) behavior
* All existing tests pass with no regressions (serviceability package: 0.274s)
* Verified the SDK package builds cleanly

### DIFF
--- a/smartcontract/sdk/go/serviceability/state.go
+++ b/smartcontract/sdk/go/serviceability/state.go
@@ -149,11 +149,15 @@ const (
 )
 
 func (d DeviceDeviceType) String() string {
-	return [...]string{
+	s := [...]string{
 		"hybrid",
 		"transit",
 		"edge",
-	}[d]
+	}
+	if int(d) >= len(s) {
+		return fmt.Sprintf("unknown(%d)", d)
+	}
+	return s[d]
 }
 
 type DeviceStatus uint8
@@ -266,7 +270,7 @@ const (
 )
 
 func (i InterfaceStatus) String() string {
-	return [...]string{
+	s := [...]string{
 		"invalid",
 		"unmanaged",
 		"pending",
@@ -274,7 +278,11 @@ func (i InterfaceStatus) String() string {
 		"deleting",
 		"rejecting",
 		"unlinked",
-	}[i]
+	}
+	if int(i) >= len(s) {
+		return fmt.Sprintf("unknown(%d)", i)
+	}
+	return s[i]
 }
 
 func (i InterfaceStatus) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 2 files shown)


ben-dz:
Panic behavior was better.

PR #3654, opened 2026-05-02
feat: move dns lookup to run independently

### LINKED ISSUE #3544: geolocation: move geoprobe-agent DNS lookups out of measurement loop
PR https://github.com/malbeclabs/doublezero/pull/3534 added support for targets being defined by DNS hostname in addition to IP address, but this adds potential for non-deterministic delays to the main measurement loop. Each DNS timeout can add several seconds to the loop. 

Proposal: Move DNS lookups into a separate loop, which runs on startup and every 5 minutes. If the measurement loop runs before one of the DNS names has been resolved to an IP address, that target is skipped. 

Related: How can we surface DNS resolution errors to users?

0xzrf:
## Summary of Changes
* Describe what changed in the PR

   Introduced a `DeliveryDNSRefresher` that runs DNS for result-destination `host:port` strings outside the measurement loop: it refreshes on a 5-minute ticker, once at startup, and whenever the desired set of destinations changes (after outbound or ICMP target discovery updates). The measurement/send path now uses `DNSCache.LookupDeliveryUDPAddr`, which never performs `LookupHost` for hostnames—it only returns an address if the cache is already populated and still valid; literal public IPs are still resolved synchronously with the existing `ValidateScope()` checks. `sendCompositeOffsets` uses `deliveryDNS.Lookup` instead of `dnsCache.Resolve`, so slow or timing DNS no longer runs inside the RTT/measurement cycle. If a hostname is not ready yet, that target is skipped for that cycle with a warning log.

* Explain why the change is necessary
    Due to:  https://github.com/malbeclabs/doublezero/issues/3544
* Is there supporting documentation or external resources that explain the change? No
* Is a CHANGELOG.md update needed? No

## Testing Verification
* `go test` on `controlplane/telemetry/internal/geoprobe` for `DNSCache_*`, `DeliveryDNS_*`, and `LookupDelivery*` tests: **pass** (covers startup refresh, coalesced refresh after `SetDesiredHostPorts`, literal-IP lookup without waiting, domain miss before refresh, cache hit/miss/TTL, private IP rejection, invalid `host:port`).
* `GOOS=linux go build ./controlplane/telemetry/cmd/geoprobe-agent/`: succeeds (agent is `//go:build linux`).
* Note: `go test ./controlplane/telemetry/cmd/geoprobe-agent/...` may fail on macOS because main_test.go is not tagged linux while main.go is—run agent package tests on Linux / CI for a full green check.

### DIFF
--- a/controlplane/telemetry/cmd/geoprobe-agent/main.go
+++ b/controlplane/telemetry/cmd/geoprobe-agent/main.go
@@ -30,16 +30,17 @@ import (
 )
 
 const (
-	defaultTWAMPListenPort       = 8925
-	defaultSignedTWAMPListenPort = 8924
-	defaultUDPListenPort         = 8923
-	defaultProbeInterval         = 30 * time.Second
-	defaultTWAMPSenderTimeout    = 1 * time.Second
-	defaultTWAMPReflectorTimeout = 1 * time.Second
-	defaultMaxOffsetAge          = 1 * time.Hour
-	defaultEvictionInterval      = 30 * time.Minute
-	defaultVerifyInterval        = 29 * time.Second
-	discoveryInterval            = 60 * time.Second
+	defaultTWAMPListenPort            = 8925
+	defaultSignedTWAMPListenPort      = 8924
+	defaultUDPListenPort              = 8923
+	defaultProbeInterval              = 30 * time.Second
+	defaultTWAMPSenderTimeout         = 1 * time.Second
+	defaultTWAMPReflectorTimeout      = 1 * time.Second
+	defaultMaxOffsetAge               = 1 * time.Hour
+	defaultEvictionInterval           = 30 * time.Minute
+	defaultVerifyInterval             = 29 * time.Second
+	discoveryInterval                 = 60 * time.Second
+	defaultDeliveryDNSRefreshInterval = 5 * time.Minute
 )
 
 var (
@@ -612,6 +613,9 @@ func main() {
 
 	// Run main measurement loop. This runs regardless of whether trusted parents
 	// are configured at startup, since they may be added dynamically at runtime.
+	deliveryDNS := geoprobe.NewDeliveryDNSRefresher(log, defaultDeliveryDNSRefreshInterval)
+	go deliveryDNS.Start(ctx, defaultDeliveryDNSRefreshInterval)
+
 	go func() {
 		ml := &measurementLoop{
 			ctx:                ctx,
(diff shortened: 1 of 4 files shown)


controlplane/telemetry/cmd/geoprobe-agent/main.go
```diff
@@ -612,6 +613,9 @@ func main() {
 
 	// Run main measurement loop. This runs regardless of whether trusted parents
 	// are configured at startup, since they may be added dynamically at runtime.
+	deliveryDNS := geoprobe.NewDeliveryDNSRefresher(log, defaultDeliveryDNSRefreshInterval)
+	go deliveryDNS.Start(ctx, defaultDeliveryDNSRefreshInterval)
```
ben-dz, line 618: The TTL and the refresh interval are set to the same value, so they will beat against each other, with expiring TTLs before refreshes happen. Ideally TTL is ~2.5x Refresh interval to be sure there are multiple opportunities for refresh before TTL expires.

controlplane/telemetry/internal/geoprobe/delivery_dns_refresher.go
```diff
@@ -0,0 +1,98 @@
...
+	return &DeliveryDNSRefresher{
+		cache:   NewDNSCache(ttl),
+		log:     log,
+		desired: make(map[string]struct{}),
+		trigger: make(chan struct{}, 1),
+	}
+}
+
+// Cache returns the underlying DNS cache for tests.
+func (r *DeliveryDNSRefresher) Cache() *DNSCache {
```
ben-dz: This is only used in the test.go, so isn't necessary– the test caller has access to it directly.

ben-dz (changes requested):
Other than these two changes, I think this looks good.

0xzrf:
@ben-dz apologies for the late reply

Will make the suggested changes

vihu:
@0xzrf Can you please rebase this branch?

0xzrf:
@vihu done

ben-dz:
Hey @0xzrf This looks good, and we now are able to run CI on it, so we will be able to merge it. I'm ready to approve it, but our branch protection policy requires that your commits are signed. if you already have commit signing set up:

`git rebase --exec 'git commit --amend --no-edit -S' HEAD~2 && git push --force-with-lease` should do it.

0xzrf:
@ben-dz I've made the changes, rebasing with `main` branch ended up being a non-verified commit, so had to make a last commit that was verified(apologies for flooding the PR with unrelated commits)

ben-dz:
The branch protection requires all the commits to be signed, but you should be able to run that rebase command (HEAD~3 instead of 2), and force-push over the non-signed commits with 3 signed ones.

0xzrf:
@ben-dz done

ben-dz (approved):

PR #3670, opened 2026-05-05
serviceability: migrate processors to new_interfaces

### LINKED ISSUE #3658: serviceability: migrate processors to new_interfaces
Part of the forward-compatible Device interfaces refactor.
Design: https://gist.github.com/elitegreg/e1d27c97034656a980fa8d7628ae16c5

Depends on #3665 (Device with new_interfaces vec).

## Scope

Move the smart contract processors off `into_current_version()` / `into_v3()` / per-variant matching and onto direct field access on `NewInterface` via `device.new_interfaces`. The Device serializer keeps the legacy vec in sync, so processors only mutate `new_interfaces`.

## Tasks

- [ ] `processors/device/interface/{create,update,activate,delete,reject,remove,unlink}.rs`: read/mutate `device.new_interfaces` directly.
- [ ] `processors/link/{accept,activate,closeaccount,create,delete,update}.rs`: same.
- [ ] `processors/topology/backfill.rs`: mutate `device.new_interfaces[i].flex_algo_node_segments` directly. Note that segments are dropped on serialize into the V2-projected legacy vec; the new vec is the source of truth.
- [ ] Drop calls to `into_current_version()` / `into_v3()` in these processors.
- [ ] Verify `device.interfaces` is no longer touched directly anywhere in `processors/` after this issue lands.

## Tests

- [ ] All existing processor tests must pass.
- [ ] Add a backfill test that round-trips `flex_algo_node_segments` through the new vec.

## Out of scope

- Renaming `NewInterface` / `new_interfaces` to canonical names — separate issue.
- Removing `migrate_interfaces` processor — separate issue.

elitegreg:
## Summary

- Migrate `device/interface/{create,update,activate,delete,reject,remove,unlink}`, `link/{accept,activate,closeaccount,create,delete,update}`, and `topology/backfill` processors to read and mutate `Device::new_interfaces` directly, dropping all `into_current_version()` / `into_v3()` / `find_interface_legacy()` calls from `processors/`.
- `Device::push_interface` now takes a `NewInterface` directly, avoiding a fallible `TryFrom<&InterfaceV2>` at the call site.
- `BackfillTopology` no longer mirrors `flex_algo_node_segments` into the legacy in-memory `interfaces` vec — segments live only in `new_interfaces` and are intentionally dropped on the V2-projected on-disk legacy slot. The custom `Device::BorshSerialize` continues to project `new_interfaces` → V2 for byte-compatibility with older readers.
- `Device::find_interface_legacy` is retained as scaffolding for CLI callers; no `processors/` code calls it any more. CLI migration will follow in a separate change.

Part of the forward-compatible Device interfaces refactor — depends on #3666 and #3667. Closes #3658.

## Testing Verification

- New unit test `test_flex_algo_segments_roundtrip_through_new_interfaces` (in `state/device.rs`) seeds a Vpnv4 loopback with a populated `flex_algo_node_segments`, round-trips through borsh, and asserts segments survive in `new_interfaces` while the V2-projected legacy slot drops them.
- Existing `test_topology_backfill_populates_vpnv4_loopbacks` integration test continues to pass — it already asserted on `device.new_interfaces[0].flex_algo_node_segments`, so removing the legacy in-memory mirror is invisible to it.
- `make rust-fmt`, `make rust-lint`, and `make rust-test` all pass on this branch.
- Sanity grep over `processors/` for legacy paths returns only the out-of-scope `migrate_interfaces.rs`.


smartcontract/programs/doublezero-serviceability/src/processors/device/interface/create.rs
```diff
@@ -239,7 +242,10 @@ pub fn process_create_device_interface(
         ip_net,
         node_segment_idx,
         user_tunnel_endpoint: value.user_tunnel_endpoint,
-    })?;
+        flex_algo_node_segments: vec![],
+    };
+    new_iface.size = new_iface.compute_on_disk_size()?;
```
elitegreg: I'm not sure this is strictly necessary as the serializer calculates it. Possibly remove?

elitegreg: Good catch — confirmed the `NewInterface` `BorshSerialize` derives the size from the actual body bytes and ignores the struct field. Dropped the pre-stamp here and in the new round-trip test in 03cf4a91.

ben-dz (approved):

PR #3678, opened 2026-05-06
device-health-oracle: detect link impairment from monitoring data

### LINKED ISSUE #2652: device-health-oracle: Add activation criterion: a pending link has <burn-in slots> of data in the DoubleZero ledger telemetry program with zero loss in both directions

nikw9944:
## Summary of Changes

* Add bidirectional `LinkHealth` transitions to the device-health-oracle. The DHO previously had zero `LinkCriterion`s configured — links auto-advanced from Pending to ReadyForService and stayed there forever, even when ISIS was down or every packet was being dropped. This PR adds the demotion path (and a slow recovery path back).
* `ReadyForService → Impaired` triggers on the most recent `link_rollup_5m` bucket (fast detection, single bucket). `Impaired → ReadyForService` requires every distinct bucket in the recovery window to be clean (slow recovery, anti-flap).
* Recovery window reuses the existing `--drained-slot-count` (~30 min default) — no new burn-in flag introduced. New `--link-loss-threshold` flag (default 5.0) controls the per-direction loss cutoff.
* `LinkHealth` is reported as a signal only; the serviceability program does not gate `link.status` on it.
* Fixes #2652

### Implementation notes

* Stale-data floor: a "latest" bucket older than 15 min (3× rollup cadence) is treated as no data. Without this, a frozen telemetry pipeline at the moment of an ISIS flap would keep the link `Impaired` forever even after the link recovered.
* Recovery query deduplicates late-arriving rows by `argMax(..., ingested_at)` per bucket, so a corrected re-write doesn't keep a link `Impaired` after every distinct bucket reads as clean.
* `--link-loss-threshold` is validated at startup (must be finite and in `[0, 100]`) — a misconfiguration would otherwise produce an onchain write storm.
* `MetricCriterionResults` is now incremented for link criteria too (parity with device criteria) so impairment/recovery rates are graphable.
* Backwards compatible: deployments without ClickHouse continue to behave exactly as before — no demotion, no recovery, no impairment criterion is wired up.

## Testing Verification

* New unit tests cover all six transition cases (Pending/RFS/Impaired × pass/fail), the impairment-mode and recovery-mode criterion paths, the stale-bucket floor, threshold boundaries, no-data handling in both modes, ClickHouse error propagation, and the `LinkBurnIn` helper.
* New ClickHouse tests assert the actual SQL shape: argMax-per-bucket dedupe, `provisioning = false` filter, double-quoted database name, and `?` bind parameters.
* Manually validated both queries against mainnet `link_rollup_5m` data via the DoubleZero MCP — known impaired link `DzHDqj3cdi77eMLWKemdhfr6YZJeHHGxuysvAdekniC` returns `bad=5/total=5` over the last 30 min and the latest bucket reads `isis_down=true, a_loss=100, z_loss=100`, matching what the issue reports.
* `make go-test`, `golangci-lint`, and `go vet` all clean.

### DIFF
--- a/controlplane/device-health-oracle/cmd/device-health-oracle/main.go
+++ b/controlplane/device-health-oracle/cmd/device-health-oracle/main.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"log/slog"
+	"math"
 	"net"
 	"net/http"
 	"os"
@@ -56,6 +57,14 @@ func main() {
 	if *showVersion {
 		fmt.Printf("version: %s, commit: %s, date: %s\n", version, commit, date)
 		os.Exit(0)
+	}
+
+	// Reject obviously broken thresholds early. A negative or NaN threshold would
+	// silently flag every link as impaired (or never), triggering an onchain
+	// write storm or hiding real failures — better to fail fast at startup.
+	if math.IsNaN(*linkLossThreshold) || math.IsInf(*linkLossThreshold, 0) || *linkLossThreshold < 0 || *linkLossThreshold > 100 {
+		fmt.Fprintf(os.Stderr, "invalid --link-loss-threshold %v: must be in [0, 100]\n", *linkLossThreshold)
+		os.Exit(1)
 	}
 
 	logLevel := slog.LevelInfo
--- a/controlplane/device-health-oracle/cmd/device-health-oracle/main.go
+++ b/controlplane/device-health-oracle/cmd/device-health-oracle/main.go
@@ -44,6 +44,7 @@ var (
 	slackWebhookURL         = flag.String("slack-webhook-url", "", "The Slack webhook URL to send alerts")
 	provisioningSlotCount   = flag.Uint64("provisioning-slot-count", defaultProvisioningSlotCount, "Burn-in slot count for new devices/links (~20 hours at 200000)")
 	drainedSlotCount        = flag.Uint64("drained-slot-count", defaultDrainedSlotCount, "Burn-in slot count for reactivated devices/links (~30 min at 5000)")
+	linkLossThreshold       = flag.Float64("link-loss-threshold", 5.0, "Per-direction packet loss percentage above which a link is considered impaired (link_rollup_5m a_loss_pct/z_loss_pct)")
 	version                 = "dev"
 	commit                  = "none"
 	date                    = "unknown"
@@ -131,6 +132,8 @@ func main() {
 
 	// Initialize ClickHouse-dependent criteria.
 	var deviceCriteria []worker.DeviceCriterion
+	var linkImpairmentCriteria []worker.LinkCriterion
+	var linkRecoveryCriteria []worker.LinkCriterion
 	if chAddr := os.Getenv("CLICKHOUSE_ADDR"); chAddr != "" {
 		chDB := os.Getenv("CLICKHOUSE_DB")
 		if chDB == "" {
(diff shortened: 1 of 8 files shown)


controlplane/device-health-oracle/internal/worker/link_health.go
```diff
@@ -0,0 +1,156 @@
...
+	return c.checkImpairment(ctx, link, pubkey)
+}
+
+func (c *LinkHealthCriterion) checkImpairment(ctx context.Context, link serviceability.Link, pubkey string) (bool, string) {
+	r, found, err := c.checker.LinkHealthRecent(ctx, pubkey)
+	if err != nil {
+		c.log.Error("Failed to query link health recent",
+			"link", pubkey, "code", link.Code, "error", err)
+		return false, fmt.Sprintf("clickhouse query failed: %v", err)
+	}
```
ben-dz, line 82: **[HIGH]** A query error here fails the impairment criterion, and in the ReadyForService branch of `LinkHealthEvaluator.Evaluate` a failed criterion demotes the link to Impaired. A transient ClickHouse outage (network blip, CH restart, query timeout — anything after the successful startup ping) would therefore demote every RFS link onchain in a single tick, which is the write storm the `--link-loss-threshold` validation guards against. Once ClickHouse recovers, the recovery check immediately passes (the window's buckets are genuinely clean), so you also get a matching wave of RFS re-writes plus false operator signal in between.

Errors here should be treated like the `!found` case below: hold current health (`return true, ""`), keep the error log, and ideally bump a distinct metric so outages don't show up as impairment in `MetricCriterionResults{link_health_impairment, fail}`.

(The mirrored error path in `checkRecovery` is correct as-is — there a fail means "stay Impaired", which is the right behavior on error.)

controlplane/device-health-oracle/internal/worker/link_health_test.go
```diff
@@ -0,0 +1,240 @@
...
+				},
+			}
+			c := NewLinkHealthCriterion(LinkHealthModeImpairment, checker, 5.0, testLogger())
+			passed, _ := c.Check(context.Background(), serviceability.Link{})
+			assert.Equal(t, tt.expected, passed)
+		})
+	}
+}
+
+func TestLinkHealthCriterion_Impairment_QueryError_Fails(t *testing.T) {
```
ben-dz, line 118: **[HIGH — companion to the `checkImpairment` error-path comment]** `TestLinkHealthCriterion_Impairment_QueryError_Fails` asserts the demote-on-error behavior flagged in link_health.go — if the error path changes to hold health, this test should assert `passed == true` instead.

controlplane/device-health-oracle/internal/worker/link_health.go
```diff
@@ -0,0 +1,156 @@
...
+
+	r, found, err := c.checker.LinkHealthWindowAllClean(ctx, pubkey, start, now, c.lossThreshold)
+	if err != nil {
+		c.log.Error("Failed to query link health recovery window",
+			"link", pubkey, "code", link.Code, "error", err)
+		return false, fmt.Sprintf("clickhouse query failed: %v", err)
+	}
+	if !found {
+		return false, "no rollup data in recovery window"
+	}
```
ben-dz, line 140: **[MEDIUM]** The recovery gate requires every *present* bucket in the window to be clean, but doesn't require the window to be covered or recent:

1. **Frozen pipeline:** if rollups stop flowing after an impairment, the dirty buckets age out of the trailing `[DrainedStart, Now]` window while older clean buckets remain → the link recovers on stale data, and then *stays* RFS because `checkImpairment` treats the stale latest bucket as no-data. That contradicts the "neither demote nor recover on stale data" intent documented on `LinkHealthRecentResult`.
2. **Sparse ingestion:** a single clean bucket in the 30-minute window satisfies "all clean", so the anti-flap dwell can effectively shrink to one bucket.

`expectedMinutes` is already computed here; consider requiring approximate coverage (e.g. `Total >= expectedMinutes/5 - 1`) and/or having the window query also return `max(bucket_ts)` and applying the same `linkHealthRecentMaxAge` floor to it.

controlplane/device-health-oracle/cmd/device-health-oracle/main.go
```diff
@@ -164,6 +180,8 @@ func main() {
 	}
 	linkEvaluator := &worker.LinkHealthEvaluator{
 		ReadyForServiceCriteria: nil,
```
ben-dz, line 182: **[LOW/suggestion]** `ReadyForServiceCriteria` stays nil, so Pending/Unknown links vacuously promote to RFS even when the latest bucket already shows `isis_down` or 100% loss — and then demote one tick later (two onchain writes plus a misleading intermediate RFS). Since the impairment criterion already passes on no-data, wiring `linkImpairmentCriteria` in here would block promotion of known-impaired links without changing behavior for links with no telemetry. The issue spec calls the current promotion path "existing behavior", so fine to defer to a follow-up.

ben-dz (changes requested):
Solid implementation that matches #2652. What looks good:

- Evaluator state machine is clean and well-documented; the no-criteria guards preserve existing behavior for deployments without ClickHouse.
- `argMax(..., ingested_at)` dedupe per bucket is the right call for late-arriving rewrites, and the SQL-shape tests actually pin it.
- Stale-bucket floor on the impairment side is well-reasoned (15 min = 3× cadence) and tested with an adversarial fixture (stale bucket showing 100% loss must not demote).
- Startup validation of `--link-loss-threshold` (NaN/Inf/range) is a nice touch.
- Test coverage is genuinely thorough: all six transition cases, threshold boundaries (at-threshold passes, above fails), both no-data modes, bind-arg ordering, db-name quoting.

Main ask is the ClickHouse error-handling comment on `checkImpairment` (one-line fix plus a test flip); the recovery freshness/coverage comment is worth addressing or explicitly accepting; the other two are minor.

nikw9944:
Superseded by https://github.com/malbeclabs/doublezero/pull/4272

PR #3682, opened 2026-05-07
geolocation: add cross-language borsh fixture tests

### LINKED ISSUE #3306: geolocation: add cross-language fixture tests for Borsh deserialization
## Problem

The geolocation Go SDK tests only verify round-trip consistency within Go itself. They do not prove the Go Borsh deserialization matches the actual Rust Borsh output. If there's a subtle encoding difference (e.g., `Ipv4Addr` serialization, `String` length prefix handling, enum discriminant encoding), the Go tests would pass but deserialization of real onchain data would fail.

The serviceability and telemetry SDKs both have fixture-based cross-language tests that catch this class of bug. The geolocation SDK should follow the same pattern.

## What to do

Add a Rust fixture generator and Go fixture tests following the established pattern:

1. **Rust generator** (`sdk/geolocation/testdata/fixtures/generate-fixtures/src/main.rs`):
   - Serialize `GeolocationProgramConfig`, `GeoProbe`, and `GeolocationUser` (with targets) using Borsh
   - Write paired `.bin` + `.json` files (binary data + field metadata)

2. **Go fixture test** (`sdk/geolocation/go/fixture_test.go`):
   - Load `.bin`/`.json` pairs
   - Deserialize binary using Go SDK
   - Validate each field against the JSON metadata

3. Wire into `make generate-fixtures` and `make sdk-test`

### Reference implementations
- `sdk/serviceability/testdata/fixtures/generate-fixtures/src/main.rs`
- `sdk/serviceability/go/fixture_test.go`

## Context

Identified during architecture review of #3305. Tracked separately so the read SDK can merge without blocking on this.

0xzrf:
## Summary of Changes
* Added **cross-language fixture coverage** for the **geolocation Go SDK** by introducing a Rust fixture generator and Go fixture tests.
* The Rust generator (`sdk/geolocation/testdata/fixtures/generate-fixtures`) serializes real onchain Rust structs via **Borsh** and writes paired **`.bin` + `.json`** fixtures for:
  - `GeolocationProgramConfig`
  - `GeoProbe`
  - `GeolocationUser` (including targets and `result_destination`)
* Added Go fixture tests (`sdk/geolocation/go/fixture_test.go`) that load those fixtures, deserialize using the Go SDK, and validate **field-by-field** values against the JSON sidecars.
* Wired the new generator and tests into repo workflows:
  - `make generate-fixtures` now generates geolocation fixtures
  - `make sdk-test` now runs `go test ./sdk/geolocation/go/...`
  - excluded the generator crate from the root Cargo workspace to match existing fixture-generator conventions
* **Why this is necessary:** Go-only round-trip tests can miss subtle Rust↔Go Borsh incompatibilities; fixture tests ensure the Go SDK can deserialize **authoritative Rust-produced bytes**, preventing regressions and onchain decoding failures.
* **Metrics exposed:** None.
* **Supporting docs / external resources:** Issue: [#3306](https://github.com/malbeclabs/doublezero/issues/3306).
* **CHANGELOG.md update needed:** No.

## Testing Verification
* `make generate-fixtures`
* `go test ./sdk/geolocation/go/...`
* `make sdk-test`


controlplane/telemetry/cmd/geoprobe-agent/main.go
```diff
@@ -868,7 +868,7 @@ func (ml *measurementLoop) runCycle() {
 
 	rttData := make(map[geoprobe.ProbeAddress]uint64)
 
-	if len(ml.targets) > 0 {
+	if len(ml.targets) > 0
```
ben-dz: Missing opening brace: `if len(ml.targets) > 0` should be `if len(ml.targets) > 0 {`. `go build ./controlplane/telemetry/cmd/geoprobe-agent/` fails with `main.go:872:21: syntax error: cannot use twampResults, err := ... as value`. This breaks `make build`/`make ci` and is an accidental edit unrelated to the fixture tests — the base branch has the brace. Verified it does not mask a behavioral change (base block is byte-identical except the brace). Restore the `{`, or drop this file from the PR.

Makefile
```diff
@@ -136,6 +136,7 @@ rust-program-accounts-compat:
 .PHONY: sdk-test
 sdk-test:
 	go test ./sdk/borsh-incremental/go/...
+	go test ./sdk/geolocation/go/...
```
ben-dz, line 143: This adds `go test ./sdk/geolocation/go/...` to `sdk-test`, but the generated `.bin`/`.json` fixtures are not committed (git ls-tree shows only the generate-fixtures/ subdir). This is the established repo pattern left incomplete: geolocation's fixture_test.go is a near-verbatim copy of sdk/revdist/go/fixture_test.go, which runs unconditionally and reads the committed fixtures in sdk/revdist/testdata/fixtures/ (serviceability/telemetry commit theirs too, incl. device_legacy/device_future_version). fixture_test.go's loadFixture t.Fatalf's on missing files, and the CI sdk-test job (.github/workflows/sdk.yml) has no Rust toolchain and runs no `make generate-fixtures`. So the three fixture tests fail on every fresh checkout/CI, breaking the whole sdk-test gate. Fix: run `cd sdk/geolocation/testdata/fixtures/generate-fixtures && cargo run` and commit the six generated files. Prefer committing over build-in-CI: frozen golden bytes are reviewable, so an unintended Rust wire change shows up as a fixture diff; regenerating every run would mask it.

sdk/geolocation/go/fixture_test.go
```diff
@@ -0,0 +1,214 @@
...
+	t.Helper()
+	for _, f := range expected {
+		val, ok := got[f.Name]
+		if !ok {
+			t.Errorf("field %s: not found in Go struct map", f.Name)
+			continue
+		}
+		assertField(t, f, val)
+	}
+}
```
ben-dz, line 152: assertFields is one-directional: it errors only when a JSON field is absent from the Go `got` map. A struct field present in neither the JSON meta nor the `got` map is silently unasserted, so a future added field on GeolocationUser/GeoProbe would not fail this test. The FixtureMeta also has no struct_size (unlike revdist) and the test never checks the decoder drained the whole .bin. Consider asserting `len(meta.Fields) == len(got)` and/or pinning total byte length so drift/early-stop is caught.

sdk/geolocation/go/fixture_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+func assertField(t *testing.T, f fieldValue, got any) {
+	t.Helper()
+
+	switch f.Type {
+	case "u8":
+		want, _ := strconv.ParseUint(f.Value, 10, 8)
```
ben-dz: `want, _ := strconv.ParseUint(...)` discards the parse error here (and at the u16/u32/u64 sites). A malformed numeric value in the hand-authored JSON silently becomes 0, and if the decoded Go value is also 0 the assertion passes falsely. Since the hand-typed JSON is the drift-prone half of this test, prefer t.Fatalf on a parse error.

ben-dz (dismissed):
The fixture-test design is sound, but the branch does not build or pass CI.

**C1 (Critical):** `controlplane/telemetry/cmd/geoprobe-agent/main.go:871` is missing the opening `{` on `if len(ml.targets) > 0`, so `go build` fails. Accidental, out-of-scope edit to a production binary; breaks `make build`/`make ci`. Does not mask a behavioral change (base is byte-identical except the brace). Restore `{` or drop the file.

**C2 (Critical):** the generated `.bin`/`.json` fixtures are not committed (only the generate-fixtures crate is), yet the PR wires `go test ./sdk/geolocation/go/...` into `make sdk-test`, and `loadFixture` t.Fatalf's on missing files. This is the established repo pattern left incomplete: geolocation's `fixture_test.go` is a near-verbatim copy of `sdk/revdist/go/fixture_test.go`, which runs unconditionally and consumes the committed fixtures in `sdk/revdist/testdata/fixtures/` (serviceability and telemetry commit theirs too). The CI `sdk-test` job has no Rust toolchain and runs no `make generate-fixtures`, so all three fixture tests fail and take down the SDK test gate. Run the generator and commit the six files. Committing (vs build-in-CI) is correct: golden bytes are frozen and reviewable, so an unintended Rust wire change shows up as a fixture diff — regenerating every CI run would move producer and expected bytes together and mask it.

Medium/Low: no cross-language coverage of the incremental/old-account trailing-field fallback (M1); `assertFields` doesn't verify total field coverage or full buffer drain (M2); no CI drift-check keeps committed fixtures fresh — a repo-wide gap that is this PR's staleness risk, best fixed with `make generate-fixtures` + `git diff --exit-code` in a Rust-enabled job (M3); discarded `strconv.ParseUint` errors (L1); no generate-fixtures ordering dep (L2); missing generator `.gitignore` (L3). Both Critical items block merge.

- M1 (Medium) — sdk/geolocation/go/state.go:150-153 and 454-457: GeoProbe.TargetUpdateCount and GeolocationUser.ResultDestination are #[incremental] appended fields with Go fallback logic (decode error -> default), but the fixtures only cover the all-fields-present shape, so the fallback paths have zero cross-language coverage. The serviceability generator has direct precedent (device_legacy / device_future_version). Add truncated geo_probe_legacy / geolocation_user_legacy fixtures (fast-follow acceptable).
- M3 (Medium, repo-wide follow-up) — Nothing keeps the committed fixtures in sync with the Rust structs: `make generate-fixtures` runs in no workflow and there is no drift guard. If someone edits a Rust struct and forgets to regenerate, the committed fixture goes stale and the Go/py/ts tests pass against old bytes — the staleness risk raised in review. This already affects revdist/serviceability/telemetry, so it is a repo-wide follow-up rather than a geolocation blocker, but this PR is a natural moment to add a Rust-enabled CI job running `make generate-fixtures` then `git diff --exit-code sdk/*/testdata/fixtures`. That yields freshness AND reviewable goldens — better than regenerating every run.
- L2 (Low) — Makefile:138-139: `make sdk-test` has no ordering dependency on `generate-fixtures`, so a Rust struct change silently tests against stale fixtures until someone regenerates by hand. Matches sibling-SDK convention, so acceptable — noted to confirm intent.
- L3 (Low) — sdk/geolocation/testdata/fixtures/generate-fixtures/ is missing a `.gitignore` (with `target/`) that sibling fixture generators commit; a stray `cargo run` build dir could be accidentally staged.
- No CHANGELOG entry required — this is test-only tooling with no runtime/SDK-consumer behavior change, consistent with prior fixture-test PRs.

0xzrf:
@ben-dz 
Agreed on M1 — current fixtures only cover the full wire shape, so the Go fallbacks for TargetUpdateCount / ResultDestination aren’t exercised cross-language. 

I’ll follow up with truncated geo_probe_legacy / geolocation_user_legacy fixtures (same pattern as serviceability’s device_legacy) if that's okay

ben-dz:
This looks good to me, but the 2577cb0 commit needs to be signed for branch protection to let me merge it. 
`git rebase --force-rebase --gpg-sign origin/main && git push --force-with-lease`

will couple a rebase onto the latest `main` with signing.

0xzrf:
@ben-dz done
Could you verify the changes?
Thanks

sdk/geolocation/testdata/fixtures/generate-fixtures/src/main.rs
```diff
@@ -0,0 +1,386 @@
...
+}
+
+fn write_fixture(dir: &Path, name: &str, data: &[u8], meta: &FixtureMeta) {
+    fs::write(dir.join(format!("{name}.bin")), data).unwrap();
+    let json = serde_json::to_string_pretty(meta).unwrap();
+    fs::write(dir.join(format!("{name}.json")), json).unwrap();
+    println!("wrote {name}.bin ({} bytes) and {name}.json", data.len());
+}
+
+fn main() {
```
nikw9944, line 55: Carryover of prior-round M1: GeoProbe.TargetUpdateCount and GeolocationUser.ResultDestination are borsh-incremental appended fields with Go fallback decode paths (sdk/geolocation/go/state.go:151-153, 454-457), and both decoders tolerate extra trailing bytes — but the fixtures only cover the all-fields-present shape, so the fallback/legacy paths have zero cross-language coverage. Serviceability has direct precedent (device_legacy, device_future_version). Add geo_probe_legacy / geolocation_user_legacy fixtures (serialize, truncate the appended field's bytes) asserting the Go defaults (0, ""), plus optionally a *_future_version fixture with unknown trailing bytes. Fast-follow acceptable per the prior review.

sdk/geolocation/go/fixture_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+	assertFields(t, meta.Fields, map[string]any{
+		"AccountType":          cfg.AccountType,
+		"BumpSeed":             cfg.BumpSeed,
+		"Version":              cfg.Version,
+		"MinCompatibleVersion": cfg.MinCompatibleVersion,
+	})
+}
+
+func TestFixtureGeoProbe(t *testing.T) {
```
nikw9944, line 75: The fixtures are decode-only: the Go Serialize methods are never checked against Rust bytes, so the encode direction can silently drift from the Rust wire format. After the field assertions pass, re-serialize the decoded struct and bytes.Equal it against the fixture bytes — this also doubles as the missing buffer-drain/total-length check from the prior review round.

.github/workflows/sdk.yml
```diff
@@ -46,6 +46,19 @@ jobs:
...
+  # Committed .bin/.json goldens must match Rust Borsh output. Regenerating every
+  # sdk-test run would move producer and expected bytes together and mask drift;
+  # this Rust-enabled job regenerates then fails on any fixture diff.
+  sdk-fixture-drift:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@1.90.0
+      - uses: Swatinem/rust-cache@v2
```
nikw9944, line 61: The generator crates are workspace-excluded with their own target/ dirs, and Swatinem/rust-cache@v2 by default caches only the root workspace target plus ~/.cargo — so this job rebuilds four full Solana dependency trees on every PR (~8 min cold locally vs the 15-min timeout, no path filter). Set rust-cache's `workspaces` input to the four sdk/*/testdata/fixtures/generate-fixtures dirs (or share a cached CARGO_TARGET_DIR), and consider more timeout margin.

sdk/geolocation/go/fixture_test.go
```diff
@@ -0,0 +1,229 @@
...
+		"Target2IP":                 user.Targets[2].IPAddress,
+		"Target2LocationOffsetPort": user.Targets[2].LocationOffsetPort,
+		"Target2TargetPK":           user.Targets[2].TargetPK,
+		"Target2GeoProbePK":         user.Targets[2].GeoProbePK,
+
+		"ResultDestination": user.ResultDestination,
+	})
+}
+
+func assertFields(t *testing.T, expected []fieldValue, got map[string]any) {
```
nikw9944, line 139: assertFields uses t.Errorf (continues past a missing field) but assertField/assertEq use t.Fatalf, so the first value mismatch aborts the test and hides any further mismatches. Optional: use t.Errorf in assertEq so one run reports all field mismatches. Cosmetic — sibling SDK tests share the same shape.

nikw9944 (commented):
All blocking items from the previous review round are resolved at head 37cdac57: the stray geoprobe-agent edit is gone, the six golden fixtures are committed (verified byte-identical to fresh generator output), ParseUint errors are fatal, a .gitignore was added, a field-count check was added, and a new sdk-fixture-drift CI job regenerates and diffs the goldens (verified green today for all four SDKs, so it won't trip on pre-existing sibling drift). go test / go vet pass, and every field of the three Rust structs is covered by the Go assertions. Remaining findings are medium/low: the borsh-incremental fallback paths (GeoProbe.TargetUpdateCount, GeolocationUser.ResultDestination) still have no legacy/truncated fixture coverage (agreed fast-follow), the tests are decode-only (no re-serialize round-trip / buffer-drain check), the drift job's generator builds are uncached against a 15-minute timeout, and check-fixtures won't catch new untracked fixture files. Diff size is dominated by the generated Cargo.lock and test tooling; no production code change.

- sdk.yml has no `permissions:` block, so the new job inherits the default GITHUB_TOKEN scope. Exposure is theoretical (checkout only; pull_request trigger gives fork code a read-only token; PR caches can't overwrite base-branch caches) and the omission is pre-existing across the workflow. Defense-in-depth follow-up: add `permissions: contents: read` at the workflow top level. (low)
- Pre-existing, no action for this PR: GeoProbe.Deserialize validates Code/ParentDevices lengths only after decoding (sdk/geolocation/go/state.go:138-148), relying on the decoder's remaining-bytes checks, whereas GeolocationUser pre-validates the target count against the buffer size. Optionally align GeoProbe with the GeolocationUser pattern in a follow-up. (low)
- Verified as part of review: regenerating all four SDKs' fixtures produces byte-identical output and a clean tree, so the new sdk-fixture-drift job merges green; go test / go vet pass; every Rust struct field is covered by the Go assertions; struct_size omission and JSON conventions match the serviceability/telemetry precedent. The field-count equality check is stronger than revdist's — worth backporting to sdk/revdist/go/fixture_test.go in a follow-up.

0xzrf:
@ben-dz @nikw9944 
I've added the review fixes + signed all the prev commits

lmk if there's anything more to add for this PR
Thanks.

nikw9944 (approved):

nikw9944:
We seem to have a problem with CI - `e2e (shard 1) Expected — Waiting for status to be reported`. Working on a fix.

nikw9944:
Woohoo! Congrats @0xzrf and sorry that took so long!

PR #3688, opened 2026-05-08
smartcontract: address PR #3648 review

elitegreg:
## Summary

Address the four unresolved Copilot review threads on #3648:

- Reject duplicate topology PDAs in `CreateDeviceInterface` so a caller can't allocate two flex-algo SR IDs for the same topology — mirrors the existing dedupe in `UpdateDeviceInterface`.
- Compute `topology_count` via `u8::try_from(topology_names.len())` in the Rust SDK's `CreateDeviceInterfaceCommand` and `UpdateDeviceInterfaceCommand` so passing more than 255 topology names errors out instead of silently wrapping.
- In `UpdateDeviceInterfaceCommand`, return an error when topology reconciliation is requested but the `OnChainAllocation` feature flag is off, instead of silently dropping the request.
- Normalize CLI `--topologies` names with `to_ascii_uppercase` to match the canonical form used to derive topology PDAs.

Stacks on top of #3648 (base `bc/rfc18-devnet-fixes`).

Related RFC: rfcs/rfc-0018-flex-algo.md

## Testing Verification

- New `test_create_device_interface_rejects_duplicate_topology_accounts` in `interface_onchain_allocation_test.rs` exercises the dedupe path on `CreateDeviceInterface`.
- `cargo test -p doublezero-serviceability` — full smartcontract suite passes, including the existing `test_update_topologies_*` cases.
- `cargo test -p doublezero_sdk device::interface` and `cargo test -p doublezero_cli device::interface` cover the SDK / CLI changes.

### DIFF
--- a/smartcontract/cli/src/device/interface/update.rs
+++ b/smartcontract/cli/src/device/interface/update.rs
@@ -172,12 +172,13 @@ impl UpdateDeviceInterfaceCliCommand {
             .map_err(|e| eyre::eyre!("Invalid status: {}", e))?;
 
         // clap's `num_args = 0..` produces Some(vec!["", ...]) for `--topologies ""`,
-        // so strip empty/whitespace-only entries.
+        // so strip empty/whitespace-only entries. Topology PDAs are derived from
+        // the canonical uppercase name, so normalize here too.
         let topology_names = self.topologies.as_ref().map(|names| {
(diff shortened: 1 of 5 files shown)


juan-malbeclabs (approved):
LGTM

Issue #3411, opened 2026-03-30
smartcontract: enforce 9000-byte MTU on links and device interfaces


elitegreg:
## Summary

Change the enforced MTU for links and non-CYOA/non-DIA device interfaces from 2048 to 9000 bytes. CYOA/DIA interfaces remain at 1500. Add onchain validation using the existing `InvalidMtu` error code, and make the CLI `--mtu` flag optional with correct defaults.

## Motivation

The network should use jumbo frames (9000 bytes) by default for all fabric-facing links and interfaces. The previous 2048-byte MTU was a placeholder that does not reflect the intended production configuration. Enforcing this at both the CLI and onchain level prevents misconfigured links and interfaces from being registered.

### Onchain (smartcontract program)
- Validate MTU in link create/update processors: must be 9000, else return `InvalidMtu` (error 46)
- Validate MTU in device interface create/update processors: must be 9000 for non-CYOA/non-DIA interfaces, 1500 for CYOA/DIA interfaces

### CLI
- `doublezero link create wan/dzx`: `--mtu` becomes optional, defaults to 9000, rejects other values
- `doublezero link update`: `--mtu` validates 9000 if provided
- `doublezero device interface create`: defaults to 9000 (non-CYOA/DIA) or 1500 (CYOA/DIA), rejects other values
- `doublezero device interface update`: same validation rules
- Remove `validate_parse_mtu` range validator (no longer meaningful)

### Tests
- Update all test fixtures from 2048/1500 to 9000 for non-CYOA interfaces and links
- Add onchain tests for `InvalidMtu` rejection
- Update test script (`start-test.sh`) MTU values

Issue #3386, opened 2026-03-26
controller: IS-IS adjacency failure caused by duplicate parent interface rendering with inconsistent MTU


elitegreg:
### Summary

The IS-IS adjacency between `nyc-dz001` and `lax-dz001` on `Switch1/1/1.1002` fails to establish on testnet. The root cause is two interacting bugs in the controller's config rendering: (1) the parent interface `Switch1/1/1` is rendered multiple times when a device has multiple subinterfaces, and (2) the Go SDK's V1 interface deserializer does not default the MTU field, causing different parent entries to render with different MTU values depending on the onchain interface version.

### Observed behavior

The controller generates configs for `nyc-dz001` and `lax-dz001` with duplicate `interface Switch1/1/1` blocks, each with a different MTU. On Arista EOS, the last block wins, and because the subinterfaces sort in different order on each device, the effective MTU differs between the two sides.

**NYC** (subinterfaces: `.1000`, `.1002` — last parent wins with `mtu 1500`):
```
interface Switch1/1/1
   mtu 2048
   no switchport
!
interface Switch1/1/1
   mtu 1500
   no switchport
```

**LAX** (subinterfaces: `.1002`, `.1003` — last parent wins with `mtu 2048`):
```
interface Switch1/1/1
   mtu 1500
   no switchport
!
interface Switch1/1/1
   mtu 2048
   no switchport
```

Both sides have `isis hello padding` enabled on the subinterfaces, which pads IS-IS hellos to the interface MTU. The parent interface MTU caps what can actually transit the wire, so mismatched parent MTUs between the two sides can prevent IS-IS hello exchange and block adjacency formation.

### Bug 1: Duplicate parent interface rendering

**Location:** `controlplane/controller/internal/controller/server.go:174-180`

```go
if intf.IsSubInterface {
    parent, err := intf.GetParent()
    if err != nil {
        c.log.Error(...)
        continue
    }
    d.Interfaces = append(d.Interfaces, parent)
}
```

`processDeviceInterfacesAndPeers` iterates all onchain interfaces for a device. For every subinterface, it calls `GetParent()` and appends a new parent interface to the device's interface list. When a device has multiple subinterfaces on the same physical port (e.g., `Switch1/1/1.1000` and `Switch1/1/1.1002`), the parent `Switch1/1/1` is appended multiple times.

[...]

PR #3696, opened 2026-05-13
controller: enforce interface MTU during config render

### LINKED ISSUE #3690: controller: enforce interface MTU during config render instead of trusting onchain state
## Summary

Make the controller compute the rendered interface MTU from the interface's role rather than from onchain `Interface.mtu` / `Link.mtu`, and render each parent interface exactly once with the max MTU of its subinterfaces. This makes the rendered config robust to bad / stale / V1 onchain state, even after #3411's onchain validation.

## Motivation

- #3411 added onchain validation: device interface MTU must be 1500 for CYOA/DIA and 9000 otherwise; link MTU must be 9000. This prevents *new* bad writes but does not protect against:
  - V1 interfaces that have not been re-serialized to V2 (Go SDK leaves `Mtu = 0`; see #3386)
  - Any future regression in onchain validation
  - Past migrations or one-off scripts that wrote wrong MTUs
- #3386 showed that mismatched / wrong parent MTUs cause IS-IS adjacency to fail silently. Testnet and mainnet-beta have been manually fixed, but the controller-side root cause (trusting onchain MTU, plus duplicate parent rendering) still exists.

The rendered config should be a pure function of onchain *topology* (which interface is CYOA, DIA, a link, etc.) — not of the raw MTU bytes onchain.

### 1. Compute MTU from interface role, not from onchain `Mtu`

In `controlplane/controller/internal/controller/`:

- Add Go constants mirroring the Rust ones in `smartcontract/programs/doublezero-serviceability/src/state/interface.rs:7-9`:
  - `InterfaceMtu = 9000` (fabric-facing physical / subinterface / link)
  - `CyoaDiaInterfaceMtu = 1500` (CYOA or DIA interface)
- Add a helper, e.g. `func mtuForInterface(i Interface) uint16`, returning `CyoaDiaInterfaceMtu` if `i.IsCYOA || i.IsDIA`, else `InterfaceMtu`. Apply it to every rendered physical interface (subinterface, parent, link, CYOA/DIA, plain physical).
- `server.go:410-412` — drop the `if link.Mtu > 0 { d.Interfaces[i].Mtu = uint16(link.Mtu) }` override. The controller now sets MTU from role, not from `link.Mtu`.
- `models.go:152-168` (`Interface.GetParent`) — stop copying `i.Mtu` into the parent. Parent MTU is computed later in the dedup step (see below).
- `toInterface` (`models.go:~74-110`) — set `Mtu` from the role-based helper rather than from `iface.Mtu`. (Keep reading the field from the SDK for telemetry / logging if useful, but do not propagate it to the rendered model.)

### 2. Render each parent interface exactly once with max(subinterface MTU)

In `controlplane/controller/internal/controller/server.go:173-196` (`processDeviceInterfacesAndPeers`):

- Replace the current pattern (which appends a fresh parent for every subinterface) with one that keeps a map keyed by parent name. For each subinterface, look up the parent; if absent, create it via `GetParent()`; if present, set `parent.Mtu = max(parent.Mtu, subInterfaceMtu)`.
- After the loop, append each unique parent to `d.Interfaces` exactly once. Sort as today.
- Because MTU now comes from role, in practice all subinterfaces on the same physical parent will have the same MTU (CYOA/DIA cannot be subinterfaces of a fabric parent — confirm during implementation). The `max` rule is the safe, future-proof behavior and matches the intent in #3386.

### 3. Template

`controlplane/controller/templates/tunnel.tmpl:87-95` currently falls back to `mtu 2048` when `.Mtu == 0`. Change the fallback to `mtu 9000` so that any path that bypasses the role-based computation still produces a safe fabric MTU rather than the placeholder 2048. Once the controller always sets a real MTU this branch should be unreachable, but the default is now defense-in-depth in the correct direction.

### 4. Tests

- `controlplane/controller/internal/controller/render_test.go` and `server_test.go` currently use `Mtu: 2048` throughout. Update fixtures so subinterfaces / links use `9000` and CYOA/DIA use `1500`.
- Add focused tests for `processDeviceInterfacesAndPeers`:
  - Device with multiple subinterfaces on the same physical parent → exactly one parent rendered.
  - Subinterfaces with mixed onchain MTUs (including `0`) → rendered MTU is determined by role, not onchain value; parent MTU is max of children.
  - Device with CYOA and DIA interfaces → rendered MTU is 1500.
  - Device with V1 leftovers (Go SDK `Mtu = 0`) → rendered MTU is 9000 / 1500 by role, never 2048.

## Out of scope

- Fixing the Go SDK V1 deserializer to default `Mtu = 1500`. That is tracked in #3386 (Bug 2) and is still worth doing for non-controller consumers, but is no longer load-bearing for the controller once this change lands.
- Changing onchain validation. #3411 already covers that.

## Related

- #3386 — root-cause analysis of the IS-IS outage (duplicate parent + V1 MTU=0)
- #3411 — onchain MTU enforcement (the validation side; this issue is the rendering side)

## Verification

- Unit tests in `controlplane/controller/internal/controller/` pass.
- Run the controller against a test fixture with V1 interfaces (`Mtu = 0`) and confirm rendered config has `mtu 9000` on all link subinterfaces and parent interfaces, `mtu 1500` on CYOA/DIA, and only one block per parent name.
- Diff rendered configs for testnet / mainnet-beta before vs after — should be a no-op now that all onchain MTUs are correct.

elitegreg:
Closes #3690.

## Summary

- Compute rendered interface MTU from interface *role* (CYOA/DIA → 1500, fabric → 9000) instead of from onchain `Interface.Mtu` / `Link.Mtu`. The rendered config is now a pure function of onchain topology, so V1 interfaces that deserialize with `Mtu = 0` (#3386) and any other stale onchain value no longer leak into device config.
- Render each subinterface parent exactly once. Previously the controller appended a fresh parent block for every subinterface, producing duplicate `interface Switch1/1/2` blocks. Parent MTU is now `max(subinterfaceMtu)` so the parent matches the wire MTU of its children.
- Drop the `if link.Mtu > 0 { d.Interfaces[i].Mtu = uint16(link.Mtu) }` override in `processDeviceInterfacesAndPeers` — link MTU now flows from role, not from onchain `Link.Mtu`.
- Change the `tunnel.tmpl` `mtu` fallback from `2048` to `9000` as defense-in-depth.

#3411 added onchain validation against bad new writes; this PR is the rendering-side companion that protects against past bad state and any future regression. Together with #3386, this closes out the root cause of the testnet / mainnet-beta IS-IS adjacency outage.

## Testing Verification

- New focused tests added to `controlplane/controller/internal/controller/`:
  - `TestProcessDeviceInterfacesAndPeers_ParentDedup` — multiple subinterfaces on `Switch1/1/2` produce exactly one parent block. Confirmed to fail on `main` (3 parents) and pass on this branch (1).
  - `TestProcessDeviceInterfacesAndPeers_RoleBasedMTU` — onchain `Mtu` values of 0, 2048, and 9216 are all ignored; rendered MTU is 9000 for fabric, 1500 for CYOA/DIA. Confirmed to fail on `main` and pass on this branch.
  - `TestMtuForInterface`, `TestToInterface_RoleBasedMTU`, `TestInterface_GetParent_DoesNotCopyMTU` — unit-level coverage of the helper and the V1 path.
- Existing `TestStateCache` and `TestEndToEnd` exercise the change end-to-end: their raw onchain `Interface.Mtu` / `Link.Mtu` fixtures are left at `2048` and the expected post-processed `Interface.Mtu` / rendered `mtu` lines are updated to `9000`, so they now also assert that the controller ignores onchain MTU.
- Golden fixtures (`fixtures/interfaces.txt`, `fixtures/e2e.tmpl`, `base.config*.txt`) updated to reflect the new role-based MTUs.
- `gofmt`, `go vet`, and `golangci-lint --build-tags qa,e2e ./controlplane/controller/...` clean.


controlplane/controller/internal/controller/server.go
```diff
@@ -173,24 +173,37 @@ func WithFeaturesConfig(cfg *FeaturesConfig) Option {
...
-		// Create a parent interface since they're not defined on-chain.
 		if intf.IsSubInterface {
 			parent, err := intf.GetParent()
 			if err != nil {
 				c.log.Error("failed to get parent interface for subinterface", "device pubkey", devicePubKey, "iface", intf.Name, "error", err)
 				continue
 			}
-			d.Interfaces = append(d.Interfaces, parent)
+			if existing, ok := parents[parent.Name]; ok {
+				if intf.Mtu > existing.Mtu {
```
nikw9944: I believe you could simplify this with `existing.Mtu = max(existing.Mtu, intf.Mtu)`

controlplane/controller/internal/controller/templates/tunnel.tmpl
```diff
@@ -90,7 +90,7 @@ interface {{ .Name }}
       {{- if .Mtu }}
    mtu {{ .Mtu }}
       {{- else }}
```
nikw9944, line 92: Since this should never happen now, would it be better to error here to help us find regressions?

controlplane/controller/internal/controller/models.go
```diff
@@ -93,21 +111,23 @@ func toInterface(iface serviceability.Interface) (Interface, error) {
...
-	}, nil
-
+	}
+	if ifType == InterfaceTypePhysical {
+		out.Mtu = mtuForInterface(out)
+	}
+	return out, nil
 }
 
 func NewInterface(
```
nikw9944: You didn't update NewInterface, which makes sense because it's no longer used. Want to keep things tidy by removing it while you're touching the file or nah?

nikw9944 (approved):

PR #3703, opened 2026-05-14
rfcs: add RFC-21 to extend tenants to multicast use cases

juan-malbeclabs:
## Summary

- Adds RFC-21, proposing the extension of `Tenant` from a unicast-only construct to a per-use-case resource grouping primitive.
- Tenants declare which traffic classes they participate in (unicast, multicast, or both) and only allocate a VRF ID when they need one.
- Multicast groups gain an optional `tenant_pk` on create and update; mandatory enforcement comes in a follow-up RFC.
- Tenant administrators can manage their own access passes without going through foundation, sentinel, or feed authority.

## Why

- Every tenant today consumes one of the 1024 global VRF IDs, even when it exists only to group multicast resources. The pool becomes scarce for tenants that actually carry unicast traffic.
- Access passes require foundation/sentinel/feed authority, which blocks self-service onboarding for the team running a use case.
- One consistent construct per use case is cleaner than introducing a separate grouping primitive for multicast.

## Testing Verification

RFC document only; no implementation in this PR. Implementation lands in follow-up PRs scoped to the smartcontract, SDKs, and CLI.

[RFC-21](https://github.com/malbeclabs/doublezero/blob/264330032/rfcs/rfc21-tenant-capabilities.md)


rfcs/rfc21-tenant-capabilities.md
```diff
@@ -0,0 +1,159 @@
...
+- **VRF pool exhaustion.** Multicast-only tenants no longer consume VRF IDs, which is a net improvement for pool availability. The legacy fallback (treat `vrf_id != 0` as UNICAST on read) preserves the existing accounting for already-allocated IDs.
+- **Schema invariants.** `validate()` rejects `capabilities == 0`, unknown bits, and any inconsistency between the UNICAST bit and `vrf_id`. Unit tests cover each case.
+- **Direct `vrf_id` writes via `UpdateTenant`.** The existing `TenantUpdateArgs.vrf_id` field is preserved for compatibility; the validation layer enforces the UNICAST/`vrf_id` invariant, so an inconsistent direct write is rejected.
+
+## Backward Compatibility
+
+- The `Tenant` account layout is extended only at the end. Borsh deserialization of pre-RFC accounts returns `capabilities = 0`, which the program treats as `UNICAST` whenever `vrf_id != 0`.
+- `TenantCreateArgs` and `TenantUpdateArgs` already derive `BorshDeserializeIncremental`. Older CLIs that do not pass `capabilities` get the default (treated as `UNICAST` on create), so existing tooling continues to work for unicast tenants.
```
vihu, line 149: Compatibility nit (maybe a footgun): this section says older CLIs that do not pass `capabilities` get the default treated as `UNICAST`, but `BorshDeserializeIncremental` defaults a missing trailing `u8` to `Default::default()` (`0`) unless an explicit `#[incremental(default = ...)]` is set.

Since the RFC also says `capabilities != 0`, can we make the implementation requirement explicit here? For example: `TenantCreateArgs.capabilities` should use `#[incremental(default = TENANT_CAPABILITY_UNICAST)]` or be decoded through an `Option`/normalization path so old tenant-create instructions remain unicast-compatible.

rfcs/rfc21-tenant-capabilities.md
```diff
@@ -0,0 +1,159 @@
...
+| Operator workflow | Multicast-only tenants no longer consume VRF IDs; tenant admins can manage their own access passes. |
+| Foundation tooling| No change for foundation, sentinel, or feed callers; their existing flows are unaffected.  |
+
+## Security Considerations
+
+- **Authorization widening for `set_access_pass`.** Tenant administrators gain the ability to call `set_access_pass`, but only when every referenced tenant is one they administer. This is enforced both at the top-level gate (to accept the call) and at the per-tenant check (already present on add, mirrored to remove). Foundation-level callers retain their existing privileges.
+- **VRF pool exhaustion.** Multicast-only tenants no longer consume VRF IDs, which is a net improvement for pool availability. The legacy fallback (treat `vrf_id != 0` as UNICAST on read) preserves the existing accounting for already-allocated IDs.
+- **Schema invariants.** `validate()` rejects `capabilities == 0`, unknown bits, and any inconsistency between the UNICAST bit and `vrf_id`. Unit tests cover each case.
```
vihu, line 143: Can we clarify where the legacy `capabilities == 0 && vrf_id != 0 => UNICAST` normalization happens? The RFC says `validate()` rejects `capabilities == 0`, but also says pre-RFC accounts deserialize with `capabilities == 0`.

To avoid follow-up ambiguity, I think the RFC should say whether `Tenant::try_from` materializes legacy tenants as `capabilities = UNICAST`, whether validation uses `effective_capabilities()`, or whether writes explicitly normalize before `try_acc_write`.

rfcs/rfc21-tenant-capabilities.md
```diff
@@ -0,0 +1,159 @@
...
+- `(capabilities & UNICAST != 0)` if and only if `vrf_id != 0`.
+- Legacy tenants (written before this RFC) deserialize with `capabilities == 0`. On read, when `vrf_id != 0`, the program interprets capabilities as `UNICAST`. The next write persists the explicit byte.
+
+### Lifecycle
+
+- **Create.** Caller passes the desired `capabilities`. The VRF ID is allocated from the global pool only when `UNICAST` is set; otherwise `vrf_id` is `0` and the VRF resource extension is not mutated.
+- **Update.** Capabilities can be added (additive only) via `add_capabilities`. Adding `UNICAST` allocates a VRF ID. Removing capabilities is not supported in this RFC; the operator must delete and recreate the tenant.
+- **Delete.** VRF ID is deallocated only when `vrf_id != 0`.
+
+### Tenant administrators and access passes
```
vihu, line 71: This section probably needs more clarification: the tenant-admin path for `set_access_pass` should probably be scoped more tightly than “every tenant referenced by this call.” Today `set_access_pass` also updates access-pass type, expiry, flags, owner-on-create, client IP, and the user-payer funding path. If a tenant admin is authorized through tenant membership only, can they mutate those non-tenant fields too? I think the RFC should explicitly require at least one referenced/administered tenant and define whether the tenant-admin path may only add/remove tenants, or whether it can update the full access-pass record. It should also say how existing/mixed `tenant_allowlist` entries are handled.

rfcs/rfc21-tenant-capabilities.md
juan-malbeclabs: The idea is that `capabilities = 0` should represent `UNICAST` for backward compatibility, until the tenant is updated and rewritten with the new capability value set to `1`.

rfcs/rfc21-tenant-capabilities.md
juan-malbeclabs: Tenant administrators can create, modify, and delete access passes associated with that tenant.

We still need to think through what should happen when an access pass already exists for the same `IP + user_payer` and an administrator wants to modify it, or when the access pass includes a multicast group the administrator does not have permission to manage.

Conceptually, though, the goal is for tenant administrators to be able to manage access passes and associate them with resources that belong to the tenant they administer.

rfcs/rfc21-tenant-capabilities.md
juan-malbeclabs: The goal of the RFC is not to serve as a detailed implementation plan, but rather to define the scope and the changes in how resources are associated, allow operation without requiring a VRF, and enable administrator management outside of the Foundation allowlist.

PR #3737, opened 2026-05-20
tools/twamp/signed: agent-side challenge-response inbound probing

ben-dz:
## Summary of Changes
- Adds an opt-in nonce handshake to signed-TWAMP inbound probing. The reflector now issues a fresh 8-byte random nonce on every Probe 0 and places it in `Reply0.SinceLastRxNs` (previously always 0). On Probe 1, if the sender echoed that nonce in `Sec || Frac` and re-signed the packet, the reflector sets bit 7 of `Reply1.NumOffsets` (`Challenged` flag) — cryptographic proof the sender actually received Reply 0 before sending Probe 1, defeating the pre-emit-Probe-1 attack on `SinceLastRxNs`.
- Extends `ReplyPacket` with a `Challenged` field encoded as the top bit of byte 212 (lower 7 bits remain the offset count 0–5). The flag is covered by the Ed25519 signature, so a MITM cannot flip it.
- Backwards compatible in both directions: legacy senders never echo the nonce, so bit 7 stays 0 and the existing `NumOffsets ≤ 5` unmarshal check still passes; legacy senders also ignore `Reply0.SinceLastRxNs`, so populating it with a nonce is invisible to them.
- Updates RFC16 to document the new wire-format semantics (a new "Challenge-Response Inbound Probing" subsection under Detailed Design, plus byte-table entries for `Sec`/`Frac`/`SinceLastRxNs`/`NumOffsets`). Folds in the wallet-bound-probing use case that was previously in Future Work.
- **Agent-side only.** The follow-up PR adds a `--challenged` flag to `geoprobe-target-sender` to actually exercise the new flow; until then the reflector still answers ordinary unchallenged probes exactly as before.

## Diff Breakdown
| Core logic   |     2 | +61 / -11   |  +50 |
| Tests        |     3 | +328 / -27  | +301 |
| Docs         |     1 | +24 / -15   |   +9 |
| **Total**    |     6 | +413 / -53  | +360 |

Test-heavy: 73% of added lines are reflector / packet-codec tests covering the new behavior and the cohabiting bit-packed wire-format byte.

<details>

- `tools/twamp/pkg/signed/reflector_linux.go` — adds `nonce` to per-sender state, generates a fresh nonce on each Probe 0, verifies the echoed nonce on Probe 1 (with `state.nonce != 0` guard so a `crypto/rand` failure doesn't spuriously authenticate), clears the nonce on pair completion / rate-limit reset / stale-pair reset so the documented "0 outside a pair" invariant holds unconditionally
- `tools/twamp/pkg/signed/reflector_test.go` — adds tests for non-zero nonce in Reply 0, distinct nonces across pairs (gated by \`testing.Short\`), Reply 1 challenged-on-match, Reply 1 unchallenged-on-mismatch; updates pre-existing assertions that expected Reply 0 SinceLastRxNs to be 0
- `tools/twamp/pkg/signed/packet_test.go` — adds round-trip tests for \`Challenged=true\`, \`Challenged=false\`, and the combined case (challenged + non-zero offset count) so the bit-cohabits-with-count path is exercised
- `tools/twamp/pkg/signed/packet.go` — adds \`Challenged bool\` to \`ReplyPacket\`, named \`numOffsetsChallengedBit\`/\`numOffsetsCountMask\` constants, encodes/decodes the flag as the top bit of byte 212, threads \`challenged bool\` through \`NewReplyPacket\`
- `rfcs/rfc16-geolocation-verification.md` — new Challenge-Response Inbound Probing subsection (flow, why-unchallenged-is-permitted, backwards-compat, wallet-bound use case); byte-table entries updated for \`Sec\`/\`Frac\`/\`SinceLastRxNs\`/\`NumOffsets\`; resolves a stale "Reply 0 SinceLastRxNs is 0" note in the flow narrative; removes the obsolete Future Work entry that this PR implements

</details>

## Testing Verification
- \`go test -count=1 ./tools/twamp/pkg/signed/...\` green (20s); \`./controlplane/telemetry/cmd/geoprobe-agent/...\` green
- \`golangci-lint run\` on both packages: 0 issues
- New unit tests directly assert the wire-format byte (\`buf[212]&0x80\`, \`buf[212]&0x7F\`) so the bit-packing contract is tested at the byte level
- End-to-end reflector tests (in-process UDP via the existing Linux harness) exercise both the challenged and unchallenged Probe 1 paths against a real reflector goroutine

Related: [RFC16](rfcs/rfc16-geolocation-verification.md) — see the new "Challenge-Response Inbound Probing" subsection under Detailed Design.
Related Stacked PR: #3738


tools/twamp/pkg/signed/packet.go
nikw9944: Update comments to reflect added scope?
```
Sec          uint32   // Bytes 4-7: NTP timestamp seconds; on challenged Probe 1, upper 4 bytes of echoed Reply 0 nonce
Frac         uint32   // Bytes 8-11: NTP timestamp fractional; on challenged Probe 1, lower 4 bytes of echoed Reply 0 nonce
```

tools/twamp/pkg/signed/packet.go
```diff
@@ -251,7 +265,7 @@ func UnmarshalReplyPacket(buf []byte) (*ReplyPacket, error) {
 	return r, nil
 }
 
-func NewReplyPacket(probe *ProbePacket, signer Signer, geoprobePubkey [32]byte, offsets [][]byte, slot uint64, lat, lng float64, sinceLastRxNs, rttNs uint64) (*ReplyPacket, error) {
+func NewReplyPacket(probe *ProbePacket, signer Signer, geoprobePubkey [32]byte, offsets [][]byte, slot uint64, lat, lng float64, sinceLastRxNs, rttNs uint64, challenged bool) (*ReplyPacket, error) {
```
nikw9944, line 268: Maybe out of scope for this PR but there are so many args here that they probably deserve their own struct. Perhaps only extract the locationoffset info (slot, lat, lng, sinceLastRxNs, rttNs) into a struct

tools/twamp/pkg/signed/reflector_linux.go
```diff
@@ -258,14 +263,28 @@ func (r *LinuxReflector) Run(ctx context.Context) error {
...
+				}
+				sinceLastRxNs = state.nonce
+			} else {
+				// Variable-time compare is fine: the nonce is a freshness check, not a secret.
+				// Its full value is published in cleartext in Reply 0 to the legitimate sender.
+				presentedNonce := (uint64(probe.Sec) << 32) | uint64(probe.Frac)
+				if state.nonce != 0 && presentedNonce == state.nonce {
+					challenged = true
+				}
+				state.nonce = 0 // pair complete; field invariant says nonce is 0 outside a live pair
```
nikw9944: Clarify comment by adding: `lastTxTime intentionally preserved; will be overwritten on next Reply Tx`

nikw9944:
Claude:

The new tests mix `t.Fatalf` and `require.True/require.NotZero` for failures that other tests in the same file express via`assert.*` + `require.NoError`. Examples:

- reflector_test.go:587 uses t.Fatalf for nonce equality, while reflector_test.go:637–642 uses t.Fatalf for Challenged checks — but reflector_test.go:691 uses assert.False for the same kind of check.
- packet_test.go:578–579 uses require.NotZero / require.Zero on raw bytes; nearby packet_test.go tests use assert.Equal(byte(0), …).

Functionally identical; stylistically inconsistent. Worth a pass to align on require.* for "must hold or test is meaningless" and assert.* for "we want to keep checking after a failure."

nikw9944 (approved):

tools/twamp/pkg/signed/packet.go
ben-dz: Not doing as out of scope for now.

PR #3738, opened 2026-05-20
telemetry/geoprobe-target-sender: opt-in --challenged inbound probing

ben-dz:
> Stacked on #3737 (RFC16 + reflector). Merge the base PR first.

## Summary of Changes
- Adds an opt-in `--challenged` flag to `geoprobe-target-sender` (default `false`). When set, the sender extracts the nonce from `Reply0.SinceLastRxNs`, writes it into `Probe1.Sec || Frac`, signs Probe 1, and only then transmits — proving to the reflector (PR #3737) that it actually received Reply 0 before issuing Probe 1.
- Default behavior (`--challenged=false`) is **byte-identical on the wire** to the pre-PR2 sender. The existing `ProbePair` body is moved verbatim into a private `probePairUnchallenged`; a new `probePairChallenged` adds the deferred-sign path; the exported `ProbePair` dispatches on a constructor flag.
- Threads `Challenged` through the existing telemetry: startup `slog.Info`, per-pair `slog.Debug`, JSON output (with a `"challenged": bool` field that is intentionally **not** `omitempty` so downstream consumers always see the mode), and the human-readable text output.
- Trade-off baked into the new path: signing Probe 1 after Reply 0 receipt inflates `Reply1.SinceLastRxNs` by the sender's compute time, in exchange for cryptographic proof of Reply 0 receipt. This is the documented RFC16 trade-off — TEE senders should stay unchallenged; bare-metal senders use `--challenged`.

## Stacking & rollout
- During the rollout window (new sender against an old reflector that doesn't issue nonces), `Reply0.SinceLastRxNs == 0`, the sender echoes 0 in `Sec/Frac`, and the agent's `state.nonce != 0` guard keeps `Reply1.Challenged == false`. Sender doesn't crash and the operator sees `challenged=false` in the per-pair log, indistinguishable from "agent hasn't been upgraded yet" — which is exactly the desired signal.

## Diff Breakdown (vs PR1 base branch)
| Core logic   |     5 | +97 / -8    |  +89 |
| Tests        |     2 | +122 / -6   | +116 |
| **Total**    |     7 | +218 / -13  | +205 |

Test-heavy. The core-logic diff is dominated by the new `probePairChallenged` method (deferred-sign flow) and the `newChallengedProbePacket` helper; everything else is plumbing (constructor signature, dispatcher, CLI flag, telemetry surfacing).

<details>

- `tools/twamp/pkg/signed/sender_linux.go` — renames existing `ProbePair` body to `probePairUnchallenged` (verbatim, no logic change), adds `probePairChallenged` using `sendAndRecv` for both probes, adds a 4-line `ProbePair` dispatcher that branches on the new `challenged` field of `LinuxSender`
- `tools/twamp/pkg/signed/sender_test.go` — two new subtests inside `TestSender_Linux`: `ProbePair challenged mode` (asserts `Reply1.Challenged == true` and non-zero timing fields against a real reflector goroutine) and `ProbePair unchallenged mode is still unchallenged` (asserts the byte-identical legacy path against the same reflector); 5 existing call sites updated with `false`
- `tools/twamp/pkg/signed/packet.go` — new unexported `newChallengedProbePacket(seq, signer, nonce)` helper that takes the nonce as a parameter (does NOT generate one — that's the reflector's job in PR #3737), sets `Sec = uint32(nonce >> 32)`, `Frac = uint32(nonce & 0xFFFFFFFF)`, and signs the 44-byte payload
- `controlplane/telemetry/cmd/geoprobe-target-sender/main_test.go` — `TestProbeOutput_ChallengedFieldRoundTrip` (true case: round-trip plus raw `"challenged":true` byte assertion) and `TestProbeOutput_ChallengedFalseStillSerialized` (false case: asserts `"challenged":false` appears so an accidental `omitempty` regression fails fast)
- `controlplane/telemetry/cmd/geoprobe-target-sender/main.go` — `--challenged` flag declaration, threaded through `signed.NewSender`; new `Challenged bool` field on the `probeOutput` JSON struct (no `omitempty`); challenged status added to startup log, per-pair `slog.Debug`, JSON output, and text output ("Challenged Inbound:" line)
- `tools/twamp/pkg/signed/sender.go` — `NewSender` constructor gains trailing `challenged bool` parameter
- `tools/twamp/pkg/signed/stub_fallback.go` — non-Linux `NewLinuxSender` signature updated to match

</details>

## Testing Verification
- `go test -count=1 ./tools/twamp/pkg/signed/...` green (20s); `./controlplane/telemetry/cmd/geoprobe-target-sender/...` green
- `golangci-lint run` on both packages: 0 issues
- Sender-side challenged-mode test exercises the full Probe 0 → Reply 0 → extract-nonce → sign-Probe-1 → Reply 1 round-trip against a real in-process Linux reflector goroutine, asserting `Reply1.Challenged == true`. The unchallenged-mode test exercises the same harness with `challenged=false` and asserts `Reply1.Challenged == false`, confirming default behavior is preserved.
- JSON serialization is tested for both `Challenged=true` and `Challenged=false` so the no-`omitempty` choice is regression-protected.

Related: [RFC16 "Challenge-Response Inbound Probing"](rfcs/rfc16-geolocation-verification.md) (added in PR #3737) — see the "Why unchallenged inbound is still permitted" paragraph for the latency trade-off rationale.


tools/twamp/pkg/signed/sender_linux.go
nikw9944, line 1: ASYMMETRIC VERIFICATION BETWEEN CHALLENGED AND UNCHALLENGED PATHS

Files: tools/twamp/pkg/signed/sender_linux.go:140-311 (unchallenged) and :319-357 (challenged)

The two ProbePair paths have meaningfully different observability under failure:

- Unchallenged path never verifies signatures inside the library. It validates only MinReplyPacketSize <= n <= MaxReplyPacketSize, then for Reply 0 Seq match and later Sec/Frac match. The reply is returned to the caller regardless of signature validity, which then sets reply0_sig_valid / reply1_sig_valid in the JSON output. A malicious / corrupted reply is reported as "received, sig=false".
- Challenged path delegates both probes to sendAndRecv(..., verify=true, ...). tryRecv silently drops replies whose probe-sig, authority pubkey, or reply-sig fail (return 0, nil, nil, false), then keeps polling until the deadline elapses. The operator sees timeout instead of reply_sig_valid=false.

Both outcomes are defensible, but the asymmetry is worth either calling out in a code comment or aligning. Suggest at minimum a comment in probePairChallenged documenting that bad-sig replies are silently treated as not-received (and considering whether the challenged path should also surface the unverified reply so operators can distinguish "no packet" from "bogus packet").

nikw9944 (approved):

PR #3742, opened 2026-05-21
smartcontract/sdk: set max CU and heap frame on all serviceability transactions

elitegreg:
## Summary

- Every serviceability transaction sent through `DZClient` now prepends `set_compute_unit_limit(MAX_COMPUTE_UNIT_LIMIT = 1_400_000)` and `request_heap_frame(MAX_HEAP_FRAME_BYTES = 256 KiB)`, replacing the default 200K CU / 32 KiB heap.
- `DoubleZeroClient::execute_transaction_with_compute_unit_limit` is removed; the three per-instruction `SET_GLOBAL_CONFIG_COMPUTE_UNIT_LIMIT` / `ASSIGN_TOPOLOGY_NODE_SEGMENTS_COMPUTE_UNIT_LIMIT` constants that selectively opted into the higher CU limit are deleted along with it.
- Safe because the serviceability program runs on a dedicated private Solana cluster — there is no fee or resource-contention impact from raising every transaction to the protocol max.

## Testing Verification

- `cargo test -p doublezero_sdk --lib` (147 tests pass, including the rewritten globalconfig/set and topology backfill mock expectations).


juan-malbeclabs (approved):
LGTM

PR #3750, opened 2026-05-21
smartcontract/sdk: skip preflight simulation when submitting transactions

elitegreg:
## Summary of Changes
- Drop the explicit `simulate_transaction` call in `DZClient::execute_transaction_inner`. Previously every successful tx was simulated twice (once here, then again by `send_and_confirm_transaction`'s default preflight); now it's submitted once.
- Submit with `RpcSendTransactionConfig { skip_preflight: true }` via `send_and_confirm_transaction_with_spinner_and_config`, eliminating preflight entirely.
- Preserve the existing error UX: on failure, extract the `TransactionError` from `ClientErrorKind::TransactionError` or `SendTransactionPreflightFailure`, fetch program logs via `get_transaction_with_config`, and surface them via the same `SimulationError` / `SimulationTransactionError` types so callers and CLI output are unchanged.
- Tradeoff: failing transactions now land onchain and burn fees instead of failing for free at simulation. The `get_transaction` fetch is only hit on the cold path.

## Diff Breakdown
| Core logic   |     1 | +72 / -26   |  +46 |

Single-file refactor of the transaction submission path in the Rust SDK.

<details>

- `smartcontract/sdk/rs/src/client.rs` — rewrites `execute_transaction_inner` to skip preflight and recover program logs from `get_transaction` on the failure path.

</details>

## Testing Verification
- `cargo test -p doublezero_sdk --lib` — all 147 unit tests pass.
- `cargo clippy -p doublezero_sdk -- -Dclippy::all -Dwarnings` clean.
- Failure-path behavior not exercised against a live cluster; recommend a manual run of a known-failing CLI command (e.g. duplicate-resource create) to confirm program logs still render and `DoubleZeroError` mapping is intact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+### Changes
+
+- SDK (Rust)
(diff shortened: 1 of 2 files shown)


juan-malbeclabs (approved):
LGTM

PR #3769, opened 2026-05-25
feat(e2e): fix flaky settlement test

bgm-malbeclabs:
## Summary of Changes
Unicast QA tests will no longer hang on a stuck Multicast tunnel, regardless of what's causing the stuck state.
## Testing Verification
* Updated e2e tests


juan-malbeclabs (approved):
LGTM

PR #3775, opened 2026-05-27
cli: restore default value for device interface create --bandwidth

thijsvanemmerik:
## Summary

- `--bandwidth` on `doublezero device interface create` became mandatory after PR #3077, even though that PR's description stated it was now optional.
- Root cause: PR #3077 dropped `default_value = "0"` from the clap attribute on `bandwidth`. Because the field is `u64` (not `Option<u64>`), clap then treats omission as a missing required argument. Only `cir` actually became optional in that PR.
- Restores `default_value = "0bps"` to match the existing `cir` attribute on the next line. `0bps` is a valid input for `validate_parse_bandwidth`.

## Testing Verification

- Added regression test `test_cli_device_interface_create_bandwidth_is_optional` that uses `clap::try_parse_from` without `--bandwidth` and asserts parsing succeeds with `bandwidth == 0`.
- Verified the test fails on pre-fix code with clap error `MissingRequiredArgument` for `--bandwidth <BANDWIDTH>`, and passes after the fix.
- All 336 `doublezero_cli` unit tests pass.


ben-malbeclabs:
I think we need to scope this out a bit more:

- We have bandwidth in two places: links and interfaces
- We want bandwidth on the CYOA interfaces not connected to a link to be able to track capacity
- We want bandwidth for the interface to match the link
- Bandwidth is not an attribute of a loopback interface

We could place a CLI check that prevents: 
`interface[a|z] bandwidth < link bandwidth`

juan-malbeclabs:
@thijsvanemmerik @ben-malbeclabs  If that sounds good to everyone, we can move forward with this change and I can send a follow-up PR with the validations Ben suggested. Then we can merge them in sequence.

thijsvanemmerik:
Sounds good Juan, thanks for picking up the follow-up. Happy to keep this PR as the minimal loopback regression fix.

Ben, your scoping is right. The two checks sit naturally in different commands: require non-zero bandwidth for CYOA and DIA interfaces at `device interface create` and `device interface update`, and enforce `interface[a|z].bandwidth >= link.bandwidth` at `link wan-create` / `dzx-create`.

juan-malbeclabs:
New PR: https://github.com/malbeclabs/doublezero/pull/3778

juan-malbeclabs (approved):
LGTM -> https://github.com/malbeclabs/doublezero/pull/3778 bandwidth validation

PR #3774, opened 2026-05-27
sdk: serviceability go executor CreateUser/DeleteUser

elitegreg:
## Summary

Adds the Solana-side primitives the device-stress orchestrator (#3746) needs as a library-only change — no new binary.

- `CreateUser` (variant 36) and `DeleteUser` (variant 42) executor methods on `smartcontract/sdk/go/serviceability`, with account-list construction mirroring the Rust SDK; both wait for post-confirmation visible state so callers get a meaningful `t_activate` to record.
- PDA helpers `GetUserPDA`, `GetAccessPassPDA`, `GetTunnelIdsPDA`, `GetDzPrefixBlockPDA` — seeds mirrored from `smartcontract/programs/doublezero-serviceability/src/pda.rs`.
- Rust fixture generator extended to emit `user_create_args.{bin,json}` and `user_delete_args.{bin,json}`; Go tests load them as the cross-language wire-format contract.

Part 1 of #3746. Closes #3770.

`PlanReconcile` / `ReconcilePlan` were originally scoped here but are orchestrator policy, not an SDK primitive, so they now land alongside the orchestrator binary in part 2 of #3746.

## Testing Verification

- Golden-byte parity: Go `buildCreateUserInstruction` / `buildDeleteUserInstruction` produce the same borsh body as the Rust-generated fixtures (`user_create_args.bin`, `user_delete_args.bin`).
- Account ordering verified against the Rust SDK's expected `AccountMeta` lists (`smartcontract/sdk/rs/src/commands/user/create.rs:209` and `delete.rs:363`).
- New PDA helpers cross-checked against an independent recomputation of the seed bytes; `GetTunnelIdsPDA` / `GetDzPrefixBlockPDA` verified to use 8-byte little-endian index.
- `make go-build go-lint go-test` all green.

Issue #3746, opened 2026-05-21
stress: implement tools/stress/device-orchestrator


elitegreg:
## Why

Drives the sweep. Manages onchain user records on the live devnet ledger (alternate serviceability program), records timestamps end-to-end, and runs the agent on the DUT via SSH so its log goes to a known location.

See [Notion design](https://www.notion.so/malbeclabs/GRE-Tunnel-Capacity-Study-Design-34ffef22bebe816a87dadcacc01c2081) for full context.

## Depends on

- #3741 (agent log lines parsed by the orchestrator)

## Scope (single Go binary; PR may split — flag at review)

Create `tools/stress/device-orchestrator/` (new Go module under the existing `tools/` tree alongside `gnmi-tunnel`, `twamp`, etc.).

- **CLI flags** (per doc): `--target-user-count`, `--users-per-batch`, `--hold-seconds` (default 180), `--dut-pubkey`, `--dut-ssh-host`, `--dut-ssh-key`, `--rpc-url`, `--program-id`, `--keypair`, `--controller` (IP:PORT, passed to agent), `--abort-file`, `--working-dir`.
- **Reconcile-to-target loop**: on each batch iteration, query the alternate-program user list, create or delete to reach the next target. Use `smartcontract/sdk/go/serviceability` directly — do not shell out to the `doublezero` CLI.
- **Hold**: pause `--hold-seconds` between batches for observer samples.
- **Deprovision phase**: reverse-order delete, same timestamp set, distinct provision/deprovision event names.
- **Agent runner**: `ssh` to DUT, run `doublezero-agent ... -verbose`, stream stdout/stderr into `<working-dir>/orchestrator.agent.log`. Parse `"Committing config session due to diffs detected: <diff>"` to extract `+ interface Tunnel<ID>` and record `t_pre_commit_log`; parse the commit-success line (from #3741 scope extension) to record `t_agent_applied`.
- **Outputs** in `--working-dir`:
  - `orchestrator-config.json` — all CLI options.
  - `orchestrator-runlog.json` — one row per event: `{run_id, user_index, user_pubkey, tunnel_id, event, t_ns, n_after_event}`. Events: `submit | confirm | activate | pre_commit_log | applied | deprovision_*`.
  - `orchestrator.agent.log` — tailed agent output.
- **Abort handling**: poll `--abort-file`; on appearance, finish the current user, stop, dump partial outputs, exit non-zero.

## Acceptance

[...]

PR #3776, opened 2026-05-27
tools/stress: orchestrator skeleton (CLI, sweep, runlog, abort)

### LINKED ISSUE #3771: stress: device-orchestrator skeleton — CLI, sweep loop, runlog, abort (part 2 of #3746)
Part 2 of #3746. Depends on part 1 (sub-issue #3770) for `PlanReconcile` + user CRUD.

Standalone binary that runs the sweep end-to-end against a real devnet RPC but **stubs the agent runner behind an interface**. The issue's acceptance run (0 → 8 in batches of 2) works except `pre_commit_log` / `applied` events (those land in part 3).

## Scope

New tree `tools/stress/device-orchestrator/`, modelled on `tools/twamp/`:

- `cmd/device-orchestrator/main.go` — every flag from #3746's CLI list; dumps `orchestrator-config.json` on start.
- `pkg/runlog/` — append-only JSON-lines writer for `orchestrator-runlog.json` with the row schema `{run_id, user_index, user_pubkey, tunnel_id, event, t_ns, n_after_event}`.
- `pkg/sweep/` — sweep loop driven by `PlanReconcile`, batches of `--users-per-batch`, hold via `--hold-seconds`, deprovision phase in reverse order. Records `submit | confirm | activate | deprovision_*` events.
- `pkg/abort/` — poller watching `--abort-file` on a ticker; signals the sweep loop via context cancel between user iterations (NOT mid-instruction).
- `pkg/agent/` — Go interface `AgentRunner { Start(ctx) error; Events() <-chan AgentEvent }`. Ship a `noopAgent` impl now; SSH impl lands in part 3.
- `Makefile` mirroring `tools/twamp/Makefile`.

## Tests

- `sweep_test.go`: fake `ReconcileExecutor` + fake clock + `noopAgent`; assert event order + runlog rows.
- `abort_test.go`: tempdir, touch the file, assert cancellation within one batch.
- `runlog_test.go`: schema round-trip.

## Completion / CI

Binary builds via root `make go-build`; new package's tests run under `make go-test`. No live-RPC test in CI — manual only. PR title prefix `tools/stress: orchestrator skeleton`.

elitegreg:
## Summary

Adds the device-stress orchestrator skeleton at `tools/stress/device-orchestrator/`, for the GRE Tunnel Capacity Study. Stacked on top of #3774 (part 1, SDK user CRUD). Part 2 of #3746. Closes #3771.

- **`cmd/device-orchestrator`** — every flag from #3746's CLI list (`--target-user-count`, `--users-per-batch`, `--hold-seconds`, `--dut-pubkey`, `--dut-ssh-host`, `--dut-ssh-key`, `--rpc-url`, `--program-id`, `--keypair`, `--controller`, `--abort-file`, `--working-dir` plus `--client-ip-base`, `--tunnel-endpoint`, `--tenant-pubkey`, `--run-id`, `--log-level`, `--dry-run`). Dumps `orchestrator-config.json` on start.
- **`pkg/reconcile`** — `PlanFor(current, target, ownerFilter)` returns a deterministic `Plan{ToCreate, ToDelete}` delta. Lifted from the part-1 SDK PR per the discussion — it's orchestrator policy, not an SDK primitive.
- **`pkg/sweep`** — provision-then-reverse-deprovision loop driven by `PlanFor`; batches of `--users-per-batch` with `--hold-seconds` between batches; reverse-creation-order deprovision tracked by the sweep itself; emits `submit | confirm | activate | deprovision_*` runlog rows.
- **`pkg/runlog`** — append-only JSONL writer for `orchestrator-runlog.json` with the row schema `{run_id, user_index, user_pubkey, tunnel_id, event, t_ns, n_after_event}`.
- **`pkg/abort`** — ticker-based watcher of `--abort-file`; cancels a derived ctx so the sweep finishes the in-flight user before exiting non-zero, then still tears down what was created.
- **`pkg/agent`** — `Runner` interface (`Start(ctx) error; Events() <-chan Event`) with a no-op implementation. The SSH-backed runner and the `pre_commit_log` / `applied` row generation land in **part 3**.
- **`pkg/exec`** — `Live` impl of `sweep.Executor` wrapping `serviceability.{Client, Executor}`; picks deterministic per-user IPs (`base + idx`) and forwards `DevicePubkey` / `TenantPubkey` to `UserCreateArgs`.
- **`Makefile`** mirrors `tools/twamp/Makefile` (build, test, lint).

## Testing Verification

- `pkg/sweep`: fake `Executor` + fake `Clock` + no-op `Agent` drive a 0→4 sweep in batches of 2. Asserts ordered `submit`/`confirm`/`activate` x4, reverse-order `deprovision_submit`/`deprovision_confirm`/`deprovision_activate` x4, `Hold` fires exactly once (between batches, not after reaching target), and `n_after_event` increments at `activate` / decrements at `deprovision_activate`.
- `pkg/sweep` abort case: failing the 3rd create still drives deprovision over the first two users so the orchestrator never leaks state on abort.
- `pkg/abort`: tempdir + touch the sentinel + assert the derived ctx cancels within 1s; empty-path watch is a no-op that still propagates parent cancellation.
- `pkg/runlog`: round-trip rows, auto-fill `t_ns`, reject writes after `Close`, `Open(path)` truncates existing content.
- `pkg/reconcile`: table-driven 0→N / N→0 / partial / foreign-only / mixed / negative / tie-break-by-pubkey.
- Smoke test: `make build` produces `bin/device-orchestrator`; `./bin/device-orchestrator --dry-run --target-user-count 4 --users-per-batch 2 --working-dir /tmp/orch` writes a valid `orchestrator-config.json` without contacting RPC.
- `make go-build go-lint go-test` all green.

## Out of scope

- SSH-driven agent runner that parses `Committing config session due to diffs detected: <diff>` and the commit-success line into `pre_commit_log` / `applied` events. Lands in part 3 of #3746.
- Live-RPC end-to-end test in CI (the binary is exercisable manually with the `dz-local` devnet).


tools/stress/device-orchestrator/pkg/runlog/runlog.go
```diff
@@ -0,0 +1,101 @@
...
+		return fmt.Errorf("marshal runlog row: %w", err)
+	}
+	buf = append(buf, '\n')
+	if _, err := w.w.Write(buf); err != nil {
+		return fmt.Errorf("write runlog row: %w", err)
+	}
+	return nil
+}
+
+// Close flushes and closes the underlying file.
```
nikw9944: Maybe ask Claude to "remove any comments on code that would be obvious to an experienced developer"

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ All notable changes to this project will be documented in this file.
...
+- Tools
+  - Add `tools/stress/device-orchestrator/` — the device-stress orchestrator skeleton for the GRE Tunnel Capacity Study. The binary parses every flag from #3746's CLI list, dumps `orchestrator-config.json` on start, runs a provision-then-reverse-deprovision sweep against a live serviceability program, and emits the runlog row schema `{run_id, user_index, user_pubkey, tunnel_id, event, t_ns, n_after_event}` to `orchestrator-runlog.json` for each `submit | confirm | activate | deprovision_*` event. The agent runner is stubbed behind a `pkg/agent.Runner` interface (no-op impl ships now; the SSH-backed runner that emits `pre_commit_log` / `applied` lands in part 3). The sweep cooperates with an abort sentinel file: when the file appears the in-flight user completes and the orchestrator deprovisions everything it created before exiting non-zero. `PlanReconcile` / `Plan` (lifted from the part-1 SDK PR) now lives at `tools/stress/device-orchestrator/pkg/reconcile/` as orchestrator policy rather than SDK primitive. Part 2 of #3746 ([#3771](https://github.com/malbeclabs/doublezero/issues/3771)).
```
nikw9944: This is craaaaaazy verbose

tools/stress/device-orchestrator/pkg/sweep/sweep.go
```diff
@@ -0,0 +1,262 @@
...
+
+	created, err := provision(ctx, &cfg)
+	if err != nil && !errors.Is(err, context.Canceled) {
+		return err
+	}
+	// Always attempt deprovision so an abort during provision still cleans up
+	// what the sweep created. Use a fresh context for the deprovision phase if
+	// the original was cancelled, since the operator wants the tear-down to
+	// finish before exit. We respect the parent context's lifetime via the
+	// outer Run's error return — callers that want a hard stop pass a deadline.
```
nikw9944: This says "use a fresh context... if the original was cancelled" but then uses the original context anyway

tools/stress/device-orchestrator/pkg/sweep/sweep_test.go
```diff
@@ -0,0 +1,321 @@
...
+	require.NoError(t, sweep.Run(context.Background(), cfg))
+	require.NoError(t, w.Close())
+
+	rows := readRows(t, path)
+	assert.Empty(t, rows)
+	assert.Zero(t, exec.createN.Load())
+	assert.Zero(t, exec.deleteN.Load())
+}
+
+func TestRun_AbortBetweenUsersStillCleansUp(t *testing.T) {
```
nikw9944, line 240: Shouldn't this test either call cancel() or touch the abort sentinel via abort.Watch?

tools/stress/device-orchestrator/pkg/sweep/sweep.go
```diff
@@ -0,0 +1,262 @@
...
+			if err := ctx.Err(); err != nil {
+				return created, err
+			}
+			idx := activeCount
+			submitAt := cfg.Clock.Now()
+			if err := emit(cfg, idx, "", 0, runlog.EventSubmit, submitAt, activeCount); err != nil {
+				return created, err
+			}
+
+			res, err := cfg.Executor.CreateUser(ctx, idx)
```
nikw9944: If the context is cancelled between submit and poll, can the SDK return an error even though the transaction landed onchain?

tools/stress/device-orchestrator/pkg/sweep/sweep.go
```diff
@@ -0,0 +1,262 @@
...
+
+	Executor Executor
+	Agent    agent.Runner
+	Runlog   *runlog.Writer
+	Clock    Clock
+	Logger   *slog.Logger
+}
+
+func (c *Config) validate() error {
+	switch {
```
nikw9944, line 83: Add a check for OwnerFilter=0?

tools/stress/device-orchestrator/pkg/exec/exec.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// The SDK's CreateUser blocks on signature finalization and post-confirm
+	// account visibility; we don't get distinct stage timestamps today, so
+	// confirm and activate both anchor at the post-call wallclock. A future
+	// SDK refactor can split these.
+	tunnelID, err := l.fetchTunnelID(ctx, userPDA)
+	if err != nil {
+		// Surface the tunnel ID as 0; the sweep records the create as successful
+		// because the on-chain User already exists.
+		tunnelID = 0
```
nikw9944: I don't think this does anything since fetchTunnelID() always returns err=nil

tools/stress/device-orchestrator/cmd/device-orchestrator/main.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	return slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: lvl}))
+}
+
+func dumpJSON(path string, v any) error {
+	f, err := os.Create(path)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
```
nikw9944: We're not catching Close() errors - it could fail silently in some edge cases (like full filesystem)

tools/stress/device-orchestrator/pkg/agent/agent.go
```diff
@@ -0,0 +1,73 @@
...
+	ch := make(chan Event)
+	return &noop{log: log, events: ch}
+}
+
+type noop struct {
+	log    *slog.Logger
+	events chan Event
+}
+
+func (n *noop) Start(ctx context.Context) error {
```
nikw9944, line 66: Only one of these can safely run at a time - a 2nd one would panic on close(n.events). Guard it with sync.Once?

tools/stress/device-orchestrator/cmd/device-orchestrator/main.go
```diff
@@ -0,0 +1,277 @@
...
+		TunnelEndpoint: tunnelIP,
+		UserType:       serviceability.UserTypeIBRL,
+		CyoaType:       serviceability.CyoaTypeGREOverDIA,
+		DzPrefixCount:  1,
+	})
+	if err != nil {
+		return err
+	}
+
+	runlogPath := filepath.Join(*workingDir, "orchestrator-runlog.json")
```
nikw9944: Rename to `orchestrator-runlog.jsonl`?

tools/stress/device-orchestrator/pkg/sweep/sweep.go
```diff
@@ -0,0 +1,262 @@
...
+	case c.Hold < 0:
+		return errors.New("sweep: Hold must be >= 0")
+	case c.RunID == "":
+		return errors.New("sweep: RunID is required")
+	case c.Executor == nil:
+		return errors.New("sweep: Executor is required")
+	case c.Runlog == nil:
+		return errors.New("sweep: Runlog is required")
+	}
+	if c.Clock == nil {
```
nikw9944, line 106: It's a little funky to set these defaults in a validator function, we should probably do it in a constructor (`DefaultConfig`?) or when calling Run().

tools/stress/device-orchestrator/pkg/sweep/sweep.go
```diff
@@ -0,0 +1,262 @@
...
+		c.Clock = RealClock{}
+	}
+	if c.Logger == nil {
+		c.Logger = slog.Default()
```
nikw9944, line 110: Regardless of where we set the default, how about adding the run ID to all logs with `c.Logger.With("run_id", c.RunID)`?

tools/stress/device-orchestrator/pkg/sweep/sweep.go
```diff
@@ -0,0 +1,262 @@
...
+			if err := emit(cfg, idx, pkStr, res.TunnelID, runlog.EventActivate, res.ActivatedAt, activeCount); err != nil {
+				return created, err
+			}
+		}
+
+		runningTarget = nextTarget
+		if runningTarget >= cfg.Target {
+			break
+		}
+		if cfg.Hold > 0 {
```
nikw9944: Do we want this to sleep even when ToCreate == 0?

tools/stress/device-orchestrator/cmd/device-orchestrator/main.go
```diff
@@ -0,0 +1,277 @@
...
+		return err
+	}
+	defer f.Close()
+	enc := json.NewEncoder(f)
+	enc.SetIndent("", "  ")
+	return enc.Encode(v)
+}
+
+func requireFlags(required map[string]string) error {
+	var missing []string
```
nikw9944, line 265: Multiple missing-flag errors will not display in a deterministic order.  Fix: sort by flag name first

tools/stress/device-orchestrator/cmd/device-orchestrator/main.go
```diff
@@ -0,0 +1,277 @@
...
+		return fmt.Errorf("write orchestrator-config.json: %w", err)
+	}
+	logger.Info("orchestrator-config.json written", "path", configPath)
+
+	if *dryRun {
+		logger.Info("dry-run: skipping sweep")
+		return nil
+	}
+
+	if err := requireFlags(map[string]string{
```
nikw9944: Consider moving this above dumpJSON() so we don't write the file unless the flags are good

nikw9944 (approved):

tools/stress/device-orchestrator/pkg/runlog/runlog.go
elitegreg: Done — trimmed the obvious method comments in runlog.go (Path/Close).

CHANGELOG.md
elitegreg: Condensed to a single paragraph.

tools/stress/device-orchestrator/pkg/sweep/sweep.go
elitegreg: Good catch — the code did use the original ctx. Fixed: deprovision now runs under `context.WithoutCancel(ctx)` so teardown actually completes after an abort, and the comment matches.

tools/stress/device-orchestrator/pkg/sweep/sweep_test.go
elitegreg: Rewritten to cancel the real sweep ctx after the 2nd create (via an afterCreate hook) instead of faking an executor error, so it exercises the actual abort-at-boundary path.

tools/stress/device-orchestrator/pkg/sweep/sweep.go
elitegreg: Yes — that was the risk. Create/Delete now run under `context.WithoutCancel(ctx)` so an abort can't interrupt an in-flight tx and orphan a user; abort is observed at the iteration boundary instead.

tools/stress/device-orchestrator/pkg/sweep/sweep.go
elitegreg: Added `OwnerFilter.IsZero()` to validate().

tools/stress/device-orchestrator/pkg/exec/exec.go
elitegreg: Right, it was dead. Removed fetchTunnelID and set TunnelID=0 inline with a part-3 note.

tools/stress/device-orchestrator/cmd/device-orchestrator/main.go
elitegreg: dumpJSON now captures the Close() error so a flush failure isn't swallowed.

tools/stress/device-orchestrator/pkg/agent/agent.go
elitegreg: Guarded Start with sync.Once.

tools/stress/device-orchestrator/cmd/device-orchestrator/main.go
elitegreg: Renamed to orchestrator-runlog.jsonl.

tools/stress/device-orchestrator/pkg/sweep/sweep.go
elitegreg: Moved the defaults into applyDefaults(); validate() is now mutation-free.

tools/stress/device-orchestrator/pkg/sweep/sweep.go
elitegreg: Done — `cfg.Logger = cfg.Logger.With("run_id", cfg.RunID)` in Run().

tools/stress/device-orchestrator/pkg/sweep/sweep.go
elitegreg: Now skips the hold when the batch's plan.ToCreate == 0.

tools/stress/device-orchestrator/cmd/device-orchestrator/main.go
elitegreg: Sorted the missing names before formatting.

tools/stress/device-orchestrator/cmd/device-orchestrator/main.go
elitegreg: Moved requireFlags above dumpJSON (dry-run still exempt) so a bad invocation no longer writes the config file.

PR #3777, opened 2026-05-27
ci(e2e): add trusted fork PR e2e dispatch

### LINKED ISSUE #3695: Proposal: Making our CI Fork Friendly
Right now, doublezero is effectively in a CI blackout for external contributors. Because our current workflows are Infrastructure-First, tightly coupled to specialised 16-core runners, Linux-specific `netlink` calls, and sensitive `DOCKERHUB_TOKENS` we’ve (rightly) had to disable them for forks to protect our infra and secrets.

**The Goal:** Logic-First Testing
I’d like to propose we shift toward a Logic-First architecture. By decoupling the core business logic from the execution environment, we can create a Safe-Zone CI that is:
- Secret-Free: Zero dependency on API keys.
- Platform-Agnostic: Runs on standard ubuntu-latest (GitHub-hosted).
- Policy-Compliant: Safe to run on every fork PR immediately.

### 1. Architectural Decoupling (Go & Rust)
We need to lean into Dependency Inversion to separate the Brain (Logic) from the Hands (OS/Ledger):
- In Go: Introduce interfaces for networking components (e.g., a RouteManager interface). Instead of calling the Linux Kernel directly, we call the interface. This lets us test complex routing logic on macOS or standard runners using in-memory mocks.
- In Rust: Isolate Logic crates from those requiring a full Solana validator or heavy hardware. We should be able to lint and unit-test the core crates on standard runners without the workspace choking.

### 2. Bifurcating the Workflows
I suggest splitting the CI into two tiers:
- Tier 1: Core CI (The Safe Zone): Covers Lints, Unit Tests for api/, config/, sdk/, controller/, and mocked networking logic. It runs on ubuntu-latest and triggers on all Fork PRs.
- Tier 2: Infrastructure CI (The Heavy Tier): E2E, Container builds, Solana validator tests, and PIM tests. These stay on our self-hosted/16c runners and remain restricted to internal branches or maintainer-triggered runs.

### 3. Standardizing Build Tags
We can use `//go:build !infra` and Rust feature flags to ensure the Core runner never tries to execute a test it isn't equipped for.

## What we get out of this?
- We get Fork CI back without exposing a single secret or runner.
- Developers can finally run a lot of the suite on a Mac instead of needing a specific Linux box.

vihu:
# Summary

Adds maintainer-gated trusted e2e automation for external fork PRs, so default fork PR CI no longer fails on unavailable GHCR write permissions or repository secrets.

## Changes

- Add `trusted-fork-e2e` workflow for `/run-e2e` PR comments and manual dispatch.
- Require the requester to have `write`, `maintain`, or `admin` repo permission.
- Dispatch trusted `e2e.yml` and/or `shreds-e2e.yml` runs with the exact PR head SHA.
- Add `workflow_dispatch` support to `e2e.yml` and `shreds-e2e.yml`.
- Skip privileged e2e setup on external fork `pull_request` runs with a clear notice.
- Validate trusted workflow checkouts still match the requested PR head SHA.
- Skip Docker Hub login in Go container tests when fork PR secrets are unavailable.
- Add `actionlint` config for existing custom runner labels.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,5 @@
+self-hosted-runner:
+  labels:
+    - doublezero-k8s-ci
+    - ubuntu-24.04-16c-64gb
+config-variables: null
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -3,6 +3,17 @@ on:
   push:
     branches: [ main, 'hotfix/**' ]
   pull_request:
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: Pull request number to test
+        required: true
+      head_sha:
+        description: Expected PR head SHA
+        required: true
+      image_tag:
+        description: GHCR image tag for this trusted run
+        required: true
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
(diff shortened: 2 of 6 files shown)


elitegreg (approved):
Looks good, but how do we test it?

vihu:
> Looks good, but how do we test it?

I think we can probably try forking with one of our own personal accounts and make some yolo PR. We won't need to merge it, just check that we can execute the e2e workflow

PR #3778, opened 2026-05-27
cli/serviceability: validate CYOA/DIA and link bandwidth invariants

juan-malbeclabs:
## Summary

Reject misconfigured device interfaces and links by enforcing two new bandwidth invariants in both the CLI (immediate, fee-free feedback) and the on-chain serviceability program (source-of-truth so the invariants hold regardless of the client):

- CYOA/DIA device interfaces require a non-zero `--bandwidth` at `device interface create` and `device interface update`. On update the check fires only when the transaction is changing CYOA, DIA, or bandwidth, so legacy zero-bandwidth CYOA/DIA interfaces already onchain can still be updated for unrelated fields without first being repaired.
- Link create requires `interface[a|z].bandwidth >= link.bandwidth` at `link wan-create` and `link dzx-create` (DZX only checks side A; side Z is external).

All on-chain rejections surface as `DoubleZeroError::InvalidBandwidth` (`Custom(31)`); CLI rejections surface as targeted `eyre` errors before submitting a transaction.

## Base branch

Targets [`tve/cli-bandwidth-default-value`](https://github.com/malbeclabs/doublezero/tree/tve/cli-bandwidth-default-value) because the CLI changes build on top of `--bandwidth` being optional again (default `0bps`). When that branch lands on main, this branch can be retargeted with no replay required.

## Testing Verification

- CLI unit tests: 7 new tests covering each rejection path (CYOA-zero, DIA-zero on create; flip-to-CYOA without bandwidth and clear-bandwidth-on-CYOA on update; WAN side A / side Z insufficient; DZX side A insufficient) plus updates to 3 pre-existing tests that built links over zero-bandwidth interfaces. `cargo test -p doublezero_cli --lib` — 21 targeted tests pass.
- Onchain integration tests: new `tests/bandwidth_validation_test.rs` exercises each rejection end-to-end through the program-test bank (4 tests, all assert `Custom(31)`). Pre-existing `link_onchain_allocation_test.rs`, `link_wan_test.rs`, and `topology_test.rs` updated where they built physical interfaces with `bandwidth=0` next to non-zero link bandwidth.
- `make rust-fmt` and `make rust-lint` clean on the full workspace; full `cargo test -p doublezero-serviceability --tests` green.

## Notes for reviewers

- The on-chain update check is deliberately conditional. Quote from the comment in `processors/device/interface/update.rs`:
  > Only enforce when the transaction is changing CYOA, DIA, or bandwidth, so legacy zero-bandwidth CYOA/DIA interfaces created before this rule can still be updated for unrelated fields without first being repaired.
- DZX links do not validate side Z bandwidth onchain or in the CLI because side Z is external (no `Interface` to inspect). This mirrors the existing CYOA/MTU treatment for DZX.


elitegreg (approved):

thijsvanemmerik:
Hey Juan, small thing on DZX: side Z's interface only gets set when contributor Z runs `link accept`, so the create-time bandwidth check only catches side A. A too-small side Z interface would still slip through. Worth adding the same check in `link accept`?

juan-malbeclabs:
@thijsvanemmerik Good catch. I added the same validation in `link accept` as well, so side Z now gets checked against the link bandwidth when the interface is assigned.

PR #3780, opened 2026-05-27
tools/stress: agent SSH + log parser

### LINKED ISSUE #3772: stress: device-orchestrator SSH agent runner + log parser (part 3 of #3746)
Part 3 of #3746. Depends on part 2 (sub-issue #3771) for the `AgentRunner` interface, and on #3741 for the commit-success log line.

Real `AgentRunner` implementation that completes the #3746 acceptance criteria.

## Scope

- `pkg/agent/ssh.go` — dial `--dut-ssh-host` with `--dut-ssh-key` (golang.org/x/crypto/ssh), exec `doublezero-agent ... -verbose`, tee remote stdout/stderr into `<working-dir>/orchestrator.agent.log` while feeding it to the parser.
- `pkg/agent/parser.go`:
  - Scan for `Committing config session due to diffs detected: <diff>` → extract every `+ interface Tunnel<ID>` in the diff and emit one `pre_commit_log` event per ID with `t_ns = time.Now()` at line-read time.
  - Scan for the commit-success line from #3741 → emit `applied` events. Coordinate exact format with #3741's PR before merging.
- Wire `pkg/agent/ssh.go` into `cmd/device-orchestrator/main.go` as the default `AgentRunner`. Keep `noopAgent` reachable via `--no-agent` for offline testing.

## Tests

- `parser_test.go`: golden log fixtures (multi-tunnel diff, deprovision diff, commit-success line) — fully offline.
- SSH client itself is integration-only; gate behind a build tag or env var so CI doesn't dial.

## Completion / CI

`make go-build go-lint go-test` green; manual devnet acceptance run per #3746: 0→8 in batches of 2, hold 10s, full five-event timestamps for at least one user end-to-end, abort-file respected within one batch. PR title prefix `tools/stress: agent SSH + log parser`.

elitegreg:
## Summary

Completes the device-stress orchestrator (#3746) by replacing the no-op `AgentRunner` with the live SSH-driven runner and the log parser that turns agent diff/commit lines into `pre_commit_log` / `applied` runlog rows. Stacked on top of #3776 (part 2, orchestrator skeleton). Part 3 of #3746. Fixes #3772.

- **`pkg/agent/parser.go`** — `Parser.Parse(line) []Event` tracks two log lines from `controlplane/agent/pkg/arista/eapi.go`:
  - `Committing config session due to diffs detected: <diff>` → extracts every `+ interface Tunnel<ID>` and emits one `pre_commit_log` event per ID; the diff's `-` lines (deprovisions) are ignored.
  - `Configuration session finalized with command '... commit'` → emits one `applied` event per pending tunnel; the `... abort` variant clears the buffer without emitting.
- **`pkg/agent/ssh.go`** — `NewSSH(cfg) Runner` dials `--dut-ssh-host` with `--dut-ssh-key`, execs `doublezero-agent -verbose` (appending `-controller <addr>` when `--controller` is set), and tees remote stdout/stderr into `<working-dir>/orchestrator.agent.log` while running every line through `Parser`. The session is closed on ctx cancel; the events channel closes after both stream readers exit so consumers never see a half-emitted event. Host-key verification uses `ssh.InsecureIgnoreHostKey` (documented; targets are ephemeral cEOS containers).
- **`pkg/sweep`** — gained an agent-event consumer goroutine and a `tunnelRegistry` populated as users are created; consumer attributes each `agent.Event` back to a `user_index` via tunnel ID lookup and appends `pre_commit_log` / `applied` rows. Unknown tunnels are debug-logged and dropped. The agent is started under a derived context that the sweep cancels after deprovision, so a clean run still shuts the agent down rather than leaking goroutines.
- **`pkg/exec.fetchTunnelID`** — now implemented: `GetAccountInfo` on the user PDA + `DeserializeUser` + return `TunnelId`. `exec.Config` gains an `RPC` field; the cmd binary passes the same `*solanarpc.Client` used by `Client`/`Executor`.
- **`cmd/device-orchestrator`** — new flags `--dut-ssh-user` (default `admin`) and `--no-agent` (offline testing). SSH runner is the default when `--dut-ssh-host` and `--dut-ssh-key` are both set; with either missing, the cmd falls back to the no-op runner and warns.

## Testing Verification

- `pkg/agent/parser_test.go`: golden line fixtures for single-tunnel diff, multi-tunnel diff (mixed +/-), deprovision-only diff, commit-success after multi-tunnel diff, abort-clears-buffer, stray commit-with-no-pending, two consecutive provision cycles, oversized tunnel ID skipped, `Tunnel5000` vs `Tunnel500` boundary.
- `pkg/sweep/sweep_test.go`: new `scriptedAgent` + a `deleteGate` on the fake executor lets the test emit agent events while deprovision is blocked; asserts `pre_commit_log` / `applied` rows are written for the two registered tunnels and the unregistered tunnel 999 is dropped. Tests pass under `-race`.
- `pkg/exec/exec_test.go`: stub RPC returns a hand-encoded `User` body with `TunnelId = 4242`; `fetchTunnelID` reads it correctly. Missing-account path returns an error containing `not found`.
- Smoke test: `make build` produces `bin/device-orchestrator`; `--dry-run` writes `orchestrator-config.json` containing `dut_ssh_user`, `no_agent`, and the rest of the flag set.
- `make go-build go-lint go-test` all green.

## Out of scope

- Integration test that actually SSH-dials a container. The SSH client is deterministic plumbing over `golang.org/x/crypto/ssh`; CI never opens an SSH session. Acceptance is via the manual devnet run per #3746.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
+- Tools
+  - Complete the device-stress orchestrator (part 3): replace the stubbed agent runner with an SSH-backed runner that execs `doublezero-agent -verbose` on the DUT and tees its output to `orchestrator.agent.log`, and a log parser that turns the agent's commit-diff lines into `pre_commit_log` / `applied` runlog events. Adds `--dut-ssh-user` and `--no-agent` flags.
+
 ## [v0.25.1](https://github.com/malbeclabs/doublezero/compare/client/v0.24.0...client/v0.25.1) - 2026-06-01
 
 ### Breaking
--- a/tools/stress/device-orchestrator/cmd/device-orchestrator/main.go
+++ b/tools/stress/device-orchestrator/cmd/device-orchestrator/main.go
@@ -137,12 +137,20 @@ func run() error {
 	// doesn't leave a config file behind. A dry-run is exempt: its whole job is
 	// to dump the resolved config without needing the live-RPC flags.
 	if !*dryRun {
-		if err := requireFlags(map[string]string{
+		required := map[string]string{
 			"--dut-pubkey": *dutPubkey,
 			"--rpc-url":    *rpcURL,
 			"--program-id": *programID,
 			"--keypair":    *keypairPath,
-		}); err != nil {
+		}
+		if !*noAgent {
+			// Agent telemetry is required unless explicitly disabled; don't
+			// silently degrade to a no-op run that records no pre_commit_log /
+			// applied rows.
+			required["--dut-ssh-host"] = *dutSSHHost
+			required["--dut-ssh-key"] = *dutSSHKey
+		}
+		if err := requireFlags(required); err != nil {
 			return err
 		}
 	}
@@ -288,19 +296,17 @@ func requireFlags(required map[string]string) error {
 // selectAgentRunner picks between the SSH-backed runner and the no-op, based
 // on the CLI flags:
 //
-//   - --no-agent → noop (operator opted out)
-//   - --dut-ssh-host + --dut-ssh-key set → SSH runner (default for live runs)
-//   - otherwise → noop with a warning (operator forgot the flags)
+//   - --no-agent → noop (operator explicitly opted out, e.g. offline testing)
+//   - otherwise → SSH runner
+//
+// The SSH flags are validated as required upstream when --no-agent is unset, so
+// this never silently falls back to the no-op runner.
 //
 // The SSH runner tees remote stdout/stderr into <working-dir>/orchestrator.agent.log.
 // The exec'd command appends --controller iff the operator passed --controller.
 func selectAgentRunner(noAgent bool, sshHost, sshKey, sshUser, controllerAddr, workingDir string, logger *slog.Logger) agent.Runner {
 	if noAgent {
 		logger.Info("agent: --no-agent set; using no-op runner")
-		return agent.NewNoop(logger)
-	}
-	if sshHost == "" || sshKey == "" {
-		logger.Warn("agent: --dut-ssh-host and --dut-ssh-key not both set; falling back to no-op runner (pre_commit_log / applied events will not be recorded)")
 		return agent.NewNoop(logger)
 	}
 
--- a/tools/stress/device-orchestrator/cmd/device-orchestrator/main.go
+++ b/tools/stress/device-orchestrator/cmd/device-orchestrator/main.go
@@ -44,6 +44,7 @@ type orchestratorConfig struct {
 	DUTPubkey       string `json:"dut_pubkey"`
 	DUTSSHHost      string `json:"dut_ssh_host"`
 	DUTSSHKey       string `json:"dut_ssh_key"`
+	DUTSSHUser      string `json:"dut_ssh_user"`
 	RPCURL          string `json:"rpc_url"`
 	ProgramID       string `json:"program_id"`
 	KeypairPath     string `json:"keypair"`
@@ -53,6 +54,7 @@ type orchestratorConfig struct {
 	ClientIPBase    string `json:"client_ip_base"`
 	TunnelEndpoint  string `json:"tunnel_endpoint"`
 	TenantPubkey    string `json:"tenant_pubkey,omitempty"`
+	NoAgent         bool   `json:"no_agent"`
 }
 
 func main() {
@@ -82,6 +84,8 @@ func run() error {
 		runID           = flag.String("run-id", "", "Run identifier written into every runlog row; auto-generated if empty.")
 		logLevel        = flag.String("log-level", "info", "slog level: debug|info|warn|error.")
 		dryRun          = flag.Bool("dry-run", false, "Validate flags and dump orchestrator-config.json without contacting the RPC.")
+		dutSSHUser      = flag.String("dut-ssh-user", "admin", "SSH user for the DUT.")
+		noAgent         = flag.Bool("no-agent", false, "Use the no-op AgentRunner even when SSH flags are set (offline testing).")
 	)
 	flag.Parse()
 
@@ -117,6 +121,7 @@ func run() error {
 		DUTPubkey:       *dutPubkey,
 		DUTSSHHost:      *dutSSHHost,
 		DUTSSHKey:       *dutSSHKey,
+		DUTSSHUser:      *dutSSHUser,
 		RPCURL:          *rpcURL,
 		ProgramID:       *programID,
 		KeypairPath:     *keypairPath,
@@ -126,6 +131,7 @@ func run() error {
 		ClientIPBase:    *clientIPBase,
 		TunnelEndpoint:  *tunnelEndpoint,
 		TenantPubkey:    *tenantPubkey,
+		NoAgent:         *noAgent,
 	}
 	// Validate required flags before writing anything, so a bad invocation
 	// doesn't leave a config file behind. A dry-run is exempt: its whole job is
@@ -180,6 +186,7 @@ func run() error {
 	liveExec, err := exec.New(exec.Config{
 		Client:         client,
 		Executor:       executor,
+		RPC:            rpc,
 		DevicePubkey:   dutPK,
 		TenantPubkey:   tenantPK,
 		ClientIPBase:   baseIP,
@@ -206,6 +213,8 @@ func run() error {
 	ctx, abortCancel := abort.Watch(rootCtx, *abortFile, abort.DefaultPollInterval, logger)
 	defer abortCancel()
 
+	agentRunner := selectAgentRunner(*noAgent, *dutSSHHost, *dutSSHKey, *dutSSHUser, *controllerAddr, *workingDir, logger)
+
 	cfg := sweep.Config{
 		RunID:         *runID,
 		Target:        *targetUserCount,
@@ -213,7 +222,7 @@ func run() error {
 		Hold:          time.Duration(*holdSeconds) * time.Second,
 		OwnerFilter:   signer.PublicKey(),
 		Executor:      liveExec,
-		Agent:         agent.NewNoop(logger),
+		Agent:         agentRunner,
 		Runlog:        rlw,
 		Clock:         sweep.RealClock{},
 		Logger:        logger,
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -4,71 +4,29 @@ All notable changes to this project will be documented in this file.
...
...hile feeding the stream through `pkg/agent/parser.go`. The parser tracks two log lines from `controlplane/agent/pkg/arista/eapi.go`: `Committing config session due to diffs detected: <diff>` (extracting `+ interface Tunnel<ID>` matches and emitting one `pre_commit_log` event per ID) and `Configuration session finalized with command '... commit'` (emitting one `applied` event per pending tunnel; the `... abort` variant clears the buffer without emitting). The sweep grows a goroutine that consumes agent events and writes `pre_commit_log` / `applied` runlog rows by looking up each event's tunnel ID against a `tunnelID → user_index` map populated as users are created; unknown tunnels are debug-logged and dropped. `pkg/exec.fetchTunnelID` now reads the on-chain user account post-create to surface the assigned `TunnelId` into the runlog. New CLI flags: `--dut-ssh-user` (default `admin`) and `--no-agent` for offline testing. Host-key verification uses `ssh.InsecureIgnoreHostKey` because the orchestrator targets ephemeral cEOS containers; documented at `pkg/agent/ssh.go:SSH`. Part 3 of #3746, completes the five-event coverage ([#3772](https://github.com/malbeclabs/doublezero/issues/3772)).
```
nikw9944: Make these changelog entries much shorter are remove references to issue/PR IDs?

tools/stress/device-orchestrator/pkg/agent/parser_test.go
```diff
@@ -0,0 +1,134 @@
...
+	return func() time.Time { return at }
+}
+
+func TestParser_SingleTunnelDiffThenCommit(t *testing.T) {
+	t.Parallel()
+
+	now := time.Unix(1_700_000_000, 0)
+	p := agent.NewParser(agent.WithClock(fixedClock(now)))
+
+	events := p.Parse(`2026/05/27 12:00:01 Committing config session due to diffs detected: + interface Tunnel500   ip address 169.254.0.1/30`)
```
nikw9944, line 23: streamLines() in ssh.go splits on \n, so these tests would not notice if we're ignoring "interface TunnelXXX" lines after the first \n. Which Claude thinks is exactly what's happening. Each "interface TunnelXXX" should be on a separate line

tools/stress/device-orchestrator/pkg/agent/ssh.go
```diff
@@ -0,0 +1,228 @@
...
+
+	if err := session.Start(s.cfg.Command); err != nil {
+		s.shutdown()
+		return fmt.Errorf("ssh agent: start %q: %w", s.cfg.Command, err)
+	}
+	if s.cfg.Logger != nil {
+		s.cfg.Logger.Info("ssh agent started", "host", s.cfg.Host, "command", s.cfg.Command, "log_path", s.cfg.LogPath)
+	}
+
+	parser := NewParser()
```
nikw9944, line 157: This parser is sent to 2 different goroutines below, even though parser.go says "A single Parser is goroutine-safe only against the calling Parse goroutine"

tools/stress/device-orchestrator/pkg/agent/ssh.go
```diff
@@ -0,0 +1,228 @@
...
+	if err != nil {
+		return fmt.Errorf("ssh agent: load key %s: %w", s.cfg.KeyPath, err)
+	}
+
+	clientCfg := &ssh.ClientConfig{
+		User:            s.cfg.User,
+		Auth:            []ssh.AuthMethod{ssh.PublicKeys(signer)},
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+	}
+	client, err := ssh.Dial("tcp", s.cfg.Host, clientCfg)
```
nikw9944, line 105: Set a timeout here to make it more obvious when a device is unresponsive?

tools/stress/device-orchestrator/pkg/sweep/sweep.go
```diff
@@ -122,42 +115,129 @@ type createdUser struct {
...
+	}()
+
+	created, err := provision(ctx, &cfg, registry)
 	if err != nil && !errors.Is(err, context.Canceled) {
-		return err
+		// On a non-cancel error from provision we still want deprovision to
+		// run (clean up what was created); the consumer keeps draining in
+		// parallel so any straggling agent events for already-created users
+		// still land in the runlog.
+		_ = err
```
nikw9944: Does this if{} actually do anything?

tools/stress/device-orchestrator/pkg/exec/exec.go
```diff
@@ -112,6 +122,26 @@ func (l *Live) DeleteUser(ctx context.Context, userPDA solana.PublicKey) (sweep.
...
+	}
+	if info == nil || info.Value == nil {
+		return 0, fmt.Errorf("user account %s not found", userPDA)
+	}
+	data := info.Value.Data.GetBinary()
+	if len(data) == 0 {
+		return 0, fmt.Errorf("user account %s empty", userPDA)
+	}
+	var u serviceability.User
+	serviceability.DeserializeUser(serviceability.NewByteReader(data), &u)
```
nikw9944, line 144: Sanity check that `u.AccountType == serviceability.UserType`

tools/stress/device-orchestrator/pkg/agent/ssh.go
```diff
@@ -0,0 +1,228 @@
...
+		for _, ev := range parser.Parse(line) {
+			select {
+			case events <- ev:
+			case <-ctx.Done():
+				return
+			}
+		}
+	}
+	if err := scanner.Err(); err != nil && log != nil {
+		log.Warn("ssh agent: stream ended with error", "err", err, "stream", label)
```
nikw9944: Should we fail the run if the agent errors out, rather than just logging a warning?

nikw9944 (changes requested):

CHANGELOG.md
elitegreg: Done — collapsed the entry to a single short bullet and dropped the #3746/#3772 references.

tools/stress/device-orchestrator/cmd/device-orchestrator/main.go
```diff
@@ -276,6 +285,39 @@ func requireFlags(required map[string]string) error {
...
+// on the CLI flags:
+//
+//   - --no-agent → noop (operator opted out)
+//   - --dut-ssh-host + --dut-ssh-key set → SSH runner (default for live runs)
+//   - otherwise → noop with a warning (operator forgot the flags)
+//
+// The SSH runner tees remote stdout/stderr into <working-dir>/orchestrator.agent.log.
+// The exec'd command appends --controller iff the operator passed --controller.
+func selectAgentRunner(noAgent bool, sshHost, sshKey, sshUser, controllerAddr, workingDir string, logger *slog.Logger) agent.Runner {
+	if noAgent {
```
nikw9944, line 308: Is there a reason to keep noop mode?

tools/stress/device-orchestrator/pkg/agent/parser_test.go
elitegreg: Good catch — confirmed the agent logs the diff via a single `log.Printf` (eapi.go:274), so only the marker line gets the timestamp prefix and every `+ interface Tunnel<ID>` arrives on a following line. The old parser only scanned the marker line and silently dropped the rest.

Fixed: the parser now opens a diff block on the `Committing...` marker and scans every subsequent line for `+ interface Tunnel<ID>` until the session finalizes (commit/abort). Rewrote the multi-tunnel test to feed realistic multi-line input (one interface per `Parse` call, as `streamLines` delivers it) — it fails against the old parser and passes now — and added a test that diff-shaped lines outside a block are ignored.

tools/stress/device-orchestrator/pkg/agent/ssh.go
elitegreg: Right — that was a real data race (and the two readers also wrote the log file concurrently). Restructured so the two reader goroutines now only scan raw lines and forward them onto a shared channel; a single consumer goroutine tees to the log and runs `parser.Parse`. So exactly one goroutine touches the parser and the log file, and stdout/stderr stay interleaved in arrival order (which also keeps the new multi-line diff parsing correct). Verified with `go test -race`.

tools/stress/device-orchestrator/pkg/agent/ssh.go
elitegreg: Added a `DialTimeout` to `SSHConfig` (default 10s) and set it on `ssh.ClientConfig.Timeout`, which bounds the TCP connect + handshake. An unresponsive device now fails fast with a clear dial error instead of hanging on the OS default.

tools/stress/device-orchestrator/pkg/sweep/sweep.go
elitegreg: No — the `if` body was a no-op (`_ = err`), with the real error check happening below at the `return err`. Removed the dead block and folded its rationale into the comment above the `deprovision` call (which already explained teardown always runs).

tools/stress/device-orchestrator/pkg/exec/exec.go
elitegreg: Added — `fetchTunnelID` now checks `u.AccountType == serviceability.UserType` after deserializing and returns an error if the PDA resolves to a non-user account, so we never return a garbage `TunnelId` on a layout/account mismatch.

tools/stress/device-orchestrator/pkg/agent/ssh.go
elitegreg: Yes — changed it to fail the run. The agent is required telemetry, so a broken stream now aborts the run rather than silently degrading to missing runlog rows:

- `scanLines` returns its `scanner.Err()` (and logs at Error, not Warn). The reader goroutines record it via a new `SSH.setStreamErr`, but only if `ctx.Err() == nil`, so an error provoked by our own teardown (closing the session on ctx cancel) is not misreported as a stream failure.
- Added `Err() error` to the `Runner` interface (noop returns nil). After `consumeAgentEvents` drains the closed Events channel, the consumer checks `Agent.Err()`; if non-nil it cancels a new `WithCancelCause` run context, which aborts the provisioning loop mid-run.
- `deprovision` still runs (it derives from the original `ctx`, not the run context), so cleanup happens regardless, and `Run` returns the agent error in preference to the resulting `context.Canceled`.

Added `TestRun_AgentStreamErrorFailsRun` (fails the stream after the first create) verifying `Run` surfaces the error and all created users are deprovisioned. Verified with `go test -race`.

tools/stress/device-orchestrator/cmd/device-orchestrator/main.go
elitegreg: Yes — noop is the explicit offline-testing path (and the unit tests use it as a stand-in runner), so I kept it. But the wiring around it had a footgun, made worse now that the agent is required telemetry: if you forgot the SSH flags, `selectAgentRunner` silently fell back to noop with just a warning, producing a run with no `pre_commit_log` / `applied` rows.

Tightened it so noop is only reachable when explicitly requested:
- When `--no-agent` is unset, `--dut-ssh-host` and `--dut-ssh-key` are now validated as required (hard error) alongside the other required flags.
- `selectAgentRunner` dropped the silent fallback branch: `--no-agent` → noop, otherwise → SSH runner.

So you either get real telemetry or you have explicitly opted out — never a silent degrade.

nikw9944 (approved):

PR #3781, opened 2026-05-28
feat(controller): enable eos-native gnmi provider

vihu:
# Summary

- Enables the Arista `eos-native` gNMI provider in the controller EOS config template
- Renders only the local `management api gnmi` provider block needed by the telemetry agent tunnel path
- Avoids opening remote gRPC access on DIA interfaces
- Updates controller render fixtures, e2e expected config fixtures, and focused render coverage

## Testing

Validated controller rendering behavior for:

- the gNMI management block is rendered with `provider eos-native`
- no remote gNMI listener config is rendered (`transport grpc`, `port 57400`, `listen-addresses`)
- no explicit TCP/57400 control-plane ACL permit is added
- controller render fixtures match the simplified config

Closes malbeclabs/infra#1414

### DIFF
--- a/controlplane/controller/internal/controller/render_test.go
+++ b/controlplane/controller/internal/controller/render_test.go
@@ -12,6 +12,49 @@ import (
 	"github.com/malbeclabs/doublezero/smartcontract/sdk/go/serviceability"
 )
 
+func TestRenderGNMIManagementProvider(t *testing.T) {
+	data := templateData{
+		Strings:                  StringsHelper{},
+		MulticastGroupBlock:      "239.0.0.0/24",
+		TelemetryTWAMPListenPort: 862,
+		LocalASN:                 65342,
+		UnicastVrfs:              []uint16{1},
+		Device: &Device{
+			PublicIP:        net.IP{7, 7, 7, 7},
+			Vpn4vLoopbackIP: net.IP{14, 14, 14, 14},
+			IsisNet:         "49.0000.0e0e.0e0e.0000.00",
+			ExchangeCode:    "tst",
+			BgpCommunity:    10050,
+			Interfaces:      []Interface{},
+		},
+	}
+
+	got, err := renderConfig(data)
+	if err != nil {
+		t.Fatalf("renderConfig() error: %v", err)
+	}
+
+	want := `management api gnmi
+   provider eos-native
+!`
+	if count := strings.Count(got, "management api gnmi"); count != 1 {
+		t.Fatalf("rendered %d gNMI management blocks; want 1\nconfig:\n%s", count, got)
+	}
+	if !strings.Contains(got, want) {
+		t.Fatalf("rendered config missing gNMI provider block %q\nconfig:\n%s", want, got)
+	}
+	for _, forbidden := range []string{
+		"   85 permit tcp any any eq 57400",
+		"transport grpc",
+		"port 57400",
+		"listen-addresses",
+	} {
+		if strings.Contains(got, forbidden) {
+			t.Fatalf("rendered remote gNMI server config %q; want only local eos-native provider\nconfig:\n%s", forbidden, got)
+		}
+	}
+}
+
 func TestRenderConfig(t *testing.T) {
 	tests := []struct {
 		Name        string
--- a/controlplane/controller/internal/controller/templates/tunnel.tmpl
+++ b/controlplane/controller/internal/controller/templates/tunnel.tmpl
@@ -84,6 +84,9 @@ ip access-list MAIN-CONTROL-PLANE-ACL
 system control-plane
    ip access-group MAIN-CONTROL-PLANE-ACL in
 !
+management api gnmi
+   provider eos-native
+!
 {{- range .Device.Interfaces }}
 interface {{ .Name }}
    {{- if .IsPhysical }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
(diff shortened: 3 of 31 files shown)


ben-malbeclabs (approved):

bgm-malbeclabs (commented):
Three findings from a recall-mode review — multi-DIA duplication is the load-bearing one; the others are worth a look before merge.

controlplane/controller/internal/controller/templates/tunnel.tmpl
```diff
@@ -84,6 +85,19 @@ ip access-list MAIN-CONTROL-PLANE-ACL
...
+   transport grpc default
+      port 57400
+{{- if $.Device.MgmtVrf }}
+      vrf {{ $.Device.MgmtVrf }}
+{{- end }}
+      listen-addresses {{ $iface.Ip.Addr }}
+   provider eos-native
+!
+{{- end }}
+{{- end }}
```
bgm-malbeclabs: **Multiple DIA interfaces on one device render duplicate top-level `management api gnmi` blocks.**

The entire block from `management api gnmi` through `!` is inside `range $iface := .Device.Interfaces`, so a device with two `IsDIA && Ip.IsValid` interfaces emits two separate top-level sections — each with its own `transport grpc default` and `port 57400`. On EOS those settings are singletons under `management api gnmi`, so last-write-wins: all but the last DIA interface's `listen-addresses` is silently dropped.

Nothing caps DIA-per-device today: `smartcontract/programs/doublezero-serviceability/src/processors/device/interface/create.rs` (lines ~156–181) validates CYOA/DIA pairing, MTU, and bandwidth but never counts existing DIA interfaces on the device, and `controlplane/controller/internal/controller/server.go` (lines ~197–222) appends every interface without filtering. So this triggers the moment an operator creates a second DIA on the same box.

Fix sketch: hoist the block out of the range and emit one `listen-addresses` line per DIA inside a single shared `transport grpc default`:

```
{{- $hasDia := false }}
{{- range $iface := .Device.Interfaces }}{{- if and $iface.IsDIA $iface.Ip.IsValid }}{{- $hasDia = true }}{{- end }}{{- end }}
{{- if $hasDia }}
management api gnmi
   transport grpc default
      port 57400
{{- if $.Device.MgmtVrf }}
      vrf {{ $.Device.MgmtVrf }}
{{- end }}
{{- range $iface := .Device.Interfaces }}{{- if and $iface.IsDIA $iface.Ip.IsValid }}
      listen-addresses {{ $iface.Ip.Addr }}
{{- end }}{{- end }}
   provider eos-native
!
{{- end }}
```

controlplane/controller/internal/controller/templates/tunnel.tmpl
```diff
@@ -84,6 +85,19 @@ ip access-list MAIN-CONTROL-PLANE-ACL
 system control-plane
    ip access-group MAIN-CONTROL-PLANE-ACL in
 !
+{{- range $iface := .Device.Interfaces }}
+{{- if and $iface.IsDIA $iface.Ip.IsValid }}
+management api gnmi
+   transport grpc default
```
bgm-malbeclabs: **Plaintext gNMI on a publicly-reachable address.** `transport grpc default` has no `ssl profile` (and no explicit `no ssl profile` to mark the choice as intentional), and the new ACL line `   85 permit tcp any any eq 57400` (above) permits TCP/57400 from `any` source. `listen-addresses` is the DIA interface IP — the public-internet-facing CYOA/DIA address (that's why `bgpMartianNets` excludes it from BGP). End result: anyone with TCP reachability to the DIA address can speak gNMI without encryption.

If AAA is configured at the EOS layer, writes are still protected, but credentials and config payloads travel in plaintext. Worth confirming the security model — and, at minimum, making the no-TLS choice explicit in the rendered config so future readers see it was deliberate.

controlplane/controller/internal/controller/render_test.go
```diff
@@ -12,6 +12,63 @@ import (
...
+		t.Fatalf("rendered config missing gNMI management server block %q\nconfig:\n%s", want, got)
+	}
+	if strings.Contains(got, "listen-addresses 192.0.2.10/31") {
+		t.Fatal("gNMI listen-addresses rendered CIDR prefix; want host address only")
+	}
+	if !strings.Contains(got, "   85 permit tcp any any eq 57400") {
+		t.Fatal("control-plane ACL missing explicit TCP/57400 permit")
+	}
+
+	baseData.Device.Interfaces[0].IsDIA = false
```
bgm-malbeclabs: **Coverage gap (cheap to close):** the negative case flips `IsDIA = false` but never exercises `MgmtVrf = ""` with `IsDIA = true`. A future template edit that drops the `{{- if $.Device.MgmtVrf }}` guard would render `      vrf ` (empty value) for every DIA-without-mgmt-VRF device; today the only fixture that exercises that combination is `interfaces.txt`, whose golden diff doesn't immediately point at the gNMI block.

Suggest a third sub-case here that clears `MgmtVrf` on `baseData.Device` and asserts `      vrf ` is absent from `got`.

controlplane/controller/internal/controller/templates/tunnel.tmpl
vihu: Tagging @ben-malbeclabs for advice here

vihu:
@bgm-malbeclabs I have reworked this PR, with significant simplifications as suggested by @ben-malbeclabs

ben-malbeclabs:
> @bgm-malbeclabs I have reworked this PR, with significant simplifications as suggested by @ben-malbeclabs

Thanks, I approve.

PR #3788, opened 2026-05-28
installer: enable shell autocomplete on installation of doublezero

ben-dz:
## Summary of Changes
- Generate bash, zsh, and fish completion scripts as a post-build hook for both the testnet and mainnet-beta `doublezero` CLI builds.
- Package the generated completions into the nfpm output so they install to the standard system paths (`/usr/share/bash-completion/completions/`, `/usr/share/zsh/site-functions/`, `/usr/share/fish/vendor_completions.d/`).
- Recommend `bash-completion` as a package dependency so deb/rpm installers pull it in when available.

## Testing Verification
- built a snapshot, deployed it and tested funcitonality.


juan-malbeclabs (approved):
LGTM

PR #3800, opened 2026-05-29
geoprobe/mincache: guarded backup windowed-minimum

ben-dz:
## Summary of Changes
Addresses a negative behavior seen in another use of the MinCache approach:
- Reworks the geoprobe `MinCache` best/backup eviction into a **guarded-backup** pattern. A backup is now collected only while `best` is within its final `maxAge/2` ("guard") window, so when `best` expires the promoted value is always a recent-window minimum instead of a stale fallback that could be nearly `maxAge` old.
- A new record low resets `best` and clears `backup`, keeping the backup anchored to the current best's window rather than carrying forward an unrelated older sample.
- Expiry now promotes in a loop, so a backup that is itself already expired can't be promoted into `best`.
- `Best` / `BestRttNs` become pure read-through accessors (via a shared `lowestLocked` helper) that return the lower of the two non-expired slots without mutating or promoting.

## Diff Breakdown
| Core logic |     1 | +44 / -41   |  +3  |
| Tests      |     1 | +157 / -51  | +106 |
| **Total**  |     2 | +201 / -92  | +109 |

Small net change to the cache logic, with the bulk of the diff being expanded test coverage for the new guard/promote/reset semantics.

<details>

- [`controlplane/telemetry/internal/geoprobe/mincache.go`](https://github.com/malbeclabs/doublezero/pull/3800/files#diff-ce2b4e69e347fab3def697160e5441f4b7f24570be4f0bbbe33b8d6e8afb9c29) — guarded-backup collection, loop-based promotion on expiry, record-low reset, and read-through `lowestLocked` accessors
- [`controlplane/telemetry/internal/geoprobe/mincache_test.go`](https://github.com/malbeclabs/doublezero/pull/3800/files#diff-d620366cfc2ddf245a1aaa012161981a829a4ea25ae69396bd5f27d8342f17fc) — tests for guard-window backup capture, no-stale-promotion, record-low reset, and read-through semantics

</details>

## Testing Verification
- Unit tests in `mincache_test.go` cover the guard-window cases: a backup outside the guard is rejected, a backup inside the guard is promoted on `best` expiry, an expired backup is not promoted, and a new record low clears the backup.
- Verified `Best`/`BestRttNs` return the lower non-expired slot without triggering promotion (read-through semantics).


controlplane/telemetry/internal/geoprobe/mincache.go
```diff
@@ -77,97 +78,99 @@ func NewMinCache[T any](maxAge time.Duration, rttFunc func(T) uint64) *MinCache[
...
-	// Capture previous best before any mutation.
 	var info UpdateInfo
 	if !c.best.expiredAt(now, c.maxAge) {
 		info.PrevBestRttNs = c.best.rttNs
 		info.HadPrevBest = true
 	}
 
-	if c.best.expiredAt(now, c.maxAge) {
+	// Expire + promote: when best expires, the backup (a recent-window
+	// min) takes over. Loop in case a promoted entry is itself expired.
```
nikw9944, line 102: The comment says "Loop in case a promoted entry is itself expired", which is directionally correct but not quite accurate since the loop body only promotes the backup after verifying `!c.backup.expiredAt(...)`.

controlplane/telemetry/internal/geoprobe/mincache.go
nikw9944, line 38: "Zero if there was no live previous best; an expired best is treated as absent"

nikw9944 (approved):
After this change, info.HadPrevBest is now only set when best is not expired at call time, and info.Promoted is only set when best was expired. Therefore it appears that [`Promoted && HadPrevBest`](https://github.com/malbeclabs/doublezero/blob/main/controlplane/telemetry/cmd/geoprobe-target/main.go#L377) is in geoprob-target/main.go is now unreachable.

PR #3805, opened 2026-05-29
sdk: build CLI clients directly from CliContext (from_context)

juan-malbeclabs:
## Summary of Changes
- Add `DZClient::from_context` and `GeoClient::from_context` to the Rust SDK, building clients directly from a resolved RFC-20 `CliContext` instead of re-reading `config.yml` and re-applying moniker conversion. The context already carries fully resolved ledger RPC/WS URLs and program IDs, so the context becomes the single source of truth and the binary's previous double-resolution is removed.
- Preserve keypair precedence exactly (CLI flag > `DOUBLEZERO_KEYPAIR` > stdin > context keypair path > default): the raw `--keypair` flag is the highest-precedence source and the context keypair path is only the low-precedence fallback, so the env var still wins.
- Gate the new constructors and the `doublezero-cli-core` dependency behind a `cli-context` cargo feature so non-CLI SDK consumers (controlplane, telemetry, e2e) keep a dependency-light default build.
- Wire the `doublezero` binary to `from_context`, removing the legacy `DZClient::new(Option<String>, ...)` bridge. `DZClient::new` / `GeoClient::new` are unchanged for callers that build no `CliContext` (e.g. `controlplane/doublezero-admin`).

Related RFC: [`rfcs/rfc20-cli-standardization.md`](rfcs/rfc20-cli-standardization.md)

## Diff Breakdown
| Core logic   |     3 | +85 / -16   | +69  |
| Tests        |     1 | +62 / -0    | +62  |
| Config/build |     2 | +6 / -1     | +5   |
| Docs         |     1 | +4 / -0     | +4   |
| Generated    |     1 | +1 / -0     | +1   |
| **Total**    |     8 | +162 / -17  | +145 |

Small, focused change: two new SDK constructors plus binary wiring, with the bulk of additions being the constructors and their tests.

<details>

- `smartcontract/sdk/rs/src/client.rs` — add `DZClient::from_context` (no config re-read / moniker conversion) plus feature-gated unit tests for resolved-value consumption and keypair env-var precedence
- `smartcontract/sdk/rs/src/geolocation/client.rs` — add `GeoClient::from_context` using the context's ledger RPC URL and geolocation program ID
- `client/doublezero/src/main.rs` — construct both clients via `from_context`, drop the legacy bridge locals
- `smartcontract/sdk/rs/Cargo.toml` — optional `doublezero-cli-core` dep behind a new `cli-context` feature
- `smartcontract/sdk/rs/src/config.rs` — make `default_keypair_path` `pub(crate)` for the fallback path

</details>

## Testing Verification
- New SDK unit tests (`cargo test -p doublezero_sdk --features cli-context`): one asserts `from_context` consumes the context's RPC/WS/program-ID verbatim with no `config.yml` present, and one asserts `DOUBLEZERO_KEYPAIR` still wins over a (bogus) context keypair path, guarding the masking hazard.
- End-to-end against testnet with the rebuilt binary: `doublezero location list` (exercises `DZClient::from_context`) and `doublezero geolocation probe list` (exercises `GeoClient::from_context`) both returned live data.
- Verified the default SDK build (no `cli-context` feature) still compiles, so non-CLI consumers are unaffected.


CHANGELOG.md
```diff
@@ -47,6 +47,10 @@ All notable changes to this project will be documented in this file.
...
+- SDK (Rust)
+  - Add `DZClient::from_context` and `GeoClient::from_context`, which build clients directly from a resolved RFC-20 `CliContext` instead of re-reading `~/.config/doublezero/cli/config.yml` and re-applying moniker conversion. The context already carries the fully resolved ledger RPC/WS URLs and program IDs, so these constructors consume them verbatim, making the context the single source of truth and removing the double-resolution the binary previously incurred. Keypair precedence is preserved exactly (CLI flag > `DOUBLEZERO_KEYPAIR` > stdin > context keypair path > default): the raw `--keypair` flag is passed as the highest-precedence source and the context keypair path is used only as the low-precedence fallback, so the env var still wins. The new constructors and their `doublezero-cli-core` dependency are gated behind a `cli-context` cargo feature so non-CLI SDK consumers (controlplane, telemetry, e2e) keep a dependency-light default build. `DZClient::new` / `GeoClient::new` are unchanged for callers that do not build a `CliContext` (e.g. `controlplane/doublezero-admin`).
+- CLI
```
ben-dz: There are already CLI and SDK sections. Fold the changelog entries into them.

juan-malbeclabs: The review is addressed. The two duplicate header blocks are gone:

  - The SDK (Rust) entry (from_context constructors) now sits as a second bullet under the existing - SDK (Rust) section (next to the skip_preflight bullet).
  - The CLI entry (constructing clients via from_context in the binary) is folded into the end of the existing large RFC-20 CLI section, where the rest of the CliContext integration work
  lives.

smartcontract/sdk/rs/src/geolocation/client.rs
```diff
@@ -115,6 +115,38 @@ impl GeoClient {
             program_id,
         })
     }
+
+    /// Build a `GeoClient` from a resolved RFC-20 [`CliContext`].
+    ///
+    /// Mirrors [`crate::DZClient::from_context`]: no `config.yml` read and no
+    /// moniker conversion. The ledger RPC URL and the geolocation program ID
+    /// are taken verbatim from the context (the context already folds the
```
ben-dz, line 123: This doc claim isn't quite true today, and it hides a regression. The context build in client/doublezero/src/main.rs resolves --geo-program-id with app.geo_program_id.as_deref().and_then(|s| s.parse::<Pubkey>().ok()) — a raw pubkey parse. So a moniker (devnet/testnet/mainnet-beta/local) fails parse::<Pubkey>(), gets swallowed by .ok() in main.rs, and never reaches ctx.geolocation_program_id, which then falls back to the env default. A typo'd pubkey is silently dropped the same way.

Before this PR, GeoClient::new(url, app.geo_program_id, ..) ran the value through convert_geo_program_moniker, so --geo-program-id testnet worked. Switching to from_context(ctx) drops that conversion, so the flag silently no-ops for monikers. (--program-id has the identical parse and is already affected on main.)

ben-dz (approved):

PR #3810, opened 2026-05-29
cli: honor build-configured default environment when no config.yml is present

ben-dz:
## Summary
- The RFC-20 context-build in the `doublezero` binary resolved the no-config / no-`--env` fallback environment via `Environment::default()` — which is always `Devnet`, independent of the build. The legacy `DZClient::new` path instead keyed off the compile-time `DEFAULT_ENVIRONMENT` (`Testnet` by default, `MainnetBeta` under the `default-mainnet-beta` feature). The migration silently dropped that.
- **Impact:** a plain (testnet) build with no `config.yml` and no `--env` silently targeted Devnet's URLs and program IDs instead of Testnet. Setting `DZ_LEDGER_RPC_URL` made it worse, not better — that env var overrides only the URLs inside `Environment::config()`, so the binary would hit the testnet ledger with Devnet program IDs.
- Expose `doublezero_sdk::default_environment()` (the build-feature-aware default) and route the binary's environment resolution through it, restoring the legacy behavior exactly.
- Extract the previously-inline resolution into a testable `resolve_environment` helper — the lack of a seam here is why nothing caught the regression.

Deliberately did **not** change `Environment::default()`'s `#[default] Devnet` derive, which is referenced elsewhere (e.g. the `build_without_env` placeholder in `doublezero-cli-core`); routing the binary through `DEFAULT_ENVIRONMENT` keeps the change surgical.

Related RFC: [`rfcs/rfc20-cli-standardization.md`](rfcs/rfc20-cli-standardization.md)

## Testing Verification
- New unit tests in `client/doublezero` assert: no `--env` + no config resolves to `default_environment()` (fails before the fix — returned Devnet); a persisted config's `program_id` selects its environment; a persisted config without a `program_id` falls back to the build default; an explicit `--env` flag wins.
- `make rust-lint` clean (rustfmt + clippy); full `doublezero` bin test suite passes (137 tests).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file.
 
 - SDK (Go)
   - Add CreateUser / DeleteUser to the serviceability executor with cross-language wire-format fixtures and four new PDA helpers (GetUserPDA, GetAccessPassPDA, GetTunnelIdsPDA, GetDzPrefixBlockPDA)
+- CLI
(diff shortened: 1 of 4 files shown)


elitegreg (approved):

juan-malbeclabs (approved):
LGTM

PR #3811, opened 2026-05-29
cli: derive environment from program ID in config set

juan-malbeclabs:
## Summary of Changes

- `doublezero config set --program-id <PUBKEY>` now derives the full environment configuration (RPC URL, WS URL, geo program ID) when the program ID matches a known environment (testnet/devnet/mainnet-beta/local), so setting the program ID alone is enough to switch environments
- Explicit `--url`/`--ws`/`--geo-program-id` flags still take precedence over the derived values
- An unrecognized program ID keeps the previous behavior of setting only the program ID field

## Why

The environment a CLI invocation targets is resolved from the configured serviceability program ID via `Environment::from_program_id(...).unwrap_or_default()`. Previously `--program-id` set only that field, leaving the RPC/WS URLs and geo program ID untouched, so it was not a complete way to switch environments. This makes `--program-id` a first-class equivalent to `--env` for selecting a known environment, which is useful for hosts that provision the CLI by program ID.

## Diff Breakdown
| Core logic   |     1 | +20 / -1    | +19  |
| Tests        |     1 | +90 / -0    | +90  |
| Docs         |     1 | +2 / -0     | +2   |
| **Total**    |     2 | +112 / -1   | +111 |

Almost entirely tests around a small change to the `config set` environment-resolution branch.

<details>

- `smartcontract/cli/src/config/set.rs` — add a branch that, when `--program-id` resolves to a known environment via `Environment::from_program_id`, fills in RPC URL, WS URL, and geo program ID from that environment's config (respecting explicit overrides); plus two new unit tests
- `CHANGELOG.md` — document the new `config set --program-id` behavior under Unreleased

</details>

## Testing Verification

- Added `test_cli_config_set_program_id_known_env_derives_full_config` (testnet program ID derives RPC/WS/geo) and `test_cli_config_set_program_id_known_env_respects_explicit_url` (explicit `--url` wins, geo still derived); both written test-first and watched fail before implementing
- Existing `test_cli_config_set_program_id` (unrecognized pubkey) still passes, confirming backward compatibility; all 14 config-set tests green
- Manually verified against the built binary: testnet pubkey derives testnet geo program ID, devnet pubkey derives devnet geo program ID, and an unknown pubkey leaves the rest of the config unchanged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file.
 
 - SDK (Go)
   - Add CreateUser / DeleteUser to the serviceability executor with cross-language wire-format fixtures and four new PDA helpers (GetUserPDA, GetAccessPassPDA, GetTunnelIdsPDA, GetDzPrefixBlockPDA)
+- CLI
+  - Honor the build-configured default environment (`Testnet` by default, `MainnetBeta` under the `default-mainnet-beta` feature) when neither `--env` nor a persisted `config.yml` selects one. The RFC-20 context-build previously fell back to `Environment::default()`, which is always `Devnet` regardless of the build, so a testnet build with no config silently targeted Devnet's ledger URLs and program IDs. The binary now resolves the fallback through the new `doublezero_sdk::default_environment()`, matching the legacy `DZClient::new` defaults (`default_program_id`, `ClientConfig::default`) which already key off the compiled-in environment ([#3810](https://github.com/malbeclabs/doublezero/pull/3810))
 
 ## [v0.25.0](https://github.com/malbeclabs/doublezero/compare/client/v0.24.0...client/v0.25.0) - 2026-05-29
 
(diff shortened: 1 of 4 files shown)


juan-malbeclabs:
Closing this PR. The `--program-id` auto-derivation conflicts with TestE2E_BackwardCompatibility, which intentionally sets a known-environment program ID against a local clone ledger while keeping the local RPC URL. Deriving the public RPC URL from the program ID broke that case ("no available credits").

No code change is needed for the original goal: `doublezero config set --env testnet` already sets the full environment (RPC URL, WS URL, program ID, geo program ID) unambiguously. The testnet QA hosts should run that (ideally via the qa_agent / doublezero_client ansible role) to align the CLI with the testnet daemon.

juan-malbeclabs:
Closed because the `--program-id` auto-derivation conflicts with TestE2E_BackwardCompatibility (it sets a known-env program ID against a local clone ledger while keeping the local RPC URL; deriving the public URL from the program ID caused the "no available credits" failure).

No code change is needed for the goal: `doublezero config set --env testnet` already sets the full environment. The testnet QA hosts should run that (via the qa_agent / doublezero_client ansible role) to align the CLI with the testnet daemon.

PR #3812, opened 2026-05-31
cli: let per-field flags override the --env base

juan-malbeclabs:
## Summary

- Make `--env` a base that the per-field flags override, instead of being mutually exclusive with them. `--env <name>` resolves the whole network (ledger URL, WS URL, Solana L1 URL, serviceability + geolocation program IDs); `--url`, `--ws`, `--solana-url`, `--program-id`, `--geo-program-id` each override only their own value on top (precedence: explicit flag > `--env`). This applies to both the global `doublezero` flags and `doublezero config set`.
- This unblocks `--env local --program-id <X>`, which previously failed: the global flag rejected it at the clap layer (`ArgumentConflict`), and `config set` printed `Invalid flag combination` and exited without writing.
- Adopt the new behavior in e2e: the manager and client containers configure the CLI with `--env local` plus per-field overrides for the container ledger URLs and the deployed serviceability program ID (and geolocation on the manager).

## Why

Per RFC-20's override hierarchy (`explicit CLI flag > env var > value from --env`), the per-field flags are meant to layer on top of `--env`, not conflict with it. The implementation enforced mutual exclusion, which contradicted the documented contract and prevented configuring a known environment while overriding a single field (e.g. pointing `local` at a container ledger or a cloned program). The `CliContextBuilder` already resolved values last-write-wins, so the global flag only needed its `conflicts_with_all` removed; `config set` now resolves the env base then applies per-field overrides.

The e2e override of `--program-id` is required because the sentinel multicast-publisher stack deploys serviceability at a generated keypair and the compatibility tests clone testnet/mainnet program IDs; for standard stacks the env defaults already match the fixed localnet pubkeys.

RFC: [rfcs/rfc20-cli-standardization.md](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc20-cli-standardization.md) (§override hierarchy).

## Testing Verification

- `config set --env local --url <container> --ws <container> --program-id <generated>` writes a `config.yml` with the overridden URLs and the generated program ID while serviceability/geolocation defaults come from `local` — confirmed with the compiled binary for the standard, generated-keypair, and cloned-program cases.
- `TestE2E_SentinelMulticastPublisherCreatesPublishers` passes. This is the one stack that deploys serviceability at a generated keypair, so it exercises the `--env local` + `--program-id` override end-to-end: device, multicast group, and three IBRL users created, all multicast publishers verified `Activated`.
- Updated unit tests: global-flag parsing (`env_combines_with_url/_ws/_solana_url/_program_id/_geo_program_id`) and `config set` override tests (`test_cli_config_set_env_with_program_id_overrides`, `..._geo_program_id_overrides`).


ben-dz (approved):
lgtm

PR #3813, opened 2026-06-01
geolocation/cli: show codes instead of pubkeys in user get and probe get

ben-dz:
## Summary of Changes
- `geolocation user get`: the Targets table `probe` column now shows the human-readable probe code instead of the raw geoprobe pubkey, resolved via `list_geo_probes` with a pubkey fallback. JSON gains a `probe` code field while retaining the canonical `geoprobe_pk`.
- `geolocation probe get`: the `exchange` field now shows the exchange code instead of the raw pubkey, resolved via `list_exchanges` with a pubkey fallback — making it consistent with `geolocation probe list`, which already does this.

## Diff Breakdown
| Core logic   |     2 | +25 / -9    |  +16 |
| Tests        |     2 | +168 / -0   | +168 |
| **Total**    |     2 | +193 / -9   | +184 |

Two small CLI display fixes in the geolocation crate; the bulk of the diff is test coverage for code resolution and pubkey fallback.

<details>

- `crates/doublezero-geolocation-cli/src/user/get.rs` — resolve `geoprobe_pk` → probe code in the Targets table; keep `geoprobe_pk` in JSON, add `probe` code field; tests for code rendering and fallback
- `crates/doublezero-geolocation-cli/src/probe/get.rs` — resolve `exchange_pk` → exchange code, matching `probe list`; tests for code rendering and fallback

</details>

## Testing Verification
- `user get`: table asserts the probe code (`ams-probe-01`) and not the pubkey; new fallback test when the probe isn't found; compact-JSON test asserts both `probe` code and retained `geoprobe_pk`
- `probe get`: table asserts the exchange code (`ams`) and not the pubkey; new fallback test when the exchange isn't found; JSON/compact tests updated to expect the code
- All `user::get` and `probe::get` tests pass; `make rust-lint` clean in the dev container

### DIFF
--- a/crates/doublezero-geolocation-cli/src/probe/get.rs
+++ b/crates/doublezero-geolocation-cli/src/probe/get.rs
@@ -36,8 +36,7 @@ struct GeoProbeGetDisplay {
     pub code: String,
     #[serde(serialize_with = "serializer::serialize_pubkey_as_string")]
     pub owner: Pubkey,
-    #[serde(serialize_with = "serializer::serialize_pubkey_as_string")]
-    pub exchange: Pubkey,
+    pub exchange: String,
     pub public_ip: Ipv4Addr,
     pub port: u16,
     #[serde(serialize_with = "serialize_pubkey_vec_as_string_array")]
@@ -64,6 +63,12 @@ impl GetGeoProbeCliCommand {
             pubkey_or_code: self.probe,
         })?;
 
+        let exchanges = client.list_exchanges()?;
+        let exchange = exchanges
+            .get(&probe.exchange_pk)
+            .map(|ex| ex.code.clone())
+            .unwrap_or_else(|| probe.exchange_pk.to_string());
+
         let parent_devices_display = format!(
             "[{}]",
             probe
@@ -77,7 +82,7 @@ impl GetGeoProbeCliCommand {
             account: pubkey,
             code: probe.code,
             owner: probe.owner,
-            exchange: probe.exchange_pk,
+            exchange,
             public_ip: probe.public_ip,
             port: probe.location_offset_port,
             parent_devices: probe.parent_devices,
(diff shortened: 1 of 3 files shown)


crates/doublezero-geolocation-cli/src/probe/get.rs
```diff
@@ -36,8 +36,7 @@ struct GeoProbeGetDisplay {
     pub code: String,
     #[serde(serialize_with = "serializer::serialize_pubkey_as_string")]
     pub owner: Pubkey,
-    #[serde(serialize_with = "serializer::serialize_pubkey_as_string")]
-    pub exchange: Pubkey,
+    pub exchange: String,
```
juan-malbeclabs: This changes exchange from Pubkey to String and stores the resolved code in it, but no field retains the exchange pubkey anymore. So probe get --json drops the pubkey entirely, aconsumer reading .exchange as a pubkey now gets a code and has no way to recover the pubkey from the output.
This is also inconsistent with the companion user get change in this PR, which keeps geoprobe_pk and adds a separate probe code field, staying backward compatible. Let's mirror that here: keep the pubkey in the JSON (e.g. exchange stays the pubkey and add an exchange_code, or rename the pubkey to exchange_pk and add exchange), and let the human table show only the code. As written this is a silent breaking change to machine-readable output.

crates/doublezero-geolocation-cli/src/probe/get.rs
```diff
@@ -64,6 +63,12 @@ impl GetGeoProbeCliCommand {
             pubkey_or_code: self.probe,
         })?;
 
+        let exchanges = client.list_exchanges()?;
```
juan-malbeclabs: The new list_exchanges() call is propagated with ?, so a failure there now fails the whole command. Previously these verbs always succeeded and rendered the pubkey. Since the "not found in the map" case already falls back to the pubkey, an error from the list call should be treated the same way (best-effort, fall back to the pubkey) rather than aborting, the code resolution is a display nicety, not a hard requirement for the command to produce useful output.

crates/doublezero-geolocation-cli/src/user/get.rs
```diff
@@ -68,6 +72,8 @@ impl GetGeolocationUserCliCommand {
             pubkey_or_code: self.user,
         })?;
 
+        let probes = client.list_geo_probes(ListGeoProbeCommand)?;
```
juan-malbeclabs: The new list_geo_probes() call is propagated with ?, so a failure there now fails the whole command. Previously these verbs always succeeded and rendered the pubkey. Since the "not found in the map" case already falls back to the pubkey, an error from the list call should be treated the same way (best-effort, fall back to the pubkey) rather than aborting, the code resolution is a display nicety, not a hard requirement for the command to produce useful output.

juan-malbeclabs (changes requested):

crates/doublezero-geolocation-cli/src/probe/get.rs
ben-dz: added separate exchange_code.

crates/doublezero-geolocation-cli/src/user/get.rs
ben-dz: is now best-effort, falls back.

crates/doublezero-geolocation-cli/src/probe/get.rs
ben-dz: is now best effort and falls back.

juan-malbeclabs (approved):
LGTM

PR #3814, opened 2026-06-01
tools/stress/device-observer: EOS syslog + tail-based readers

### LINKED ISSUE #3795: 3747-3: EOS syslog + tail-based readers (~250 LOC code)
Parent issue: #3747

Split from plan review of #3747 (estimated ~250 lines).

**PR stack:** this issue's branch should target `nikw9944/doublezero-3794` (the previous sub-issue's branch).

---

- Creates `internal/tailer/tailer.go` (shared poll-based tailer with
  open + offset + recreate/truncate detection + partial-line buffering).
- Creates `internal/loggingtail/eos.go` (`show logging last <window>` poller
  with parse + dedupe + NDJSON output).
- Creates `internal/loggingtail/agent.go` (uses `tailer.Tailer` against
  `<working-dir>/orchestrator.agent.log`; emits typed events).
- Creates `internal/runlog/reader.go` (uses `tailer.Tailer` against
  `<working-dir>/orchestrator-runlog.json`; owns pending-submit maps and
  bounded duration rings; exposes `ProvisionDurations(window)` and
  `DeprovisionDurations(window)`).
- Updates README with EOS-log / agent-log / runlog sections.

nikw9944:
## Summary

* Add `internal/tailer`, a shared poll-based file tailer with inode-based rotation detection, size-based truncation detection, partial-line buffering, and a 1 MiB oversize-line cap
* Add `internal/loggingtail/eos.go`: `EOSPoller` polls `show logging last <N> seconds` over eAPI each tick, parses Arista syslog lines, deduplicates across the rolling overlap window using the full raw line as key, and appends NDJSON rows to `observer.eos_logging.json`
* Add `internal/loggingtail/agent.go`: `AgentTail` tails `orchestrator.agent.log`, appends NDJSON to `observer.agent_log.json`, and exposes `Snapshot()` with `LastLineAt` and per-pattern match counts for the abort decider
* Add `internal/runlog/reader.go`: `Reader` tails `orchestrator-runlog.jsonl`, pairs `submit`/`activate` and `deprovision_submit`/`deprovision_activate` events into bounded duration rings (cap 1024), exposes `ProvisionDurations(window)` / `DeprovisionDurations(window)`, and caps pending-submit maps at 4096 entries with oldest-eviction
* Wire the new constructors in `cmd/device-observer/main.go`; the EOS lookback is derived inside `NewEOS` as `max(2*interval, 30s)`

## Testing

* Add unit tests for all new functionality

### DIFF
--- a/tools/stress/device-observer/README.md
+++ b/tools/stress/device-observer/README.md
@@ -137,14 +137,16 @@ Row schema:
   "time": "May 29 12:34:56",
   "severity": "3",
   "facility": "BGP",
-  "message": "BGP-3-NOTIFICATION: ..."
+  "message": "peer 10.0.0.1 went down"
 }
 ```
 
-Lines that don't match the default Arista syslog format (timestamp +
-`FACILITY-SEV-MNEMONIC:` tag) still land in the file with empty `severity`
-and `facility` and the full line under `message`, so unusual formats are
-not silently dropped.
+The parser extracts `facility` and `severity` from the Arista
+`FACILITY-SEV-MNEMONIC:` tag; `message` contains only the text after the
+colon. The mnemonic (e.g. `NOTIFICATION`) is not preserved in any field.
+Lines that don't match the default format still land in the file with
+empty `severity` and `facility` and the full line under `message`, so
+unusual formats are not silently dropped.
 
 ## Agent log output
 
--- a/tools/stress/device-observer/README.md
+++ b/tools/stress/device-observer/README.md
@@ -1,7 +1,7 @@
 # device-observer
 
 > Alpha — eAPI sampler, Prometheus scrape, and log tailers are wired up.
-> The abort decider (PR #3796) is still a no-op stub in this revision.
+> The abort decider is still a no-op stub in this revision.
 
 `device-observer` samples an Arista cEOS device-under-test (DUT) during the
 GRE Tunnel Capacity Study sweep. On every tick of `--sample-interval` it
@@ -60,7 +60,7 @@ observer reads them but does not write them:
 | `orchestrator.agent.log`        | orchestrator | observer |
 
 The abort sentinel at `--abort-file` (default `<working-dir>/abort`) is
-written by the observer (in PR #3796) and read by the orchestrator.
+written by the observer's abort decider and read by the orchestrator.
 
 Filenames use the observer's local clock formatted as ISO 8601 UTC with
 nanosecond precision, with `:` replaced by `-` for filesystem portability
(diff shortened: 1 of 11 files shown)


tools/stress/device-observer/internal/loggingtail/eos.go
```diff
@@ -1,13 +1,171 @@
...
+	tNS := p.now().UTC().UnixNano()
+	parsed, scanErr := parseEOSLog(out, tNS)
+	if scanErr != nil {
+		p.logger.Warn("eos logging parse truncated", "err", scanErr)
+	}
+	p.mu.Lock()
+	next := make(map[string]struct{}, len(parsed))
+	var fresh []eosLine
+	for _, line := range parsed {
+		key := line.Time + "|" + line.Message
```
elitegreg: The dedupe key `line.Time + "|" + line.Message` uses the syslog timestamp at 1-second resolution, so two identical messages in the same second collapse into a single row. Unparseable lines have an empty `Time`, so identical unparseable lines anywhere in the overlap window dedupe to one as well. If repeated/identical log lines are expected and need to be preserved, the key won't distinguish them — flagging in case that matters for the abort decider's counting.

tools/stress/device-observer/cmd/device-observer/main.go
```diff
@@ -110,12 +108,15 @@ func run() error {
 		return err
 	}
 
+	// 2 × interval (min 30 s) so consecutive show-logging windows overlap.
+	eosLookback := max(2**sampleInterval, 30*time.Second)
```
elitegreg: Lookback is computed twice. Here `main.go` sets `eosLookback := max(2**sampleInterval, 30*time.Second)` and passes it in, but `NewEOS` then re-clamps with `max(lookback, 2*interval, eosMinLookback)` — the same policy. Since the constructor already owns the clamp (and receives `interval`), the caller-side computation is redundant and the two can drift. Consider dropping the `lookback` parameter from `NewEOS` (derive it entirely inside), or having `main.go` pass a raw/zero value and let the constructor own the floor.

tools/stress/device-observer/README.md
```diff
@@ -113,12 +115,94 @@ so the file can be consumed with line-oriented tools (`jq -c`, ClickHouse
...
+
+Row schema:
+
+```json
+{
+  "t_ns": 1748520896123456789,
+  "time": "May 29 12:34:56",
+  "severity": "3",
+  "facility": "BGP",
+  "message": "BGP-3-NOTIFICATION: ..."
```
elitegreg: This `message` example is inaccurate. The parser's regex captures `facility`/`severity` from the `FACILITY-SEV-MNEMONIC:` tag and puts only the text *after* the colon into `message` — and it drops the mnemonic entirely. `TestParseHappyPath` confirms `BGP-3-NOTIFICATION: bgp event` → `message == "bgp event"`. So this row is misleading: `"message": "BGP-3-NOTIFICATION: ..."` should be the body only (e.g. `"message": "bgp event"` / `"..."`), and it's worth noting the mnemonic isn't preserved in any field.

elitegreg (approved):

tools/stress/device-observer/internal/loggingtail/eos.go
nikw9944: Good call, updated to use the raw log line as the key, as I do think we would like to know if duplicate lines are emitted

tools/stress/device-observer/cmd/device-observer/main.go
nikw9944: Good call, the constructor now owns it.

PR #3820, opened 2026-06-02
tools/stress: fixes surfaced by running the 1024-user stress harness

nikw9944:
## Summary of Changes
- Add a post-deprovision agent quiescence wait so the orchestrator doesn't kill its SSH session mid-commit. Before this, a 1024-user sweep "succeeded" while leaving 96–192 tunnels stranded on the device because the agent was still applying the final deprovision config when the orchestrator exited.
- Lift the per-device user cap from 525 to ~4096 in `User::validate`. The previous `tunnel_id > 1024` check artificially limited each device to 525 users (start of 500 + 1024 max) despite the TunnelIds resource extension being sized for [500, 4596).
- Surface program errors from finalized transactions in the Go SDK. `waitForTransactionFinalized` was breaking out of the poll loop on `ConfirmationStatus=finalized` without inspecting `status.Err`, so program-level rejections were masked as "post-confirm visibility timeouts." Companion fix: `waitForAccountGone` now treats `ErrNotFound` as success rather than a contradictory "still present: not found."
- New `device_tunnel_gap` observer abort trigger that fires when orchestrator-reported active users exceed device-reported tunnels by ≥4 after a 120s grace, catching silent under-provisioning (e.g. controller slot-cap truncation).
- Pipeline `UsersPerBatch` creates/deletes via `sync.WaitGroup` so a batch of N runs in ~14 s instead of N×14 s (Solana finalization is the bottleneck).
- Observer tuning: 60s startup grace for counter/pattern triggers (absorbs first-apply race), agent_silence threshold raised to 120s (legitimate slow-applies at scale), CPU parser rewritten to consume the structured eAPI shape that EOS actually returns.

## Diff Breakdown
| Core logic   |     8 | +872 / -81    |  +791 |
| Tests        |     8 | +1343 / -29   | +1314 |
| Scaffolding  |     2 | +109 / -56    |   +53 |
| Docs         |     2 | +86 / -12     |   +74 |
| Fixtures     |     1 | +1 / -0       |    +1 |
| **Total**    |    21 | +2411 / -178  | +2233 |

Tests dominate by line count (decider trigger coverage alone is +645); core logic centers on the observer's abort state machine and the orchestrator's sweep loop. Both `main.go` files are pure CLI wiring around new options.

<details>

- [`tools/stress/device-observer/internal/abort/decider.go`](https://github.com/malbeclabs/doublezero/pull/3820/files#diff-bc81ea0a86e5c50de44b161b329e2548a1b39be080e33b76f69a3a88fa838274) — `device_tunnel_gap` trigger, 60s counter/pattern startup grace, `agent_silence` raised to 120s with matching gap-grace alignment
- [`tools/stress/device-orchestrator/pkg/sweep/sweep.go`](https://github.com/malbeclabs/doublezero/pull/3820/files#diff-a864c355728c685805dc490247cb61914299637f4269f384ac807cbd3d2e8ee3) — pipelined `UsersPerBatch` creates/deletes, `quiescenceTracker` + `waitForAgentQuiescence` helper
- [`tools/stress/device-observer/internal/sample/eos.go`](https://github.com/malbeclabs/doublezero/pull/3820/files#diff-a79fa5cfa77b75a18f704d1612270bd09f8469b1eccf2372fa507050a9acf2cb) — structured-eAPI CPU parser (replaces dead text-envelope path) and tunnel-count parser feeding the new gap trigger
- [`tools/stress/device-observer/internal/runlog/reader.go`](https://github.com/malbeclabs/doublezero/pull/3820/files#diff-e225a16cd01dc878f1b5c128215568d9524f7e36fae52d1fe3751b79d19d222b) — `n_after_event` capture + `ActiveUserCount()` accessor for the gap trigger
- [`smartcontract/sdk/go/serviceability/executor.go`](https://github.com/malbeclabs/doublezero/pull/3820/files#diff-7be6c1e59c995ffb57df9bcf3926738f52220f83f5b4e4747fc0dd2d993d1bc8) — `waitForTransactionFinalized` checks `status.Err`, `waitForAccountGone` treats `ErrNotFound` as success
- [`tools/stress/device-orchestrator/pkg/agent/agent.go`](https://github.com/malbeclabs/doublezero/pull/3820/files#diff-87d42ec9f02032a4cd0c2869586a5c8f015cfbfbc2a9dfb25db1bdc18eba2acd) + [`parser.go`](https://github.com/malbeclabs/doublezero/pull/3820/files#diff-276c05c8880aa7db6773188e04cfcd5ba927e14131431d6d14d8c9a849a72df7) — new `EventCommit` activity signal on every commit-finalize so deprovision quiescence is observable (pure-removal diffs emit no Applied)
- [`smartcontract/programs/doublezero-serviceability/src/state/user.rs`](https://github.com/malbeclabs/doublezero/pull/3820/files#diff-0e24a7e1473ca2f2de965d23e72a3b5212b2d7d3255e6042659604eb08ee19f6) — `tunnel_id` validation aligned with TunnelIds extension `[500, 4596)`

</details>

## Testing Verification
- Validated end-to-end against a containerized 1024-user sweep: provision + deprovision completes with `0` tunnels left on the device and all 1024 `applied` events captured in the runlog. The same run without the quiescence wait stranded 96–192 tunnels and dropped 32 events.
- Added a new `TestRun_WaitsForAgentQuiescenceAfterDeprovision` covering the wait blocks for ≥ quiet window, and `TestRun_SkipsQuiescenceWaitWhenNoAppliedObserved` for the noop/--no-agent path.
- Added `TestParser_DeprovisionCommitEmitsCommitOnly` proving the agent parser surfaces `EventCommit` on pure-removal commits — the structural gap that broke the first quiescence-wait iteration.
- Added test fixture `ceos-show-processes-top-once.json` covering the cEOS structured CPU shape the rewritten parser targets.


tools/stress/device-orchestrator/pkg/sweep/sweep_test.go
```diff
@@ -515,3 +550,213 @@ func TestRun_CancellationStopsBetweenUsers(t *testing.T) {
...
+			t.Fatal("sweep did not reach deprovision within 1s")
+		}
+		time.Sleep(time.Millisecond)
+	}
+	// Mark the agent as active right before releasing deprovision so the
+	// quiescence wait sees a recent Applied.
+	ag.Emit(agent.Event{Kind: agent.EventApplied, TunnelID: 500, At: time.Now()})
+
+	gateReleased := time.Now()
+	close(gate)
```
elitegreg, line 728: **Flaky under the parallel suite (~5/6 failures reproduced).** Passes in isolation, but fails repeatedly when the whole package runs:

```
sweep_test.go: "221.135µs" is not greater than or equal to "50ms"
    Run returned 221.135µs after gate release; expected ≥ 50ms (quiet window)
```

Root cause is a test-only race, not a production bug. `ag.Emit(EventApplied)` on the line above followed immediately by `close(gate)` has nothing guaranteeing the consumer goroutine drained that event and called `tracker.markEvent` before `waitForAgentQuiescence` reads `tracker.lastEvent()`. Under CPU contention the consumer isn't scheduled in time, the tracker is empty, `ok == false`, and the wait is *skipped* — Run returns in ~200µs. The run log confirms it: `run-quiesce` printed `"no agent events observed; skipping agent quiescence wait"`.

Suggested deterministic fix: after `ag.Emit(...)`, poll the runlog file for the `applied` row before `close(gate)`. In `consumeAgentEvents`, `tracker.markEvent` runs *before* the runlog write, so the row's presence proves the tracker was marked, closing the race.

tools/stress/device-orchestrator/pkg/sweep/sweep.go
```diff
@@ -266,11 +341,65 @@ func consumeAgentEvents(cfg *Config, registry *tunnelRegistry) {
...
+		now := cfg.Clock.Now()
+		last, _ = tracker.lastEvent()
+		sinceLast := now.Sub(last)
+		if sinceLast >= cfg.AgentQuietWindow {
+			cfg.Logger.Info("sweep: agent quiesced",
+				"quiet_for", sinceLast,
+				"wait_elapsed", now.Sub(start))
+			return
+		}
+		if cfg.AgentQuiescenceTimeout > 0 && !now.Before(deadline) {
```
elitegreg: Minor: the deadline is only enforced when `AgentQuiescenceTimeout > 0`. `validate()` permits `AgentQuiescenceTimeout == 0` alongside a positive `AgentQuietWindow`, so a never-quiescing agent would loop here until `ctx` cancellation rather than honoring a hard cap. `main.go` defaults to 300s so it's not hit in practice — worth either a guard in `validate()` or a doc note on the `AgentQuiescenceTimeout` field that 0 means unbounded (relies on ctx).

smartcontract/sdk/go/serviceability/executor.go
```diff
@@ -674,6 +684,16 @@ func (e *Executor) waitForTransactionFinalized(ctx context.Context, sig solana.S
 		}
 	}
 
+	// A finalized status carries a non-nil Err when the transaction executed
+	// but the program returned an error (e.g. the doublezero-serviceability
+	// "tunnel_id > 1024" rejection at 525 users/device). Without this check
```
elitegreg: Nit: comment still references `"tunnel_id > 1024" rejection at 525 users/device`. Fine as historical context for the bug this fixes, but slightly stale now that the cap is 4596 in this same PR.

elitegreg (approved):
Reviewed the full stack of fixes — well-motivated and clearly documented; each change traces to a concrete failure from the 1024-user sweep. I verified the central claims against the code:

- `user.rs`: the new `tunnel_id >= 4596` cap matches the allocator's half-open `[500, 4596)` range exactly (`id_allocator.rs` rejects `id >= range.1`). ✅
- `executor.go`: `finalStatus.Err` check has no nil-deref risk (only set on the single `break` path); `ErrNotFound`-as-closure is correct. ✅
- Observer `device_tunnel_gap` trigger + locking are sound; CPU parser matches the new fixture. ✅
- `sweep.go` batch pipelining is concurrency-safe — ran `-race`, clean; provision goroutines use `WithoutCancel` so aborts can't orphan creates. ✅

Approving. One non-blocking item worth addressing: the new `TestRun_WaitsForAgentQuiescenceAfterDeprovision` is flaky under the parallel package suite (~5/6 failures reproduced) — details inline. Production code is fine; it's a test-only synchronization gap. Two smaller notes inline as well.

PR #3821, opened 2026-06-02
tools/stress: local containerized harness script

nikw9944:
Depends on: #3820

## Summary of Changes
- Add `tools/stress/scripts/run-stress-local.sh`: an end-to-end harness that brings up the e2e ledger / manager / funder via `dev/dzctl`, layers a stress-test device on top of the e2e device image, creates the device + Loopback255/256 interfaces + prepaid access passes onchain, and launches `device-orchestrator` and `device-observer` against it. Tolerates the broken upstream `doublezero geolocation init` step: continues if the smart contract is already initialized, starts the controller itself if `dzctl` bailed before reaching it (with `-max-user-tunnel-slots` derived from `TARGET_USERS`), and creates the `dz-local-cyoa` Docker network if missing. Access-pass creation fans out via `xargs -P 16` so high user counts don't sit serialized for minutes at the validator.
- Add `tools/stress/docker/device/Dockerfile`: thin layer on `dz-local/device:dev` that installs an agent wrapper, adds a `stress` system user with `/bin/bash`, and grants it passwordless sudo via `/etc/sudoers.Eos` (cEOS overlays that template onto `/etc/sudoers` at boot and does not source `/etc/sudoers.d/`, so anything we'd write to either gets silently wiped).
- Add `tools/stress/docker/device/agent-wrapper.sh`: shim invoked as `doublezero-agent` over the SSH agent path. Re-execs through `sudo` so the agent runs as root: the agent shells out to `ip netns exec default /usr/bin/Cli` to compute configure-session diffs, which requires `CAP_SYS_ADMIN`, and without root the agent loops forever on `Operation not permitted`. The wrapper also injects `-pubkey` (from `/etc/doublezero/agent/pubkey`, planted by the run script) and `-metrics-enable -metrics-addr :50100`. Port 50100 sits in the controller-pushed `MAIN-CONTROL-PLANE-ACL`'s default `50000-50100` permit range, which is needed because that ACL is fully redefined on every agent apply.
- Add `tools/stress/scripts/README.md`: operator-facing usage, env-var knobs, the `--no-agent` smoke-test path, why the device image differs from the e2e device image, why the metrics port is 50100, teardown.

## Diff Breakdown
| Scaffolding  |     3 | +623 / -0   | +623 |
| Docs         |     1 | +79 / -0    | +79  |
| **Total**    |     4 | +702 / -0   | +702 |

All scaffolding — no production code, no tests. The run script is the bulk of the diff; the Dockerfile and wrapper are small shims.

<details>

- [`tools/stress/scripts/run-stress-local.sh`](https://github.com/malbeclabs/doublezero/pull/3821/files#diff-41614446143de970) — 549-line bash harness: dzctl build/start with tolerant fallbacks, stress device image build, network introspection, fallback CYOA network + controller container creation with `-max-user-tunnel-slots`, SSH keypair generation, container start + CYOA attach, EOS startup-config render with default-network IP for Management0 routing, onchain device create + loopback interface registration + parallelized access-pass seeding, orchestrator + observer launch.
- [`tools/stress/docker/device/Dockerfile`](https://github.com/malbeclabs/doublezero/pull/3821/files#diff-95c7cab634324697) — 38-line layer on `dz-local/device:dev`: install the agent wrapper, create the `stress` bash user, grant passwordless sudo via `/etc/sudoers.Eos`.
- [`tools/stress/docker/device/agent-wrapper.sh`](https://github.com/malbeclabs/doublezero/pull/3821/files#diff-f3fba5bdd61d363b) — 36-line shim that re-execs through sudo (agent needs root for `ip netns exec`), injects `-pubkey` from disk, and pins the prometheus listener on a controller-ACL-permitted port.

</details>

## Testing Verification
Validated end-to-end against the local containerized devnet (with the #3820 fix stack applied):

- **`--no-agent` smoke**: 4-user sweep, all six lifecycle events × 4 users, observer captured 17+ EOS sample ticks with no warnings.
- **Full SSH-agent path**: 8-user sweep (batch=4, hold=60), 24 runlog rows, 28 deprovisions, 728 prometheus metric rows captured, 38 k agent-log lines streamed via SSH, no abort triggers fired.
- **1024-user run**: completed cleanly in 885 s wall-clock (batch=64, hold=30) with 1024 provisions + 1024 deprovisions, zero agent errors, p95 per-user provision 14.1 s, no degradation across the sweep.
- **Self-sufficient fresh-controller path**: removed only the controller container, re-ran; the script's fallback started a new controller with the correct `-max-user-tunnel-slots`, the loopback-already-exists idempotent path triggered, sweep proceeded normally.

### DIFF
--- a/tools/stress/scripts/run-stress-local.sh
+++ b/tools/stress/scripts/run-stress-local.sh
@@ -166,16 +166,54 @@ fi
 SSH_PUBKEY="$(cat "${SSH_KEY_PATH}.pub")"
 
 # ---------------------------------------------------------------------------
-# Phase 5: render EOS startup-config (no doublezero-agent daemon)
+# Phase 5: start the stress device container (it will block on its
+# entrypoint's wait-for-startup-config loop until phase 6 copies the config)
+# ---------------------------------------------------------------------------
+if docker inspect "$CONTAINER_NAME" >/dev/null 2>&1; then
+    log "removing existing stress device container"
+    docker rm -f "$CONTAINER_NAME" >/dev/null
+fi
+
+log "starting stress device container: $CONTAINER_NAME"
+docker run -d \
+    --name "$CONTAINER_NAME" \
+    --hostname "device-${DEVICE_CODE}" \
+    --privileged \
+    --network "$DEFAULT_NETWORK" \
+    --label "dz.malbeclabs.com/type=devnet" \
+    --label "dz.malbeclabs.com/deploy-id=${DEPLOY_ID}" \
+    "$STRESS_IMAGE" >/dev/null
+
+# Network ordering matters with containerized EOS: the first network
+# attached is the management interface (Management0/eth0), then subsequent
+# networks correspond to Ethernet1+ in order. So default → eth0, cyoa → eth1.
+docker network connect --ip "$CYOA_IP" "$CYOA_NETWORK" "$CONTAINER_NAME"
+
+# Inspect for the default-network IP / gateway / prefix Docker assigned —
+# the agent's path to the controller goes through Management0, so the EOS
+# config needs all three.
+DEFAULT_IP="$(docker inspect "$CONTAINER_NAME" \
+    --format "{{(index .NetworkSettings.Networks \"${DEFAULT_NETWORK}\").IPAddress}}")"
+DEFAULT_PREFIX="$(docker inspect "$CONTAINER_NAME" \
+    --format "{{(index .NetworkSettings.Networks \"${DEFAULT_NETWORK}\").IPPrefixLen}}")"
+DEFAULT_GATEWAY="$(docker inspect "$CONTAINER_NAME" \
+    --format "{{(index .NetworkSettings.Networks \"${DEFAULT_NETWORK}\").Gateway}}")"
+log "default network: ip=${DEFAULT_IP}/${DEFAULT_PREFIX} gw=${DEFAULT_GATEWAY}"
+
+# ---------------------------------------------------------------------------
+# Phase 6: render EOS startup-config and unblock the container's init
 # ---------------------------------------------------------------------------
 STARTUP_CONFIG_PATH="${WORKING_DIR}/startup-config"
 CYOA_CIDR_PREFIX="${CYOA_SUBNET##*/}"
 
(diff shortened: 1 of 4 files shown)


tools/stress/scripts/run-stress-local.sh
```diff
@@ -0,0 +1,549 @@
...
+
+# Derive an IP inside the CYOA subnet (host octet = DEVICE_HOST_ID) and a
+# globally-routable /29 dz_prefix at a non-overlapping host offset. Mirrors
+# the rules in e2e/internal/devnet/device.go.
+read -r CYOA_IP DZ_PREFIX < <(python3 - <<PY
+import ipaddress
+net = ipaddress.ip_network("$CYOA_SUBNET")
+host_id = $DEVICE_HOST_ID
+ip = net.network_address + host_id
+last = (host_id + 128) if (host_id + 128) < 256 else ((host_id - 128) // 8) * 8
```
elitegreg: With the default `DEVICE_HOST_ID=50`, this yields `last=178`, so `dz_prefix` becomes `<net>.178/29` — which is not a /29 network boundary (`.176` would be). It works today only because the prefix is constructed via `ipaddress.ip_address(...)` (no network validation) and the serviceability CLI currently accepts the unaligned value, so `device create` succeeds (confirmed by the 1024-user run).

It is a latent footgun, though: if `--dz-prefixes` ever validates network alignment, or someone changes `DEVICE_HOST_ID`, this silently breaks. Worth either snapping `last` down to a /29 boundary (e.g. `last &= ~0x7`) or a short comment noting the prefix is intentionally only required to be non-overlapping, not aligned.

elitegreg (approved):
Scaffolding-only PR (no production code). Verified the harness assumptions against the codebase: all orchestrator/observer flags exist, `-max-user-tunnel-slots` is real, the agent path matches the e2e device image, and the access-pass IP math is identical to the orchestrator's `ipForIndex`, so passes line up with provisioned users at scale.

The inline workaround comments (broken `geolocation init`, 128-slot silent truncation, `:50100` ACL reasoning, `/etc/sudoers.Eos`, root/CAP_SYS_ADMIN, CGNAT rejection) are excellent and won't rot into mystery.

One non-blocking note left inline re: the unaligned `dz_prefix`.

PR #3822, opened 2026-06-02
geolocation-cli: route prompts through writer, add TTY guard, rename update-payment

ben-dz:
## Summary of Changes

* Route confirmation prompts in `init.rs`, `user/delete.rs`, and `probe/delete.rs` through the injected `Write` parameter instead of `eprint!`, complying with RFC-20 rule 7
* Add `std::io::IsTerminal` TTY guard so non-interactive contexts (CI, scripts) get a clear error directing them to pass `--yes` instead of hanging on stdin
* Rename `UserCommands::UpdatePayment` to `UpdatePaymentStatus` so the CLI subcommand (`user update-payment-status`) matches the file and SDK naming
* Fixes malbeclabs/infra#1461

## Testing Verification

* All 44 tests in `doublezero-geolocation-cli` pass, including 3 new tests verifying non-TTY rejection for each destructive verb (`init`, `user delete`, `probe delete`)
* `cargo clippy -p doublezero-geolocation-cli -- -Dclippy::all -Dwarnings` passes with no warnings
* `make rust-fmt` applied with no formatting changes required

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
+- CLI
+  - Change `geolocation user update-payment` to `update-payment-status` for clarity. 
 - Tools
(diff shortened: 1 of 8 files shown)


juan-malbeclabs (approved):
LGTM

PR #3824, opened 2026-06-02
cli: RFC-20 serviceability cleanups

ben-dz:
## Summary of Changes

* Remove direct `std::env::args()` and `std::env::var()` reads from `requirements.rs` per RFC-20 §67 (modules MUST NOT read env/argv); the keypair-source flag is now computed once at startup in the binary and plumbed through a new `CliCommand::has_keypair_source()` trait method
* Convert four `eprintln!` calls to `tracing::error!` in `requirements.rs` (`check_id`, `check_balance`, `check_allowlist`) and `helpers.rs` (`print_error`) for structured logging
* Add `--json` flag to `globalconfig featureflags get` using `render_record` and a new `FeatureFlagsDisplay` struct
* Add `--json` flag to `accesspass user-balances` with pretty-printed JSON serialization of the existing `UserBalanceDisplay` collection
* Related RFC: `rfcs/rfc20-cli-standardization.md`
* Fixes malbeclabs/infra#1460

## Testing Verification

* All 301 `doublezero-serviceability-cli` unit tests pass, including the new `test_cli_globalconfig_featureflags_get_json` test that validates JSON output structure and content
* `cargo clippy -p doublezero-serviceability-cli -p doublezero -- -Dclippy::all -Dwarnings` passes with zero warnings
* `cargo fmt --check` passes
* `cargo check --tests -p doublezero-serviceability-cli -p doublezero` verifies test compilation across both crates
* The `doublezero-admin` binary (which also uses `CliCommandImpl::new`) continues to compile without changes due to the backward-compatible builder pattern (`with_keypair_source` is opt-in, defaulting to `false`)


juan-malbeclabs (approved):
LGTM

PR #3828, opened 2026-06-03
daemon-cli: scaffold doublezero-daemon-cli crate with DaemonClient trait and types

ben-dz:
## Summary of Changes

* Scaffold new `doublezero-daemon-cli` library crate (`crates/doublezero-daemon-cli/`) as PR 1 of the RFC-20 daemon verb extraction
* Extract `DaemonClient` trait (renamed from `ServiceController`) with `#[automock]`, all response/request types, and `DaemonClientImpl` (Unix socket HTTP) from `client/doublezero/src/servicecontroller.rs`
* Define `LedgerClient` trait — narrow SDK interface for daemon verbs, starting with `get_environment()` and expanded as verbs migrate
* Create empty `DaemonCommand` enum to be populated as verbs migrate in subsequent PRs
* Fix inconsistent error handling: `v2_status()` and `routes()` now use `parse_daemon_response` and check HTTP status codes consistently with other endpoints
* Related RFC: `rfcs/rfc20-cli-standardization.md`
* Fixes malbeclabs/infra#1487

**PR stack** (this is PR 1 of 8):
1. **Scaffold crate + traits** ← this PR
2. Move enable/disable verbs
3. Move status + helpers
4. Move latency/routes + utilities
5. Move disconnect
6. Move connect
7. Move multicast daemon ops
8. Cleanup

**Note:** This PR exceeds the ~500-line guideline (591 lines added) because it includes the full response type definitions and DaemonClient implementation extracted from `servicecontroller.rs`. The existing `servicecontroller.rs` is left unchanged during the transition — type duplication is intentional and resolved in PR 8 (cleanup).

## Testing Verification

* `cargo test -p doublezero-daemon-cli` — 7 serde round-trip tests pass for all response types (LatencyRecord, RouteRecord, StatusResponse, V2StatusResponse, MulticastGroups) plus DaemonClientImpl socket path resolution
* `cargo test -p doublezero` — all 141 existing binary tests pass unchanged
* `cargo clippy -p doublezero-daemon-cli -- -Dclippy::all -Dwarnings` — clean
* `cargo +nightly fmt -- --check` — clean

### DIFF
--- a/crates/doublezero-daemon-cli/src/client.rs
+++ b/crates/doublezero-daemon-cli/src/client.rs
@@ -14,7 +14,7 @@ use std::{fmt, fs::File, path::Path, sync::OnceLock};
 use tabled::{derive::display, Tabled};
 
 pub(crate) const DEFAULT_SOCKET_PATH: &str = "/var/run/doublezerod/doublezerod.sock";
-const NANOS_TO_MS: f64 = 1000000.0;
+const NANOS_TO_MS: f64 = 1_000_000.0;
 static GLOBAL_SOCKET_PATH: OnceLock<String> = OnceLock::new();
 
 // ---------------------------------------------------------------------------
@@ -163,6 +163,25 @@ pub struct MulticastGroups {
     pub subscriber: Vec<String>,
 }
 
+/// A single multicast group the user participates in, with the group's onchain
+/// details and the user's role(s). A user that is both publisher and subscriber
+/// of a group appears once with both booleans set.
+#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq, Eq)]
+pub struct Subscription {
+    #[serde(default)]
+    pub pubkey: String,
+    #[serde(default)]
+    pub code: String,
+    #[serde(default)]
+    pub multicast_ip: String,
+    #[serde(default)]
+    pub max_bandwidth: u64,
+    #[serde(default)]
+    pub publisher: bool,
+    #[serde(default)]
+    pub subscriber: bool,
+}
+
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct V2ServiceStatus {
     #[serde(flatten)]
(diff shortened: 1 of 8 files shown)


crates/doublezero-daemon-cli/src/client.rs
```diff
@@ -0,0 +1,539 @@
...
+            .collect()
+            .await
+            .map_err(|e| eyre!("Unable to read response body: {e}"))?
+            .to_bytes();
+        parse_daemon_response::<Vec<RouteRecord>>(&data, "/routes")
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
juan-malbeclabs, line 418: The new crate’s serde tests here are slimmer than the ones in `servicecontroller.rs` (`test_*_json_output_format`), which assert full field presence as a JSON API-contract check. That’s fine while both copies coexist, but when `servicecontroller.rs` is deleted in the PR 8 cleanup those stricter contract assertions will be lost. Consider porting them into this module (or explicitly deferring to PR 8) so the JSON-output coverage isn’t quietly dropped.

juan-malbeclabs (approved):
LGTM. Clean, well-scoped RFC-20 scaffolding PR. Verified the type/impl extraction is a faithful copy of servicecontroller.rs, and the v2_status()/routes() error-handling fixes correctly match the shape of status(). Only follow-up is the inline note about porting the stricter JSON-contract tests before servicecontroller.rs is deleted in PR 8.

crates/doublezero-daemon-cli/src/client.rs
ben-dz: replaced the four slim roundtrip tests with contract tests named to match the old file (test_latency_record_json_output_format, test_route_record_json_output_format, ..._with_nulls, test_status_response_json_output_format). Rather than porting ~230 lines of per-field assert! boilerplate verbatim, each test asserts exact-object equality against a serde_json::json! literal — strictly stronger than the originals (one assertion covers field presence, values, renames, explicit-null serialization of Option fields, and unexpected extra fields), then deserializes the expected JSON back to keep the daemon-response direction covered. The four tests with no old-file counterpart (V2 status, multicast/subscription defaults, socket path) are unchanged.

PR #3829, opened 2026-06-03
tools/stress: physical-device harness driver

### LINKED ISSUE #3748: stress: implement tools/stress/scripts/test-setup
## Why

The doc lists eight setup steps that are operator-driven but tedious to do by hand. A small script keeps the run reproducible and the working directory consistent.

See [Notion design](https://www.notion.so/malbeclabs/GRE-Tunnel-Capacity-Study-Design-34ffef22bebe816a87dadcacc01c2081) for full context.

## Scope

Bash or small Go binary at `tools/stress/scripts/test-setup`. Takes `--working-dir`, `--dut-host`, `--dut-platform` (7130LBR | 7280CR3A), `--run-id`. Performs:

1. `show run` → `<wd>/running-config.pre.txt`.
2. Deploy serviceability program to the devnet ledger, recording its program ID (refer to `e2e/internal/devnet/smartcontract_init.go`). Include 1 exchange + 1 location + any other Device-record prerequisites.
3. Start the stress controller with `MaxUserTunnelSlots = 1024` (uses the configurable flag from the controller issue), pointed at the devnet ledger. Record PID and port in `<wd>/stress-controller.json`.
4. `configure checkpoint save pre-hwstress-<run-id>` on the DUT.
5. Verify or install the agent binary appropriate to the DUT platform. **Do not** start it as a daemon — the orchestrator runs it over SSH.
6. Generate device keypair → `<wd>/device.keypair.json`.
7. Register Device account onchain using the keypair from step 6.
8. Create access passes for each client IP using payer `DZfHfcCXTLwgZeCRKQ1FL1UuwAwFAZM93g86NMYpfYan`.

## Acceptance

- Re-runnable for a new `--run-id` against a clean DUT without manual intervention.
- All artifacts land in `--working-dir`.
- Failure at any step is fatal; the script exits non-zero and prints the offending step number.

Tracker: #3744.

nikw9944:
## Summary of Changes
- New `tools/stress/scripts/run-stress-physical.sh` driver that runs the same orchestrator + observer (from PR #3796/#3820/#3821) against a physical Arista EOS device instead of the containerized cEOS. Idempotent serviceability init, idempotent device + loopback create, controller launch via `go run` against the devnet ledger, parallel access-pass setup, orchestrator + observer launch with a port-conflict fail-fast and a trap that cleans up the controller on any pre-launch error.
- Four additive flags on the device-orchestrator so the SSH-exec'd agent command can be made self-contained on a DUT with no wrapper script: `--agent-binary`, `--agent-command-prefix`, `--agent-pubkey`, `--agent-metrics-addr`. All default to empty so the containerized harness path is byte-for-byte unchanged.
- ssh-agent fallback in the orchestrator's SSH runner: when the configured key file is passphrase-protected (Go's `ssh.ParsePrivateKey` can't prompt), the runner connects to `$SSH_AUTH_SOCK` and uses the agent's loaded signers. Empty-agent and unreachable-agent paths fail fast with actionable diagnostics instead of opaque SSH handshake errors.
- Post-deprovision agent quiescence wait now engages on the abort path too (not just clean success) and enforces a minimum window after deprovision returns. Without this, the observer's abort sentinel (legitimate or false-positive) could kill the SSH session mid-commit and strand tunnels in the device's running-config even though onchain deprovision completed.
- README appendix on physical-DUT prerequisites: SSH user with passwordless sudo + bash shell, the agent binary on disk, EOS SDK RPC daemon in the management VRF, eAPI http-commands with credentials, and the management netns. Includes the env-var knob table.

## Diff Breakdown
| Category     | Files | Lines (+/-)     | Net   |
|--------------|-------|-----------------|-------|
| Core logic   |     2 | +121 / -26      |  +95  |
| Scaffolding  |     2 | +534 / -5       |  +529 |
| Tests        |     1 | +91  / -0       |  +91  |
| Docs         |     1 | +142 / -0       |  +142 |
| **Total**    |     6 | +888 / -31      |  +857 |

Mostly scaffolding (the new driver script is ~500 of the 857 net lines); core-logic changes are concentrated in two files (ssh-agent fallback + quiescence-wait predicate) and accompanied by a test rewrite covering the new branch.

<details>

- [`tools/stress/scripts/run-stress-physical.sh`](https://github.com/malbeclabs/doublezero/pull/3829/files#diff-06acb461af69f0333245bea3535f3b302fa7d3bf382ac3e6d776e91c21733616) — new driver: idempotent init, device + loopbacks, controller startup, parallel access passes, orchestrator + observer launch, port-conflict fail-fast.
- [`tools/stress/scripts/README.md`](https://github.com/malbeclabs/doublezero/pull/3829/files#diff-2d60594907d94eba27e2702565f6b2c3314d1457d0256304346542715c8c4e77) — physical-harness section with prereqs and the env-var knob table.
- [`tools/stress/device-orchestrator/pkg/sweep/sweep_test.go`](https://github.com/malbeclabs/doublezero/pull/3829/files#diff-3901c06a779aac608c7ba63c230d1be60b3cfc0a24c59c632ddfc362c8b16201) — new tests for the abort-path quiescence wait and the elapsed-since-wait-start floor.
- [`tools/stress/device-orchestrator/pkg/agent/ssh.go`](https://github.com/malbeclabs/doublezero/pull/3829/files#diff-2329dc0bd36a1b91ad8d3ab3aa3aba79afd4c59f7554e7aa96279c25ee8330fd) — `loadAuthMethod` adds an ssh-agent fallback that's used when the key file is passphrase-protected or empty; conn is stashed on the SSH struct and closed in `shutdown`.
- [`tools/stress/device-orchestrator/pkg/sweep/sweep.go`](https://github.com/malbeclabs/doublezero/pull/3829/files#diff-a864c355728c685805dc490247cb61914299637f4269f384ac807cbd3d2e8ee3) — post-deprovision quiescence wait now engages on the abort path (`context.WithoutCancel` for the wait), and requires both `silent ≥ window` and `elapsed-since-wait-start ≥ window`.
- [`tools/stress/device-orchestrator/cmd/device-orchestrator/main.go`](https://github.com/malbeclabs/doublezero/pull/3829/files#diff-9b9d11f1e05b024cb4a921b112759fbad66da00cd07c80357666a91084b9b20d) — four additive flags threaded through `orchestratorConfig` and `selectAgentRunner`.

</details>

## Testing Verification
- `TestRun_WaitsForAgentQuiescenceEvenOnAbort` exercises both axes of the new branch: that the wait engages when ctx is cancelled (abort path), and that the elapsed-since-wait-start floor blocks even when the absolute "silent for ≥ window" predicate would return instantly.
- `TestRun_WaitsForAgentQuiescenceAfterDeprovision` and `TestRun_SkipsQuiescenceWaitWhenNoAppliedObserved` continue to assert the success and no-event paths respectively.
- Validated end-to-end against chi-dn-dzd5: 4-user sweep completes cleanly with 0 stranded tunnels; 512-user sweep provisions + deprovisions all 512 onchain. The 1024-user ceiling is a hardware limit on this DUT model (Tunnel ID names cap at 1023; with `StartUserTunnelNum=500` that's 524 usable users on chi-dn-dzd5 specifically — different physical DUTs have higher caps).

### DIFF
--- a/tools/stress/scripts/run-stress-physical.sh
+++ b/tools/stress/scripts/run-stress-physical.sh
@@ -15,15 +15,20 @@
 #   tools/stress/scripts/run-stress-physical.sh \
 #       --target-users 4 --users-per-batch 2 --hold 0
 #
-# Required env (or defaults shown):
+# Required env:
+#   DZ_PROGRAM_ID       serviceability program id (no default — operator
+#                       passes the stress-test program id, e.g. from the
+#                       private infra repo)
+# Env (defaults shown — override per operator):
 #   DZ_RPC_URL          devnet RPC (default: doublezerolocalnet pool)
-#   DZ_PROGRAM_ID       serviceability program id (default: GXxf6xgC...)
-#   DUT_HOST            device IP or hostname (default: 10.0.0.141)
+#   DUT_HOST            device IP or hostname
 #   DUT_SSH_USER        SSH user on the DUT
 #   DUT_SSH_KEY         SSH private key path
-#   CONTROLLER_BIND_ADDR  controller listen address (default: 10.0.0.141's gateway from the DUT)
-#   AGENT_BINARY        path to doublezero-agent on the DUT (default: /usr/local/bin/doublezero-agent)
-#   SOLANA_KEYPAIR      operator's keypair (default: $HOME/.config/solana/id.json)
+#   CONTROLLER_BIND_ADDR  controller listen address (default: 0.0.0.0)
+#   AGENT_BINARY        path to doublezero-agent on the DUT
+#   SOLANA_KEYPAIR      operator's keypair (default: $HOME/.config/doublezero/id.json
+#                       — the same default the doublezero CLI uses, so `dz` calls
+#                       and orchestrator share the operator's authority)
 set -euo pipefail
 
 SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
@@ -33,7 +38,10 @@ WORKSPACE_DIR="$(cd -- "${SCRIPT_DIR}/../../.." &> /dev/null && pwd)"
 # Config (env-overridable)
 # ---------------------------------------------------------------------------
 DZ_RPC_URL="${DZ_RPC_URL:-https://doublezerolocalnet.rpcpool.com/8a4fd3f4-0977-449f-88c7-63d4b0f10f16}"
-DZ_PROGRAM_ID="${DZ_PROGRAM_ID:-GXxf6xgCdngKsRPGg3svaoKUQskVnu4mTuUoJ5PhNUau}"
+# DZ_PROGRAM_ID has no default: the stress-test serviceability program ID is
+# kept in the private infra repo, not here. Operators export it before
+# running the script.
+DZ_PROGRAM_ID="${DZ_PROGRAM_ID:?DZ_PROGRAM_ID is required (stress-test serviceability program id; see infra repo)}"
 
 DUT_HOST="${DUT_HOST:-10.0.0.15}"
 DUT_SSH_USER="${DUT_SSH_USER:-nik}"
@@ -69,7 +77,11 @@ CONTROLLER_BIND_ADDR="${CONTROLLER_BIND_ADDR:-0.0.0.0}"
 CONTROLLER_ADVERTISE_ADDR="${CONTROLLER_ADVERTISE_ADDR:-}"
 CONTROLLER_LISTEN_PORT="${CONTROLLER_LISTEN_PORT:-7000}"
 
-SOLANA_KEYPAIR="${SOLANA_KEYPAIR:-$HOME/.config/solana/id.json}"
+# Default to the doublezero CLI's keypair location so `dz` (which reads this
+# location implicitly) and the orchestrator's --keypair both use the same
+# operator authority. The standard solana CLI default at ~/.config/solana/id.json
+# may be a different key and would need an explicit override.
+SOLANA_KEYPAIR="${SOLANA_KEYPAIR:-$HOME/.config/doublezero/id.json}"
 
 DEPLOY_DIR="${WORKSPACE_DIR}/dev/.deploy/stress-physical"
 WORKING_DIR="${DZ_STRESS_WORKING_DIR:-${DEPLOY_DIR}/run}"
--- a/tools/stress/scripts/run-stress-physical.sh
+++ b/tools/stress/scripts/run-stress-physical.sh
@@ -296,6 +296,26 @@ if ss -ltn "sport = :$CONTROLLER_LISTEN_PORT" 2>/dev/null | tail -n +2 | grep -q
 fi
 
 log "starting controller (listen=${CONTROLLER_BIND_ADDR}:${CONTROLLER_LISTEN_PORT}, max-slots=${TARGET_USERS})"
+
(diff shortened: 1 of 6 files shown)


tools/stress/scripts/run-stress-physical.sh
```diff
@@ -0,0 +1,376 @@
...
+    dz device create \
+        --contributor co01 \
+        --code "$DEVICE_CODE" \
+        --location "$DEVICE_LOCATION" \
+        --exchange "$DEVICE_EXCHANGE" \
+        --public-ip "$DEVICE_PUBLIC_IP" \
+        --dz-prefixes "$DEVICE_DZ_PREFIX" \
+        --mgmt-vrf mgmt
+    DEVICE_PK="$(dz device get --code "$DEVICE_CODE" --json | jq -r .account)"
+    dz device update --pubkey "$DEVICE_PK" --max-users "$TARGET_USERS" \
```
elitegreg: `dz device update --max-users "$TARGET_USERS"` only runs inside the `if ! dz device get …` create branch. On a rerun with a larger `--target-users`, the onchain device cap stays at its original value and silently caps the sweep — even though the freshly-launched controller picks up the new `--max-user-tunnel-slots`. Consider moving the `update --max-users` outside the create guard so reruns scale the cap.

tools/stress/scripts/run-stress-physical.sh
```diff
@@ -0,0 +1,376 @@
...
+}
+trap cleanup_controller EXIT
+
+# Wait for the controller's listen port to accept connections (gRPC handshake).
+log "waiting for controller to accept connections"
+for _ in $(seq 1 30); do
+    if nc -z -w 1 127.0.0.1 "$CONTROLLER_LISTEN_PORT" 2>/dev/null; then
+        log "controller listening on :${CONTROLLER_LISTEN_PORT}"
+        # Disarm the trap — controller is up and should outlive this script.
+        trap - EXIT
```
elitegreg: `trap - EXIT` disarms the cleanup trap the moment the gRPC port is up, but several phases still follow (access-pass setup, build, orchestrator/observer launch). If any of those fail under `set -e`, the `go run` controller is orphaned rather than cleaned up. The pid file makes it recoverable, but disarming only after the orchestrator has launched would avoid the leak on partial failure.

elitegreg (approved):
Reviewed. The orchestrator change is minimal and cleanly backward compatible — with all three new flags at their defaults the agent command is byte-identical to the existing containerized path. The driver script is a faithful sibling of `run-stress-local.sh`. Approving.

Two items worth addressing (left inline):
- `--max-users` is only set in the device-create branch, so reruns with a larger `--target-users` won't scale the onchain cap.
- The controller `EXIT` trap disarms as soon as the gRPC port is up, orphaning the `go run` controller if a later phase fails.

tools/stress/scripts/run-stress-physical.sh
nikw9944: Done in 4e3441dc — moved `dz device update --max-users` outside the `if ! dz device get` guard so every run resyncs the cap. Reruns with a larger `--target-users` now scale the onchain cap too.

tools/stress/scripts/run-stress-physical.sh
nikw9944: Done in 4e3441dc — the disarm now happens right after the orchestrator launches (line 388), not after the gRPC port comes up. The trap covers access-pass setup, build, and orchestrator+observer launch, so a `set -e` failure in any of those tears the controller down instead of orphaning it.

PR #3838, opened 2026-06-04
daemon-cli: move enable/disable verbs into doublezero-daemon-cli crate

ben-dz:
⚠️ PR 2 in Stack 
Must Land after: https://github.com/malbeclabs/doublezero/pull/3828

PR stack (this is PR 2 of 8):

Scaffold crate + traits
Move enable/disable verbs  ← this PR
Move status + helpers
Move latency/routes + utilities
Move disconnect
Move connect
Move multicast daemon ops
Cleanup

## Summary
* Move `enable` and `disable` verbs from the binary (`client/doublezero/src/command/`) into the `doublezero-daemon-cli` module crate per RFC-20 — PR 2 of 8 in the daemon-cli extraction stack
* Replace `println!`/`eprintln!` with `writeln!(out, ...)`/`tracing::warn!` for testability and log hygiene
* Add shared `check_daemon()` pre-flight helper using `DaemonClient` + `LedgerClient` traits
* Wire binary dispatch through `DaemonCommand` flatten with `LedgerAdapter` bridging `CliCommand` → `LedgerClient`

## Testing Verification
* `cargo test -p doublezero-daemon-cli` — 16 tests pass (9 new verb tests + 7 existing)
* `cargo test -p doublezero` — 135 binary tests pass unchanged
* `cargo clippy -p doublezero-daemon-cli -- -Dclippy::all -Dwarnings` — clean
* `cargo +nightly fmt -- --check` — clean
* Grep verified: no `println!`/`eprintln!` in new crate execute paths
* Architecture and security reviews posted to issue — no critical/high findings

PR #3839, opened 2026-06-04
daemon-cli: move status verb and shared output helpers into crate

ben-dz:
⚠️ PR 3 in Stack
Must Land after: https://github.com/malbeclabs/doublezero/pull/3838

PR stack (this is PR 3 of 8):

Scaffold crate + traits
Move enable/disable verbs 
Move status + helpers ← this PR
Move latency/routes + utilities
Move disconnect
Move connect
Move multicast daemon ops
Cleanup

## Summary of Changes
* Move `status` verb from `client/doublezero/src/command/status.rs` into `crates/doublezero-daemon-cli/src/status.rs`, adopting the RFC-20 execute signature with `DaemonClient`, `LedgerClient`, and `Write` trait parameters
* Create `helpers.rs` with a writer-based `show_output` helper that replaces `println!` with `writeln!(out, ...)` for testability — shared across daemon-control verbs
* Replace `ServiceController`/`CliCommand` trait dependencies with `DaemonClient`/`LedgerClient` and `check_daemon` pre-flight check
* Propagate `build_status` errors via `?` instead of swallowing them with `print_error` — fail-closed behavior, mitigated by retry logic (3 retries with exponential backoff)
* Migrate V2ServiceStatus serde wire-format tests to `client.rs` to preserve coverage
* Fixes malbeclabs/infra#1489

Related: [RFC-20](rfcs/rfc20-cli-standardization.md) — PR 3 of 8 in the daemon-cli extraction stack. Depends on PR 2 (enable/disable).

## Testing Verification
* `cargo test -p doublezero-daemon-cli` — 31 tests pass, covering tunnel up/down, empty-service disconnected synthesis, reconciler disabled, multicast groups display, JSON output format contract, lowest-latency emoji formatting, retry on transient errors, and daemon-not-running error path
* `cargo test -p doublezero` — 120 binary tests pass unchanged
* `cargo clippy -p doublezero-daemon-cli -p doublezero -- -Dclippy::all -Dwarnings` — clean
* `cargo +nightly fmt -- --check` — clean


crates/doublezero-daemon-cli/src/status.rs
```diff
@@ -0,0 +1,787 @@
...
+impl Status {
+    pub async fn execute<D: DaemonClient, L: LedgerClient, W: Write>(
+        self,
+        _ctx: &CliContext,
+        daemon: &D,
+        ledger: &L,
+        out: &mut W,
+    ) -> eyre::Result<()> {
+        check_daemon(daemon, ledger).await?;
+        let responses = self.build_status(daemon, ledger).await?;
```
juan-malbeclabs, line 69: Behavioral change worth an explicit sign-off: propagating `build_status` via `?` makes `status` **fail-closed**, so `/v2/status` being unreachable after the 3 retries now exits non-zero, where the old `print_error` path printed and exited 0. The blast radius is narrow (`check_daemon` already exited non-zero for daemon-down/permission/env-mismatch, and `build_status` only errors on the `v2_status` retry exhaustion), and fail-closed is the more correct behavior — but the code comments here reference the QA agent / global monitor depending on status output, so please confirm those tolerate a non-zero exit before this lands.

crates/doublezero-daemon-cli/src/status.rs
```diff
@@ -0,0 +1,787 @@
...
+                        subscriber: true,
+                    },
+                ]
+            );
+        });
+    }
+
+    /// Test JSON output format contract — validates the exact field names and nesting.
+    #[test]
+    fn test_status_json_output_format() {
```
juan-malbeclabs, line 552: This migrated `test_status_json_output_format` drops a few field-presence/value assertions the original had (explicit `tunnel_src`/`tunnel_dst`/`doublezero_ip` presence + values). It still covers the overall shape, but given the original comment framed this as the JSON API-contract guard, consider restoring the full field set so the contract check stays as strict as before.

juan-malbeclabs (approved):
LGTM. build_status is a faithful copy of the old command_impl (only client→ledger.get_environment()), helpers::show_output matches util::show_output byte-for-byte modulo println!→writeln!(out), and the retry backoff is preserved. End-to-end tests through execute() with real stdout parsing are a solid coverage upgrade. Two inline notes: (1) confirm downstream monitoring tolerates the new fail-closed non-zero exit on status, and (2) consider restoring the trimmed JSON-contract assertions. Neither blocks.

crates/doublezero-daemon-cli/src/status.rs
ben-dz: Confirmed — I swept every subprocess caller of `doublezero status`. The only fatal-on-error consumers are the global monitor (`telemetry/global-monitor/internal/dz/status.go` → `gm/source.go`, startup) and the QA agent (`e2e/internal/rpc/agent.go` `fetchStatus`), and both already errored in this exact case: pre-change, the endpoint-unreachable path wrote the error to stderr and exited 0 with **empty stdout**, so their `json.Unmarshal` already failed. This PR only shifts their failure mode from unmarshal-error to exit-code-error — same outcome, no new breakage.

The synthetic "disconnected" entry is only produced when `/v2/status` succeeds with zero services; that path still exits 0, so no caller loses parseable output it was previously getting. All remaining callers tolerate a non-zero exit explicitly (`|| true` in `docker-entrypoint.sh` and the install scripts, `|| echo '[]'` in `allocation_lifecycle_test.go`, error-ignoring e2e diagnostics).

crates/doublezero-daemon-cli/src/status.rs
ben-dz: Restored in 97aa05f9 — the assertion block is back to the original verbatim: presence checks for all `response` fields (including `tunnel_src`/`tunnel_dst`/`doublezero_ip`/`user_type`) plus the full value set (`network`, `tunnel_name`, `lowest_latency_device`, `multicast_groups == ""`, and the tunnel/IP values). The nulls variant also regains its trailing `multicast_groups` check.

PR #3841, opened 2026-06-05
tools/stress: harden orchestrator + observer for physical EOS

### LINKED ISSUE #3835: stress: device stress test per-run reporting tool

nikw9944:
**Stack:** 1 of 4 — base `main`. Followed by [#stress-physical-script-polish], [#stress-reporter-analyze], [#stress-reporter-cli].

## Summary of Changes
- **Orchestrator agent-log parser** learns the real-Arista-EOS unified-diff shape (\`interface TunnelN\` as a context line with \`+   property\` lines below it) on top of the existing cEOS \`+interface TunnelN\` shape. Without this, physical-hardware runs emitted zero \`applied\` events.
- **Two new agent activity events**: \`EventConfigReceived\` (from \`Received N bytes...\`) and \`EventCommitAborted\` (from \`session ... abort\`). The first keeps the post-deprovision quiescence wait alive across the multi-second diff-check window; the second clears the pending-commit flag during steady-state polling so the wait doesn't pin to its 5-min timeout.
- **\`quiescenceTracker\` grows a sticky pending-commit flag** — set on \`EventConfigReceived\`, cleared on the matching terminal event — so the wait blocks across the entire diff-check, not just the visible activity beats.
- **\`waitForAppliedToCatchUp\`** blocks the provision→deprovision boundary until the agent's applied count covers the orchestrator's provisioned-user count. New \`--apply-catch-up-timeout-seconds\` (default 300s) bounds the wait; \`--no-agent\` auto-zeros it.
- **Observer tunnel counter** switches from \`show gre tunnel static\` (returns only routing-fabric tunnels, never user tunnels on either platform tested) to \`show interfaces description\` filtered on the \`USER-UCAST-\` prefix the controller renders on every user-tunnel block. Robust against the controller's tunnel-id range (legacy 500+; current 1+).

## Diff Breakdown
| Category    | Files | Lines (+/-)   | Net   |
|-------------|-------|---------------|-------|
| Core logic  |     5 | +474   / -49  | +425  |
| Scaffolding |     2 | +33    / -2   | +31   |
| Tests       |     3 | +254   / -24  | +230  |
| **Total**   |    10 | +761   / -75  | +686  |

<details>

- [\`tools/stress/device-orchestrator/pkg/agent/parser.go\`](https://github.com/malbeclabs/doublezero/pull/new/nikw9944/stress-orchestrator-real-eos/files) — per-section state machine that opens a section on \`[ ]?interface TunnelN\`, promotes it on the first \`+   property\` line, and closes on \`!\` or the next section header. cEOS path preserved.
- [\`tools/stress/device-orchestrator/pkg/sweep/sweep.go\`](https://github.com/malbeclabs/doublezero/pull/new/nikw9944/stress-orchestrator-real-eos/files) — pending-commit flag on the tracker; \`waitForAppliedToCatchUp\` with timeout; consumeAgentEvents handles all 5 event kinds.
- [\`tools/stress/device-observer/internal/sample/eos.go\`](https://github.com/malbeclabs/doublezero/pull/new/nikw9944/stress-orchestrator-real-eos/files) — description-prefix filter for the tunnel counter.
- [\`tools/stress/device-orchestrator/pkg/agent/agent.go\`](https://github.com/malbeclabs/doublezero/pull/new/nikw9944/stress-orchestrator-real-eos/files) — \`EventConfigReceived\` and \`EventCommitAborted\` constants.
- [\`tools/stress/device-orchestrator/cmd/device-orchestrator/main.go\`](https://github.com/malbeclabs/doublezero/pull/new/nikw9944/stress-orchestrator-real-eos/files) — \`--apply-catch-up-timeout-seconds\` flag + \`--no-agent\` auto-zero.

</details>

## Testing Verification
- Unit tests cover the new real-EOS state machine, the activate↔applied edge cases, the per-cycle join, both new wait semantics (pending-commit blocking + apply catch-up + its timeout path), and the description-prefix tunnel counter.
- Replay-validated against archived agent logs from chi-dn-dzd5: the parser now yields 524/524 (was 0/0) PreCommitLog + Applied events; cEOS 992-tunnel run still yields 992/992 (no regression).
- End-to-end on physical hardware: 524-user runs on chi-dn-dzd5 + chi-dn-dzd9 and 1023-user runs on both DUTs completed with 100% on-device coverage and clean teardown.

### DIFF
--- a/tools/stress/device-observer/README.md
+++ b/tools/stress/device-observer/README.md
@@ -47,7 +47,7 @@ The observer writes the following files into `--working-dir`:
 | ---------------------------------------- | --------- | ----------------------------------------------- |
 | `observer-config.json`                   | observer  | selected flag values (excludes `eapi_pass` and `eapi_port`) + PID + start timestamp |
 | `show-hardware-capacity-<ts>.json`       | observer  | one per tick                                    |
-| `show-gre-tunnel-static-<ts>.json`       | observer  | one per tick                                    |
+| `show-interfaces-description-<ts>.json`  | observer  | one per tick                                    |
 | `show-processes-top-once-<ts>.json`      | observer  | one per tick                                    |
 | `show-logging-errors-<ts>.log`           | observer  | one per tick                                    |
 | `show-logging-critical-<ts>.log`         | observer  | one per tick                                    |
@@ -288,7 +288,7 @@ cancels the observer's root context so the process exits.
 | `lock_not_taken`             | agent-log substring `not overriding lock since its age` | same                                                                                     |
 | `agent_silence`              | `AgentTail.Snapshot().LastLineAt`                     | `LastLineAt` non-zero AND `now - LastLineAt > 120 s` (suppressed before any line is seen). Threshold accommodates the legitimate silent window during a large config apply at high user counts — at ~290 users the agent processes ~22k lines per cycle without intermediate logs. |
 | `ledger_heartbeat_stale`     | mtime of `<working-dir>/orchestrator.ledger_heartbeat` | file present AND `now - mtime > 30 s` (absent file is suppressed forward-compatibly)      |
-| `device_tunnel_gap`          | runlog `n_after_event` vs sampler `greTunnels` map length | orchestrator active-user count exceeds the device's tunnel count by `≥ 4` AND `now - last_activate ≥ 120 s` (suppressed before the first activate is seen). Grace matches `agent_silence` — past that, a stuck agent fires `agent_silence` first; a persistent gap with a healthy agent is the controller's slot cap or similar truncation. |
+| `device_tunnel_gap`          | runlog `n_after_event` vs sampler's `show interfaces description` filtered to interfaces whose description begins `USER-UCAST-` | orchestrator active-user count exceeds the device's tunnel count by `≥ 4` AND `now - last_activate ≥ 120 s` (suppressed before the first activate is seen). Grace matches `agent_silence` — past that, a stuck agent fires `agent_silence` first; a persistent gap with a healthy agent is the controller's slot cap or similar truncation. |
 
 **Startup grace.** The four counter and log-pattern triggers
 (`apply_config_errors`, `get_config_errors`, `diff_timeout`,
--- a/tools/stress/device-observer/internal/abort/decider.go
+++ b/tools/stress/device-observer/internal/abort/decider.go
@@ -99,8 +99,9 @@ type Sources struct {
 	// been seen.
 	ActiveUserCount func() (count int, lastActivate time.Time, ok bool)
 	// TunnelCount returns the most recently observed number of user
-	// tunnels on the device (from `show gre tunnel static`). `ok=false`
-	// before the first successful sample.
+	// tunnels on the device (parsed from `show interfaces description`,
+	// filtered to interfaces whose description begins `USER-UCAST-`).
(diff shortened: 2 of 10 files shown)


elitegreg (changes requested):
**[blocking] `waitForAgentQuiescence` can incorrectly skip the wait due to a race with the agent-event consumer.**
   - File: `tools/stress/device-orchestrator/pkg/sweep/sweep.go`
   - In `waitForAgentQuiescence`, the early return on `!tracker.lastEvent()` can fire even when an event was already emitted but not yet consumed by `consumeAgentEvents`.
   - This leaves `pendingCommit`/`lastEvent` unset and can cause immediate teardown while a commit is still in flight.
   - I can reproduce flaky behavior with:
     - `go test ./tools/stress/device-orchestrator/pkg/sweep -run TestRun_QuiescenceBlocksOnPendingCommit -count=100`
   - Suggestion: avoid the immediate "no events observed" fast-path, or add a short synchronization/settling step before deciding there were no events.

nikw9944:
Good catch — fixed in c5b27d0d.

Root cause was what you diagnosed: `tracker.lastEvent()` returns `ok=false` while `lastEventNanos == 0`, but the tracker is updated by `consumeAgentEvents` *after* it reads each event off the channel. If the agent emits an event right before the orchestrator returns from `deprovision()`, the event can be sitting in the channel buffer while the main goroutine reads tracker state and concludes 'no events' — racing with the consumer goroutine.

The fix has two parts:

1. **Explicit `Config.NoAgent` signal** instead of inferring 'no agent' from tracker state. `main.go` already knows when `--no-agent` is set (it auto-zeros `ApplyCatchUpTimeout` on that path), so it now also sets `cfg.NoAgent`. `waitForAgentQuiescence` short-circuits on the explicit flag — no race because it's an operator-provided declaration, not an inferred state.

2. **Drop the `if !ok { return }` fast-path** in the wait loop. With (1) in place, the loop's existing exit condition handles 'no events yet' correctly: `lastEvent` of zero yields a huge `sinceLast` (always > window), so the wait gates on `elapsed >= AgentQuietWindow` and blocks at least one quiet window before returning. A slow consumer goroutine gets the full quiet window to catch up.

Reproduction now stable. Tested with your exact command:

```
$ go test -run TestRun_QuiescenceBlocksOnPendingCommit -count=100 ./tools/stress/device-orchestrator/pkg/sweep
ok      .../pkg/sweep    100.171s

$ go test -race -run TestRun_QuiescenceBlocksOnPendingCommit -count=200 ./tools/stress/device-orchestrator/pkg/sweep
ok      .../pkg/sweep    201.569s
```

The downstream PRs (#3842, #3843, #3844) have been rebased onto the new tip.

elitegreg (approved):
Verified the fix in c5b27d0d. The racy `!tracker.lastEvent()` fast-path is gone, replaced by the explicit `Config.NoAgent` flag wired from `main.go`. Real-agent runs now always enter the wait loop and block at least one quiet window, so the consumer goroutine has time to drain any in-flight event and set `pendingCommit` before quiescence is declared. That closes the window the race lived in.

Ran the reproduction under the race detector — clean:

```
go test -race -run 'TestRun_QuiescenceBlocksOnPendingCommit|TestRun_SkipsQuiescenceWaitWhenNoAgent' -count=30 ./pkg/sweep
ok
```

Thanks for the quick turnaround. Approving.

PR #3850, opened 2026-06-06
telemetry: retry metrics-listener bind on EADDRNOTAVAIL/EADDRINUSE

nikw9944:
## Summary of Changes

* Wrap the device-telemetry-agent's Prometheus metrics-listener bind in a context-aware retry loop with capped exponential backoff (1s → 30s, indefinite until shutdown).
* `EADDRNOTAVAIL` and `EADDRINUSE` are retried (transient classes); all other errors surface immediately so a real configuration problem is not silently retried.
* Fixes the silent-failure mode from the 2026-05-18 incident on `chi-dn-dzd1`/`dzd2`/`dzd3` where the agent lost a race against `ns-management` IP assignment, the bind returned `EADDRNOTAVAIL`, and the daemon kept running for ~18 days with `up{component="device_telemetry_agent"} == 0`.
* Fixes malbeclabs/infra#1542

Companion docs PR with the design rationale: malbeclabs/infra#1546 (work plan in `docs/work-plan-1542.md`).

## Testing Verification

* `go test ./controlplane/telemetry/cmd/telemetry/` — all tests pass, including new ones in `metrics_listener_test.go`:
  * `TestListenWithRetry_SucceedsAfterTransientEADDRNOTAVAIL` — recovers after 2 transient EADDRNOTAVAIL failures, then succeeds.
  * `TestListenWithRetry_RetriesEADDRINUSE` — recovers after a transient EADDRINUSE failure.
  * `TestListenWithRetry_SurfaceNonRetryableErrorImmediately` — non-bind errors surface on attempt 1 (no silent retry).
  * `TestListenWithRetry_ReturnsOnContextCancel` — cancellation during backoff short-circuits within 500ms; cancel-path error wraps both the last bind error and `ctx.Err()`.
  * `TestNextBackoff` — verifies the documented 1s, 2s, 4s, 8s, 16s, 30s, 30s, ... sequence including the cap clamp.
  * `TestIsRetryableBindError` — classifier table covers nil, EADDRNOTAVAIL, EADDRINUSE, generic, and ECONNREFUSED.
* `go build ./controlplane/telemetry/cmd/telemetry/` and `go vet ./controlplane/telemetry/...` both clean.
* Manual e2e exercise (racing the namespace IP on real Arista cEOS) is not part of the unit-test harness — that path will be observed in prod after the next agent release rolls out, watching `up{component="device_telemetry_agent"}` and the `Component: Service Down` alert on the previously affected DZD devices.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
+- Telemetry
+  - Retry `EADDRNOTAVAIL` / `EADDRINUSE` failures on the device-telemetry-agent's Prometheus metrics-listener bind with capped exponential backoff (1s → 30s, indefinite until context cancel), so a startup race against `ns-management` IP assignment no longer leaves the agent running silently without a `/metrics` endpoint ([malbeclabs/infra#1542](https://github.com/malbeclabs/infra/issues/1542))
+
 ## [v0.26.0](https://github.com/malbeclabs/doublezero/compare/client/v0.25.1...client/v0.26.0) - 2026-06-05
 
 ### Breaking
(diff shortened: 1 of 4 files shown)


nikw9944:
Not a high priority at the moment

Issue #3856, opened 2026-06-09
Dynamic Seat Allocation (RFC 21) — tracking issue


elitegreg:
## Dynamic Seat Allocation (RFC 21) — tracking issue

A DoubleZero AccessPass is currently bound to a single `(client_ip, user_payer)` pair, so a buyer must
know the client's public IP up front and one purchase serves exactly one machine. This effort introduces
**dynamic seat allocation**: a buyer purchases one entitlement that admits several users, and a node
onboards without registering its IP or device ahead of time. It adds per-category user caps to the
AccessPass (enforced for seat passes), reuses the dynamic IP-less AccessPass primitive, and adds a
**`ConnectionTicket`** account to the Feed Subscription Program. The ticket custodies USDC and carries the
seat's caps, its funding authority, and the service key; the Feed Oracle turns each ticket into an
EdgeSeat AccessPass and funds a per-connection v1 `ClientSeat`/`PaymentEscrow` for each node that
connects. The serviceability `SetAccessPass` airdrop is scaled by the seat's caps so the service keypair
is funded to provision its own users through the standard `doublezero connect` workflow. No new endpoint
is added to the Feed Oracle.

**RFC:** https://github.com/malbeclabs/doublezero/blob/rfc-virtual-accesspass/rfcs/rfc21-dynamic-seat-allocation.md _(repoint to `main` once merged)_

Delivered as a sequence of <500-line PRs across two repositories. Sections D–E are tracked in
`malbeclabs/infra`; their code (Feed Subscription Program / Feed Oracle) lives in
`malbeclabs/doublezero-shreds`.

## Sub-issues (delivery order)

- [x] **A — serviceability: per-category user caps + scaled airdrop on AccessPass** — malbeclabs/doublezero#3852 _(no deps; blocks the rest)_
- [x] **B — sdk: UNSPECIFIED-first AccessPass lookup** — malbeclabs/doublezero#3853 _(after A)_
- [x] **D — shred-subscription: ConnectionTicket account + connection-ticket instructions** — malbeclabs/infra#1575 _(no deps; implemented in malbeclabs/doublezero-shreds#465)_
- [x] **E.1–2 — shred-oracle: EdgeSeat AccessPass creation from ConnectionTickets (feed-group allowlist + auto-funding) + fund per-connection seats** — malbeclabs/infra#1576 _(after A + D)_

**Client (RFC §C): no code.** The operator runs the standard `doublezero connect` with the funded service
[...]

PR #3859, opened 2026-06-09
serviceability: per-category user caps + scaled airdrop on AccessPass

### LINKED ISSUE #3852: serviceability: per-category user caps + scaled airdrop on AccessPass
Part of the **Dynamic Seat Allocation** effort (RFC §A). Tracking issue: #3856.

RFC: https://github.com/malbeclabs/doublezero/blob/rfc-virtual-accesspass/rfcs/rfcXX-dynamic-seat-allocation.md (repoint to `main` once merged).

## Why
Bound how many users a single `EdgeSeat` AccessPass admits, split by category (IBRL/unicast vs. multicast), and fund the seat keypair to provision those users. Both the cap enforcement and the airdrop scaling are gated on `EdgeSeat`, so existing Prepaid/validator/RPC passes are unaffected.

## Scope
- Append four fields to `AccessPass` in `smartcontract/programs/doublezero-serviceability/src/state/accesspass.rs`: `unicast_user_count: u16`, `max_unicast_users: u16`, `multicast_user_count: u16`, `max_multicast_users: u16`. Apply read-defaults in the manual `TryFrom<&[u8]>` (counts → `0`, maxes → `1`), mirroring the existing `unwrap_or(Ipv4Addr::UNSPECIFIED)` pattern, so pre-existing accounts stay deserialize-compatible.
- Add methods `try_add_user(UserType)` and `remove_user(UserType)` — EdgeSeat-only (no-op + success for all other types). They must **not** touch `connection_count` (that increment/decrement stays where it is).
- Add error variants `AccessPassMaxUnicastUsersExceeded` and `AccessPassMaxMulticastUsersExceeded` in `error.rs` (distinct from the device-scoped `MaxUnicastUsersExceeded`).
- Hook `accesspass.try_add_user(user_type)?` next to `accesspass.connection_count += 1` in `processors/user/create_core.rs`. Remove the block that locks an unspecified-IP pass to its first user's IP, and remove the now-unused `IS_DYNAMIC` flag.
- Hook `accesspass.remove_user(user_type)` alongside the `saturating_sub(1)` in `processors/user/delete.rs`.
- Add `max_unicast_users` / `max_multicast_users` with `#[incremental(default = 1)]` to `SetAccessPassArgs` in `processors/accesspass/set.rs`; persist on create, update on set, leave live counts untouched; drop the set-time logic that derived the removed `IS_DYNAMIC` flag.
- **Scale the SetAccessPass airdrop.** `process_set_access_pass` already airdrops to `user_payer` enough to rent-exempt the `User` accounts a connect creates (`AIRDROP_USER_RENT_LAMPORTS_BYTES`) plus `globalstate.user_airdrop_lamports`, topping the account up to that target. For `EdgeSeat` passes, scale the target by `max_unicast_users + max_multicast_users` so the seat keypair holds enough SOL to pay for every node's `create_user`. Non-EdgeSeat passes keep today's fixed airdrop; the existing top-up semantics (`saturating_sub(user_payer.lamports())`) are preserved.
- CLI: add `--max-unicast-users` / `--max-multicast-users` to `smartcontract/cli/src/accesspass/set.rs`; render the four new fields in `get.rs` and `list.rs`.
- SDKs: add the four fields to the AccessPass layouts in `smartcontract/sdk/go/serviceability/state.go`, `sdk/serviceability/python/serviceability/state.py`, `sdk/serviceability/typescript/serviceability/state.ts`; regenerate fixtures with `make generate-fixtures`.

## Acceptance
- An EdgeSeat pass rejects the (N+1)th user per category with the new errors; non-EdgeSeat passes are unaffected.
- Creating/updating an EdgeSeat pass airdrops a balance scaled by the caps to `user_payer`; non-EdgeSeat passes get the existing fixed airdrop.
- Old accounts and older binary fixtures still deserialize.
- `make rust-test` and `make sdk-test` pass.

**Depends on:** none. **Blocks:** the SDK lookup and oracle issues.

elitegreg:
Closes #3852

Part of the **Dynamic Seat Allocation** effort (RFC §A, tracking #3856). RFC: https://github.com/malbeclabs/doublezero/blob/rfc-virtual-accesspass/rfcs/rfcXX-dynamic-seat-allocation.md

## Summary of Changes
- Add per-category seat caps to `AccessPass` (`unicast_user_count`/`max_unicast_users`/`multicast_user_count`/`max_multicast_users`), enforced **only** for `EdgeSeat` passes via new `try_add_user`/`remove_user` helpers hooked into user create/delete. The (N+1)th user per category is rejected with new errors `AccessPassMaxUnicastUsersExceeded` (89) / `AccessPassMaxMulticastUsersExceeded` (90); all other pass types are unaffected.
- Scale the `SetAccessPass` airdrop by the per-category caps when `allow_multiple_ip` is set, so a multi-IP seat keypair holds enough SOL to provision every node it admits. Passes without the flag keep today's fixed 1× airdrop regardless of their max fields.
- Remove the legacy "lock a dynamic pass to its first user's IP" behavior and the now-unused `IS_DYNAMIC` flag. A pass at the UNSPECIFIED (`0.0.0.0`) PDA stays dynamic for its lifetime; seat counting replaces IP-locking as admission control. **Behavioral change** — see below.
- Add `--max-unicast-users` / `--max-multicast-users` to the `doublezero access-pass set` CLI and render the new counts/caps in `get`/`list`.
- Add the four fields to the Go, Python, and TypeScript AccessPass layouts (also filling the pre-existing `tenant_allowlist` read gap in Go/Python so the new trailing fields decode correctly), and regenerate fixtures (plus a new `access_pass_edge_seat` fixture).

**Behavioral change:** dynamic (UNSPECIFIED-IP) access passes no longer lock to the first connecting client IP. Pre-existing accounts and older binary fixtures still deserialize (counts default to 0, caps to 1).

## Diff Breakdown
| Core logic   |     5 | +210 / -24   |  +186 |
| Scaffolding  |    28 | +294 / -6    |  +288 |
| Tests        |    16 | +515 / -10   |  +505 |
| Fixtures     |     8 | +223 / -2    |  +221 |
| **Total**    |    57 | +1242 / -42  | +1200 |

Small, focused program change; most of the diff is mechanical field-plumbing through SDK/CLI struct literals plus tests and regenerated fixtures. This exceeds DoubleZero's ~500-line guideline, but the onchain fields and the read-only SDK layouts must land together to stay byte-compatible, so it is kept as one PR.

<details>

- `smartcontract/programs/doublezero-serviceability/src/state/accesspass.rs` — four new fields, read-defaults in `TryFrom`, `try_add_user`/`remove_user` (EdgeSeat-only), `IS_DYNAMIC` removal, unit tests
- `smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set.rs` — `max_*` args (`#[incremental(default = 1)]`), persist on create/update, `allow_multiple_ip`-gated airdrop scaling
- `smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs` — drop IP-lock block, skip IP check for UNSPECIFIED passes, hook `try_add_user`
- `smartcontract/programs/doublezero-serviceability/src/processors/user/delete.rs` — hook `remove_user` alongside connection_count decrement
- `smartcontract/programs/doublezero-serviceability/src/error.rs` — `AccessPassMaxUnicastUsersExceeded` / `AccessPassMaxMulticastUsersExceeded`
- `smartcontract/sdk/go/serviceability/{state,deserialize}.go`, `sdk/serviceability/python/serviceability/state.py`, `sdk/serviceability/typescript/serviceability/state.ts` — AccessPass layout + `tenant_allowlist`/cap reads

</details>

## Testing Verification
- New unit tests for `try_add_user`/`remove_user` cap and no-op behavior, and `TryFrom` read-defaults (caps → 1, counts → 0).
- `test_edge_seat_user_caps_enforced`: an EdgeSeat pass admits one user per category and rejects the next with `Custom(89)`/`Custom(90)`; unicast and multicast are independent.
- `test_set_accesspass_airdrop_scales_with_allow_multiple_ip`: a flag-on pass airdrops `(max_unicast + max_multicast)`× the base; a flag-off pass gets the unscaled base.
- Updated the dynamic-pass delete tests to assert `client_ip` stays UNSPECIFIED (no IP-locking).
- `make rust-test` (serviceability program + SDK + CLI) and the Go/Python SDK suites pass against regenerated fixtures, including the new EdgeSeat fixture. TypeScript fixture tests were added but not run locally (no `bun` available in this environment).


smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set.rs
```diff
@@ -335,11 +352,21 @@ pub fn process_set_access_pass(
...
         .unwrap()
         .minimum_balance(AIRDROP_USER_RENT_LAMPORTS_BYTES)
-        .saturating_add(globalstate.user_airdrop_lamports)
+        .saturating_add(globalstate.user_airdrop_lamports);
+    let multiplier = if value.allow_multiple_ip {
+        (value.max_unicast_users as u64).saturating_add(value.max_multicast_users as u64)
+    } else {
+        1
+    };
+    let deposit = base_target
```
juan-malbeclabs, line 368: The airdrop multiplier is keyed on `allow_multiple_ip`, but the seat caps are only enforced for `EdgeSeat` passes (`try_add_user`/`remove_user` are no-ops otherwise). These two conditions are decoupled, which produces two inconsistent cases:

- **EdgeSeat at the UNSPECIFIED PDA without `allow_multiple_ip`**: since the `0.0.0.0` PDA admits any client IP by construction (per the new `create_core.rs` logic), it can create up to `max_unicast_users + max_multicast_users` users, but the airdrop stays at 1x → the keypair is under-funded for the rent of all the `User` accounts it admits. This directly contradicts the comment ("scale the target ... so the keypair holds enough SOL to pay for every create_user it admits").
- **Non-EdgeSeat with `allow_multiple_ip`**: gets the scaled airdrop (≥2x with the 1/1 defaults) but the caps are never enforced, so it over-funds something with no seat limit.

Suggestion: key the multiplier on `EdgeSeat` (the type that actually determines capacity), or validate `EdgeSeat ⇒ allow_multiple_ip`. As a side effect, note that existing non-EdgeSeat `allow_multiple_ip` passes now jump from 1x to 2x airdrop with no change in real capacity.

smartcontract/sdk/go/serviceability/deserialize.go
```diff
@@ -381,6 +381,11 @@ func DeserializeAccessPass(reader *ByteReader, ap *AccessPass) {
 	ap.MGroupPubAllowlist = reader.ReadPubkeySlice()
 	ap.MGroupSubAllowlist = reader.ReadPubkeySlice()
 	ap.Flags = reader.ReadU8()
+	ap.TenantAllowlist = reader.ReadPubkeySlice()
+	ap.UnicastUserCount = reader.ReadU16()
+	ap.MaxUnicastUsers = reader.ReadU16()
+	ap.MulticastUserCount = reader.ReadU16()
+	ap.MaxMulticastUsers = reader.ReadU16()
```
juan-malbeclabs: Default divergence between the Rust program and the SDKs for not-yet-migrated accounts. Rust `TryFrom` defaults `max_*_users` to **1** (`unwrap_or(1)`), matching what the program effectively enforces. But when reading an old account (which lacks these 8 bytes), `ByteReader.ReadU16` returns **0** on EOF, so `MaxUnicastUsers`/`MaxMulticastUsers` come back as `0`. The Python (`DefensiveReader`) and TypeScript (`readU16`) layouts behave the same way — the dataclass/struct defaults get overwritten with the `0` read from EOF.

Net effect: an SDK consumer sees `max_unicast_users = 0` (looks like "no seats") while the on-chain program admits 1. It self-heals once the account is rewritten, but until then dashboards/monitoring can misreport caps. Consider normalizing `0 → 1` in the SDK decoders (or documenting the discrepancy) so SDK reads match the program’s effective default.

elitegreg: Good catch — fixed in fbc166ea. The Go/Python/TS decoders now default `max_unicast_users`/`max_multicast_users` to **1** when the field bytes are absent (EOF), matching the program’s `unwrap_or(1)`. It’s gated on EOF (`Remaining()/remaining` check) rather than a blanket `0 → 1`, so a migrated account that explicitly stores `0` is preserved. The counts keep defaulting to `0` (matching `unwrap_or(0)`). Added a `…LegacyCapDefaults` regression test in each SDK that decodes a pre-migration account (fixture minus the 8 trailing cap bytes) and asserts caps read back as 1.

juan-malbeclabs (approved):
LGTM

PR #3860, opened 2026-06-09
sdk: UNSPECIFIED-first AccessPass lookup in GetAccessPassCommand

### LINKED ISSUE #3853: sdk: UNSPECIFIED-first AccessPass lookup in GetAccessPassCommand
Part of the **Dynamic Seat Allocation** effort (RFC §B). Tracking issue: #3856.

RFC: https://github.com/malbeclabs/doublezero/blob/rfc-virtual-accesspass/rfcs/rfcXX-dynamic-seat-allocation.md (repoint to `main` once merged).

## Why
The read path should find a shared **dynamic** seat pass before the exact-IP pass, aligning the read side with the onchain `create_user` path (which already tries both the exact-IP PDA and the `UNSPECIFIED` PDA). This is a read-only change; it does not alter onchain behavior. It is the key client-side enabler for seats: with it, the standard `doublezero connect` resolves a node's dynamic seat pass (at `accesspass(UNSPECIFIED, user_payer)`), and no other client changes are needed.

## Scope
- In `smartcontract/sdk/rs/src/commands/accesspass/get.rs`, change `GetAccessPassCommand::execute` so that when `client_ip != UNSPECIFIED` it first tries `get_accesspass_pda(UNSPECIFIED, user_payer)` and falls back to the exact-IP PDA only if that account is absent.

## Acceptance
- A dynamic seat pass resolves first; existing per-IP passes still resolve via the fallback.
- The `connect` command's `check_accesspass` (in `client/doublezero/src/command/connect.rs`) benefits automatically since it goes through this command.

**Depends on:** #3852 (ordered after the AccessPass caps work).

elitegreg:
## Summary

- Make `GetAccessPassCommand::execute` resolve a shared **dynamic-seat** AccessPass (the `UNSPECIFIED`/`0.0.0.0` PDA) before the exact-IP pass, falling back to the exact-IP PDA only when the dynamic pass is absent.
- Aligns the SDK read path with the onchain `create_user` path, which already accepts either the exact-IP PDA or the `UNSPECIFIED` PDA. This is a read-only change with no onchain behavior impact.
- `connect`'s `check_accesspass` benefits automatically, since it routes through this command — so `doublezero connect` can now resolve a node's dynamic seat pass with no other client changes.

When `client_ip` is already `UNSPECIFIED`, a single lookup is performed (no redundant double query).

Part of the Dynamic Seat Allocation effort (RFC-21, see #3858). Closes #3853.

## Dependency

Stacks on #3859 (AccessPass user caps, branch `gm/accesspass-user-caps`) and is based on that branch. Retarget to `main` once #3859 merges.

## Testing Verification

- Added unit tests for `GetAccessPassCommand` covering: dynamic pass preferred over exact-IP, fallback to exact-IP when the dynamic pass is absent, `None` when neither exists, and a single lookup when `client_ip == UNSPECIFIED`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file.
   - `access-pass set` gains `--max-unicast-users` / `--max-multicast-users`; `get`/`list` show the per-category counts and caps.
 - SDK
   - Decode the four new `AccessPass` cap fields (and the previously-missing `tenant_allowlist`) in the Go, Python, and TypeScript layouts.
+  - `GetAccessPassCommand` resolves a shared dynamic-seat AccessPass (the `UNSPECIFIED` PDA) before the exact-IP pass, matching the onchain `create_user` lookup. (#3853)
 
 ## [v0.26.0](https://github.com/malbeclabs/doublezero/compare/client/v0.25.1...client/v0.26.0) - 2026-06-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ All notable changes to this project will be documented in this file.
   - `access-pass set` gains `--max-unicast-users` / `--max-multicast-users`; `get`/`list` show the per-category counts and caps.
 - SDK
   - Decode the four new `AccessPass` cap fields (and the previously-missing `tenant_allowlist`) in the Go, Python, and TypeScript layouts.
-  - `GetAccessPassCommand` resolves a shared dynamic-seat AccessPass (the `UNSPECIFIED` PDA) before the exact-IP pass, matching the onchain `create_user` lookup. (#3853)
+  - `GetAccessPassCommand` and the multicast allowlist resolver both resolve a shared dynamic-seat AccessPass (the `UNSPECIFIED` PDA) before the exact-IP pass, matching the onchain `create_user` lookup. (#3853)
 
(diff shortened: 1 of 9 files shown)


juan-malbeclabs (approved):
Approving — pending resolution of the item below (follow-up is fine; not blocking the merge).

**Divergent AccessPass resolution strategies in the SDK**

This PR makes `GetAccessPassCommand::execute` prefer the dynamic (`UNSPECIFIED`) PDA first, unconditionally. But `multicastgroup/allowlist/mod.rs::resolve_accesspass_pda` does the opposite: it tries the exact-IP PDA first and only falls back to the dynamic PDA when `ap.allow_multiple_ip()` is true.

So after this change the SDK has two resolvers with opposite precedence and different handling of `allow_multiple_ip`:

- `GetAccessPassCommand` — dynamic first, always (no flag check).
- `resolve_accesspass_pda` — static first, dynamic only if `allow_multiple_ip()`.

The new `GetAccessPassCommand` behavior is actually the one more aligned with the onchain semantics, which state that a pass stored at the `UNSPECIFIED` PDA is valid for any client IP "by construction" (`create_core.rs`), independent of `allow_multiple_ip`. That makes `resolve_accesspass_pda`'s flag gate look like the odd one out.

Not blocking this PR, but please open a follow-up to unify the two resolvers (or document why allowlist ops intentionally differ) so the chosen AccessPass doesn't depend on which verb you went through.

PR #3861, opened 2026-06-09
client: auto-allocate IP on connect when behind NAT

elitegreg:
## Summary of Changes

- `doublezero connect ibrl` now auto-enables allocated-IP mode (`IBRLWithAllocatedIP`) when the host's default-route source address is a private RFC1918 address, i.e. the host is behind NAT. Plain IBRL can't work in that case because the client can't advertise its own private address.
- The daemon (`doublezerod`) is the source of truth: its client-IP discovery already does the default-route lookup, and now records whether that source was RFC1918 and exposes it on `/v2/status` as `behind_nat`.
- The CLI reads `behind_nat` and flips the user type, printing an informational message explaining the auto-enable. It is skipped when `-a`/`--allocate-addr` is already set or when the user supplied `--client-ip`, so an explicit choice always wins.
- Detection is strictly RFC1918 (10/8, 172.16/12, 192.168/16) via Go's `net.IP.IsPrivate()`; CGNAT and other reserved ranges do not trigger it.

## Diff Breakdown

| Core logic   |     3 | +171 / -20    |  +151 |
| Scaffolding  |     8 | +36 / -2      |  +34  |
| Tests        |     2 | +117 / -1     |  +116 |
| **Total**    |    13 | +324 / -23    |  +301 |

Most of the diff is tests and mechanical wiring; the behavior change is concentrated in `clientip.go` (NAT detection) and `connect.rs` (the auto-allocate decision). The `connect.rs` line count is dominated by new tests and a fixture refactor.

<details>

- [`client/doublezero/src/command/connect.rs`](https://github.com/malbeclabs/doublezero/pull/3861/files#diff-0e50f0119aa96b168465129d4b6f763c1125ae34defc1d682d9858bf9aac8184) — flip `IBRL → IBRLWithAllocatedIP` when `behind_nat` and no client IP was supplied; print the explanatory message; new tests + fixture support
- [`client/doublezerod/internal/runtime/clientip.go`](https://github.com/malbeclabs/doublezero/pull/3861/files#diff-cc5b9be9fd6937d7e0633c63424bc3fc8484f2567a1ebf9bb968951742e28ddc) — `DiscoverClientIP` returns `behindNAT`; default-route source lookup split out and made injectable; RFC1918 source flags NAT while the public IP still comes from external discovery
- [`client/doublezero/src/command/helpers.rs`](https://github.com/malbeclabs/doublezero/pull/3861/files#diff-11f78aa3fd6a78046dfdc8874a07dc06a16ee55948756ba564eee40423ad4717) — add `resolve_client_ip_with_nat`, sharing IP-parse logic with `resolve_client_ip`
- [`client/doublezerod/internal/manager/manager.go`](https://github.com/malbeclabs/doublezero/pull/3861/files#diff-27424874932194e8d91a18571b7c6a3bf1f1bfa057986e442db5324e6fced881) — `behindNAT` field + `WithBehindNAT` option
- [`client/doublezero/src/servicecontroller.rs`](https://github.com/malbeclabs/doublezero/pull/3861/files#diff-1db99e1cb86c0b5cdde04b7a5d368e21308127a7f57570b55628811ff6491b22) — `behind_nat` field on `V2StatusResponse` (`serde(default)` for older-daemon compat)
- [`client/doublezerod/internal/manager/http.go`](https://github.com/malbeclabs/doublezero/pull/3861/files#diff-5db980b5d3ad21e383edbd8d3ecb5a1f9c9db61590b1cf0f2364f7ea4df60f80) — `behind_nat` JSON field, populated in `ServeV2Status`

</details>

## Testing Verification

- Rust: added connect tests covering the three cases — auto-allocates when behind NAT, stays plain IBRL when `--client-ip` is supplied, and stays allocated when `-a` is already set. Full `doublezero` bin suite passes (144 tests).
- Go: added a table test for discovery (RFC1918 → NAT-flagged + external fallback, public → used directly, CGNAT → not flagged, no-route → not flagged) and a `ServeV2Status` test asserting `behind_nat` round-trips. `internal/runtime` passes; `go vet` clean.
- Note: `TestServeV2Status_Enrichment` fails locally on multicast provisioning (`socket: operation not permitted`, needs sudo for raw sockets) — verified it fails identically on `origin/main`, so it is pre-existing and unrelated.

## Note for reviewers

Behavior change: a host that previously connected as plain `IBRL` and now connects from behind NAT (without `-a`) will be steered to a new `IBRLWithAllocatedIP` user.


juan-malbeclabs (approved):
## Conditional approval ✅ (pending one fix)

Solid, well-tested change with the right separation of concerns: the daemon owns NAT detection and the CLI just reacts to `behind_nat`. Backward compat (`serde(default)`, additive JSON field) and the injectable `defaultRouteSource` refactor are clean.

**Approving conditional on resolving the `--client-ip` guard below.**

### Blocking: `self.client_ip.is_none()` guard contradicts the deprecation of CLI `--client-ip`

In `connect.rs`:

```rust
if behind_nat && self.client_ip.is_none() {
    // ... upgrade IBRL -> IBRLWithAllocatedIP
}
```

The CLI `--client-ip` flag is documented as **deprecated and ignored** (`connect.rs:76`), and the runtime warns "will be ignored. Set --client-ip on the daemon instead." But this guard treats a non-`None` CLI `client_ip` as "the user wants to control the address" and **skips** the NAT auto-upgrade.

Net effect: a user who passes the deprecated, otherwise-ignored CLI flag while behind NAT ends up as **plain IBRL behind NAT — exactly the broken state this PR exists to prevent** — while being told the flag has no effect.

The legitimate override is already handled daemon-side: `DiscoverClientIP` returns `behindNAT=false` whenever an explicit IP is supplied, so `behind_nat` can only be true when the IP was auto-discovered. That makes the CLI-side `self.client_ip.is_none()` check redundant for the real override path and actively harmful for the deprecated one.

**Suggested fix:**
- Drop the `self.client_ip.is_none()` condition and gate solely on `behind_nat`.
- Update the info message — `"set --client-ip on the daemon to override"` is the only true override now; the CLI flag should not be presented (implicitly) as one.
- Update `test_connect_command_ibrl_behind_nat_skipped_when_client_ip_set`, which currently codifies this behavior.

### Non-blocking
- CGNAT (100.64.0.0/10) is deliberately excluded but such hosts also can't do plain IBRL. If intentional, a one-line comment explaining why CGNAT is not flagged would prevent a future "bug" report.

Once the `--client-ip` guard is reconciled, this is good to merge.

elitegreg:
Thanks @juan-malbeclabs — good catch, you're right that the daemon already encodes the override.

**Blocking (fixed in c83fdee1):**
- Dropped the `self.client_ip.is_none()` condition; the auto-upgrade now gates solely on `behind_nat`. Since `DiscoverClientIP` returns `behindNAT=false` for any explicit daemon IP, the only true override (`--client-ip` on the daemon) is honored daemon-side, and the deprecated CLI flag no longer drops a NAT'd host back into plain IBRL.
- Updated the code comment to explain the daemon-side gating. The info message already pointed only at the daemon flag (`set --client-ip on the daemon to override`), so it stands as-is.
- Repurposed `test_connect_command_ibrl_behind_nat_skipped_when_client_ip_set` → `..._ignores_deprecated_cli_client_ip`: it now asserts the deprecated CLI `--client-ip` does **not** suppress the upgrade (still steers to `IBRLWithAllocatedIP`).

**Non-blocking (CGNAT):** Intentional — added a comment at the `IsPrivate()` check noting CGNAT (100.64.0.0/10) is deliberately out of scope (RFC1918 only) and where to widen it if we decide to cover it later.

(Also rebased onto the `main` merge that landed on the branch.)

juan-malbeclabs (approved):

PR #3875, opened 2026-06-10
ci(e2e): skip e2e on docs and version-bump-only PRs

ben-dz:
## Summary of Changes
- Skip the e2e and shreds-e2e suites on PRs where the suites add no signal — version bumps and docs/RFC-only changes — so they don't burn ~20 minutes of self-hosted runner time building images and running the full suite.
- A new gate step in each workflow's `setup` job inspects the PR's changed files via the API. A PR qualifies only if every file is either markdown (`*.md`, any path; renames/copies must come from markdown too, since a rename also deletes the source path) or `Cargo.toml`/`Cargo.lock` with diffs touching nothing but `version = "..."` lines, with the two Cargo files changing together. A dependency bump fails the check — `cargo update` changes Cargo.lock checksum/source lines, an inline requirement change in Cargo.toml doesn't match the version-line shape, and a requirement change that leaves the lock untouched fails the pairing rule — as does any code file or an API-truncated patch. All of those fall through to a normal e2e run. Markdown is verified inert: no `.md` files are used as test fixtures, embedded via `include_str!`, or read by e2e code.
- Because the gated matrix jobs never create their shard check runs, the gate reports the required `e2e (shard 1-5)` / `shard-e2e (shard 1-4)` contexts as successful on the PR head via the checks API (same pattern the trusted fork-PR flow already uses), so branch protection doesn't block the merge.
- The gate runs only on same-repo `pull_request` events; fork PRs, pushes to main/hotfix, and trusted `workflow_dispatch` runs are unaffected. Any gate error fails open to running e2e.

## Diff Breakdown
| Config/build |     2 | +152 / -0   | +152 |

Workflow-only change; the same gate logic is added to both e2e workflows.

<details>

- `.github/workflows/e2e.yml` — skip-gate step + early exit in the existing `run-e2e` gate; reports `e2e (shard 1-5)` checks when skipping
- `.github/workflows/shreds-e2e.yml` — same gate; reports `shard-e2e (shard 1-4)` checks when skipping

</details>

## Testing Verification
- Ran the gate's file-classification logic (extracted verbatim into a node harness) against real PR data: #3874 (0.27.1 version bump) correctly classifies as skippable; #3870 (code change) correctly runs e2e.
- Synthetic dependency-bump cases all run e2e: `cargo update`-shaped Cargo.lock diff (checksum lines); dependency requirement change in Cargo.toml, both inline and multi-line-table form with the lock untouched; lock-only version-line change.
- Synthetic markdown cases: new RFC added, RFC+README edits, markdown deleted, markdown renamed from markdown, and version bump + RFC edit all skip; markdown plus a code file, a file renamed from `.rs` to `.md`, and a Cargo.lock with an API-omitted patch all run e2e.
- `actionlint` passes on both workflows.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -43,13 +43,75 @@ jobs:
     permissions:
       packages: write
       contents: read
+      checks: write
     outputs:
       run-e2e: ${{ steps.gate.outputs.run-e2e }}
       matrix: ${{ steps.shard.outputs.matrix }}
     steps:
+      # A version-bump PR changes only version lines in Cargo.toml/Cargo.lock
+      # plus CHANGELOG.md; e2e adds no signal there. When skipping, report the
+      # required shard checks as successful on the PR head, because the gated
+      # matrix job never creates its check runs and branch protection would
+      # otherwise block the merge. Keep in sync with shreds-e2e.yml.
+      - name: Check for version-bump-only PR
+        id: bump
+        if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
+        uses: actions/github-script@v7
+        env:
+          CHECK_NAME: e2e
+          # SHARD_COUNT round-robin shards + 1 dedicated shard; must match the
+          # required status check contexts in the main ruleset.
+          CHECK_SHARDS: "5"
+        with:
+          script: |
+            try {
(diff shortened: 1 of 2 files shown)


elitegreg (approved):
How about any *.md file?

PR #3878, opened 2026-06-11
serviceability: require the device's contributor on device/interface mutations

ben-dz:
## Summary of Changes
- `UpdateDevice`, `CreateDeviceInterface`, `UpdateDeviceInterface`, and `DeleteDeviceInterface` now verify that the supplied contributor account is the one the target device belongs to (`device.contributor_pk`), returning `InvalidContributorPubkey` on mismatch. Previously they only checked that the contributor account was owned by the payer, so a contributor passing their own contributor account with the wrong device pubkey could mutate a device belonging to a different contributor — an easy way to fat-finger an unexpected device.
- This mirrors the binding check `DeleteDevice` and the link processors already perform, closing a consistency gap. Foundation allowlist members bypass the check, matching the existing behavior of those processors.
- No client-side changes needed: the Rust SDK already passes `device.contributor_pk` as the contributor account for all of these instructions.

## Diff Breakdown
| Core logic   |     4 | +32 / -0    |  +32 |
| Tests        |     1 | +420 / -0   | +420 |
| **Total**    |     5 | +452 / -0   | +452 |

A small, uniform guard added to four processors; the bulk of the diff is new integration test coverage.

<details>

- `smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs` — binding check in `process_update_device`
- `smartcontract/programs/doublezero-serviceability/src/processors/device/interface/create.rs` — binding check in `process_create_device_interface`
- `smartcontract/programs/doublezero-serviceability/src/processors/device/interface/update.rs` — binding check in `process_update_device_interface`
- `smartcontract/programs/doublezero-serviceability/src/processors/device/interface/delete.rs` — binding check in `process_delete_device_interface`

</details>

## Testing Verification
- New integration test file `device_contributor_binding_test.rs`: negative tests assert each of `UpdateDevice`, `CreateDeviceInterface`, `UpdateDeviceInterface`, and `DeleteDeviceInterface` rejects a mismatched contributor with `InvalidContributorPubkey` (Custom(10)) and leaves the device untouched; a positive test asserts a foundation allowlist member can still act with a non-matching contributor.
- Existing `test_device_update_multicast_counts_ignored_for_non_foundation_payer` continues to cover the non-foundation, matching-contributor success path.
- Full `doublezero-serviceability` suite passes (262 unit + all integration tests), confirming no existing flow regresses.

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/processors/device/interface/create.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/device/interface/create.rs
@@ -182,6 +182,14 @@ pub fn process_create_device_interface(
 
     let mut device: Device = Device::try_from(device_account)?;
 
+    // The supplied contributor must be the one the device belongs to,
+    // unless the payer is on the foundation allowlist.
+    if !globalstate.foundation_allowlist.contains(payer_account.key)
+        && device.contributor_pk != *contributor_account.key
+    {
+        return Err(DoubleZeroError::InvalidContributorPubkey.into());
+    }
+
     if device.find_interface(&name).is_ok() {
         return Err(DoubleZeroError::InterfaceAlreadyExists.into());
     }
--- a/smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs
@@ -226,6 +226,14 @@ pub fn process_update_device(
 
     let mut device: Device = Device::try_from(device_account)?;
 
+    // The supplied contributor must be the one the device belongs to,
+    // unless the payer is on the foundation allowlist.
+    if !globalstate.foundation_allowlist.contains(payer_account.key)
+        && device.contributor_pk != *contributor_account.key
+    {
+        return Err(DoubleZeroError::InvalidContributorPubkey.into());
+    }
+
     // Only allow updates from the foundation allowlist
     if globalstate.foundation_allowlist.contains(payer_account.key) {
         if let Some(contributor_pk) = value.contributor_pk {
--- a/smartcontract/programs/doublezero-serviceability/tests/device_contributor_binding_test.rs
+++ b/smartcontract/programs/doublezero-serviceability/tests/device_contributor_binding_test.rs
@@ -1,0 +1,420 @@
+//! Verifies that device mutation instructions only accept the contributor the
+//! device belongs to, so a contributor working with their own contributor
+//! account cannot target another contributor's device by mistake. Foundation
+//! allowlist members may act on any device.
+
+use doublezero_serviceability::{
+    instructions::*,
+    pda::*,
+    processors::{
+        contributor::create::ContributorCreateArgs,
+        device::{
+            create::DeviceCreateArgs,
+            interface::{
+                create::DeviceInterfaceCreateArgs, delete::DeviceInterfaceDeleteArgs,
+                update::DeviceInterfaceUpdateArgs,
+            },
+            update::DeviceUpdateArgs,
+        },
+        *,
+    },
+    resource::ResourceType,
+    state::{
+        device::*,
+        interface::{InterfaceCYOA, InterfaceDIA, LoopbackType, RoutingMode, INTERFACE_MTU},
+    },
+};
+use solana_program_test::*;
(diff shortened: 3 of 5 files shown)


smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs
```diff
@@ -226,6 +226,14 @@ pub fn process_update_device(
 
     let mut device: Device = Device::try_from(device_account)?;
 
+    // The supplied contributor must be the one the device belongs to,
+    // unless the payer is on the foundation allowlist.
+    if !globalstate.foundation_allowlist.contains(payer_account.key)
```
nikw9944, line 231: Nit: `process_update_device` now evaluates `globalstate.foundation_allowlist.contains(payer_account.key)` in three places (here, line 238, and line 352). The link processors hoist this into a `payer_in_foundation` local (e.g. `link/delete.rs:127`); doing the same here would read better and save a few compute units. Purely optional.

smartcontract/programs/doublezero-serviceability/src/processors/device/interface/create.rs
```diff
@@ -182,6 +182,14 @@ pub fn process_create_device_interface(
 
     let mut device: Device = Device::try_from(device_account)?;
 
+    // The supplied contributor must be the one the device belongs to,
+    // unless the payer is on the foundation allowlist.
+    if !globalstate.foundation_allowlist.contains(payer_account.key)
+        && device.contributor_pk != *contributor_account.key
+    {
+        return Err(DoubleZeroError::InvalidContributorPubkey.into());
```
nikw9944, line 190: Note (no action needed): the same class of binding mismatch yields `InvalidContributor` in `link/create.rs` and `NotAllowed` in `link/suspend.rs`/`resume.rs`/`delete.rs`/`accept.rs`/`update.rs`. `InvalidContributorPubkey` matches the closest analogue (`device/delete.rs`), which is the right call here — just flagging the pre-existing codebase-wide inconsistency in case anyone wants to unify it later.

smartcontract/programs/doublezero-serviceability/tests/device_contributor_binding_test.rs
```diff
@@ -0,0 +1,420 @@
...
+        ),
+        AccountMeta::new(
+            get_resource_extension_pda(program_id, ResourceType::SegmentRoutingIds).0,
+            false,
+        ),
+    ]
+}
+
+fn assert_invalid_contributor(result: Result<(), BanksClientError>) {
+    // DoubleZeroError::InvalidContributorPubkey = Custom(10)
```
nikw9944, line 256: Optional: hardcoding `Custom(10)` matches the existing convention in this suite (e.g. `user_bgp_status_test.rs:308`), so this is fine as-is — but deriving the code from the enum (`DoubleZeroError::InvalidContributorPubkey as u32` style) would survive future renumbering.

nikw9944 (commented):
The fix is correct and well tested: the binding check is placed uniformly after the existing ownership check in all four processors, the foundation bypass matches the link suspend/resume/delete pattern, and the SDK already passes `device.contributor_pk` for all four instructions so no legitimate client flow regresses. Verified locally: the new test file passes (5/5) and the full `doublezero-serviceability` suite passes (522 tests).

One finding that can't be attached inline because the file isn't part of this diff:

**`ResumeDevice` has the identical gap and is left out of the sweep** (`src/processors/device/resume.rs:75-90`). The processor takes a contributor account, checks only `contributor.owner == payer` (or foundation), and never compares `device.contributor_pk` to the supplied contributor — so a non-foundation payer who owns *any* contributor account can flip *any* suspended device to `Activated`. That's exactly the bug class this PR closes, on an instruction that mutates device state in a way that matters operationally. `ResumeDevice` is live (variant 25, dispatched in `entrypoint.rs`); the lack of a current SDK builder for it doesn't protect anything since anyone can craft the transaction. A sweep for processors that read a `Device` and take a `contributor_account` without the binding check flags exactly this one remaining file. The fix is the same 8 lines plus one test — recommend including it in this PR, or explicitly deferring it with a follow-up.

elitegreg (approved):

PR #3902, opened 2026-06-15
doublezerod: gate multicast publisher heartbeat on BGP session up

ben-dz:
## Summary

- Gate the multicast **publisher heartbeat** (the packet stream that registers the source at the DZD and builds the `(S,G)`) on the BGP session reaching `Up`, instead of starting it inline in `MulticastService.Setup`. This closes the reprovision race that leaves a **wedged `(S,G)`** on the publisher's DZD — present but missing the MSDP notify (`N`) flag and with no OIF — after a host-side network reload.
- The gate rides the **existing** BGP session lifecycle (`BGPSessionTimeout = 30s`) and the 10s reconcile loop — no new timeout or tunable is introduced.
- `Teardown` cancels and awaits the readiness watcher before closing the heartbeat, so a pending start can't race with or follow the close.
- Incremental group updates (add/remove groups **without** a disconnect/reconnect) now respect readiness: `UpdateGroups` records the new group set and updates the running sender in place, or defers to the watcher when the heartbeat hasn't started yet — avoiding a nil-conn panic in the pre-`Up` window.

## Why

The underlying wedge is an EOS/DZD MSDP bug, and DZD/EOS upgrades depend on independent contributors. This is the client-side mitigation we can ship unilaterally: by registering the source only after the tunnel/BGP plumbing is up, we stop creating the overlap that produces the wedge in the first place, rather than relying on operators to notice and reconnect.

## Testing Verification

- `TestMulticastService_HeartbeatGatedOnBGPUp` — heartbeat stays silent while the session is `Pending`, then starts once it reaches `Up`.
- `TestMulticastService_HeartbeatGroupChangeBeforeBGPUp` — a group change before BGP `Up` does not touch the not-yet-started sender (0 `UpdateGroups` calls) and the eventual start uses the updated group set (regression guard for the no-disconnect group-update path).
- Reworked `TestMulticastService_UpdateGroups_AddPubGroup` to bring BGP up first, exercising the running-sender in-place update path.
- `go test -race` green across `internal/services` and `internal/manager`; `golangci-lint` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
+- Client
+  - Gate the multicast publisher heartbeat on the BGP session reaching Up
 - Onchain programs
   - Validate the device `mgmt_vrf` field against the account-code charset (`[A-Za-z0-9:_-]`) and a 32-byte length cap, matching the device `code` field. Empty (the default VRF) is still accepted.
 - Controller
--- a/client/doublezerod/internal/services/multicast.go
+++ b/client/doublezerod/internal/services/multicast.go
@@ -1,10 +1,12 @@
 package services
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"log/slog"
 	"net"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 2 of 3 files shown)


ben-dz:
Closing, since this didn't solve our problem.

PR #3909, opened 2026-06-16
serviceability: expose public fn to calculate access pass airdrop target

elitegreg:
## Summary

- Extract a public, pure `airdrop_user_target_lamports(rent, user_airdrop_lamports, multiplier)` from `airdrop_user_credits` in the serviceability program, so the feed oracle can compute the expected access pass airdrop target without duplicating the program's math.
- The function takes `&Rent` rather than calling `Rent::get()` internally, so it works both on-chain (caller passes `Rent::get()?`) and off-chain where `Rent` is fetched over RPC (`getMinimumBalanceForRentExemption` or the rent sysvar account).
- `airdrop_user_credits` now delegates the target calculation to the new function, then subtracts the current balance and transfers the difference. Its signature and behavior are unchanged, so the three existing call sites are untouched.

## Testing Verification

- Added a unit test asserting the target equals `minimum_balance(AIRDROP_USER_RENT_LAMPORTS_BYTES) + user_airdrop_lamports` for `multiplier = 1`, and that result scaled by `N` for `multiplier = N`.
- Existing accesspass and multicast allowlist tests still pass (`cargo test -p doublezero-serviceability`), confirming the refactor preserves the airdrop transfer behavior.

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/mod.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/mod.rs
@@ -11,6 +11,23 @@ use solana_program::{
 // `User` account size assumes a single publisher and subscriber pubkey registered (266 bytes each).
 pub const AIRDROP_USER_RENT_LAMPORTS_BYTES: usize = 266 * 3; // 266 bytes per User account x 3 accounts = 798 bytes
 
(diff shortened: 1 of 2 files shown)


juan-malbeclabs (approved):
LGTM

PR #3911, opened 2026-06-17
client: add --no-wait flag to doublezero disconnect

elitegreg:
## Summary of Changes
- Add a `--no-wait` flag to `doublezero disconnect` that skips the daemon tunnel-teardown poll, returning as soon as the onchain user deletion is confirmed.
- Useful for scripting and automation that does not need to block (up to ~60s) on local tunnel removal by `doublezerod`.
- Onchain delete confirmation (`poll_for_user_closed`) still runs under `--no-wait`; only the daemon deprovision wait is skipped.

## Diff Breakdown
| Category    | Files | Lines (+/-) | Net |
|-------------|-------|-------------|-----|
| Core logic  |     1 | +26 / -17   |  +9 |

Single-file CLI change: a new flag plus a conditional around the existing daemon poll, with the test fixture updated for the new field.

<details>

- `client/doublezero/src/command/disconnect.rs` — add `no_wait` field to `DecommissioningCliCommand`; gate `poll_for_daemon_deprovisioned` behind `!no_wait`, printing a notice when skipped.

</details>

## Testing Verification
- `cargo test -p doublezero disconnect` — all 15 existing disconnect tests pass.
- Verified `doublezero disconnect --help` lists the new `--no-wait` flag.

### DIFF
--- a/client/doublezero/src/command/disconnect.rs
+++ b/client/doublezero/src/command/disconnect.rs
@@ -38,6 +38,10 @@ pub struct DecommissioningCliCommand {
     /// Allocate a new address for the user
     #[arg(short, long, default_value_t = false)]
     pub verbose: bool,
+    /// Skip waiting for the daemon to tear down the tunnel(s); exit once the
+    /// onchain user deletion is confirmed.
+    #[arg(long, default_value_t = false)]
+    pub no_wait: bool,
     #[arg(value_enum)]
     pub dz_mode: Option<DzMode>,
 }
@@ -60,23 +64,27 @@ impl DecommissioningCliCommand {
         let (_, gstate) = client.get_globalstate(GetGlobalStateCommand)?;
         self.delete_users(client, client_ip, gstate.feed_authority_pk, &spinner)?;
 
-        // Wait for daemon to deprovision the tunnel(s)
-        let user_type_filter: Option<&str> = match self.dz_mode {
-            Some(DzMode::IBRL) => Some("IBRL"),
-            Some(DzMode::Multicast) => Some("Multicast"),
-            None => None,
-        };
-        match self
-            .poll_for_daemon_deprovisioned(&controller, user_type_filter, &spinner)
-            .await
-        {
-            Ok(()) => {
-                spinner.println("    Tunnel confirmed removed");
-            }
-            Err(e) => {
-                spinner.println(format!(
-                    "    Daemon deprovisioning in progress (will complete automatically): {e}"
-                ));
+        if self.no_wait {
+            spinner.println("    Skipping tunnel teardown wait (--no-wait)");
+        } else {
+            // Wait for daemon to deprovision the tunnel(s)
+            let user_type_filter: Option<&str> = match self.dz_mode {
+                Some(DzMode::IBRL) => Some("IBRL"),
+                Some(DzMode::Multicast) => Some("Multicast"),
+                None => None,
+            };
+            match self
+                .poll_for_daemon_deprovisioned(&controller, user_type_filter, &spinner)
+                .await
+            {
+                Ok(()) => {
+                    spinner.println("    Tunnel confirmed removed");
+                }
+                Err(e) => {
+                    spinner.println(format!(
+                        "    Daemon deprovisioning in progress (will complete automatically): {e}"
+                    ));
+                }
             }
         }
 
@@ -260,6 +268,7 @@ mod tests {
             device: None,
             client_ip: None,
             verbose: false,
+            no_wait: false,
             dz_mode: None,
         }
     }
--- a/client/doublezero/src/command/disconnect.rs
+++ b/client/doublezero/src/command/disconnect.rs
@@ -38,8 +38,10 @@ pub struct DecommissioningCliCommand {
     /// Allocate a new address for the user
     #[arg(short, long, default_value_t = false)]
     pub verbose: bool,
-    /// Skip waiting for the daemon to tear down the tunnel(s); exit once the
-    /// onchain user deletion is confirmed.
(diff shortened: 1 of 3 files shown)


client/doublezero/src/command/disconnect.rs
```diff
@@ -38,6 +38,10 @@ pub struct DecommissioningCliCommand {
     /// Allocate a new address for the user
     #[arg(short, long, default_value_t = false)]
     pub verbose: bool,
+    /// Skip waiting for the daemon to tear down the tunnel(s); exit once the
+    /// onchain user deletion is confirmed.
+    #[arg(long, default_value_t = false)]
```
vihu, line 45: What are the concerns in making this `true` by default?

elitegreg: Would need to test the disconnect and immediately reconnect case to make sure it works right.

elitegreg: Also, it may give a false sense that doublezero is disconnected, but you are actually still routing until the tunnel comes down.

juan-malbeclabs, line 45: Note that `--no-wait` only skips the daemon teardown poll. The onchain path (`delete_users` → `poll_for_user_closed`) still runs with exponential backoff up to ~127s per deleted user, so this can still block when the RPC is slow to reflect the deletion. Worth calling out in the help text so scripting/automation callers don't expect a guaranteed fast exit.

client/doublezero/src/command/disconnect.rs
```diff
@@ -60,23 +64,27 @@ impl DecommissioningCliCommand {
...
-        {
-            Ok(()) => {
-                spinner.println("    Tunnel confirmed removed");
-            }
-            Err(e) => {
-                spinner.println(format!(
-                    "    Daemon deprovisioning in progress (will complete automatically): {e}"
-                ));
+        if self.no_wait {
+            spinner.println("    Skipping tunnel teardown wait (--no-wait)");
```
juan-malbeclabs: This message, combined with the unconditional `✅  Deprovisioning Complete` printed on line 91, gives a false sense that DoubleZero is fully disconnected. With `--no-wait` the GRE tunnel is usually still up and BGP still advertising routes until the reconciler picks up the onchain deletion (~10s poll + teardown). Two asks:

1. Make this message honest about the async teardown (suggestion below).
2. Don't print `Deprovisioning Complete` under `--no-wait` — gate line 91 on `!self.no_wait`, or print a distinct message there.

```suggestion
            spinner.println(
                "    Onchain user deletion confirmed. The daemon will tear down the \
                 tunnel(s) shortly — traffic may still route over DoubleZero until then.",
            );
```

juan-malbeclabs (commented):
A few inline notes. The change is correct and low-risk — no correctness bug, and the disconnect→reconnect case is safe (the reconciler is driven by onchain state and reprovisions atomically, and the onchain delete is still awaited before returning). The asks below are about messaging/expectations under `--no-wait`.

client/doublezero/src/command/disconnect.rs
juan-malbeclabs: We think it should stay opt-in (default false). The wait isn't purely cosmetic: on the default path, once disconnect returns successfully the daemon has confirmed the tunnel is torn down, so the caller knows traffic is no longer routing over DoubleZero. If we flipped the default, disconnect would return while the GRE tunnel is usually still up and BGP still advertising routes (the reconciler only picks up the onchain deletion on its next ~10s poll), which is exactly the false-sense-of-disconnection issue Greg raised. That trade-off is fine for scripting/automation that explicitly opts in via --no-wait, but a surprising default for interactive use — so we kept the safe/confirming behavior as the default.

elitegreg: Keeping it opt-in (default false). On the default path, a successful return means the daemon has confirmed the tunnel is torn down, so the caller knows traffic is no longer routing over DoubleZero. Flipping the default would return while the GRE tunnel is usually still up and BGP still advertising (the reconciler only picks up the onchain deletion on its next ~10s poll) — the false-sense-of-disconnection issue Greg raised. Acceptable for automation that explicitly opts in, but a surprising default for interactive use.

elitegreg: Good call — updated the help text to note the onchain deletion is still awaited (up to ~127s/user on slow RPC) and that only the local tunnel-teardown wait is skipped, so it is not a guaranteed fast exit. (419f925)

client/doublezero/src/command/disconnect.rs
elitegreg: Done in 419f925. The skip message now reads: "Onchain user deletion confirmed. The daemon will tear down the tunnel(s) shortly; traffic may still route over DoubleZero until then." and the unconditional `✅  Deprovisioning Complete` is gated to the wait path — under `--no-wait` it prints `✅  Onchain deletion complete (tunnel teardown pending)` instead.

client/doublezero/src/command/disconnect.rs
elitegreg: Reasoning-wise this should be safe: the reconciler is driven by onchain state and reprovisions atomically, so a stale tunnel left up by a --no-wait disconnect and a freshly created user converge on the next poll. I have not run the disconnect→immediate-reconnect case on a live devnet yet, though — I'll verify it there before merge and report back.

elitegreg: Addressed in 419f925 — under `--no-wait` the output now says the teardown is pending and traffic may still route over DoubleZero until the daemon tears the tunnel down, rather than claiming "Deprovisioning Complete".

juan-malbeclabs (approved):

PR #3919, opened 2026-06-20
e2e: route all devnet networks through collision-safe subnet allocation

### LINKED ISSUE #3917: e2e: flaky Docker network "Pool overlaps" — MiscNetwork/DefaultNetwork bypasses the subnet allocator
## Summary

`TestE2E_BackwardCompatibility` (and any test that brings up a devnet `DefaultNetwork`) intermittently fails at devnet network creation with a Docker address-pool collision:

```
compatibility_test.go:739:
  Received unexpected error:
  failed to create network: create network: Error response from daemon:
  invalid pool request: Pool overlaps with other one on this address space
```

This is a **test-infra flake**, not a product bug. It's distinct from the multicast flake tracked in #3907 — different test, different root cause.

## Root cause

The devnet has two network types with **inconsistent** subnet handling:

- `CYOANetwork.Create` allocates a subnet via `n.dn.subnetAllocator.FindAvailableSubnet(...)` and passes it as explicit IPAM config — collision-safe. See [e2e/internal/devnet/network_cyoa.go:75-95](../blob/main/e2e/internal/devnet/network_cyoa.go#L75-L95).
- `MiscNetwork.Create` (used as the devnet's `DefaultNetwork`) calls `dockerClient.NetworkCreate` with **no IPAM config** — so Docker auto-assigns a subnet from its default address pool. See [e2e/internal/devnet/network_misc.go:66-91](../blob/main/e2e/internal/devnet/network_misc.go#L66-L91).

When many e2e networks exist on the runner (concurrent shards + `/run-e2e` dispatched runs + leftover networks), Docker's auto-assignment picks a subnet that overlaps an existing network and `NetworkCreate` fails with "Pool overlaps". Because `MiscNetwork` bypasses the `subnetAllocator` that already exists for exactly this purpose, it's the only network type exposed to the collision.

The `CreateIfNotExists` race-handling in [network_misc.go:43-64](../blob/main/e2e/internal/devnet/network_misc.go#L43-L64) only handles the "already exists" race, not the pool-overlap failure, so the error propagates and fails the test.

## Suggested fix

Route `MiscNetwork.Create` through the same subnet allocator as `CYOANetwork`:
1. Call `n.dn.subnetAllocator.FindAvailableSubnet(ctx, n.dn.Spec.DeployID)` (as `CYOANetwork` does).
2. Pass it as explicit IPAM config in the `NetworkCreate` options (`IPAM: &dockernetwork.IPAM{ Config: []dockernetwork.IPAMConfig{{Subnet: subnetCIDR}} }`), keeping the existing MTU option and labels.

This makes all devnet networks collision-safe via one mechanism. Confirm the `subnetAllocator` accounts for already-in-use subnets on the host (inspect existing Docker networks) so concurrent runs don't hand out the same CIDR; if it only tracks subnets it allocated this process, it may also need to consult `docker network ls`/inspect.

Secondary hardening (optional): in `MiscNetwork.CreateIfNotExists`, retry on the "Pool overlaps"/"invalid pool request" daemon error with a freshly allocated subnet (a few attempts) so a transient collision self-heals.

## Reproduction / evidence

- Surfaced on PR #3900 (openssl bump) `/run-e2e` dispatch — unrelated to that change.
- Dispatched e2e.yml run **27856466765**, job `e2e (shard 1)`: `TestE2E_BackwardCompatibility/testnet/v0.20.0` failed at `compatibility_test.go:739` (the `require.NoError` on `dn.DefaultNetwork.CreateIfNotExists` at line 738). A plain re-run of the failed shard passed — confirming a timing/concurrency flake, not a code regression.
- Blast radius is broader than the compat test: any devnet whose `DefaultNetwork` (a `MiscNetwork`) is created concurrently can hit this.

## Acceptance criteria

- `MiscNetwork`/`DefaultNetwork` creation allocates its subnet via the same collision-safe path as `CYOANetwork` (no reliance on Docker auto-assignment).
- `TestE2E_BackwardCompatibility` and other devnet-bring-up tests pass reliably under concurrent shard load / repeated runs.
- No regression to MTU (9000) or testcontainers/Ryuk cleanup labels on the network.

ben-dz:
## Summary of Changes
* Fix the intermittent `Pool overlaps with other one on this address space` Docker failure that flaked `TestE2E_BackwardCompatibility` and any devnet bring-up under concurrent shard load.
* The default and inter-device link (misc) networks bypassed the collision-safe `docker.SubnetAllocator` that `CYOANetwork` already used: `MiscNetwork` relied on Docker auto-assignment (no IPAM config), and `DefaultNetwork` used a fixed deterministic `10.X.Y.0/24` hash with no overlap check and no retry. Either could pick a CIDR already taken by a leftover or concurrently-created network and fail hard.
* Route all three network types through a shared `createNetworkWithSubnet` helper that allocates a docker-overlap-checked subnet and retries on a pool-overlap daemon error. The allocator is re-seeded per attempt (`id` -> `id#N`) so even a deterministic finder converges to a *different* subnet when the conflict is invisible to the `docker network ls` scan (e.g. Docker's default address pool, or a network created without an explicit IPAM subnet).
* Each network type allocates from a disjoint range so a subnet can never overlap another network's range or an onchain `dz_prefix`: CYOA stays on `9.128.0.0/9`, default on `10.0.0.0/8`, and link networks move off Docker's default pool onto a dedicated `11.0.0.0/8` allocator. This avoids the case where a link subnet could overlap a device's onchain-advertised `9.x` dz_prefix (including `uteDzPrefix`, which can fall outside the CYOA `/23`) that the docker scan cannot see.
* MTU (9000) and the testcontainers/Ryuk cleanup labels on the misc network are preserved.
* Fixes #3917

## Testing Verification
* Added `e2e/internal/devnet/network_test.go` covering `createNetworkWithSubnet` and `isPoolOverlapErr`: success path, retry-on-pool-overlap with assertions that each retry re-allocates a fresh subnet **and** is seeded with a distinct ID (`deploy-1`, `deploy-1#1`, `deploy-1#2`), graceful give-up after max retries, non-overlap errors not retried, and allocation-error propagation.
* `go test ./e2e/internal/devnet/...` passes; `go vet` and `golangci-lint run ./internal/devnet/...` report no issues.
* Verified the new `11.0.0.0/8` link range is disjoint from every other range the e2e suite uses (`9.128.0.0/9`, `10.0.0.0/8`, `172.16.0.0/16`/`192.168.x`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ All notable changes to this project will be documented in this file.
   - Skip rendering device config when a string field would not survive as a single config token (contains control or whitespace characters).
 - Device agents
   - Reduce agent CPU usage by continuing to fetch the full config every 5 seconds but only applying when it has changed or after 60s timeout
+- E2E tests
+  - Route all devnet networks (CYOA, default, and miscellaneous) through a shared collision-safe subnet allocator to prevent overlapping subnet assignments across concurrent test runs. (#3919)
 
 ## [v0.27.1](https://github.com/malbeclabs/doublezero/compare/client/v0.27.0...client/v0.27.1) - 2026-06-10
 
--- a/e2e/internal/devnet/devnet.go
+++ b/e2e/internal/devnet/devnet.go
@@ -50,13 +50,22 @@ const (
 	defaultNetworkSubnetMask = 24
 
 	// linkNetworkBaseCIDR is the address range the devnet's inter-device link (misc) networks are
-	// allocated from. It is deliberately disjoint from the CYOA range (9.128.0.0/9) — and the
-	// onchain dz_prefixes derived from it, which are not visible to the docker-network scan — as
-	// well as from the default network (10.0.0.0/8) and the device tunnel net (172.16.0.0/16), so a
-	// link subnet can never silently overlap a routed prefix.
+	// allocated from. It must stay disjoint from every other range in play so a link subnet can
+	// never silently overlap a routed prefix:
+	//   - the CYOA range (9.128.0.0/9), including the onchain dz_prefixes, which are derived from
+	//     CYOA public IPs in 9.x and so fall inside it; these prefixes are invisible to the
+	//     docker-network scan, so only range separation keeps a link subnet off them;
+	//   - the default network range (10.0.0.0/8);
+	//   - the device tunnel net (defaultDeviceTunnelNet, 172.16.0.0/16).
+	// TestNetworkRangesAreDisjoint enforces this invariant against the constants below.
 	linkNetworkBaseCIDR = "11.0.0.0/8"
 	// linkNetworkSubnetMask is the prefix length for each link network subnet (a /24).
 	linkNetworkSubnetMask = 24
+
+	// defaultDeviceTunnelNet is the onchain device-tunnel block used when DevnetSpec.DeviceTunnelNet
+	// is left unset. It is one of the ranges the CYOA/default/link allocators must avoid; see
+	// TestNetworkRangesAreDisjoint.
+	defaultDeviceTunnelNet = "172.16.0.0/16"
 )
 
 var (
@@ -190,7 +199,7 @@ func (s *DevnetSpec) Validate() error {
 	}
 
 	if s.DeviceTunnelNet == "" {
-		s.DeviceTunnelNet = "172.16.0.0/16"
+		s.DeviceTunnelNet = defaultDeviceTunnelNet
 	}
 	return nil
 }
--- a/e2e/internal/devnet/devnet.go
+++ b/e2e/internal/devnet/devnet.go
@@ -41,6 +41,22 @@ const (
 
 	containerDoublezeroKeypairPath = "/root/.config/doublezero/id.json"
 	containerSolanaKeypairPath     = "/root/.config/solana/id.json"
+
+	// defaultNetworkBaseCIDR is the address range the devnet's default network is allocated from.
+	// It is kept separate from the CYOA network range (9.128.0.0/9) so tests that detect interfaces
+	// by IP range do not confuse the two.
+	defaultNetworkBaseCIDR = "10.0.0.0/8"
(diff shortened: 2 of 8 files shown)


ben-dz:
Addressed the review feedback in 609ab9f1:

- **M1 (range disjointness asserted only in prose):** Added `TestNetworkRangesAreDisjoint` asserting the CYOA, default, link, and default device-tunnel ranges are pairwise non-overlapping. CYOA is derived from `docker.NewDefaultSubnetAllocator(nil)` so a future change to the base CIDR is caught rather than slipping past. Exported `docker.CIDROverlap` (was `cidrOverlap`) so the test reuses the allocator's own overlap logic instead of duplicating it. Also extracted the `172.16.0.0/16` literal into a `defaultDeviceTunnelNet` constant.
- **L2/N1 (comment over-claims / dense wording):** Reworded the re-seed comment to be accurate that the re-seed is probabilistic for docker-invisible overlaps and a guaranteed skip for visible ones, and clarified the `linkNetworkBaseCIDR` disjointness rationale.
- **L3 (retry search space):** Documented on `poolOverlapMaxRetries` — the inner salted skips only widen the search for docker-*visible* conflicts; for the invisible case each re-seed yields one fresh candidate (~`poolOverlapMaxRetries`).

One note from review worth flagging as a possible follow-up: the disjointness test covers the compile-time base ranges and the *default* device-tunnel net. An explicit `DevnetSpec.DeviceTunnelNet` override is not guarded at runtime; the test docstring now states keeping such an override disjoint is the caller's responsibility. A `Validate()` check could close that gap if desired.

packethog (approved):
Reviewed the change against the surrounding code with an adversarial correctness pass: the shared allocator + pool-overlap retry, the per-attempt re-seed, the disjoint per-network-type ranges, and the `CIDROverlap` containment check are all sound, and the new range-disjointness test locks in the load-bearing invariant. `go vet` is clean and the new unit tests pass. All CI checks are green, including every e2e shard. LGTM.

PR #3931, opened 2026-06-22
controller: stop exporting metrics for pubkeys removed from the ledger

nikw9944:
## Summary of Changes
- Prune a device's per-pubkey Prometheus series (`controller_grpc_getconfig_requests_total`, render errors, duplicate tunnel pairs, and `controller_link_metrics`) when the device disappears from the on-chain ledger. `swapCache` diffs the old and new caches and calls `deleteDeviceMetrics` for each removed pubkey, so the frozen counters go stale (~5 min) instead of being scraped forever. This stops the `Network: Device Stopped Calling Controller` alert from firing indefinitely on a decommissioned device.
- Replace the per-pubkey `controller_grpc_getconfig_pubkey_errors_total` counter with a single aggregate `controller_grpc_getconfig_unknown_pubkey_total` (no `pubkey` label). Since the pubkey is caller-supplied, the old metric let an unknown caller blow up cardinality and could resurrect a just-pruned series. The unknown-pubkey path now increments the aggregate counter and returns `NotFound` without touching any per-pubkey series.
- Log unknown-pubkey check-ins at `WARN` with the device pubkey and caller address, rate-limited (via an atomic timestamp + CAS) to at most once per minute so a still-polling decommissioned box can't flood the logs.
- Register `linkMetrics` and `linkMetricInvalid` with Prometheus. They were populated in `server.go` but never registered, so they were neither scraped nor prunable; registering them makes the link-metrics prune meaningful.
- Document the metrics, pruning behavior, and unknown-pubkey handling in the controller README, and add a CHANGELOG entry.

## Testing Verification
- Added `metrics_test.go` covering: pruning of a removed device's series across all per-device vectors while a surviving device's series stay intact (`TestSwapCache_PrunesRemovedDeviceMetrics`); the aggregate counter incrementing once per call, no per-pubkey series being created, and exactly one rate-limited WARN log carrying the caller address (`TestGetConfig_LedgerAbsentPubkey`); and a pruned pubkey not being resurrected by repeated check-ins (`TestGetConfig_PrunedPubkeyNotResurrected`).
- `go build ./...`, `go vet ./internal/controller/`, `gofmt`, and `go test ./internal/controller/` all pass locally.
- CI: `go-build`, `release-validation-controller`, rust, and sdk checks green; e2e integration shards still running at time of writing with no failures.


controlplane/controller/internal/controller/metrics.go
```diff
@@ -112,3 +114,17 @@ func init() {
...
+// on-chain ledger so Prometheus can no longer scrape its now-frozen counters;
+// after a scrape interval plus the staleness window the series go stale and
+// queries return empty. DeletePartialMatch removes all series matching just the
+// pubkey label, regardless of the other (agent_version, etc.) label values.
+func deleteDeviceMetrics(pubkey string) {
+	byPubkey := prometheus.Labels{"pubkey": pubkey}
+	getConfigOps.DeletePartialMatch(byPubkey)
+	getConfigRenderErrors.DeletePartialMatch(byPubkey)
+	duplicateTunnelPairs.DeletePartialMatch(byPubkey)
+	linkMetrics.DeletePartialMatch(prometheus.Labels{"device_pubkey": pubkey})
```
packethog, line 143: `linkMetrics` (and `linkMetricInvalid`) are never registered with a Prometheus registry — they're absent from the `MustRegister` block in `init()` and not registered anywhere else. So this `DeletePartialMatch` prunes a collector that is never scraped: for `linkMetrics` the prune is effectively a no-op, and the new test asserting it is pruned exercises a series that isn't actually exposed on `/metrics`. Worth either registering `linkMetrics` here (if it's meant to be exposed) or dropping it from the prune set so it doesn't imply the series is managed/scraped.

controlplane/controller/internal/controller/metrics_test.go
```diff
@@ -0,0 +1,220 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/peer"
+	"google.golang.org/grpc/status"
+)
+
+// countSeriesWithLabel collects a metric vector directly (bypassing the global
+// registry to avoid interference from other tests) and counts the series whose
+// label `name` equals `value`. Tests use unique pubkeys so the count is exact.
+func countSeriesWithLabel(c prometheus.Collector, name, value string) int {
+	ch := make(chan prometheus.Metric, 1024)
```
packethog: nit: `countSeriesWithLabel` collects synchronously into a channel with a fixed 1024 buffer before draining it. These metric vectors are package-global and accumulate series across the package's tests, so a collector emitting more than 1024 metrics would block `Collect` indefinitely (nothing drains until after `close`). Consider draining in a goroutine concurrently with `Collect`, or sizing the buffer defensively.

packethog (changes requested):
Requesting changes — see the inline comments. The core change (pruning per-device series on ledger removal, rejecting ledger-absent pubkeys without resurrecting series) is sound, but two items are worth addressing before merge: `linkMetrics` is never registered with a Prometheus registry, so pruning it (and the test asserting that) operates on a series that isn't exposed; and the test collection helper can deadlock if a global vector ever exceeds its fixed channel buffer.

controlplane/controller/internal/controller/metrics.go
nikw9944: Good catch. `linkMetrics`/`linkMetricInvalid` were populated in `server.go` but never registered, so this prune was a no-op on an unexposed series. Rather than drop it from the prune set, I registered both in `init()` (commit 8a8e263f) — they carry real per-device data and should be scraped, so this fixes the latent gap and makes the pruning meaningful.

controlplane/controller/internal/controller/metrics_test.go
nikw9944: Fixed in 8a8e263f — `Collect` now runs in a goroutine while the main routine drains the channel, so the fixed buffer can no longer deadlock. Dropped the explicit 1024 buffer (unbuffered channel) since draining is concurrent.

controlplane/controller/internal/controller/metrics.go
```diff
@@ -109,6 +111,24 @@ func init() {
 	prometheus.MustRegister(cacheUpdateFetchErrors)
 	prometheus.MustRegister(cacheUpdateOps)
 
+	// link metrics
+	prometheus.MustRegister(linkMetrics)
```
packethog, line 115: `linkMetrics` is now registered and scraped, but its per-interface gauge series are only ever `Set` (server.go:466), never deleted or zeroed when an interface stops being an active link. In `updateStateCache` the non-link / non-active-status branch (server.go ~L424-428) and the invalid-delay branch (server.go ~L431-433) `continue` without clearing the existing `controller_link_metrics{device_code,interface,device_pubkey}` series. As a result, a surviving device that loses a link (link removed, drained to an unlisted status, or delay becomes invalid) keeps exporting its last delay value indefinitely — until the whole device is pruned. Now that this gauge is scrapeable, consider clearing the series for interfaces that are no longer active links (e.g. `linkMetrics.DeleteLabelValues(device.Code, iface.Name, d.PubKey)` in those `continue` branches) so it doesn't go stale on surviving devices.

packethog (changes requested):
Requesting changes — one inline comment on a metric-staleness gap introduced by registering `linkMetrics`. The device-removal prune itself is sound and well covered by tests.

controlplane/controller/internal/controller/server.go
```diff
@@ -419,11 +425,16 @@ func (c *Controller) updateStateCache(ctx context.Context) error {
 				d.Interfaces[i].IsLink = false
 				d.Interfaces[i].Metric = 0
 				d.Interfaces[i].LinkStatus = linkStatusUnknown
+				// Drop any stale gauge for an interface that is no longer an active
+				// link so a surviving device stops exporting its last delay value.
+				linkMetrics.DeleteLabelValues(device.Code, iface.Name, d.PubKey)
```
packethog: Clearing the gauge here covers interfaces that are still present on the rebuilt device but are no longer active links. Two paths still leak a stale `controller_link_metrics` series on a surviving device, because deletion only targets the current `(device.Code, iface.Name, d.PubKey)` tuple:

1. A device that develops a pathology `continue`s at server.go:400 — before this interface loop — so a device that previously exported link gauges and then gains a `DevicePathology` keeps its old gauge until the device is removed entirely.
2. If an interface is removed/renamed on-chain (no longer in `d.Interfaces`) or the device's `Code` changes, the tuple previously set at server.go:471 is never iterated, so it is never deleted.

Consider clearing a surviving device's link gauges by `device_pubkey` once near the top of the per-device iteration (e.g. `linkMetrics.DeletePartialMatch(prometheus.Labels{"device_pubkey": d.PubKey})`) before re-`Set`ting the active interfaces — placed before the pathology `continue` it covers all three paths uniformly.

controlplane/controller/internal/controller/metrics.go
```diff
@@ -109,6 +111,24 @@ func init() {
...
+// on-chain ledger so Prometheus can no longer scrape its now-frozen counters;
+// after a scrape interval plus the staleness window the series go stale and
+// queries return empty. DeletePartialMatch removes all series matching just the
+// pubkey label, regardless of the other (agent_version, etc.) label values.
+func deleteDeviceMetrics(pubkey string) {
+	byPubkey := prometheus.Labels{"pubkey": pubkey}
+	getConfigOps.DeletePartialMatch(byPubkey)
+	getConfigRenderErrors.DeletePartialMatch(byPubkey)
+	duplicateTunnelPairs.DeletePartialMatch(byPubkey)
+	linkMetrics.DeletePartialMatch(prometheus.Labels{"device_pubkey": pubkey})
```
packethog, line 143: `linkMetricInvalid` is now registered as well, but `deleteDeviceMetrics` doesn't prune it, so a removed device's `controller_link_metrics_invalid_total` series (labeled `link_pubkey`, `device_code`, `interface`) keep being scraped. It's keyed by `link_pubkey` rather than `device_pubkey`, so it can't be pruned here by device pubkey directly — either prune the device's link series explicitly on removal or document it as an accepted, bounded leak. (nit)

packethog (changes requested):
Thanks — the previous link-gauge staleness is now addressed for the in-loop deactivation/invalid-delay paths, and the device-removal prune, aggregate counter, and rate-limit CAS all look sound. A couple of remaining gauge-staleness gaps for surviving devices are noted inline.

controlplane/controller/internal/controller/metrics.go
nikw9944: Fixed in ca25d96e. `updateStateCache` now calls `linkMetrics.DeleteLabelValues(device.Code, iface.Name, d.PubKey)` in both `continue` branches — the non-active-link path (link removed or status not Activated/SoftDrained/HardDrained) and the invalid-delay path (`DelayNs <= 0`). A surviving device that loses a link now drops its `controller_link_metrics` series instead of exporting the last delay value indefinitely; the gauge is only kept while a valid active link is present. Added `TestStateCache_ClearsLinkMetricOnLinkLoss` to cover set-then-clear across two cache updates.

controlplane/controller/internal/controller/server.go
nikw9944: Fixed in 2cd5a553. Replaced the per-branch `DeleteLabelValues` with a single `clearDeviceLinkMetrics(devicePubKey)` (a `DeletePartialMatch` by `device_pubkey`) called at the top of each per-device iteration in `updateStateCache`, before the pathology `continue` and the interface loop. Active links are re-`Set` afterward. This covers all three paths uniformly: inactive/removed links, interfaces removed or renamed on-chain, a device `Code` change, and a device that gains a pathology and never reaches the interface loop. Added `TestStateCache_ClearsLinkMetricOnDevicePathology` for the pathology path (gauge set on a healthy update, then cleared once the device gains a pathology and skips the interface loop); `TestStateCache_ClearsLinkMetricOnLinkLoss` still covers the inactive-link path.

controlplane/controller/internal/controller/metrics.go
nikw9944: Fixed in 2cd5a553. `deleteDeviceMetrics` now also takes the device `code` and prunes `linkMetricInvalid` via `DeletePartialMatch(prometheus.Labels{"device_code": code})` on device removal (device codes are unique per device, so this is exact). `swapCache` passes the removed device's `Code` from the old cache. Extended `TestSwapCache_PrunesRemovedDeviceMetrics` (and the `seedDeviceMetrics` helper) to assert the removed device's `controller_link_metrics_invalid_total` series is pruned while a surviving device's is untouched.

controlplane/controller/internal/controller/server.go
```diff
@@ -330,6 +336,14 @@ func (c *Controller) updateStateCache(ctx context.Context) error {
 		devicePubKey := base58.Encode(device.PubKey[:])
 		d := NewDevice(ip, devicePubKey, c.maxUserTunnelSlots)
 
+		// Clear this device's link gauges before rebuilding them below. Doing it
+		// here (before the pathology continue and the interface loop) drops any
+		// series that should no longer be exported on a surviving device: a link
+		// that went inactive, an interface removed/renamed on-chain, a device code
+		// change, or a device that develops a pathology and never reaches the
+		// interface loop. Active links are re-Set further down.
+		clearDeviceLinkMetrics(devicePubKey)
```
packethog: `clearDeviceLinkMetrics` deletes all of this device's `controller_link_metrics` series here, but the active links are only re-`Set` later in the interface loop (server.go:476). The cache-update loop runs on its own ~10s cadence, independent of Prometheus scrapes, so a scrape that lands between this blanket delete and the re-`Set` will momentarily observe the device's link gauges as absent — intermittent gaps in a gauge this PR is otherwise trying to make reliable. Deleting series that are immediately recreated on every cycle is the root cause. Consider removing only the series actually going stale (delete the specific `(device_code, interface, device_pubkey)` tuples no longer active), or rebuild the link gauges inside a custom collector at scrape time, rather than blanket-deleting and recreating every device every update.

controlplane/controller/internal/controller/metrics.go
```diff
@@ -109,6 +111,41 @@ func init() {
...
+//
+// linkMetricInvalid is keyed by link_pubkey rather than device_pubkey, so it is
+// pruned by device_code instead (device codes are unique per device).
+func deleteDeviceMetrics(pubkey, code string) {
+	byPubkey := prometheus.Labels{"pubkey": pubkey}
+	getConfigOps.DeletePartialMatch(byPubkey)
+	getConfigRenderErrors.DeletePartialMatch(byPubkey)
+	duplicateTunnelPairs.DeletePartialMatch(byPubkey)
+	linkMetrics.DeletePartialMatch(prometheus.Labels{"device_pubkey": pubkey})
+	linkMetricInvalid.DeletePartialMatch(prometheus.Labels{"device_code": code})
```
packethog, line 144: This prunes `linkMetricInvalid` by `device_code`, and the comment above states device codes are unique per device — but that isn't enforced on-chain: `CreateDevice` only format-validates/normalizes the code and keys the device PDA by a global index, so two devices can share a code. If that occurs, removing one device's metrics also wipes the other's `controller_link_metrics_invalid_total` series. Impact is limited (a low-importance counter that re-increments on the next update), but the uniqueness assumption is load-bearing here — either guarantee/validate it or soften the approach and the comment. (nit)

packethog (changes requested):
The pathology / interface-churn staleness from the last round is now handled cleanly, and the device-removal prune, aggregate counter, and rate-limit CAS remain sound. Two issues with the new per-update clear-then-reSet approach are noted inline — one metric-continuity (should-fix), one edge-case assumption (nit).

controlplane/controller/internal/controller/server.go
nikw9944: Good point — fixed in 8e012379 by switching to a targeted prune. `updateStateCache` now records every `controller_link_metrics` series it sets this cycle in an `activeLinkSeries` set, and after the new cache is built `pruneStaleLinkMetrics` deletes only the series that were active in the previous cache but are absent from the new active set. Still-active links are re-`Set` in place and never deleted, so a scrape can no longer land in a delete/recreate gap. The targeted prune still covers all the staleness paths (inactive link, removed/renamed interface, code change, device pathology, device removed) because none of those produce an entry in the active set. Added `TestStateCache_PrunesOnlyStaleLinkMetric` (two active links, one goes away) asserting the surviving series stays at 1 and only the stale one is removed. I went with targeted deletion rather than a scrape-time custom collector to keep the change localized to the existing cache-update path.

controlplane/controller/internal/controller/metrics.go
nikw9944: Fixed in 8e012379 by softening the approach and the comment. The doc on `deleteDeviceMetrics` no longer claims device codes are unique; it now notes they are not enforced unique on-chain (CreateDevice only normalizes the code and keys the PDA by a global index) and documents the collision as an accepted, self-healing edge case: `controller_link_metrics_invalid_total` is a counter that re-increments on the next cache update if the surviving device still has an invalid link, so a transient over-delete corrects itself within one cycle. I kept the device_code prune (rather than leaving the series to leak on removal) since the worst case is bounded and self-correcting.

controlplane/controller/internal/controller/server.go
```diff
@@ -419,11 +431,13 @@ func (c *Controller) updateStateCache(ctx context.Context) error {
 				d.Interfaces[i].IsLink = false
 				d.Interfaces[i].Metric = 0
 				d.Interfaces[i].LinkStatus = linkStatusUnknown
+				// Gauge already cleared at the top of the loop; nothing to re-Set.
```
packethog: This comment (and the one at server.go:440) says the gauge was "already cleared at the top of the loop," but this revision removed the top-of-loop clear — stale gauges are now pruned at the end of the update via `pruneStaleLinkMetrics`. The comments are leftovers from the previous approach and now misdescribe the control flow; worth updating to point at the end-of-update prune. (nit)

controlplane/controller/internal/controller/server.go
```diff
@@ -665,6 +680,13 @@ func (c *Controller) updateStateCache(ctx context.Context) error {
 		}
 	}
 
+	// Delete only the link gauges that were active in the previous cache but are
+	// no longer active, leaving still-active series in place. Done before the swap
+	// so c.cache still holds the previous device set. Safe without a lock: only
+	// this (single) cache-update goroutine writes c.cache, and GetConfig only reads
+	// it.
+	pruneStaleLinkMetrics(c.cache.Devices, activeLinkSeries)
```
packethog, line 701: Both this prune and the device-removal prune in `swapCache` are skipped when the ledger returns zero devices: `updateStateCache` returns early at server.go:283-285 (`0 devices found on-chain`) before building/swapping the cache. So if the *last* device is removed from the ledger, its frozen metrics are never pruned and the `Device Stopped Calling Controller` alert keeps firing — the very case this PR targets. The early return is also a useful guard against a transient empty fetch causing a mass false-prune, so there's a real tradeoff; at minimum consider documenting this zero-device limitation (or distinguishing a genuine empty ledger from a transient empty result). (nit / edge)

packethog (changes requested):
The link-gauge prune rework is sound: still-active links are now updated in place and only genuinely-stale series are deleted (no scrape gap), and the unlocked read in pruneStaleLinkMetrics is race-free under the single cache-update goroutine. Two minor items remain inline — a stale comment and an edge in the zero-device path. Both are nits; no correctness issue in the new logic.

controlplane/controller/internal/controller/server.go
nikw9944: Fixed in 78278442. Updated both comments (server.go:434 and :440) to describe the current flow: these branches deliberately skip recording the series in `activeLinkSeries`, so any existing gauge is removed by `pruneStaleLinkMetrics` at the end of the update. No more reference to a top-of-loop clear.

controlplane/controller/internal/controller/server.go
nikw9944: Documented in 78278442. Added a comment on the zero-device early return explaining the tradeoff: the guard exists to avoid mass false-pruning on a transient empty fetch, and the accepted cost is that if the *last* device is genuinely removed, the cache is never rebuilt/swapped so neither `pruneStaleLinkMetrics` nor the `swapCache` device-removal prune runs, leaving that device's frozen series scraped (and the alert firing) until a device reappears. We cannot reliably distinguish a genuinely empty ledger from a transient empty result here, so I left it as a documented limitation rather than risk false-pruning every device on a flaky fetch.

packethog (approved):

PR #3932, opened 2026-06-22
e2e/qa: make mainnet-beta QA client resilient to flaky/stale Solana RPC

### LINKED ISSUE #3930: e2e/qa: make mainnet-beta QA client resilient to flaky/stale Solana RPC
## Background

The mainnet-beta QA workflows (`qa.mainnet-beta.yml`, `qa.alldevices.mainnet-beta.yml` in [malbeclabs/infra](https://github.com/malbeclabs/infra)) point the QA client at a single Solana RPC endpoint via the `SOLANA_RPC_URL` secret. Over time this endpoint has flaked in **two distinct ways**, each requiring a manual repoint of the secret to recover:

1. **Non-responsive / timeout** — the RPC stops answering (seen this past weekend with the private Helius endpoint).
2. **Stale reads** — the RPC accepts transaction submission fine, but follow-up read calls return stale account state because the node is lagging behind the cluster. This is **not** a timeout: the call succeeds and returns old data, so error-based retries don't help.

infra PR [malbeclabs/infra#1715](https://github.com/malbeclabs/infra/pull/1715) attempted a CI-side fix: a job-start `getHealth` preflight that selects the private RPC if healthy and falls back to a public RPC otherwise. That only covers "RPC already dead at job start" — it does nothing for stale-but-successful reads or for an endpoint that degrades mid-run. Per [Slack discussion](https://doublezerofoundation.slack.com/archives/C0804H5R7JP/p1781981280122059), the resilience belongs in the QA client itself, not in a one-shot CI preflight. PR #1715 will be closed in favor of this work.

## Goal

Harden the QA client's Solana RPC usage so both failure modes are tolerated automatically, eliminating the manual `SOLANA_RPC_URL` secret flipping during outages.

## Relevant code

- QA client: [`e2e/internal/qa/`](https://github.com/malbeclabs/doublezero/tree/main/e2e/internal/qa) (Go, `gagliardetto/solana-go`).
- Settlement/balance reads: `client_settlement.go` — `GetTokenAccountBalance` at `CommitmentConfirmed` wrapped in a 5-retry / 30s exponential backoff. The retry only fires on errors, so a lagging node that returns stale-but-valid data is not caught.
- Onchain state reads: `getProgramDataWithRetry` in `test.go` and the `poll.Until` patterns in `client.go`.
- There is currently **no** multi-endpoint fallback in the client; it uses the single resolved `SOLANA_RPC_URL`.

## Proposed scope

1. **Multi-endpoint fallback in the client.** Accept a primary (`SOLANA_RPC_URL`) plus a public fallback, and fail over at runtime when an endpoint times out or is unhealthy — so an endpoint that dies mid-run is replaced rather than failing the run.
2. **Stale-read handling.** Stop trusting a single read after a write:
   - Wait for the submitted transaction to be confirmed (`GetSignatureStatuses`) before reading derived state.
   - Where we read post-write state (e.g. settlement USDC balance), poll until the expected state is observed rather than reading once — extend the existing `poll.Until` approach to these reads.
   - Optionally detect lag via slot height (`getSlot`) and fail over to another endpoint when a node is behind.
3. **Configurable timeout/retry budgets** so the values aren't hardcoded to the 30s window.

## Acceptance criteria

- A single-RPC timeout/outage does not fail the QA run when a healthy fallback endpoint exists.
- Stale reads from a lagging RPC do not produce spurious assertion failures (poll-until-consistent and/or fail over).
- No manual `SOLANA_RPC_URL` secret repointing is needed during an RPC outage.
- infra PR #1715 is closed once this lands.

## Out of scope

- The CI-side preflight in infra #1715 (this work supersedes it).
- testnet/devnet QA workflows (they use the DZ ledger RPC, not `SOLANA_RPC_URL`).

ben-dz:
## Summary of Changes
* Make the mainnet-beta QA client tolerate both Solana RPC failure modes automatically, eliminating the manual `SOLANA_RPC_URL` secret repointing during outages.
* **Multi-endpoint failover.** A sticky-index failover pool (`solanaRPCPool`) implements `rpc.JSONRPCClient`, so failover is transparent to both direct RPC reads and the shreds SDK reads. On a retryable failure (timeout, network error, HTTP 429/5xx) it advances to the next endpoint; it stays sticky once failed over. Endpoints are resolved from `SOLANA_RPC_URL` plus `SOLANA_RPC_FALLBACK_URLS`, with the public mainnet RPC appended as a default fallback when the var is unset (mainnet only; testnet/devnet keep the DZ-ledger single endpoint).
* **Active slot-lag detection.** Probes each endpoint's slot height (`getSlot`) concurrently and fails over to the freshest node when the current endpoint trails the max observed slot beyond a configurable threshold — turning a "stale but valid" lagging node into an explicit failover trigger.
* **Poll-until-consistent post-write reads.** The post-write client-seat read polls until the seat is visible (tolerating `rpc.ErrNotFound`/`ErrAccountNotFound`) rather than failing on a single stale read; the settlement test already polls the balance.
* **Configurable timeout/retry budgets** (`QA_SOLANA_RPC_*`) replacing the hardcoded 30s window; defaults preserve current behavior.
* Settlement **writes** (`FeedSeatPay`/`FeedSeatWithdraw`) target the pool's health-selected endpoint but deliberately do not retry across endpoints, to avoid double-submitting a settlement transaction.
* Fixes #3930

## Testing Verification
* New unit tests (`solanarpc_test.go`, no network): failover advances on retryable errors and is sticky; exhaustion returns the last error; non-retryable/business errors surface without failover; single-endpoint passthrough; endpoint resolution dedupe and public-default-when-unset; slot-lag failover / within-threshold / all-probes-fail; read-failover retry vs non-retryable; credential redaction and error scrubbing.
* Concurrent-failover test runs under `go test -race` and asserts the sticky index converges to exactly the healthy endpoint without overshoot.
* Verified against the live error path that a missing account surfaces as `rpc.ErrNotFound` (not the shreds sentinel), so the poll-until guard now matches both.
* `go test -race -tags qa ./e2e/internal/qa/...` green; `golangci-lint` clean; the `e2e` test binary compiles under `-tags qa`.
* Full mainnet-beta QA validation happens in the infra workflow after merge.

Note: the code diff is comment-dense (~385 lines of actual code; the remainder is documentation comments).

### DIFF
--- a/e2e/internal/qa/client_settlement.go
+++ b/e2e/internal/qa/client_settlement.go
@@ -98,7 +98,7 @@ func (c *Client) ClosestDevice(ctx context.Context) (*Device, error) {
 func (c *Client) FeedSeatPrice(ctx context.Context) ([]*pb.DevicePrice, error) {
 	c.log.Debug("Querying seat prices", "host", c.Host)
 	var prices []*pb.DevicePrice
-	err := c.withWriteFailover(func(rpcURL string) error {
+	err := c.withReadFailover(func(rpcURL string) error {
 		resp, err := c.grpcClient.FeedSeatPrice(ctx, &pb.FeedSeatPriceRequest{
 			SolanaRpcUrl:               rpcURL,
 			DzLedgerUrl:                c.DZLedgerURL,
@@ -119,13 +119,14 @@ func (c *Client) FeedSeatPrice(ctx context.Context) ([]*pb.DevicePrice, error) {
 	return prices, nil
 }
 
-// withWriteFailover runs an agent-driven settlement command against the pool's
-// current Solana RPC endpoint, failing over to the next endpoint and retrying
-// on error. It is only safe for idempotent operations (reads/queries) — see
-// the note on FeedSeatPay/FeedSeatWithdraw, which deliberately do not retry to
-// avoid double-submitting a settlement transaction. With a single endpoint (or
-// no pool) it runs fn exactly once.
-func (c *Client) withWriteFailover(fn func(rpcURL string) error) error {
+// withReadFailover runs an agent-driven settlement query against the pool's
+// current Solana RPC endpoint, failing over to the next endpoint on a retryable
+// error. It is ONLY safe for idempotent operations (reads/queries): the
+// settlement WRITES (FeedSeatPay/FeedSeatWithdraw) deliberately bypass this and
+// do not retry across endpoints, since a write that timed out on submission may
+// have landed onchain and a blind retry risks double-submission. With a single
+// endpoint (or no pool) it runs fn exactly once.
+func (c *Client) withReadFailover(fn func(rpcURL string) error) error {
 	attempts := 1
 	if c.solanaRPC != nil {
 		attempts = c.solanaRPC.EndpointCount()
@@ -136,10 +137,14 @@ func (c *Client) withWriteFailover(fn func(rpcURL string) error) error {
 		if lastErr == nil {
 			return nil
 		}
-		if c.solanaRPC != nil {
+		// Only fail over on retryable (connectivity/timeout) failures; a genuine
+		// business error should surface rather than burn the remaining endpoints.
+		if c.solanaRPC != nil && isRetryableRPCErr(lastErr) {
 			c.log.Warn("Settlement query failed, failing over to next endpoint",
-				"host", c.Host, "endpoint", c.currentSolanaRPCURL(), "error", lastErr)
+				"host", c.Host, "endpoint", redactURL(c.currentSolanaRPCURL()), "error", c.solanaRPC.scrubErr(lastErr))
 			c.solanaRPC.Failover()
+		} else {
+			return lastErr
 		}
 	}
 	return lastErr
@@ -260,7 +265,10 @@ func (c *Client) GetEffectiveSeatPrice(ctx context.Context, devicePubkey string,
 	var seat *shreds.ClientSeat
 	if err := poll.Until(ctx, func() (bool, error) {
 		s, fetchErr := shredsClient.FetchClientSeat(ctx, deviceKey, clientIPBits)
-		if errors.Is(fetchErr, shreds.ErrAccountNotFound) {
+		// A missing account surfaces as rpc.ErrNotFound through the live RPC
+		// path (gagliardetto GetAccountInfo) and as shreds.ErrAccountNotFound
+		// through the shreds nil-result path; treat both as "not yet visible".
+		if errors.Is(fetchErr, shreds.ErrAccountNotFound) || errors.Is(fetchErr, rpc.ErrNotFound) {
 			c.log.Debug("Client seat not yet visible, polling", "host", c.Host)
 			return false, nil
 		}
--- a/e2e/internal/qa/client_settlement.go
+++ b/e2e/internal/qa/client_settlement.go
@@ -3,6 +3,7 @@ package qa
 import (
 	"context"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"math"
 	"strconv"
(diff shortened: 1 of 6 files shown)


e2e/internal/qa/client_settlement.go
```diff
@@ -247,13 +347,25 @@ func (c *Client) GetUSDCBalance(ctx context.Context) (uint64, error) {
...
-	solanaClient := rpc.New(c.SolanaRPCURL)
+	// Use the failover pool when present so a dead/lagging endpoint is replaced
+	// transparently; fall back to a single-endpoint client for hand-built test
+	// clients. Before reading, actively fail over off a lagging node so a
+	// stale-but-valid read can't produce a spurious assertion failure.
+	var solanaClient *rpc.Client
+	budget := rpcBudgetFromEnv()
+	if c.solanaRPC != nil {
+		c.solanaRPC.SelectHealthiestEndpoint(ctx)
+		solanaClient = c.solanaRPC.RPC()
```
packethog, line 372: Credential redaction is bypassed on the balance read path. With the pool present, this routes a potentially private, API-keyed endpoint (SOLANA_RPC_URL / SOLANA_RPC_FALLBACK_URLS) through GetUSDCBalance. solana-go embeds the full request URL — including any embedded API key — in its connectivity/HTTP error strings, but the retry body below logs the raw error (`"error", rpcErr`, ~L375) and wraps it into the returned error (`: %w`, ~L380) without scrubbing. The pool's own failover paths already route errors through `scrubErr`/`redactURL`; these two sites don't, so a failing private endpoint can leak its credential into CI logs and surfaced errors. Consider scrubbing here when the pool is present (e.g. `c.solanaRPC.scrubErr(rpcErr)`).

e2e/internal/qa/solanarpc.go
```diff
@@ -0,0 +1,401 @@
...
+		return false
+	}
+	if errors.Is(err, context.Canceled) {
+		return false
+	}
+	if errors.Is(err, context.DeadlineExceeded) {
+		return true
+	}
+
+	var httpErr *jsonrpc.HTTPError
```
packethog, line 384: isRetryableRPCErr does not classify JSON-RPC-level `*jsonrpc.RPCError` responses. `CallForInto` returns `rpcResponse.Error` (a `*jsonrpc.RPCError`) directly, so an explicit stale-node response such as `-32005 "Node is behind by N slots"` (or `-32004`) is treated as non-retryable and surfaces without failing over — even though that is exactly the stale-endpoint condition this change targets. The active slot-lag probe only covers silent staleness (a node serving valid-but-old data with no error); it doesn't help when an endpoint explicitly reports it is behind. Consider matching the relevant `*jsonrpc.RPCError` codes here so those also trigger failover.

packethog (changes requested):
Requesting changes — two items in the inline comments. The failover pool, slot-lag detection, poll-until-consistent reads, and the write-no-retry reasoning are otherwise well-structured and well-tested.

e2e/internal/qa/client_settlement.go
ben-dz: Fixed in 350d7646. `GetUSDCBalance` now scrubs the RPC error through `c.scrubRPCErr(...)` (which routes through the pool's `scrubErr`/`redactURL`) both at the debug log site and before wrapping into the returned error, so a failing private endpoint no longer leaks its credential into CI logs or surfaced errors.

e2e/internal/qa/solanarpc.go
ben-dz: Fixed in 350d7646. `isRetryableRPCErr` now matches `*jsonrpc.RPCError` codes -32005 ("node is behind by N slots") and -32004 ("block not available for slot") as retryable so an endpoint that explicitly reports it is behind triggers failover; all other JSON-RPC error codes remain non-retryable and surface. Added unit coverage for the classification plus a pool-level test asserting a -32005 response fails over to the next endpoint.

e2e/internal/qa/client_settlement.go
ben-dz: Added a Client.scrubRPCErr helper that routes through the pool's existing scrubErr/redactURL, and applied it both at the debug log site and before building the returned error (switched %w → scrubbed %s).

e2e/internal/qa/solanarpc.go
ben-dz: isRetryableRPCErr now classifies those two codes as retryable; all other JSON-RPC codes remain non-retryable (business errors surface).

packethog (approved):

PR #3936, opened 2026-06-23
e2e: fix flaky maxusers rollover and interface validation tests

juan-malbeclabs:
## Summary of Changes
- Make device selection deterministic in the maxusers rollover test by waiting until the nearby device is measured as faster than the faraway device by more than the client's 5ms latency-tolerance window, so the client can no longer connect to the wrong device on a tie.
- Match CLI validation errors case-insensitively in the interface validation test, decoupling assertions from the program's message casing (e.g. "Invalid Interface IP" vs "Invalid interface IP").
- Add a CI safety net: rerun only the failed tests via `gotestsum --rerun-fails=1` so transient flakes don't redden a shard while remaining visible/trackable; raise the job timeout above the `go test` timeout so Go dumps per-test stacks on timeout instead of GitHub killing the job with no diagnostics.

## Diff Breakdown
| Tests        |     3 | +102 / -19   |  +83 |
| Config/build |     1 | +17 / -2     |  +15 |
| **Total**    |     4 | +119 / -21   |  +98 |

Test-only and CI changes; no production code touched.

<details>

- `e2e/internal/devnet/client.go` — new `WaitForLatencyOrdering` helper that polls the daemon's `/latency` endpoint until both devices are reachable and the faster device's `min_latency_ns` is below the slower device's by more than a given margin.
- `e2e/interface_validation_test.go` — add `containsAnyFold` helper and route the validation-error assertions through it for case-insensitive matching across message/enum/hex-code forms.
- `.github/workflows/e2e.yml` — install `gotestsum`, wrap `go test` with `--rerun-fails=1 --rerun-fails-max-failures=15`, and bump the job timeout from 15m to 30m.
- `e2e/device_maxusers_rollover_test.go` — wait for deterministic latency ordering before connecting the client in IBRL mode.

</details>

## Testing Verification
- `TestE2E_DeviceMaxusersRollover` — PASS (287.70s) locally.
- `TestE2E_InterfaceValidation` — PASS (294.08s) locally, including all rejection subtests that depend on the new case-insensitive error matching.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -257,9 +257,7 @@ jobs:
       run:
         working-directory: "e2e/"
     runs-on: doublezero-k8s-ci
-    # Keep above the go test -timeout (20m) plus rerun headroom, so Go dumps its
-    # per-test stack on timeout instead of GitHub killing the job with no diagnostics.
-    timeout-minutes: 30
+    timeout-minutes: 15
     permissions:
       packages: read
       contents: read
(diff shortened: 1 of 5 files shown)


nikw9944 (changes requested):
I'm in the process of building a system that will detect flakes and propose fixes (https://github.com/malbeclabs/infra/issues/1615), the goal of which is to quickly catch and fix flakes. It currently looks for failed e2e on merge to main; this change would require it to look at green runs too, and to look for reformatted FAIL blocks. I'm also worried that the automatic rerun will also mask the flakes behind longer e2e runtimes rather than fixing the flakes. Can we postpone the rerun feature for now and revisit it once the flake-detector lands? 

The maxusers rollover test fix look good.

juan-malbeclabs:
@nikw9944 Great initiative identifying the flaky test! I've already removed it.

nikw9944 (approved):

PR #3938, opened 2026-06-23
cli: add --narrow output to device, link, and access-pass list

ben-dz:
## Summary of Changes
- Add a `--narrow` flag to `device list`, `link list`, and `access-pass list` that renders a width-reduced table for wide output — dropping low-value columns and abbreviating the rest — while leaving `--json` and the default table untouched.
- Abbreviate pubkeys to a contiguous leading-10-char prefix + `..` (copyable for prefix search in explorers); shorten the validator/RPC/`others` keys embedded in `accesspass_type`; reuse the user-list `P:/S:` multicast shortening; map the default tenant pubkey to `empty`.
- Best-effort width target: identifiers (codes, contributor, interface and topology names) stay readable; only pubkeys, embedded keys, and unknown-group/topology fallbacks are bounded.

## Diff Breakdown
| Category    | Files | Lines (+/-)   | Net    |
|-------------|-------|---------------|--------|
| Core logic  |     6 | ~+150 / -11   | ~+139  |
| Scaffolding |     3 | ~+157 / 0     | ~+157  |
| Tests       |     5 | ~+281 / 0     | ~+281  |
| **Total**   |     6 | +588 / -11    |  +577  |

Roughly half the diff is test assertions (per-list narrow checks + helper unit tests); the remainder splits between the formatting/conversion helpers (core) and the clap flags + narrow display structs + render branches (scaffolding). All six files are mixed logic+tests, so per-category file counts overlap.

<details>

- `smartcontract/cli/src/accesspass/list.rs` — `AccessPassDisplayNarrow`; shortens all pubkeys including the key inside `accesspass_type` (exhaustive over SolanaValidator/SolanaRPC/Others); multicast via the shared `narrow_groups`; default tenant → `empty`; `lst_epch`/`rem_epch`/`conns` headers.
- `smartcontract/cli/src/link/list.rs` — `LinkDisplayNarrow`; drops side device names, mtu, tunnel_net, owner; abbreviates health (`ready`) and drained status (`sft_drain`/`hrd_drain`); shortens account; narrow `topos` with a shortened unknown-topology fallback.
- `smartcontract/cli/src/device/list.rs` — `DeviceDisplayNarrow`; drops dz_prefixes, mgmt_vrf, owner (keeps the full account).
- `smartcontract/cli/src/topology/mod.rs` — `resolve_topology_names_short` shortens the pubkey shown for an unknown topology (narrow only; full list unchanged).
- `smartcontract/cli/src/util.rs` — `display_pubkey_short` now emits a contiguous leading-prefix abbreviation instead of a first..last split.
- `smartcontract/cli/src/user/list.rs` — `narrow_groups` made `pub(crate)` for reuse; unknown-group fallback shortened.

</details>

## Testing Verification
- Added per-list narrow tests asserting kept vs dropped/renamed headers and a ≤240-column bound on every rendered line, plus unit tests for the new abbreviation helpers (link status/health, `accesspass_type` including SolanaRPC/Others/EdgeSeat, `narrow_groups` unknown-group fallback, and `resolve_topology_names_short` unknown-topology fallback).
- Verified against mainnet-beta: max line width dropped from 344→225 (device), 404→237 (link), and 500→229 (access-pass).

### DIFF
--- a/smartcontract/cli/src/accesspass/list.rs
+++ b/smartcontract/cli/src/accesspass/list.rs
@@ -1,15 +1,18 @@
-use crate::doublezerocommand::CliCommand;
+use crate::{doublezerocommand::CliCommand, user::list::narrow_groups};
 use clap::Args;
 use doublezero_cli_core::CliContext;
 use doublezero_program_common::serializer;
-use doublezero_sdk::commands::{
-    accesspass::list::ListAccessPassCommand, multicastgroup::list::ListMulticastGroupCommand,
-    tenant::list::ListTenantCommand,
+use doublezero_sdk::{
+    commands::{
+        accesspass::list::ListAccessPassCommand, multicastgroup::list::ListMulticastGroupCommand,
+        tenant::list::ListTenantCommand,
+    },
+    MulticastGroup,
 };
 use doublezero_serviceability::state::accesspass::{AccessPassStatus, AccessPassType};
 use serde::Serialize;
 use solana_sdk::pubkey::Pubkey;
-use std::{io::Write, net::Ipv4Addr};
+use std::{collections::HashMap, io::Write, net::Ipv4Addr};
 use tabled::{settings::Style, Table, Tabled};
 
 #[derive(Args, Debug, Default)]
@@ -54,6 +57,11 @@ pub struct ListAccessPassCliCommand {
     /// Output as compact JSON
     #[arg(long, default_value_t = false)]
     pub json_compact: bool,
+    /// Narrow table output: shortens pubkeys, abbreviates accesspass_type, shows
+    /// the first multicast group per role (`+N` for the rest), maps the default
+    /// tenant pubkey to `empty`, and shortens the epoch/connections headers.
+    #[arg(long, default_value_t = false)]
+    pub narrow: bool,
 }
 
 #[derive(Tabled, Serialize)]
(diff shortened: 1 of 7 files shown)


smartcontract/cli/src/accesspass/list.rs
```diff
@@ -75,6 +83,99 @@ pub struct AccessPassDisplay {
...
+    fn from_display(d: &AccessPassDisplay, mgroups: &HashMap<Pubkey, MulticastGroup>) -> Self {
+        Self {
+            account: d.account,
+            accesspass_type: accesspass_type_short(&d.accesspass_type_value),
+            client_ip: d.client_ip,
+            user_payer: d.user_payer,
+            // A literal default (all-ones) tenant pubkey reads as blank; show
+            // "empty". Non-default tenants stringify differently and are
+            // unaffected; the substring can't occur inside a real code/pubkey.
+            tenant: d.tenant.replace(&Pubkey::default().to_string(), "empty"),
```
juan-malbeclabs: This doesn't do what the comment above (and the CHANGELOG) say. For the common "no tenant" pass, `tenant_allowlist` is empty, so `d.tenant` is `""` (the wide table renders it blank — see the `test_*` expected output). `"".replace(&Pubkey::default().to_string(), "empty")` is a no-op, so narrow also shows blank, never `empty`. The replace only fires in the rare case where the all-ones system pubkey is *explicitly* in the allowlist and missing from the tenants map.

Suggest mapping the empty case directly (keeping the substring replace for the rare explicit-default case):

```suggestion
            // The common "no tenant" pass has an empty allowlist and renders
            // blank in the wide table; surface it as `empty` in narrow. Also map
            // an explicit all-ones tenant pubkey for the rare case it appears in
            // the allowlist.
            tenant: if d.tenant.is_empty() {
                "empty".to_string()
            } else {
                d.tenant.replace(&Pubkey::default().to_string(), "empty")
            },
```

And add a narrow test asserting the `tenant` column contains `empty`, since this path is currently uncovered.

smartcontract/cli/src/link/list.rs
```diff
@@ -251,6 +330,10 @@ impl ListLinkCliCommand {
                     unicast_drained: link.link_flags
                         & doublezero_serviceability::state::link::LINK_FLAG_UNICAST_DRAINED
                         != 0,
+                    link_topologies_narrow: resolve_topology_names_short(
```
juan-malbeclabs: `link_topologies_narrow` is computed for every link row even when rendering `--json` or the default table, where the field is `#[tabled(skip)]` + `#[serde(skip)]` and never read. `topology_map` is already in scope in `execute()`, so the abbreviated string can be built lazily inside the `else if self.narrow` branch (or via `LinkDisplayNarrow::from_display` taking `&topology_map`) instead of being precomputed on the hot path for all callers. Minor, but it's avoidable per-row work — same applies to the extra `#[serde(skip)]`/`#[tabled(skip)]` clone fields on `AccessPassDisplay`.

juan-malbeclabs (approved):
## Review

Nice, contained change and the test coverage across the three commands (plus `topology` and `util`) is solid — output branching is correctly mutually exclusive, pubkey slicing is safe (base58 is ASCII, and `shorten_str` is char-based for non-ASCII `Others` keys). No correctness/panic issues found.

**Approving conditioned on the inline points below** — the only one I'd consider blocking is the `tenant → empty` mapping (1), which doesn't fire for the case its comment/CHANGELOG describe and is untested. (2) and (3) are nice-to-haves.

1. **`access-pass list`: the `tenant → empty` mapping is effectively dead for the common case + untested.** `d.tenant` is the joined `tenant_allowlist`; for the normal "no tenant" pass the allowlist is empty, so `d.tenant == ""` and `"".replace("111…111", "empty")` is a no-op — the cell stays blank, never `empty`. The substring replace only fires if the allowlist *literally contains* the all-ones system pubkey. So the documented behavior ("maps the default tenant pubkey to `empty`") doesn't trigger for the case it names, and the narrow test never asserts `empty` appears. See inline suggestion.

2. **`link list`: `link_topologies_narrow` is computed on every row even for `--json`/default output**, where the field is `#[tabled(skip)]` + `#[serde(skip)]` and unused. Minor wasted work; see inline.

3. **Abbreviation schemes are slightly divergent** (`shorten_str` 12/10 char-based vs `display_pubkey_short` 12/10 byte-based vs `abbreviate_name` 4+4). Defensible per-context, but worth a short note pointing future `--narrow` authors at the canonical helper. Non-blocking.

Also: please add a narrow-output test that asserts the `tenant` cell renders `empty` once (1) is fixed, since that path is currently uncovered.

smartcontract/cli/src/accesspass/list.rs
ben-dz: Fixed in 1e7342de — the empty-allowlist case now maps to `empty` as well (the substring replace is kept for an explicit all-ones default entry inside a non-empty allowlist), and added a narrow test asserting the `empty` cell renders. On mainnet this now labels 2531/2533 rows `empty` (was 30). The flag/struct doc comments are corrected to say it labels a no-tenant pass `empty`.

smartcontract/cli/src/link/list.rs
ben-dz: Link: fixed in 1e7342de — `LinkDisplay` now moves the raw topology pubkeys (owned source, so it's free) and builds the abbreviated `topos` lazily in `from_display(&topology_map)`, so `--json`/default paths do no extra per-row work.

Access-pass: the source there is `&AccessPass` (borrowed from the list result), so the skip fields can't be moved, and building the narrow rows straight from source would duplicate the epoch/flags/users/tenant/multicast formatting the converter reuses. The residual cost is cloning two usually-empty `Vec<Pubkey>` plus a one-Pubkey enum per row, bounded by list size — left as the cheaper tradeoff, happy to restructure if you'd prefer.

(Re: the divergent abbreviation schemes from the review summary — `display_pubkey_short` and the local `shorten_str` are now unified into a single canonical `util::abbreviate_prefix`; multicast group codes still use `abbreviate_name`.)

juan-malbeclabs (approved):

PR #3950, opened 2026-06-29
doublezerod: add route-liveness-backoff-max flag to fix multi-client IBRL e2e flake

### LINKED ISSUE #3949: Fix: active-mode route-liveness backoff flake in multi-client IBRL e2e tests
# Fix: active-mode route-liveness backoff flake in multi-client IBRL e2e tests

## Problem

`TestE2E_MultiClientIBRLAllocatedIP` flakes on `main` (~1.5%, 1 in ~65 e2e runs). Every observed
failure (3/3 over 45 days) is the **same** assertion — "client1 should have route to client2" —
and never the reverse. PR #3944's 60s→90s timeout bump did not fix it (a failure post-dates it).

## Root cause

The two clients use different route-liveness modes:
- **client1 = active** (`RouteLivenessEnableActive: true`)
- **client2 = passive** (`RouteLivenessEnablePassive: true`)

Active and passive install kernel routes differently:
- Passive (`manager.go:332-346`): installs the route **immediately** on BGP receipt.
- Active (`onSessionUp`, `manager.go:731-760`): **withholds** the route until the client↔client
  liveness session reaches `StateUp` (a BFD-style bidirectional handshake completes).

The test asserts on the **kernel** route (`ip r list dev doublezero0`), so the active client's
route can't appear until its handshake with the peer succeeds. That's why only the active
direction ever fails.

The handshake pacing is the defect. While the session is `Down`, `ComputeNextTx`
(`session.go:133-176`) applies exponential backoff: interval = `base × {1,2,4,8,…}` capped at
`BackoffMax`. `base ≈ TxMin = 1s`. `BackoffMax` is **never set by the daemon**, so
`ManagerConfig.Validate` fills in `defaultBackoffMax = 60s` (`manager.go:25,136-137`). The
resulting Down probe schedule (seconds after the route is registered) is:

```
t = 0, 1, 3, 7, 15, 31, 63, 123, …   →  60-second dead zone between t=63 and t=123
```

If the client↔client data path isn't usable during the early fast probes (≤~31s) — plausible on a
loaded e2e shard, where device control-plane routes are confirmed but cross-exchange forwarding /
client BGP sessions are still converging — the active client backs off and won't probe again
between t=63s and t=123s. The 90s test deadline lands inside that gap, so the handshake never
completes and the route is never installed. (Backoff resets on a received packet, but the passive
peer's session is also Down/backing-off on the same schedule, so both sides can be silent
simultaneously.) This also explains why #3944 didn't help: 60s sat in the (31,63) gap and 90s
moved into the longer (63,123) gap.

The same latent flake exists in the sibling tests `multi_client_ibrl_test.go` and
`multi_client_ibrl_liveness_test.go`, which also pair active clients with "client should have route
to peer" assertions (and carry band-aided 60/90/120s timeouts).

## Scope decision (operator-chosen)

Keep the production default at 60s (minimal blast radius). Add a tunable flag, pin a small value in
the e2e clients, and bundle e2e failure diagnostics. Do **not** change `defaultBackoffMax`.

### 1. Add a `-route-liveness-backoff-max` daemon flag
`client/doublezerod/cmd/doublezerod/main.go`

- Add a default const next to the other liveness defaults (line ~70-74):
  `defaultRouteLivenessBackoffMax = 1 * time.Minute` (matches `liveness.defaultBackoffMax`, so prod
  behavior is unchanged).
- Add a flag next to the other liveness flags (line ~48-52):
  `routeLivenessBackoffMax = flag.Duration("route-liveness-backoff-max", defaultRouteLivenessBackoffMax, "route liveness backoff max (cap on Down-state probe interval)")`
- In the `liveness.ManagerConfig{...}` literal (lines 164-185, alongside `TxMin`/`RxMin`/…), add:
  `BackoffMax: *routeLivenessBackoffMax,`

Net: prod default identical to today (60s, now explicit instead of via the Validate fallback);
operators and the e2e harness can override. Leave `liveness.defaultBackoffMax` (60s) in place as
the `BackoffMax==0` fallback in `Validate`.

### 2. Wire the value through the e2e devnet client
`e2e/internal/devnet/client.go`

- Add a field to the client `Spec` struct (near `RouteLivenessEnableActive`/`Passive`, ~line 38):
  `RouteLivenessBackoffMax time.Duration`
- In the daemon-args builder (~lines 195-203,

ben-dz:
## Summary of Changes
* Add a `-route-liveness-backoff-max` daemon flag capping the active-mode Down-state liveness probe interval; production default stays 60s (now explicit, behavior unchanged).
* Pin `RouteLivenessBackoffMax: 3s` on the multi-client IBRL e2e clients so the Down probe schedule is `0,1,3,3,3,…` (no >3s gap) and the active client's route installs once the data path is up — instead of stalling in the 60s `(63s,123s)` dead zone that outlasts the 90s deadline.
* Add e2e route-convergence failure diagnostics, extracted into a shared helper used by all three multi-client IBRL tests, so future flakes are self-explaining.
* Fixes #3949

Root cause: active route-liveness withholds the kernel route until a BFD-style handshake reaches `Up`; while `Down`, `ComputeNextTx` backs off exponentially capped at `BackoffMax` (defaulted to 60s by `ManagerConfig.Validate`). The resulting 60s gap between probes at `t=63s..123s` outlives the test deadline, so only the active direction ("client1 should have route to client2") flaked. #3944's 60→90s bump moved the deadline deeper into the gap, which is why it didn't help.

## Testing Verification
* New manager-level test (`RegisterRoute_PropagatesBackoffMax`) covers the plumbing this PR adds: `ManagerConfig.BackoffMax` flows into each session's `backoffMax`, and a zero value falls back to the 60s default via `Validate`.
* Added a `ComputeNextTx` guardrail test asserting the Down-state interval stays capped at `backoffMax` (+jitter) across 20 consecutive Down transmits. Note: the cap logic itself pre-exists on `main`, so this is a guardrail/characterization test rather than a regression of the cap behavior.
* `go vet -tags e2e ./e2e/...` clean; liveness package tests green.

### DIFF
--- a/client/doublezerod/internal/liveness/session_test.go
+++ b/client/doublezerod/internal/liveness/session_test.go
@@ -294,11 +294,12 @@ func TestClient_Liveness_Session_BackoffResetsWhenNotDown(t *testing.T) {
 	require.Equal(t, uint32(1), s.backoffFactor)
 }
 
-// TestClient_Liveness_Session_ComputeNextTx_DownIntervalCappedAtBackoffMax pins the
-// cap semantics that fix the #3949 e2e flake: while Down, the effective transmit
-// interval grows via exponential backoff but must never exceed backoffMax (plus the
-// ±10% jitter band). With a small backoffMax the Down probe schedule has no large gap,
-// so the active client's liveness handshake converges promptly once the data path is up.
+// TestClient_Liveness_Session_ComputeNextTx_DownIntervalCappedAtBackoffMax is a guardrail
+// for the cap semantics this fix relies on: while Down, the effective transmit interval
+// grows via exponential backoff but must never exceed backoffMax (plus the ±10% jitter
+// band). The cap logic already exists; pinning a small backoffMax (as the e2e harness now
+// does) keeps the Down probe schedule gap-free so the active client's liveness handshake
+// converges promptly once the data path is up.
 func TestClient_Liveness_Session_ComputeNextTx_DownIntervalCappedAtBackoffMax(t *testing.T) {
 	t.Parallel()
 	for _, backoffMax := range []time.Duration{3 * time.Second, 200 * time.Millisecond, 1 * time.Minute} {
--- a/client/doublezerod/internal/liveness/session_test.go
+++ b/client/doublezerod/internal/liveness/session_test.go
@@ -294,6 +294,41 @@ func TestClient_Liveness_Session_BackoffResetsWhenNotDown(t *testing.T) {
 	require.Equal(t, uint32(1), s.backoffFactor)
 }
 
+// TestClient_Liveness_Session_ComputeNextTx_DownIntervalCappedAtBackoffMax pins the
+// cap semantics that fix the #3949 e2e flake: while Down, the effective transmit
+// interval grows via exponential backoff but must never exceed backoffMax (plus the
+// ±10% jitter band). With a small backoffMax the Down probe schedule has no large gap,
+// so the active client's liveness handshake converges promptly once the data path is up.
+func TestClient_Liveness_Session_ComputeNextTx_DownIntervalCappedAtBackoffMax(t *testing.T) {
+	t.Parallel()
+	for _, backoffMax := range []time.Duration{3 * time.Second, 200 * time.Millisecond, 1 * time.Minute} {
+		backoffMax := backoffMax
+		t.Run(backoffMax.String(), func(t *testing.T) {
+			t.Parallel()
+			s := newSess()
+			s.state = StateDown
+			s.localTxMin = 1 * time.Second
+			s.minTxFloor = 50 * time.Millisecond
+			s.maxTxCeil = 3 * time.Second
+			s.backoffMax = backoffMax
+			s.backoffFactor = 1
+			r := rand.New(rand.NewSource(1))
+
+			// The jitter band is ±10% of the (capped) effective interval, so the
+			// observed interval can sit at most backoffMax/10 above backoffMax.
+			maxInterval := backoffMax + backoffMax/10
+			now := time.Unix(0, 0)
+			for i := 0; i < 20; i++ {
+				next := s.ComputeNextTx(now, r)
+				interval := next.Sub(now)
+				require.LessOrEqual(t, interval, maxInterval,
+					"iteration %d: Down interval %v must be capped at backoffMax %v (+jitter)", i, interval, backoffMax)
+				now = next
+			}
+		})
+	}
+}
+
 func TestClient_Liveness_Session_HandleRxIgnoredWhenAdminDown(t *testing.T) {
 	t.Parallel()
 	s := newSess()
--- a/e2e/multi_client_ibrl_allocated_ip_test.go
+++ b/e2e/multi_client_ibrl_allocated_ip_test.go
@@ -3,6 +3,7 @@
 package e2e_test
 
 import (
+	"context"
 	"fmt"
 	"log/slog"
 	"os"
@@ -304,12 +305,16 @@ func runMultiClientIBRLAllocatedIPWorkflowTest(t *testing.T, log *slog.Logger, d
 // #3949 flake slow to diagnose.
 func dumpClientRouteDiag(t *testing.T, log *slog.Logger, name string, client *devnet.Client, peerDZIP string) {
 	t.Helper()
+	// Use a fresh bounded context: the test deadline may already have cancelled
+	// t.Context(), and a wedged container must not let the dump hang.
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
 	for _, cmd := range [][]string{
 		{"doublezero", "status"},
 		{"ip", "route", "show"},
 		{"ip", "r", "list", "dev", "doublezero0"},
 	} {
-		out, err := client.Exec(t.Context(), cmd)
+		out, err := client.Exec(ctx, cmd)
 		log.Error("route-diag",
 			"client", name,
 			"pubkey", client.Pubkey,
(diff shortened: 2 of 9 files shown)


client/doublezerod/internal/liveness/session_test.go
```diff
@@ -294,6 +294,41 @@ func TestClient_Liveness_Session_BackoffResetsWhenNotDown(t *testing.T) {
 	require.Equal(t, uint32(1), s.backoffFactor)
 }
 
+// TestClient_Liveness_Session_ComputeNextTx_DownIntervalCappedAtBackoffMax pins the
```
nikw9944: This test asserts the Down-state cap in `ComputeNextTx`, but `session.go` is byte-identical to `main` in this PR — the cap logic and `ManagerConfig.Validate` defaulting already exist on main. Copying this test onto main's `session.go` and running with `-count=1` passes, so this is a characterization/guardrail test, not a regression test for the fix. The PR description's claim that it 'fails today's 60s climb to 32/60s; passes with the cap' is therefore inaccurate. The behavior this PR actually adds — the flag → `ManagerConfig.BackoffMax` → `session.backoffMax` plumbing — has no direct test coverage. Suggest rewording the comment/PR description, and optionally adding a test that runs `ManagerConfig{BackoffMax: 3s}` (and `0`) through `Validate()` and asserts the resulting session's `backoffMax` (3s, and the 60s default).

e2e/multi_client_ibrl_allocated_ip_test.go
```diff
@@ -284,3 +299,29 @@ func runMultiClientIBRLAllocatedIPWorkflowTest(t *testing.T, log *slog.Logger, d
 	require.Equal(t, devnet.ClientSessionStatusDisconnected, status[0].DoubleZeroStatus.SessionStatus)
 	log.Debug("--> Confirmed clients are disconnected and do not have a DZ IP allocated")
 }
+
+// dumpClientRouteDiag emits tunnel/route state for a client when a route-convergence
+// assertion times out. These checks otherwise emit nothing on failure, which made the
+// #3949 flake slow to diagnose.
+func dumpClientRouteDiag(t *testing.T, log *slog.Logger, name string, client *devnet.Client, peerDZIP string) {
+	t.Helper()
+	// Use a fresh bounded context: the test deadline may already have cancelled
```
nikw9944: The comment justifies the fresh `context.Background()` by saying `t.Context()` 'may already have cancelled,' but this helper runs before `t.Fatalf` while the test is still active, so `t.Context()` is not yet cancelled. The 30s-bounded fresh context is still the right call (a wedged container must not hang the dump); just reword the rationale to that actual benefit.

nikw9944 (commented):
Small, low-risk, backward-compatible change adding a `-route-liveness-backoff-max` flag, a 3s e2e pin, and route-convergence diagnostics. Production behavior is genuinely unchanged (flag default 1m matches `defaultBackoffMax`; `Validate` already defaulted 0→60s). `go vet -tags e2e ./e2e/...` is clean and the new unit test passes.

No critical, high, security, or data-loss issues. Two Medium points: (1) the new unit test (`session_test.go`) asserts behavior in `session.go`, which is byte-identical to `main` — it passes on `main` unchanged, so it is a characterization test, not a regression test for this fix, and the plumbing this PR actually adds (flag → ManagerConfig.BackoffMax → session.backoffMax) has no direct coverage; the PR description's claim that it 'fails today's 60s climb' is inaccurate. (2) Failure diagnostics were added to only one of the three affected e2e tests. Plus two Low nits.

Recommended verdict: COMMENT.

- Consistency (Medium): `dumpClientRouteDiag` + `assert.Eventually`/`t.Fatalf` was applied only to `multi_client_ibrl_allocated_ip_test.go`. The route-convergence checks in `multi_client_ibrl_test.go` (lines ~289, ~296) and `multi_client_ibrl_liveness_test.go` received the `RouteLivenessBackoffMax: 3s` pin but still use bare `require.Eventually` with no diagnostics, so the same flake there stays opaque. Consider extracting a shared route-convergence-with-diagnostics helper for all three, or note why only the allocated-IP variant needs it.
- UX (Low): A `-route-liveness-backoff-max` value below `MinTxFloor` (e.g. `30ms` vs the 50ms default) makes `ManagerConfig.Validate` (manager.go:142-144) error and the daemon refuse to start, rather than clamping. Consistent with existing `MaxTxCeil`/`MinTxFloor` validation and defensible; optionally document `must be >= route-liveness-min-tx-floor` in the flag help text.
- Security (Low, not exploitable, pre-existing): `extraArgs` is space-joined into `DZ_CLIENT_EXTRA_ARGS` and consumed unquoted in `e2e/docker/client/entrypoint.sh`. The new value is a `time.Duration` formatted with `%s` (only ever `3s`/`1m0s`/`200ms`, no shell metacharacters) set by test code, so there is no injection path; this PR does not worsen the pattern.

nikw9944 (approved):

ben-dz:
Thanks @nikw9944. Addressed:

- **Test coverage / inaccurate claim (Medium):** Added `TestClient_Liveness_Manager_RegisterRoute_PropagatesBackoffMax`, which runs `ManagerConfig{BackoffMax: 3s}` and `{BackoffMax: 0}` through `Validate`/`NewManager` and asserts the registered session's `backoffMax` (3s, and the 60s default) — direct coverage of the flag→`ManagerConfig.BackoffMax`→`session.backoffMax` seam. Reworded the `ComputeNextTx` test comment to call it a guardrail (not a regression of the cap), and fixed the PR description's "fails today's 60s climb" claim.
- **Diagnostics consistency (Medium):** Extracted `requireClientHasRoutes` + `dumpClientRouteDiag` into `multi_client_ibrl_helpers_test.go` and applied it to the client-route checks in all three IBRL tests, replacing the bare `require.Eventually` calls.
- **Flag help text (Low):** Added `must be >= route-liveness-min-tx-floor` to the flag description.
- **Context rationale comment (inline):** Reworded to the actual benefit (a wedged container must not hang the dump; `t.Context()` is still live at that point).

PR #3951, opened 2026-06-29
RFC-22: Client-Originated PIM Register for Multicast Source Origination

ben-malbeclabs:
## Summary

Adds **RFC-22: Client-Originated PIM Register for Multicast Source Origination** (Status: Draft).

When a client both publishes and subscribes to the same multicast group over one GRE tunnel (the simultaneous pub/sub case from RFC-15), the subscriber-side PIM neighbor causes the source's RPF to resolve via that neighbor, which suppresses the device's `pim border-router` source injection. The device never sets the `N` flag, never originates the MSDP SA, and the source is invisible to the rest of the anycast-RP mesh — so remote subscribers receive it only intermittently (observed on group 233.84.178.5).

This RFC proposes that `doublezerod` originate a PIM Register **beacon** for its published sources, sent periodically to the RP. Receiving the Register makes the device originate the MSDP SA regardless of border-router suppression — Arista's own lab confirmed that permitting the Register in flips the `N` flag on. The device-side change is a single additive ACL permit; `pim border-router` is retained as a backstop (belt-and-suspenders).

### Highlights
- Client `RegisterSender` reuses the existing heartbeat as the encapsulated packet; beacon semantics (no inbound Register-Stop path).
- Interface-pinned raw socket so no `10.0.0.0` route is installed in the validator's routing table.
- Default 60s, staggered cadence; CoPP headroom shown to be ample (well under 0.5% of the relevant class on the deployed 7130LBR / 7280CR3A defaults).
- Phased, backward-compatible rollout; no smart-contract change.

Full design, the packet-encapsulation breakdown, alternatives, security, and open questions are in the RFC.

Draft for review. RFC number `22` is provisional pending assignment.


ben-dz (approved):
This seems like a good plan to me. I'm sure it will evolve as implementation happens, but lets merge it, and it can evolve through PRs.

Issue #1700, opened 2025-09-22
Funder: topping off user accounts


dalalama:
Ensure DZ Ledger users have sufficient credits to perform operations.  

For users service keys we need to ensure that they have a balance greater than x credits for the connect/disconnect transaction fee. 
For contributors service keys we need to ensure that the account has  greater than y credits for smart contract operations.

PR #3953, opened 2026-06-30
serviceability: add Feed catalog account and CRUD

nikw9944:
Add a serviceability `Feed` account: a catalog entry scoped to a single metro, keyed by `(code, exchange)` and holding the multicast groups joinable there — one `feed_key` is one feed in one metro (e.g. `hyperliquid@tokyo`). Create/update/delete managed by a catalog admin (`FEED_AUTHORITY` Permission or `FOUNDATION`). Includes Rust SDK command builders and a `doublezero feed` CLI.

Part of malbeclabs/infra#1700.

**Stack (merge bottom-up):**
1. #3953 — Feed catalog account + CRUD
2. #3954 — EdgeSeat data model + `SetAccessPassFeeds`
3. #3955 — metro gate enforcement at connect
4. #3956 — Go/Python/TS SDK read support


martinsander00:
should we add `tests/feed_test.rs` covering create/update/delete success + all error/edge cases?

smartcontract/cli/src/feed/update.rs
```diff
@@ -0,0 +1,55 @@
...
+    /// Updated name for the feed
+    #[arg(long)]
+    pub name: Option<String>,
+    /// Replace the metro map with these `EXCHANGE_PK=GROUP_PK[,GROUP_PK...]` entries (repeatable).
+    /// When omitted, the metro map is left unchanged.
+    #[arg(long = "metro", value_parser = parse_metro)]
+    pub metros: Vec<(Pubkey, Vec<Pubkey>)>,
+}
+
+impl UpdateFeedCliCommand {
```
martinsander00, line 28: should we add a `--clear-metros` option?

smartcontract/programs/doublezero-serviceability/src/state/feed.rs
```diff
@@ -0,0 +1,183 @@
...
+            deserialize_with = "doublezero_program_common::serializer::deserialize_pubkey_from_string"
+        )
+    )]
+    pub owner: Pubkey, // 32
+    pub bump_seed: u8,             // 1
+    pub code: String,              // 4 + len (PDA seed, immutable)
+    pub name: String,              // 4 + len
+    pub reference_count: u32,      // 4 - number of access passes referencing this feed
+    /// `exchange_pk → group_pks`. Empty ⇒ no metro restriction.
+    pub metros: Vec<(Pubkey, Vec<Pubkey>)>,
```
martinsander00: **Strong/domain types over raw primitives.** Consider a named type instead of the bare tuple, e.g. `struct MetroGroups { exchange: Pubkey, groups: Vec<Pubkey> }`. This `(Pubkey, Vec<Pubkey>)` shape is repeated across `FeedCreateArgs`, this account, the SDK command, and the CLI, and no other serviceability state models a map this way — a named type makes the `(_, groups)` / `(exchange_pk, group_pks)` destructures read as domain concepts.

smartcontract/programs/doublezero-serviceability/src/state/feed.rs
```diff
@@ -0,0 +1,183 @@
...
+}
+
+impl Feed {
+    /// Match `exchange` against this feed's metro map. See [`FeedMetroMatch`].
+    pub fn groups_for(&self, exchange: &Pubkey) -> FeedMetroMatch<'_> {
+        if self.metros.is_empty() {
+            return FeedMetroMatch::Unrestricted;
+        }
+        match self.metros.iter().find(|(ex, _)| ex == exchange) {
```
martinsander00: **Revert on invariant violations rather than silently dedupe.** `.find()` silently resolves a duplicate exchange key to the first entry. If `metros` ever contains the same exchange twice the underlying data is wrong — better to reject duplicate exchange keys in `process_create_feed` / `process_update_feed` (revert) than to silently pick one here.

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
```diff
@@ -0,0 +1,107 @@
...
+        account_type: AccountType::Feed,
+        owner: *payer_account.key,
+        bump_seed,
+        code: code.clone(),
+        name: value.name.clone(),
+        reference_count: 0,
+        metros: value.metros.clone(),
+    };
+
+    try_acc_create(
```
martinsander00, line 98: **Program logs for state changes.** There's no success `msg!` on create (the `Created:` log inside `try_acc_create` is `#[cfg(test)]`-gated), and `update`/`delete` log nothing either. Consider an explicit log for these state changes, and assert the exact log line in the integration tests.

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
```diff
@@ -0,0 +1,107 @@
...
+    let payer_account = next_account_info(accounts_iter)?;
+    let system_program = next_account_info(accounts_iter)?;
+
+    assert!(payer_account.is_signer, "Payer must be a signer");
+    assert_eq!(
+        globalstate_account.owner, program_id,
+        "Invalid GlobalState Account Owner"
+    );
+    assert_eq!(
+        *system_program.unsigned_key(),
```
martinsander00: **Redundant/inconsistent check.** `create` asserts the system-program id but `update`/`delete` don't — inconsistent within the PR. Per `programs/CLAUDE.md` this check is unnecessary (the system interface builds its instruction with the system program as program-id and reverts automatically), so dropping it here to match `update`/`delete` and leaning on the revert would be more consistent. Minor.

smartcontract/cli/src/feed/list.rs
```diff
@@ -0,0 +1,62 @@
...
+impl ListFeedCliCommand {
+    pub async fn execute<C: CliCommand, W: Write>(
+        self,
+        _ctx: &CliContext,
+        client: &C,
+        out: &mut W,
+    ) -> eyre::Result<()> {
+        let feeds = client.list_feed(ListFeedCommand)?;
+
+        let mut displays: Vec<FeedDisplay> = feeds
```
martinsander00: Nit: prefer turbofish at the callsite over annotating the binding — `.collect::<Vec<FeedDisplay>>()` (or just let the following `sort_by` infer it) rather than `let mut displays: Vec<FeedDisplay> =`.

martinsander00 (commented):
A few pattern-based suggestions (from Karl's usual review patterns). None are correctness/security blockers — overall this looks good and cleanly scoped as Part 1. Comments inline.

smartcontract/cli/src/feed/update.rs
nikw9944: Added `--clear-metros` on `feed update` (errors if combined with `--metro`).

smartcontract/programs/doublezero-serviceability/src/state/feed.rs
nikw9944: Done — added a `MetroGroups { exchange, groups }` type used across the account, args, SDK, and CLI (borsh-identical to the old tuple).

smartcontract/programs/doublezero-serviceability/src/state/feed.rs
nikw9944: Now reject duplicate exchange keys in create/update with `InvalidArgument`.

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
nikw9944: Added non-test success logs on create/update/delete.

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
nikw9944: Dropped the redundant system-program assert to match update/delete.

smartcontract/cli/src/feed/list.rs
nikw9944: Done — turbofish on the `collect`.

nikw9944:
Added `tests/feed_test.rs` covering create/update/delete plus duplicate-code and duplicate-exchange rejection and an unauthorized caller.

nikw9944 (commented):
Automated review using the `solana-program-review` skill (distilled from karl-dz's PR feedback across the DoubleZero Solana repos). Focused on the on-chain serviceability program + tests; CLI/SDK treated as off-chain and skipped.

Baseline is solid — the handlers mirror the sibling `location`/`exchange` create processors, the PDA bump is cached, `account_type` is enforced on deserialize, and `try_acc_close` closes properly. Inline comments below, roughly in priority order (no-op reject and the test assertions are the main ones).

_Checked and cleared as house style / correct: `assert!` validation, `try_from(rest).unwrap()`, not re-deriving the PDA on update/delete, not validating the system program, account-type enforcement._

One out-of-band question: `reference_count` is created at 0 and guarded on delete, but nothing here increments/decrements it — confirm the access-pass code references a feed symmetrically (increment on attach, decrement on detach), or that it's a follow-up PR.

smartcontract/programs/doublezero-serviceability/src/processors/feed/update.rs
```diff
@@ -0,0 +1,85 @@
...
+        payer_account.key,
+        &globalstate,
+        permission_flags::FEED_AUTHORITY | permission_flags::FOUNDATION,
+    )?;
+
+    if let Some(ref metros) = value.metros {
+        reject_duplicate_exchanges(metros)?;
+    }
+
+    let mut feed = Feed::try_from(feed_account)?;
```
nikw9944, line 66: **Reject the all-`None` no-op.** `FeedUpdateArgs` is all-optional and derives `Default`, but this handler never early-reverts when nothing is set — `UpdateFeed { name: None, metros: None }` runs auth, re-serializes the account unchanged, and costs the caller credits for a no-op.

This is the exact guidance from #3279 on this repo: *"you should revert at the beginning of the instruction processor if args == UpdateGeolocationUserArgs::default() (derive Default). Then if you add more optional args, you still won't have to check if any of the values are modified."*

```rust
if value == &FeedUpdateArgs::default() {
}
```
at the top of the processor (the derive is already there).

smartcontract/programs/doublezero-serviceability/tests/feed_test.rs
```diff
@@ -0,0 +1,317 @@
...
+            AccountMeta::new(globalstate_pubkey, false),
+        ],
+        &payer,
+    )
+    .await;
+
+    // InvalidArgument is Custom(65).
+    let error_string = format!("{:?}", result.unwrap_err());
+    assert!(
+        error_string.contains("Custom(65)"),
```
nikw9944: **Pin the exact error variant + log line, not a substring.** `error_string.contains("Custom(65)")` is a substring match: it doesn't pin *which* instruction index failed or *which* check fired, and the hardcoded `65` silently rots if the error enum is renumbered. (Same pattern at lines 279 and 314.)

Per #465: *"the repository convention is to also pin the exact program-log line at an exact index, since `InvalidAccountData` is emitted by many checks and the variant alone doesn't identify which one fired."*

Assert the exact `InstructionError(0, InstructionError::Custom(DoubleZeroError::InvalidArgument as u32))` (derive the code from the enum, don't inline `65`) and `assert_eq!` the exact `msg!` line (`"Duplicate exchange in metros: {}"`) at its index.

smartcontract/programs/doublezero-serviceability/src/processors/feed/delete.rs
```diff
@@ -0,0 +1,65 @@
...
+    authorize(
+        program_id,
+        accounts_iter,
+        payer_account.key,
+        &globalstate,
+        permission_flags::FEED_AUTHORITY | permission_flags::FOUNDATION,
+    )?;
+
+    let feed = Feed::try_from(feed_account)?;
+    if feed.reference_count > 0 {
```
nikw9944, line 49: **This rejection path is untested.** `reference_count > 0` is the only real invariant on delete, but there's no negative test exercising it. Karl's testing standard: cover *every* rejection path a handler adds, not just the happy path. Add a test that creates a feed, sets `reference_count > 0`, and asserts delete is rejected with the exact `ReferenceCountNotZero` variant + log line.

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
```diff
@@ -0,0 +1,121 @@
...
+    assert_eq!(
+        globalstate_account.owner, program_id,
+        "Invalid GlobalState Account Owner"
+    );
+    assert!(feed_account.is_writable, "PDA Account is not writable");
+
+    reject_duplicate_exchanges(&value.metros)?;
+
+    let code =
+        validate_account_code(&value.code).map_err(|_| DoubleZeroError::InvalidAccountCode)?;
```
nikw9944, line 78: **`name` and the metro vecs are unbounded.** `code` is validated here, but `name` is unchecked (any length/charset) and `metros` / per-metro `groups` are unbounded `Vec`s — a caller can bloat the account (rent falls on the payer, and large inputs approach account-size limits), and `reject_duplicate_exchanges` is O(n²) over uncapped `n`. Suggest a max `name` length, a cap on `metros.len()`/`groups.len()`, and rejecting an empty `groups` vec if that's meaningless (reject degenerate input).

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+impl fmt::Debug for FeedCreateArgs {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "code: {}, name: {}, metros: {}",
+            self.code,
+            self.name,
+            self.metros.len()
```
nikw9944: **Hand-written `Debug` prints only `metros.len()`.** Per #888: *"Just showing the len doesn't really lend itself to fruitful debugging… if we don't even plan on debugging these instruction arguments, maybe we don't even implement either Debug or Display."* If the intent is to keep the `format!("{args:?}")` logs in instructions.rs short, a one-line comment saying so beats a len-only `Debug`; otherwise just `derive(Debug)`. (Same for `FeedUpdateArgs`.)

smartcontract/programs/doublezero-serviceability/src/processors/feed/delete.rs
```diff
@@ -0,0 +1,65 @@
...
+    msg,
+    pubkey::Pubkey,
+};
+
+#[derive(BorshSerialize, BorshDeserializeIncremental, PartialEq, Clone, Default)]
+pub struct FeedDeleteArgs {}
+
+impl fmt::Debug for FeedDeleteArgs {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "")
```
nikw9944: A `Debug` impl that writes `""` is the "Debug that writes nothing" anti-pattern. `FeedDeleteArgs` is a unit struct — just `#[derive(Debug)]` (or drop it) rather than hand-writing an empty one.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ All notable changes to this project will be documented in this file.
 - SDK
   - revdist Python SDK migrated to the async solana-py RPC API (solana-py 0.40.0 removed the sync `Client`). The `Client` read methods (`fetch_config`, `fetch_distribution`, etc.) are now coroutines and must be awaited; `new_rpc_client` returns an `AsyncClient`. (#3945)
 
+### Added
+
+- Serviceability
+  - `Feed` account: a catalog mapping `metro(exchange) → group-set`, managed by a catalog admin (`FEED_AUTHORITY` Permission or `FOUNDATION`) via `CreateFeed`/`UpdateFeed`/`DeleteFeed`. A feed with no metros imposes no restriction. (#1700)
```
nikw9944: **Use the PR number, not the ticket.** `(#1700)` is the infra ticket from the branch name; every other CHANGELOG entry references the PR number (`#3973`, `#3972`, …). Change to `(#3953)`. Also worth checking whether a new `### Added` block fits — the Unreleased section otherwise uses `### Breaking` / `### Changes`.

smartcontract/programs/doublezero-serviceability/src/processors/feed/update.rs
nikw9944: Added an all-`None` no-op guard (`if value == &FeedUpdateArgs::default()`) at the top of `process_update_feed`, reverting with `InvalidArgument` before any account/auth work.

smartcontract/programs/doublezero-serviceability/tests/feed_test.rs
nikw9944: Now assert the structured `InstructionError(0, Custom(code))` with `code` derived from the enum via `ProgramError::from(DoubleZeroError::…)` (no inline `65`). On the log line: I couldn't assert it — solana-program-test's native `processor!` harness doesn't surface `Program log:` lines to the BanksClient (both `process_transaction_with_metadata` and `simulate_transaction` return only the runtime `invoke`/`failed` frames), so pinning the variant + instruction index is the strongest available here. Documented that in a comment on the test helper.

smartcontract/programs/doublezero-serviceability/src/processors/feed/delete.rs
nikw9944: Added `test_feed_delete_with_reference_count_rejected`: seeds a `Feed` with `reference_count = 1` (and a foundation GlobalState) via `ProgramTest::add_account`, then asserts delete is rejected with `ReferenceCountNotZero` (Custom(13)) at instruction index 0.

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
nikw9944: Added a shared `validate_feed_inputs` used by create and update: name ≤ 64 (`NameTooLong`), ≤ 64 metros, ≤ 64 groups/metro, and reject an empty per-metro `groups` vec (`InvalidArgument`). Empty top-level `metros` stays valid (no restriction). Duplicate-exchange rejection folded in.

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
nikw9944: Switched `FeedCreateArgs` and `FeedUpdateArgs` to `#[derive(Debug)]` (dropped the len-only hand impls).

smartcontract/programs/doublezero-serviceability/src/processors/feed/delete.rs
nikw9944: Switched `FeedDeleteArgs` to `#[derive(Debug)]` (dropped the empty hand impl).

CHANGELOG.md
nikw9944: Changed to `(#3953)`; kept the `### Added` block (the other stacked PRs' entries now use their own PR numbers too).

martinsander00 (dismissed):

smartcontract/sdk/rs/src/commands/feed/get.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub fn execute(&self, client: &dyn DoubleZeroClient) -> eyre::Result<(Pubkey, Feed)> {
+        match parse_pubkey(&self.pubkey_or_code) {
+            Some(pk) => match client.get(pk)? {
+                AccountData::Feed(feed) => Ok((pk, feed)),
+                _ => Err(eyre::eyre!("Invalid Account Type")),
+            },
+            None => client
+                .gets(AccountType::Feed)?
+                .into_iter()
+                .find(|(_, v)| match v {
```
martinsander00: now that `(code, exchange)` is the key and the same code lives in many metros (your own `test_feed_same_code_different_exchange_allowed`), looking up by code alone returns whichever feed the `gets()` HashMap happens to yield first. that's just confusing for `get`, but `doublezero feed delete --pubkey hyperliquid` / `feed update --pubkey hyperliquid` will mutate or delete an arbitrary metro's feed. can we require the exchange alongside the code (or refuse when the code resolves to more than one feed) so we can't hit the wrong metro?

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ All notable changes to this project will be documented in this file.
 
 ### Breaking
 
+### Added
+
+- Serviceability
+  - `Feed` account: a catalog mapping `metro(exchange) → group-set`, managed by a catalog admin (`FEED_AUTHORITY` Permission or `FOUNDATION`) via `CreateFeed`/`UpdateFeed`/`DeleteFeed`. A feed with no metros imposes no restriction. (#3953)
```
martinsander00: this still says "a catalog mapping `metro(exchange) → group-set`" and "a feed with no metros imposes no restriction" — both are from before the per-metro pivot. a feed is one metro now, and create rejects the default exchange, so there's no no-metro feed. can we reword to match?

smartcontract/cli/src/cli/command.rs
```diff
@@ -62,6 +63,8 @@ pub enum ServiceabilityCommand {
     Location(LocationCliCommand),
     /// Manage exchanges
     Exchange(ExchangeCliCommand),
+    /// Manage feeds (metro→group-set catalog)
```
martinsander00: same here — `Manage feeds (metro→group-set catalog)` is the pre-pivot wording; a feed is one metro's group set now.

smartcontract/cli/src/cli/feed.rs
```diff
@@ -0,0 +1,28 @@
...
+    #[command(subcommand)]
+    pub command: FeedCommands,
+}
+
+#[derive(Debug, Subcommand)]
+pub enum FeedCommands {
+    /// Create a new feed (catalog entry)
+    #[clap()]
+    Create(CreateFeedCliCommand),
+    /// Update a feed's name or metro map
```
martinsander00: and here — `Update a feed's name or metro map`; there's no metro map anymore, just name/groups.

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
```diff
@@ -0,0 +1,144 @@
...
+    let (expected_pda, bump_seed) = get_feed_pda(program_id, &code, &value.exchange);
+    assert_eq!(feed_account.key, &expected_pda, "Invalid Feed PubKey");
+
+    if !feed_account.data_is_empty() {
+        return Err(ProgramError::AccountAlreadyInitialized);
+    }
+
+    // Catalog admin: FEED_AUTHORITY (Permission PDA) or FOUNDATION.
+    let globalstate = GlobalState::try_from(globalstate_account)?;
+    authorize(
```
martinsander00: small one — `authorize()` runs after the name/groups/exchange validation, the PDA check and `data_is_empty`, so an unauthorized caller can probe existence and trip validation errors before getting `NotAllowed`. exchange/create authorizes right after parsing globalstate. want to move it up to match?

smartcontract/sdk/rs/src/commands/feed/get.rs
nikw9944: Fixed. `GetFeedCommand` now refuses a bare code that resolves to more than one feed, and `feed get/update/delete` take an optional `--exchange` to derive the exact `(code, exchange)` PDA — so update/delete can no longer hit an arbitrary metro. Added an ambiguity unit test.

CHANGELOG.md
nikw9944: Reworded to describe a per-metro `(code, exchange)` feed (no no-metro case).

smartcontract/cli/src/cli/command.rs
nikw9944: Reworded to "per-metro multicast group-set catalog".

smartcontract/cli/src/cli/feed.rs
nikw9944: Reworded — "Update a feed's name or group set" (and create is now "a metro's multicast group set").

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
nikw9944: Moved `authorize()` up to right after parsing GlobalState, before the name/groups/exchange validation, PDA check, and `data_is_empty` probe.

nikw9944:
On the `reference_count` symmetry question: it's intentionally monotonic. `SetAccessPassFeeds` (#3954) increments it when a feed is newly attached to a pass, but detaching is *not* decremented — `SetAccessPassFeeds` only receives the accounts for the new feed set, not the dropped ones, so it can't safely touch a removed feed's account. An over-count only makes `DeleteFeed` harder (never unsafe), since `reference_count` solely guards delete. The *seat* counter `FeedSeat.current_users` is the one that must be symmetric, and #3955 makes that deterministic (release bound to the feed stored on the User, not a caller-supplied account). Full `reference_count` decrement-on-detach would need the dropped feed accounts passed to `SetAccessPassFeeds` — happy to do it as a follow-up if we want reclaimable deletes.

martinsander00 (approved):

PR #3954, opened 2026-06-30
serviceability: EdgeSeat feed-seat data model and SetAccessPassFeeds

nikw9944:
Rework the `AccessPass` `EdgeSeat` variant into `EdgeSeat(Vec<FeedSeat>)` (feed_key + per-feed cap) and add the `SetAccessPassFeeds` instruction so the oracle provisions feed_keys onto a pass via its `ACCESS_PASS_ADMIN` permission.

Also drops `Feed.reference_count`: feeds are not reference-counted — a deleted-but-referenced feed_key fails closed at connect, so the count can't diverge and never blocks `DeleteFeed`.

EdgeSeat and Feed are new and unused, so the borsh layout changes need no migration.

Part of malbeclabs/infra#1700.

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_feeds.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_feeds.rs
@@ -108,13 +108,24 @@ pub fn process_set_access_pass_feeds(
     // newly-referenced feeds. NOTE: feeds dropped from the pass are intentionally NOT decremented
     // here — their accounts are not passed, and an over-count only makes a feed harder to delete
     // (never unsafe), since reference_count solely guards DeleteFeed.
-    let mut new_seats = Vec::with_capacity(value.feeds.len());
+    let mut new_seats: Vec<FeedSeat> = Vec::with_capacity(value.feeds.len());
     for (seat, feed_account) in value.feeds.iter().zip(feed_accounts.iter()) {
         assert_eq!(
             *feed_account.key, seat.feed_key,
             "Feed account does not match seat feed_key"
         );
         assert_eq!(feed_account.owner, program_id, "Invalid Feed Account Owner");
+
+        // Reject a feed_key listed more than once: it would double-bump reference_count and
+        // write duplicate seats, neither of which is reclaimable.
+        if new_seats.iter().any(|s| s.feed_key == seat.feed_key) {
+            msg!(
+                "Duplicate feed_key in SetAccessPassFeeds: {}",
+                seat.feed_key
+            );
+            return Err(DoubleZeroError::InvalidArgument.into());
+        }
+
         let mut feed = Feed::try_from(*feed_account)?;
 
         let current_users = prior_seats
@@ -122,6 +133,18 @@ pub fn process_set_access_pass_feeds(
             .find(|s| s.feed_key == seat.feed_key)
             .map(|s| s.current_users)
             .unwrap_or(0);
+
+        // A re-provision must not set a cap below the live count carried over from the prior seat,
+        // or the seat would start over its cap (enforced once connect-time ticking lands).
+        if seat.max_users < current_users {
+            msg!(
+                "max_users {} below current_users {} for feed {}",
+                seat.max_users,
+                current_users,
+                seat.feed_key
+            );
+            return Err(DoubleZeroError::InvalidArgument.into());
+        }
 
         if !prior_seats.iter().any(|s| s.feed_key == seat.feed_key) {
             assert!(feed_account.is_writable, "Feed Account is not writable");
--- a/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_feeds.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_feeds.rs
@@ -1,0 +1,148 @@
+use crate::{
+    authorize::authorize,
+    error::DoubleZeroError,
+    pda::get_accesspass_pda,
+    serializer::try_acc_write,
+    state::{
+        accesspass::{AccessPass, AccessPassType, FeedSeat},
+        feed::Feed,
+        globalstate::GlobalState,
+        permission::permission_flags,
+    },
+};
+use borsh::BorshSerialize;
+use borsh_incremental::BorshDeserializeIncremental;
+use core::fmt;
+use solana_program::{
+    account_info::{next_account_info, AccountInfo},
+    entrypoint::ProgramResult,
+    msg,
+    pubkey::Pubkey,
+};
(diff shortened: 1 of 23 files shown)


smartcontract/programs/doublezero-serviceability/src/state/accesspass.rs
```diff
@@ -39,8 +58,13 @@ pub enum AccessPassType {
...
     Others(String, String), // (type_name, key)
-    // The seat is identified by the access pass `user_payer`, so the variant carries no payload.
-    EdgeSeat,
+    /// A metro-gated seat scoped to one or more feeds. Each `FeedSeat` is one SKU (feed_key +
+    /// per-feed cap). Provisioned by the oracle via `SetAccessPassFeeds`.
+    ///
+    /// Layout note (#1700): this variant previously carried no payload (#3865). EdgeSeat is new and
+    /// has no production passes, so changing the payload (same discriminant index 4) does not
+    /// orphan any deployed account; no migration is required.
+    EdgeSeat(Vec<FeedSeat>),
```
martinsander00, line 67: **Build break — fix in this PR.** Turning `EdgeSeat` into a tuple variant breaks the fixture generator, which still constructs the bare unit variant at `sdk/serviceability/testdata/fixtures/generate-fixtures/src/main.rs:1199` (`accesspass_type: AccessPassType::EdgeSeat`). That crate is in the workspace `exclude` list, so `make rust-build` / `rust-test` skip it and CI stays green — but `make generate-fixtures` compiles it and hard-fails. One-line fix there: `AccessPassType::EdgeSeat(vec![])`. (Regenerating the fixture data itself stays deferred to Part 4 — only the compile fix belongs here.)

_(Commenting here since `main.rs` isn't part of this diff; this variant change is the root cause.)_

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_feeds.rs
```diff
@@ -0,0 +1,148 @@
...
+        assert_eq!(feed_account.owner, program_id, "Invalid Feed Account Owner");
+        let mut feed = Feed::try_from(*feed_account)?;
+
+        let current_users = prior_seats
+            .iter()
+            .find(|s| s.feed_key == seat.feed_key)
+            .map(|s| s.current_users)
+            .unwrap_or(0);
+
+        if !prior_seats.iter().any(|s| s.feed_key == seat.feed_key) {
```
martinsander00: **Duplicate `feed_key` double-counts.** `prior_seats` is snapshotted once before the loop (line 105) and never updated, so if `value.feeds` lists the same `feed_key` twice (and it wasn't already on the pass) this guard passes on *both* iterations: `reference_count` is bumped twice — the second `Feed::try_from` re-reads the just-written +1 — and `new_seats` ends up with two identical `FeedSeat` entries. Nothing dedups `value.feeds`. Since `reference_count` is never decremented, the over-count isn't reclaimable either. Recommend rejecting or de-duplicating repeated `feed_key`s (track seen keys in-loop, or revert on a dup) so a feed is bumped at most once and seats aren't duplicated.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_feeds.rs
```diff
@@ -0,0 +1,148 @@
...
+            assert!(feed_account.is_writable, "Feed Account is not writable");
+            feed.reference_count = feed
+                .reference_count
+                .checked_add(1)
+                .ok_or(DoubleZeroError::InvalidIndex)?;
+            try_acc_write(&feed, feed_account, payer_account, accounts)?;
+        }
+
+        new_seats.push(FeedSeat {
```
martinsander00, line 156: **`max_users` can be set below the preserved `current_users`.** `current_users` is preserved from the prior seat while `max_users` comes from the caller, with no `max_users >= current_users` check — so a re-provision can leave a seat over its cap. Latent today (nothing ticks `current_users` until Part 3), but Part 3's connect-time enforcement will inherit this unchecked invariant. Worth a clamp/guard when that lands — flagging here so it isn't lost.

martinsander00 (commented):
A few inline notes (F1/F2/F4 from the review). F1 breaks `make generate-fixtures` and should land in this PR; F2 is a real double-count bug; F4 is a latent note for Part 3. Otherwise this looks well-scoped and the design comments are great.

smartcontract/programs/doublezero-serviceability/src/state/accesspass.rs
nikw9944: Fixed — the generator now builds `EdgeSeat(vec![])` so `make generate-fixtures` compiles (fixture data regen stays in #3956).

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_feeds.rs
nikw9944: Now reject a repeated `feed_key` in-loop with `InvalidArgument` (no double-bump, no duplicate seats).

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_feeds.rs
nikw9944: Added a guard rejecting `max_users < current_users`.

martinsander00 (approved):

martinsander00:
can we add a similar test to test_feed_create_unauthorized_caller_rejected for setaccesspassfeeds?

martinsander00:
also can we add a test in set_feeds that exercises the > 64 rejection?

nikw9944:
Added both to `set_access_pass_feeds_test.rs`: `test_cannot_set_feeds_unauthorized_caller` (a non-foundation, non-`ACCESS_PASS_ADMIN` signer → `NotAllowed`) and `test_cannot_set_feeds_exceeds_max` (`MAX_ACCESS_PASS_FEEDS + 1` feeds → `InvalidArgument`, rejected by the cap check before any Feed account is read).

martinsander00 (approved):

nikw9944:
Dropped `Feed.reference_count` entirely (per discussion with @martinsander00). It only ever incremented — `SetAccessPassFeeds` doesn't see the feeds a re-provision drops, so it diverged upward and would eventually block `DeleteFeed` on feeds no longer in use. Delete is safe without it: a still-referenced feed_key that's deleted fails closed at connect (the metro gate can't load the deleted Feed), and the oracle keeps feeds/passes in sync. Removed the field + the delete guard + the increment; `SetAccessPassFeeds` now takes the Feed accounts read-only (owner/discriminator check only).

PR #3959, opened 2026-07-01
Client-Originated PIM Register beacon (implementation for RFC-22)

ben-malbeclabs:
# Client-Originated PIM Register beacon (implementation for RFC-22)

Implementation for **RFC-22** (PR #3951). On a dual-role (publisher+subscriber) multicast tunnel, the subscriber-side PIM neighbor suppresses the device's `pim border-router` source injection, so the device never originates the MSDP SA for the published source and it goes dark to the rest of the anycast-RP mesh. This change has `doublezerod` originate a periodic PIM **Register** (a "beacon") for its published sources, so the device — receiving a Register — originates the SA regardless of border-router suppression.

## Changes
- **`pim`**: PIM Register message serialization + a `RegisterSender` beacon (periodic, no inbound Register-Stop path; checksum over the first 8 bytes per RFC 7761 §4.9.1; egress pinned to the tunnel via `ControlMessage.IfIndex` so no `10.0.0.0` route is installed). The sender is restart-safe: teardown signals its goroutine rather than closing the channel, so a later reconnect resumes the beacon.
- **`api`**: `MulticastRpAddress` on `ProvisionRequest` — defaults to `10.0.0.0`, rejects a non-IPv4 value, and is compared in `Equal`/`Diff` so an RP change reconciles.
- **`services`**: start the beacon for publishers only.
- **`manager`/`runtime`**: wire it through; the reconciler validates the built request so the RP defaults (see the bug note below).
- **`controller`**: `permit pim any host 10.0.0.0` on publisher tunnels' `SEC-USER-PUB-MCAST-IN` ACL so the client Register reaches the RP; `pim ipv4 border-router` retained as a backstop.

## How it works
The client encapsulates its existing heartbeat into a PIM Register and unicasts it to the RP over the GRE tunnel. The device, as the anycast RP, learns the source via the Register (`C` flag), sets the "may notify MSDP" (`N`) flag, and originates the SA — the exact step border-router can't do when a PIM neighbor is present.

### Unit tests
All client suites (`pim`, `api`, `services`, `manager`) and the controller golden suite pass (run in a Linux container; the client is Linux-only).

### Automated e2e (runs in CI)
`TestE2E_MulticastPublisherRegisterLifecycle` connects a client publisher-only, adds the subscriber role in place (making the tunnel dual-role, which suppresses `pim border-router`), then disconnects and reconnects dual-role. It asserts the published `(S,G)` keeps and then regains the `C` (learned via a register) and `N` (may notify MSDP) flags across those transitions. That guards both the beacon working on a dual-role tunnel and the `RegisterSender` lifecycle: a teardown followed by a reconnect must not silently disable it.

### A bug the e2e caught (and this PR fixes)
End-to-end testing surfaced that the reconciler path (`buildProvisionRequest`) never called `Validate()`, so in production `MulticastRpAddress` was `nil` → the Register failed with "missing address" and was never sent (the unit tests validated manually and missed it). Fixed by validating the reconciler-built request, with a regression test (`TestReconcile_ProvisionMulticast_DefaultsRpAddress`) that drives the real reconcile path and asserts the RP reaching the register is `10.0.0.0`.

### End-to-end on real cEOS — the Register on the wire
Captured on the client underlay (`eth0`), **`eth0-gre-run2.pcap` frame 242** shows the complete protocol stack from RFC-22: a GRE-encapsulated PIM **Register** to the RP, encapsulating the original heartbeat datagram.

```
Frame 242: 98 bytes on wire
Ethernet II, Src: a6:d0:b4:f7:af:24, Dst: 56:c3:e2:c2:b2:ee
    Type: IPv4 (0x0800)
Internet Protocol Version 4, Src: 9.210.180.100, Dst: 9.210.180.8   ← GRE delivery (underlay)
    Protocol: Generic Routing Encapsulation (47)
Generic Routing Encapsulation (IP)
    Protocol Type: IP (0x0800)
Internet Protocol Version 4, Src: 169.254.0.1, Dst: 10.0.0.0         ← the Register, unicast to the RP
    Protocol: PIM (103)
Protocol Independent Multicast
    Type: Register (1)
    PIM Options
        Flags: 0x00000000                                           ← Border=0, Null=0 (data-bearing)
Internet Protocol Version 4, Src: 148.51.120.0, Dst: 233.84.178.0    ← encapsulated original datagram
    Protocol: UDP (17)
User Datagram Protocol, Src Port: 5765, Dst Port: 5765
    Length: 12
Data (4 bytes)
    Data: 445a0001                                                  ← the heartbeat ("DZ" + version)
```

Protocol stack: `eth:ethertype:ip:gre:ip:pim:ip:udp:data` — exactly the RFC-22 packet diagram.

### End-to-end on real cEOS: the MSDP SA flood to a second RP

A two-device run (dz1 + dz2, both anycast RP 10.0.0.0) proves the SA propagates end to end. dz1 originates the source from the client Register, then floods it to dz2.

dz1 (the RP that receives the Register), `show ip mroute`:

```
233.84.178.0
  0.0.0.0, RP 10.0.0.0, flags: W
    Incoming interface: Register
  148.51.120.0, flags: SNCP
    Incoming interface: Tunnel500
    RPF route: [U] 148.51.120.0/32 via 169.254.0.1
```

`C` = learned from a DR via a register, `N` = may notify MSDP. Before this change, with border-router suppressed on a dual-role tunnel, the entry never carried `N` and nothing was originated.

dz2 (the remote anycast RP), `show ip msdp sa-cache`:

```
MSDP Source Active Cache
(148.51.120.0, 233.84.178.0), RP 10.0.0.0, heard from 9.210.180.8
```

dz2's `show ip mroute` carries flags `MPE` (`M` = learned via MSDP) and the MSDP session shows SA Count 1. That is the full chain: client Register, dz1 sets `N` and originates the SA, MSDP floods it, dz2 installs it.

### cEOS constraints (checks that are real-hardware only)
- ACL binding to the tunnel is unsupported on cEOS. The `permit pim any host 10.0.0.0` entry renders correctly in the ACL, but `ip access-group SEC-USER-PUB-MCAST-IN in` on a tunnel returns "not supported on this hardware platform" in the emulator. That is exactly why the template guards the binding behind `NoHardware`. The per-entry hit-counter check for the Register permit can only run on real hardware.
- MSDP peering used the directly-connected addresses (9.210.180.8 and 9.210.180.9), not the loopbacks. cEOS in Docker does not carry loopback-sourced MSDP or BGP TCP between the two containers; only directly-connected traffic forwards. Production peers over loopbacks on a real data plane. The re-point is a devnet-only shim to exercise the flood and does not change the control-plane behavior under test.
- dz2's mroute incoming interface is Null, the documented cEOS multicast data-plane limitation (`e2e/docs/CEOS_MULTICAST_LIMITATION.md`). The SA flood proven here is control-plane and works regardless.

## Backward compatibility
Belt-and-suspenders: `pim ipv4 border-router` stays as a backstop, and the ACL permit is additive — every old/new client × old/new device combination is safe (see RFC-22 §Backward Compatibility). No smart-contract change. New optional `mcast_rp_address` provision field defaults to `10.0.0.0`.

### DIFF
--- a/client/doublezerod/internal/api/requests.go
+++ b/client/doublezerod/internal/api/requests.go
@@ -115,6 +115,9 @@ func (p *ProvisionRequest) Equal(other *ProvisionRequest) bool {
 	if !ipSlicesEqual(p.MulticastSubGroups, other.MulticastSubGroups) {
 		return false
 	}
+	if !p.MulticastRpAddress.Equal(other.MulticastRpAddress) {
+		return false
+	}
 	return ipNetSlicesEqual(p.DoubleZeroPrefixes, other.DoubleZeroPrefixes)
 }
 
@@ -158,6 +161,9 @@ func (p *ProvisionRequest) Diff(other *ProvisionRequest) string {
 	}
 	if !ipSlicesEqual(p.MulticastSubGroups, other.MulticastSubGroups) {
 		diffs = append(diffs, fmt.Sprintf("MulticastSubGroups: %v -> %v", p.MulticastSubGroups, other.MulticastSubGroups))
+	}
+	if !p.MulticastRpAddress.Equal(other.MulticastRpAddress) {
+		diffs = append(diffs, fmt.Sprintf("MulticastRpAddress: %s -> %s", p.MulticastRpAddress, other.MulticastRpAddress))
 	}
 	if !ipNetSlicesEqual(p.DoubleZeroPrefixes, other.DoubleZeroPrefixes) {
 		diffs = append(diffs, fmt.Sprintf("DoubleZeroPrefixes: count %d -> %d", len(p.DoubleZeroPrefixes), len(other.DoubleZeroPrefixes)))
@@ -276,6 +282,8 @@ func (p *ProvisionRequest) Validate() error {
 	}
 	if p.MulticastRpAddress == nil {
 		p.MulticastRpAddress = net.IPv4(10, 0, 0, 0)
+	} else if p.MulticastRpAddress.To4() == nil {
+		return fmt.Errorf("mcast_rp_address must be an IPv4 address, got %q", p.MulticastRpAddress)
 	}
 	return nil
 }
--- a/client/doublezerod/internal/api/requests.go
+++ b/client/doublezerod/internal/api/requests.go
@@ -84,6 +84,7 @@ type ProvisionRequest struct {
 	DoubleZeroPrefixes []*net.IPNet `json:"doublezero_prefixes"`
 	MulticastSubGroups []net.IP     `json:"mcast_sub_groups"`
 	MulticastPubGroups []net.IP     `json:"mcast_pub_groups"`
+	MulticastRpAddress net.IP       `json:"mcast_rp_address"`
 	BgpLocalAsn        uint32       `json:"bgp_local_asn"`
 	BgpRemoteAsn       uint32       `json:"bgp_remote_asn"`
 }
@@ -191,6 +192,9 @@ func (p *ProvisionRequest) InfraEqual(other *ProvisionRequest) bool {
 	if p.BgpLocalAsn != other.BgpLocalAsn || p.BgpRemoteAsn != other.BgpRemoteAsn {
 		return false
 	}
+	if !p.MulticastRpAddress.Equal(other.MulticastRpAddress) {
+		return false
+	}
 	return ipNetSlicesEqual(p.DoubleZeroPrefixes, other.DoubleZeroPrefixes)
 }
 
@@ -269,6 +273,9 @@ func (p *ProvisionRequest) Validate() error {
 	}
 	if p.BgpRemoteAsn == 0 {
 		p.BgpRemoteAsn = 65001
+	}
+	if p.MulticastRpAddress == nil {
+		p.MulticastRpAddress = net.IPv4(10, 0, 0, 0)
 	}
 	return nil
 }
--- a/client/doublezerod/internal/pim/register.go
+++ b/client/doublezerod/internal/pim/register.go
@@ -1,0 +1,145 @@
+package pim
+
+import (
+	"encoding/binary"
+	"fmt"
+	"log/slog"
+	"math/rand"
+	"net"
+	"sync"
+	"time"
+
+	"golang.org/x/net/ipv4"
+)
+
+// RegisterSender periodically sends PIM Register messages (a beacon) for the
+// publisher's groups to the RP, so the RP originates the source into MSDP.
+// It ignores Register-Stop: there is no inbound path.
+type RegisterSender struct {
+	done       chan struct{}
+	wg         *sync.WaitGroup
+	mu         sync.Mutex
+	conn       RawConner
+	iface      string
+	srcOverlay net.IP
+	innerSrc   net.IP
+	rp         net.IP
+	dport      int
+	payload    []byte
+	groups     []net.IP
+	updateCh   chan []net.IP
+}
+
+func NewRegisterSender() *RegisterSender {
+	return &RegisterSender{done: make(chan struct{}), updateCh: make(chan []net.IP)}
+}
+
+func (s *RegisterSender) Start(iface string, srcOverlay, innerSrc net.IP, groups []net.IP, rp net.IP, dport int, payload []byte, interval time.Duration) error {
+	c, err := net.ListenPacket("ip4:103", "0.0.0.0")
+	if err != nil {
+		return fmt.Errorf("register: failed to listen: %v", err)
+	}
+	r, err := ipv4.NewRawConn(c)
+	if err != nil {
+		return fmt.Errorf("register: failed to create raw conn: %v", err)
+	}
+	if err := r.SetControlMessage(ipv4.FlagInterface, true); err != nil {
+		return fmt.Errorf("register: failed to enable control message: %v", err)
+	}
+	intf, err := net.InterfaceByName(iface)
+	if err != nil {
+		return fmt.Errorf("register: failed to get interface: %v", err)
+	}
+	s.srcOverlay = srcOverlay
+	s.innerSrc = innerSrc
+	s.rp = rp
+	s.dport = dport
+	s.payload = payload
+	return s.startWithConn(r, intf, groups, interval)
+}
+
+func (s *RegisterSender) startWithConn(conn RawConner, intf *net.Interface, groups []net.IP, interval time.Duration) error {
+	s.conn = conn
+	s.iface = intf.Name
+	s.groups = groups
+	s.wg = &sync.WaitGroup{}
+	s.wg.Add(1)
+	go func() {
+		defer conn.Close()
+		defer s.wg.Done()
+
+		// Stagger the first send within the interval so publishers across the
+		// fleet do not synchronize into CoPP-stressing bursts.
+		jitter := time.Duration(rand.Int63n(int64(interval)))
+		select {
+		case <-time.After(jitter):
+		case <-s.done:
+			return
+		}
+
+		s.sendAll(intf)
+		ticker := time.NewTicker(interval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
+				s.sendAll(intf)
+			case g := <-s.updateCh:
+				s.mu.Lock()
+				s.groups = g
+				s.mu.Unlock()
+			case <-s.done:
+				return
+			}
+		}
+	}()
+	return nil
+}
+
+func (s *RegisterSender) sendAll(intf *net.Interface) {
+	s.mu.Lock()
+	groups := s.groups
+	s.mu.Unlock()
+	for _, g := range groups {
+		if err := s.sendRegister(intf, g); err != nil {
+			slog.Error("failed to send pim register", "group", g, "error", err)
+		}
+	}
+}
+
+func (s *RegisterSender) sendRegister(intf *net.Interface, group net.IP) error {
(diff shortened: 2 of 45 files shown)


client/doublezerod/internal/pim/register.go
```diff
@@ -0,0 +1,152 @@
...
+func (s *RegisterSender) UpdateGroups(groups []net.IP) error {
+	select {
+	case s.updateCh <- groups:
+	case <-s.done:
+	}
+	return nil
+}
+
+func (s *RegisterSender) Close() error {
+	s.closeOnce.Do(func() { close(s.done) })
```
packethog: should-fix: `Close()` permanently closes `done` via `closeOnce`, but this `RegisterSender` is a daemon-lifetime singleton (created once in `runtime.Run`, reused by every `CreateService`/`Setup`). After the first `Teardown()` (e.g. a `disconnect`, or a reconciler full reprovision), the next `Start()` spawns a goroutine that hits the already-closed `done` in the jitter `select` and returns immediately — so no Registers are ever sent again until the daemon restarts, silently disabling the feature. The sibling `PIMServer`/`HeartbeatSender` avoid this by signalling with `done <- struct{}{}` (leaving the channel reusable) rather than closing it. Signal instead of close, or recreate `done` (and reset `wg`) at the start of `Start()`.

client/doublezerod/internal/pim/register.go
```diff
@@ -0,0 +1,152 @@
...
+	s.wg.Add(1)
+	go func() {
+		defer conn.Close()
+		defer s.wg.Done()
+
+		// Stagger the first send within the interval so publishers across the
+		// fleet do not synchronize into CoPP-stressing bursts.
+		jitter := time.Duration(rand.Int63n(int64(interval)))
+		select {
+		case <-time.After(jitter):
```
packethog, line 77: should-fix: the initial jitter `select` only handles `time.After` and `done`; it does not drain `updateCh`. Since `updateCh` is unbuffered and `UpdateGroups` blocks on `s.updateCh <- groups`, a group change arriving during the jitter window (0–`interval`, up to 60s) stalls the caller until the jitter expires. `UpdateGroups` runs on the reconciler path (`reconcileService` → `GroupUpdater.UpdateGroups`), so this can block reconciliation for up to a minute. Buffer `updateCh`, or add an `updateCh` case to the jitter select.

client/doublezerod/internal/api/requests.go
```diff
@@ -191,6 +192,9 @@ func (p *ProvisionRequest) InfraEqual(other *ProvisionRequest) bool {
 	if p.BgpLocalAsn != other.BgpLocalAsn || p.BgpRemoteAsn != other.BgpRemoteAsn {
 		return false
 	}
+	if !p.MulticastRpAddress.Equal(other.MulticastRpAddress) {
```
packethog, line 201: The `MulticastRpAddress` comparison added here has no effect on reconciliation: `reconcileService` gates first on `currentPR.Equal(&pr)`, and `Equal` does not compare `MulticastRpAddress`. An RP-only change therefore short-circuits as "identical" before `InfraEqual` is consulted, so the register destination would never be updated. For consistency, also compare `MulticastRpAddress` in `Equal` (and report it in `Diff`).

client/doublezerod/internal/api/requests.go
```diff
@@ -270,6 +274,9 @@ func (p *ProvisionRequest) Validate() error {
 	if p.BgpRemoteAsn == 0 {
 		p.BgpRemoteAsn = 65001
 	}
+	if p.MulticastRpAddress == nil {
```
packethog, line 283: nit: `Validate()` defaults a nil RP to `10.0.0.0` but doesn't constrain a caller-supplied `mcast_rp_address`. A non-IPv4 value would make `innerSrc.To4()`/`s.rp` produce a malformed Register, and any RP other than `10.0.0.0` won't be permitted by the controller ACL (`permit pim any host 10.0.0.0`, hardcoded in `tunnel.tmpl`). Consider rejecting non-IPv4 RP values here, or keying the ACL off the same RP so the two can't drift.

packethog (changes requested):
Requesting changes — see the inline comments. The main issue is a lifecycle bug that silently disables the Register beacon after the first teardown/reconnect; the others are supporting correctness and consistency findings.

client/doublezerod/internal/pim/register.go
ben-malbeclabs: Good catch, this was a real silent-disable. Fixed by mirroring `HeartbeatSender`: `Close()` now signals `done <- struct{}{}` and resets `wg` instead of `closeOnce`+`close`, so a later `Start()` restarts the goroutine cleanly. Added `TestRegisterSenderRestartsAfterClose` (Start -> Close -> Start -> assert Registers resume), run under `-race`, which fails without the fix.

client/doublezerod/internal/pim/register.go
ben-malbeclabs: Fixed. Dropped the `updateCh` channel entirely and now apply the new groups under the mutex, matching `HeartbeatSender.UpdateGroups`, so the call never blocks regardless of where the sender goroutine is. Added `TestRegisterSenderUpdateGroupsDoesNotBlock` (long interval so the goroutine sits in the jitter window; asserts `UpdateGroups` returns promptly and the groups are applied).

client/doublezerod/internal/api/requests.go
ben-malbeclabs: You're right, it was dead code. `reconcileService` gates on `Equal` first (`manager.go:570`) and returns early, so an RP-only change never reached `InfraEqual`. Now compared in `Equal` and reported in `Diff` as well. Added `TestProvisionRequestEqualDetectsRpChange`. Low real-world impact today since the RP is always `10.0.0.0`, but the check is now consistent.

client/doublezerod/internal/api/requests.go
ben-malbeclabs: `Validate()` now rejects a non-IPv4 `mcast_rp_address` (which would otherwise produce a malformed Register). Test `TestProvisionRequestValidateRejectsNonIPv4Rp`. I left the value otherwise flexible rather than hard-pinning it to `10.0.0.0`, since the ACL coupling is intentional for now; happy to also reject non-`10.0.0.0` if you'd prefer the two can't drift.

ben-malbeclabs:
> Requesting changes — see the inline comments. The main issue is a lifecycle bug that silently disables the Register beacon after the first teardown/reconnect; the others are supporting correctness and consistency findings.

Thanks @packethog.  I have tried to address each comment, re-requesting another review when you have a min please.

client/doublezerod/internal/services/multicast.go
```diff
@@ -92,6 +95,9 @@ func (s *MulticastService) Setup(p *api.ProvisionRequest) error {
 		if err := s.heartbeat.Start(tun.Name, p.DoubleZeroIP, p.MulticastPubGroups, multicast.DefaultHeartbeatTTL, multicast.DefaultHeartbeatInterval); err != nil {
 			return fmt.Errorf("error starting heartbeat sender: %v", err)
 		}
+		if err := s.register.Start(tun.Name, s.Tunnel.LocalOverlay, p.DoubleZeroIP, p.MulticastPubGroups, p.MulticastRpAddress, multicast.HeartbeatPort, multicast.HeartbeatPayload, pim.DefaultRegisterInterval); err != nil {
```
packethog, line 98: non-blocking (pre-existing): on a `Setup` error after this point (e.g. `register.Start` failing, or `bgp.AddPeer` returning a non-exists error), `provisionLocked` discards the service without calling `Teardown`, so already-started senders (heartbeat, now register, and the subscriber `pim` FSM) leak their goroutine + raw socket. This mirrors the existing heartbeat/pim ordering rather than being introduced here, but adding the register sender widens the window. Worth a follow-up to unwind started senders on a failed `Setup`, especially since the reconciler retries provisioning on failure.

controlplane/controller/internal/controller/templates/tunnel.tmpl
```diff
@@ -450,6 +450,7 @@ ip access-list SEC-USER-PUB-MCAST-IN
    permit tcp any any eq bgp
    permit ip any 224.0.0.13/32
    permit ip any {{ .MulticastGroupBlock }}
+   permit pim any host 10.0.0.0
```
packethog, line 453: non-blocking: the RP is now a validated but caller-settable field (`mcast_rp_address`), while this ACL hardcodes `host 10.0.0.0`. The reconciler always defaults the RP to `10.0.0.0`, so they only diverge if a custom RP is set via the direct provision API — in which case the client would beacon to an RP the device ACL drops. Consider keying this permit off the same RP value (or documenting that the RP is fixed at `10.0.0.0`) so the two can't drift.

packethog (approved):
The previously requested changes are addressed and verified:

- `RegisterSender.Close` now signals `done` (instead of closing it) and resets `wg`, so the daemon-lifetime singleton restarts correctly after teardown/reconnect — covered by `TestRegisterSenderRestartsAfterClose`.
- `UpdateGroups` is now mutex-only (no channel handoff), so it can't stall the reconciler during startup jitter — covered by `TestRegisterSenderUpdateGroupsDoesNotBlock`.
- `Validate` rejects non-IPv4 RP addresses, and `Equal`/`Diff`/`InfraEqual` all compare `MulticastRpAddress` consistently.

Build, `go vet`, the `pim` suite under `-race`, and full CI are green. Two minor, non-blocking notes inline — neither needs to hold this PR.

ben-malbeclabs:
@packethog added e2e test, please confirm that your approval still stands. :)

packethog:
LGTM

PR #3964, opened 2026-07-01
client: add multicast subscriptions to doublezero status output

juan-malbeclabs:
## Summary of Changes
- Add a structured `subscriptions` array to the `doublezero status` (`/v2/status`) output, placed right after `multicast_groups`, giving programmatic consumers per-group detail instead of the flattened `"P:code,S:code"` string.
- Each entry describes one multicast group the user participates in (deduplicated across roles): group `pubkey`, `code`, `multicast_ip`, `max_bandwidth`, and `publisher`/`subscriber` booleans. A user who is both publisher and subscriber of the same group appears once with both booleans set.
- The daemon (`doublezerod`) builds the list in `enrichStatuses()`, reusing the already-resolved onchain multicast groups; the existing `multicast_groups` string is left unchanged for backward compatibility (table view and current consumers).

## Diff Breakdown
| Core logic   |     3 | +123 / -2    | +121 |
| Tests        |     1 | +46 / -10    |  +36 |
| **Total**    |     4 | +169 / -12   | +157 |

Mostly a small, additive core change across the daemon and CLI, with the remainder in test coverage (including inline tests within `status.rs`).

<details>

- `client/doublezero/src/command/status.rs` — add `subscriptions` field to the status response (skipped in the table, serialized in JSON); populate it and add coverage.
- `client/doublezerod/internal/manager/http.go` — add `Subscription` struct and `Subscriptions` field; build the deduplicated per-group list in `enrichStatuses()`.
- `client/doublezerod/internal/manager/reconciler_test.go` — extend `TestServeV2Status_Enrichment` with subscription expectations and a publisher+subscriber dual-role case.
- `client/doublezero/src/servicecontroller.rs` — mirror `Subscription` struct and add `subscriptions` to `V2ServiceStatus` for deserialization.

</details>

## Testing Verification
- `cargo test -p doublezero status` — 17/17 pass, including JSON-contract assertions that `subscriptions` serializes as an array of objects after `multicast_groups`.
- Extended `TestServeV2Status_Enrichment` covers the dual-role case (single entry, both booleans `true`); the subscriber-role subtests require `CAP_NET_RAW` (PIM socket) and run under `sudo`/CI, matching the pre-existing `multicast_subscriber` subtest.

### DIFF
--- a/client/doublezero/src/command/status.rs
+++ b/client/doublezero/src/command/status.rs
@@ -783,16 +783,27 @@ mod tests {
         assert_eq!(result[0].multicast_groups, "P:solana-lv,S:solana-ams");
 
         // The structured subscriptions list is carried through verbatim.
-        assert_eq!(result[0].subscriptions.len(), 2);
-        assert_eq!(result[0].subscriptions[0].code, "solana-lv");
-        assert_eq!(result[0].subscriptions[0].multicast_ip, "233.84.178.1");
-        assert_eq!(result[0].subscriptions[0].pubkey, "pubLV");
-        assert_eq!(result[0].subscriptions[0].max_bandwidth, 1_000_000_000);
-        assert!(result[0].subscriptions[0].publisher);
-        assert!(!result[0].subscriptions[0].subscriber);
-        assert_eq!(result[0].subscriptions[1].code, "solana-ams");
-        assert!(!result[0].subscriptions[1].publisher);
-        assert!(result[0].subscriptions[1].subscriber);
+        assert_eq!(
+            result[0].subscriptions,
+            vec![
+                Subscription {
+                    pubkey: "pubLV".to_string(),
+                    code: "solana-lv".to_string(),
+                    multicast_ip: "233.84.178.1".to_string(),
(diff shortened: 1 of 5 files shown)


client/doublezero/src/command/status.rs
```diff
@@ -734,6 +781,18 @@ mod tests {
         let result = result.unwrap();
         assert_eq!(result.len(), 1);
         assert_eq!(result[0].multicast_groups, "P:solana-lv,S:solana-ams");
+
```
martinsander00, line 784: Subscription already derives PartialEq so can we assert_eq! the whole struct here instead of checking each field one by one? then if we add a field to Subscription later this test forces an update instead of silently passing. (and maybe derive Eq too while we're at it)

juan-malbeclabs: Good call. Derived Eq on Subscription and switched the test to a single assert_eq! against the full Vec<Subscription>, so adding a field now forces the test to be updated instead of passing silently.

martinsander00 (approved):

PR #3955, opened 2026-06-30
serviceability: enforce EdgeSeat feed metro gate at connect

nikw9944:
Enforce the EdgeSeat feed metro gate at multicast connect: the feed passed at connect must be provisioned on the pass and must serve the device's metro (exchange) — otherwise rejected — and the target multicast group must be in that feed's group set. The matching feed's per-feed seat is ticked (superseding the legacy `max_multicast_users` cap).

Part of malbeclabs/infra#1700.

**Stack (merge bottom-up):**
1. #3953 — Feed catalog account + CRUD
2. #3954 — EdgeSeat data model + `SetAccessPassFeeds`
3. #3955 — metro gate enforcement at connect
4. #3956 — Go/Python/TS SDK read support

PR #3966, opened 2026-07-01
smartcontract: enforce Permission flags on multicast subscribe and create-user handlers

bgm-malbeclabs:
## Summary of Changes

* Completes the least-privilege (RFC-2) Permission-account migration for the two remaining serviceability handlers the feed oracle calls, so the oracle can operate on an `ACCESS_PASS_ADMIN | USER_ADMIN` Permission account instead of `foundation_allowlist` membership — least privilege for a hot, always-online key.
* `UpdateMulticastGroupRoles`: granting roles on another owner's pass is now an `ACCESS_PASS_ADMIN` operation, while removal-only cleanup stays `USER_ADMIN` (unchanged from #3206). Splitting by operation preserves the deliberate "USER_ADMIN is removal-only here" contract and the exact historical `NotAllowed` / `Unauthorized` error variants.
* `CreateSubscribeUser` owner-override now accepts a `USER_ADMIN` Permission holder in addition to the sentinel and foundation. The Permission account is threaded through `CreateUserCoreAccounts` (`None` for plain `CreateUser`, which never overrides the owner).
* Both handlers disambiguate the optional trailing Permission account from the EdgeSeat feed/device accounts via `split_trailing_permission` (PDA match, not position), so Permission and Feed coexist unambiguously. This also fixes a latent ordering bug in `create_subscribe`'s EdgeSeat feed wiring: the SDK's `assemble_instructions` always appends `[payer, system, permission?]` last, so the feed sits *before* payer/system, not after — only latent today because `feed_pk` is `None` everywhere.
* The Permission path carries **no owns-it restriction**, so a flag holder acts across owners — required for the validator-owned users and passes the oracle manages. The change is additive: every existing caller keeps its exact current authority.
* `permission/audit.rs`: corrects the `NON_MIGRATED_SUBSYSTEMS` text — the user-create owner override now honors `USER_ADMIN` via `authorize()` (foundation via the legacy mapping, Permission holders directly), leaving only the sentinel *direct* check un-migrated.

Context: RFC-2 (malbeclabs/infra#1764), malbeclabs/doublezero-shreds#517. The subscriber-allowlist handlers (`Add`/`Remove MulticastGroupSubAllowlist`) were migrated to `authorize()` upstream in #3982 while this branch was in review, so that half is dropped here; this PR completes the set for the subscribe/create-user surface, following the established `authorize()` + `split_trailing_permission` pattern with no new authorization helper.

## Diff Breakdown

| Tests        |     4 | +550 / -3    | +547 |
| Docs         |     2 | +26 / -2     |  +24 |
| Core logic   |     3 | +87 / -36    |  +51 |
| Scaffolding  |     2 | +7 / -2      |   +5 |
| **Total**    |    11 | +670 / -43   | +627 |

Overwhelmingly tests: the authorization logic itself is a small, additive change to three processors, with the bulk of the diff proving the new cross-owner Permission paths.

<details>

- [`smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs`](https://github.com/malbeclabs/doublezero/pull/3966/files#diff-1d7d0a23bce8c73543d89242acceeb43eee45982378606a03657115f59e9fd40) — `UpdateMulticastGroupRoles` grants via `ACCESS_PASS_ADMIN` / cleanup via `USER_ADMIN`; peels the trailing Permission by PDA match, deriving device/feed from the leading list
- [`smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs`](https://github.com/malbeclabs/doublezero/pull/3966/files#diff-b825218e6fdd5871d82b5623d95b74b2460b480dd4c0e3f2cb150b98623ff342) — owner-override authorizes a `USER_ADMIN` Permission holder alongside the sentinel
- [`smartcontract/programs/doublezero-serviceability/src/processors/user/create_subscribe.rs`](https://github.com/malbeclabs/doublezero/pull/3966/files#diff-cddaff613b96778de97f99ba3220571784cbb1d1d3533c475f24ad47af738624) — parses `[feed?, payer, system, permission?]` via `split_trailing_permission`, threading both feed and Permission into `create_user_core`
- [`smartcontract/cli/src/permission/audit.rs`](https://github.com/malbeclabs/doublezero/pull/3966/files#diff-f534db970f2172a858f24b4717de2a7a99d4455b77afcfc29155e33e87d2ba30) — updates `NON_MIGRATED_SUBSYSTEMS` so `doublezero permission audit` no longer over-reports the owner override
- [`smartcontract/programs/doublezero-serviceability/src/processors/user/create.rs`](https://github.com/malbeclabs/doublezero/pull/3966/files#diff-50a1438f8af41564d2206a2bc66f4efbd8d410b02b1273d63634342dc8ba321e) — passes `permission_account: None` (plain `CreateUser` never overrides the owner)

</details>

## Testing Verification

* Added integration tests exercising the new Permission path for each wired handler, all proving the cross-owner (no owns-it) property:
  * `test_subscribe_access_pass_admin_permission_allowed` — an `ACCESS_PASS_ADMIN` holder (not owner/foundation/sentinel/feed) subscribes a user cross-owner.
  * `test_create_subscribe_user_user_admin_owner_override` — a `USER_ADMIN` holder creates a subscribe-user with `owner != payer`.
  * Preserved the #3206 contract tests: `test_subscribe_user_admin_permission_rejected` (USER_ADMIN still cannot add roles) and `test_unsubscribe_user_admin_permission_allowed` (USER_ADMIN removal cleanup).
* Retained ACCESS_PASS_ADMIN coverage for the subscriber-allowlist path (`test_multicast_subscriber_allowlist_access_pass_admin_permission`) — the oracle's exact path, which #3982 migrated without adding a test.
* Updated `feed_metro_gate_test` to the SDK-consistent account order (feed before payer/system); all 4 EdgeSeat metro-gate cases pass, confirming feed + the new Permission parsing coexist.
* Note: `test_set_accesspass_refills_depleted_user_payer` is a pre-existing failure on `main` (airdrop-refill balance assertion, unrelated to authorization) — confirmed it fails identically on a clean `origin/main` checkout.


juan-malbeclabs (commented):
.

smartcontract/programs/doublezero-serviceability/src/processors/user/create_subscribe.rs
```diff
@@ -93,6 +93,12 @@ pub fn process_create_subscribe_user(
     let payer_account = next_account_info(accounts_iter)?;
     let system_program = next_account_info(accounts_iter)?;
 
+    // Optional trailing Permission PDA for the payer, appended by the SDK when one exists on
+    // chain. dz_prefix_count fixes the account count deterministically, so anything after
+    // system_program is the Permission account. It authorizes a USER_ADMIN holder to set a custom
+    // owner (owner-override) inside create_user_core.
+    let permission_account = next_account_info(accounts_iter).ok();
```
juan-malbeclabs: 🔴 **High — collides with the EdgeSeat metro gate after rebase.**

This line reads the optional Permission PDA at the position immediately after `system_program`. On `main` (post #3955), `create_subscribe.rs` reads at that **same position**:

```rust
let feed_account = accounts_iter.next();   // EdgeSeat metro gate
```

After rebasing, a single trailing account cannot be both the payer's Permission PDA and the Feed. Worse, the scenario that motivates this PR — the oracle provisioning an EdgeSeat validator user — needs **both**: the Permission PDA (to authorize the owner-override) and the Feed (for the metro gate). Positional parsing with a single `.next()` cannot express that.

**Proposed fix:** use the same PDA-match disambiguation you already introduced in `add.rs` (`split_trailing_permission`), which tells the Permission apart by its PDA rather than by position. That lets Permission and Feed coexist unambiguously. Rebase onto `main` and re-verify the end-to-end flow (oracle + EdgeSeat) before merging.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
```diff
@@ -218,23 +218,48 @@ pub fn process_update_multicastgroup_roles(
...
         } else {
-            msg!(
-                "AccessPass user_payer {:?} does not match payer {:?}",
-                accesspass.user_payer,
-                payer_account.key
-            );
-            return Err(DoubleZeroError::Unauthorized.into());
+            permission_flags::ACCESS_PASS_ADMIN
+        };
+        let authorized = authorize(
```
juan-malbeclabs, line 255: 🔴 **High — trailing-account ordering vs EdgeSeat.**

This PR broadens `authorize()` so it runs (and consumes the Permission account) in the *grant* path too. Directionally this is compatible with `main`, which reads `device_account`/`feed_account` via `accounts_iter.next()` **after** the `authorize()` call (with a comment that they must come after so they don't consume the trailing Permission account).

But that only holds if the on-wire order is `[payer, system, permission, device, feed]`. The SDK auto-injects the Permission PDA as the **last** account (permission-SDK auto-injection contract), which would place it *after* `device`/`feed` and break the ordering assumption.

**Proposed fix:** after rebasing onto `main`, verify end-to-end the actual order the SDK/CLI produces for the oracle+EdgeSeat case, and disambiguate the Permission by PDA (like `add.rs`) instead of relying on positional order. Add a test that passes Permission + device + feed together.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/allowlist/subscriber/add.rs
```diff
@@ -52,17 +54,17 @@ pub fn process_add_multicastgroup_sub_allowlist(
...
+    // Trailing layout: [user_payer?, payer, system, permission?]. The optional user_payer
+    // (connect-credit funding, passed by clients that fund the user; older clients omit it and
+    // skip the airdrop) and the optional trailing Permission account both sit at the tail.
+    // split_trailing_permission peels payer/system off the end and identifies the Permission
+    // account by matching the payer's Permission PDA, so a bare length check cannot confuse the
+    // two. Whatever precedes payer/system is the caller's leading list — the single optional
+    // user_payer account here.
+    let all_remaining: Vec<&AccountInfo> = accounts_iter.collect();
+    let (payer_account, system_program, leading, permission_account) =
+        split_trailing_permission(program_id, &all_remaining)?;
```
juan-malbeclabs: 🟢 This is the robust, correct pattern: `split_trailing_permission` disambiguates by PDA match rather than by position, so it supports `[payer,system]`, `[user_payer,payer,system]`, `[payer,system,perm]` and `[user_payer,payer,system,perm]`. I verified all four cases and they are not spoofable (the system program can never equal a payer's PDA). `add.rs` also has no EdgeSeat trailing accounts, so there is no collision here.

This is exactly the approach worth porting to `create_subscribe.rs` and `subscribe.rs` (see comments above).

smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs
```diff
@@ -100,16 +106,25 @@ pub fn create_user_core(
...
+    // This is what lets the oracle provision validator-owned subscribe-users (owner = validator)
+    // via its USER_ADMIN grant, without foundation membership.
     let effective_owner = match owner_override {
         Some(pk) if pk != Pubkey::default() => {
-            let is_foundation = globalstate
-                .foundation_allowlist
-                .contains(core.payer_account.key);
             let is_sentinel = globalstate.sentinel_authority_pk == *core.payer_account.key;
-            if !is_foundation && !is_sentinel {
+            let has_user_admin = authorize(
```
juan-malbeclabs, line 124: 🟡 **Medium — `NON_MIGRATED_SUBSYSTEMS` is now stale** (commenting here because `audit.rs` is not in the diff).

The CLAUDE.md / Permissions rule asks to revisit `NON_MIGRATED_SUBSYSTEMS` in `smartcontract/cli/src/permission/audit.rs` when migrating an instruction to `authorize()`, and it names "the user-create owner override" explicitly. This PR migrates that override to an `authorize(USER_ADMIN)` path, but the const still says:

```rust
"user create (custom owner override requires foundation_allowlist or sentinel_authority; ...)"
```

That is no longer accurate: the override no longer rests solely on a direct GlobalState check — a `USER_ADMIN` holder now qualifies, and `foundation_allowlist` is covered via the `USER_ADMIN` legacy mapping in `check_legacy_any`. Only the **sentinel direct check** and the seat-limit bypass (qa/foundation) remain genuinely non-migrated.

**Proposed fix:** update the `NON_MIGRATED_SUBSYSTEMS` text so `doublezero permission audit` doesn't over-report. (`AUTHORIZE_GATED_FLAGS` already lists `USER_ADMIN`/`ACCESS_PASS_ADMIN` since #3206, so no change is needed there.)

smartcontract/programs/doublezero-serviceability/tests/multicastgroup_allowlist_subcriber_test.rs
```diff
@@ -503,6 +504,163 @@ async fn test_multicast_subscriber_allowlist_sentinel_authority() {
     );
 }
 
+/// An ACCESS_PASS_ADMIN Permission holder who is neither the multicast-group owner, the
+/// access-pass owner, nor a foundation/sentinel/feed authority CAN add and remove a subscriber
+/// allowlist entry by appending their Permission PDA as the trailing account. The Permission path
+/// has no owns-it restriction, so it works across owners — the mechanism the oracle relies on to
+/// reconcile allowlists on passes it does not own.
+#[tokio::test]
+async fn test_multicast_subscriber_allowlist_access_pass_admin_permission() {
```
juan-malbeclabs, line 513: 🟢 Nit (coverage): this test doesn't pass `user_payer`, so it exercises the `[payer, system, permission]` (n=3) branch of `split_trailing_permission` but never the `[user_payer, payer, system, permission]` (n=4) branch — exactly the funded-user + Permission combination the oracle will use. Worth adding a case that passes both to lock down that branch.

juan-malbeclabs (changes requested):
The authorization logic itself is solid: the migration is faithful to the `authorize()` pattern, the error-variant preservation (`NotAllowed` vs `Unauthorized`) in `UpdateMulticastGroupRoles` is careful, and `split_trailing_permission` correctly disambiguates all four possible account layouts. The tests prove the cross-owner property and preserve the #3206 contracts.

The blocker is integration, not logic. The branch is based on a merge-base that predates EdgeSeat (`#3955` / `0e488f63`, already on `main`) — I confirmed the branch does not contain those commits. Two of the migrated handlers parse their trailing accounts **positionally**, and that collides with the EdgeSeat metro-gate trailing accounts once rebased. The merge-base diff hides this because they are additions in adjacent regions that git merges textually while producing incorrect semantics.

Each point is detailed inline with a proposed fix. Summary:

- 🔴 **High** — trailing-account collision with EdgeSeat in `create_subscribe.rs` and `subscribe.rs` after rebase.
- 🟡 **Medium** — `NON_MIGRATED_SUBSYSTEMS` in `smartcontract/cli/src/permission/audit.rs` is now stale (CLAUDE.md / Permissions rule). The file is not in the diff, so I commented on `create_core.rs`.
- 🟢 Nit — missing coverage for the n=4 branch of `split_trailing_permission` (user_payer + permission together).

Recommendation: rebase onto `main` and re-verify end-to-end before merging; port the two handlers to the PDA-match approach already used in `add.rs`. The green suite reported in the PR was necessarily run against the stale base and does not cover this interaction.

smartcontract/programs/doublezero-serviceability/src/processors/user/create_subscribe.rs
bgm-malbeclabs: Fixed in 247ea4a63. Ported to `split_trailing_permission`, so the tail parses as `[feed?, payer, system, permission?]` and Feed + Permission coexist, disambiguated by PDA match. Confirmed the collision was real end-to-end: `authorize()` hard-errors with `InvalidArgument` when the trailing account it reads isn't the payer's Permission PDA (authorize.rs), so the merged positional parse would have mis-read `feed` as the Permission.

While verifying, this also surfaced a latent ordering bug in the EdgeSeat feed wiring here: the processor read `feed` *after* payer/system, but the SDK's `assemble_instructions` always appends `[payer, system, permission?]` last — so any command account (feed) sits *before* payer/system. It was only latent because `feed_pk` is `None` everywhere today. `feed_metro_gate_test` is updated to the SDK-consistent order and passes.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
bgm-malbeclabs: Hardened in 247ea4a63 with the same `split_trailing_permission` approach — the Permission is peeled off the end by PDA match, and device/feed come from the `leading` list.

One nuance from verifying end-to-end: this wasn't actually a live bug. The subscribe SDK command deliberately does **not** send device/feed (its own comment: *"leaves no slot for them ... deferred to the oracle lifecycle"*), and `execute_authorized_transaction` appends the Permission last — so with no device/feed in between, the old `authorize(accounts_iter)` read the Permission correctly. It's a latent fragility for the future oracle re-gating path rather than a current break. Hardened anyway so that path is safe by construction.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/allowlist/subscriber/add.rs
bgm-malbeclabs: Agreed. Since #3982 landed the same `split_trailing_permission` migration for the subscriber allowlist `add`/`remove` on `main`, my version here was redundant — I dropped it and took main's during the rebase. The PR and CHANGELOG entry are narrowed to `UpdateMulticastGroupRoles` + the `CreateSubscribeUser` owner-override. (I kept one `ACCESS_PASS_ADMIN` integration test for the allowlist path since #3982 migrated the code but added no test there, and it's the oracle's exact path.)

smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs
bgm-malbeclabs: Fixed in 247ea4a63 (`smartcontract/cli/src/permission/audit.rs`). Verified your read of the mapping first: `check_legacy_any` routes `USER_ADMIN` → `foundation_allowlist`, and `legacy_keys_for_flags` lists `USER_ADMIN` in the foundation-fallback set, so foundation membership is covered via the legacy mapping and a `USER_ADMIN` Permission holder qualifies directly. The `NON_MIGRATED_SUBSYSTEMS` text now says the owner override honors `USER_ADMIN` via `authorize()`, leaving only the sentinel *direct* check un-migrated (the qa/foundation seat-limit + device-status bypass still noted). The 27 CLI permission tests pass.

smartcontract/programs/doublezero-serviceability/tests/multicastgroup_allowlist_subcriber_test.rs
bgm-malbeclabs: Good catch on the intent — the funded-user + Permission (`n=4`) combination is exactly what the oracle uses. Turns out the branch is already locked down at the unit level: `test_split_trailing_permission_tail5_tenant_pair_plus_permission` (landed with the helper in #3982) exercises `[leading.., payer, system, permission]` — the "highest-risk" leading + Permission case — plus a `tail4` no-permission variant. Both `create_subscribe` and `subscribe` now route through that same helper, and `feed_metro_gate_test` exercises the `leading=[feed]` case. So rather than duplicate coverage in the allowlist integration test, I'm relying on the helper's unit tests. Happy to add the integration-level `n=4` case too if you'd prefer it explicit here.

juan-malbeclabs (approved):
All review points addressed in 247ea4a6 and verified in the tree:

- 🔴 create_subscribe.rs and subscribe.rs: ported to `split_trailing_permission`; the Permission is disambiguated by PDA and Feed/device coexist without collision.
- 🟡 audit.rs: `NON_MIGRATED_SUBSYSTEMS` updated — the owner override now honors USER_ADMIN via authorize(), leaving only the direct sentinel_authority check un-migrated.
- 🟢 add.rs: took main's version (#3982) during the rebase.
- 🟢 n=4 coverage: accepted via the helper's unit tests.

LGTM.

PR #3975, opened 2026-07-04
permission: restrict FOUNDATION grant and add pre-activation audit

juan-malbeclabs:
## Summary of Changes

- Fix a privilege-escalation gap in the permission model: a plain `PERMISSION_ADMIN` holder could grant the `FOUNDATION` flag (bit 0) to anyone, including themselves. Granting/adding `FOUNDATION` now requires the caller to be a `foundation_allowlist` member or an existing `FOUNDATION` holder (new `authorize::can_grant_foundation`, wired into permission `create`/`update`). The check runs independently of `RequirePermissionAccounts` so foundation members remain authoritative and cannot be locked out.
- Add `doublezero permission audit` — a pre-activation gate that reports which legacy keys would lose access to migrated instructions before `require-permission-accounts` is enabled (coverage gaps, non-zero exit for runbook/CI use), plus super-admin (`PERMISSION_ADMIN`/`FOUNDATION`) holders and the non-migrated subsystems that still depend on the GlobalState allowlists (so operators know not to remove those keys yet).

## Diff Breakdown

| Core logic  |     4 | +676 / -4   | +672 |
| Scaffolding |     3 | +17 / -1    |  +16 |
| Tests       |     1 | +214 / -0   | +214 |
| Docs        |     1 | +4 / -0     |   +4 |
| **Total**   |     9 | +911 / -5   | +906 |

Mostly new code: the audit command and the FOUNDATION-grant guard, with substantial test coverage (in-file unit tests plus a dedicated integration suite). The core-logic figure includes co-located `#[cfg(test)]` unit tests.

<details>

- `smartcontract/cli/src/permission/audit.rs` — new `permission audit` command: legacy→Permission coverage-gap detection (mirrors `check_legacy_any`), super-admin reporting, non-migrated legacy-dependency warning; +6 unit tests
- `smartcontract/programs/doublezero-serviceability/tests/permission_test.rs` — integration tests for the FOUNDATION-grant restriction (deny on create/update, positive path for a FOUNDATION holder)
- `smartcontract/programs/doublezero-serviceability/src/authorize.rs` — new `can_grant_foundation` helper + 5 unit tests
- `smartcontract/programs/doublezero-serviceability/src/processors/permission/{create,update}.rs` — gate the FOUNDATION grant/add behind `can_grant_foundation`

</details>

## Testing Verification

- `permission audit` logic covered by 6 CLI unit tests (gap detection with the foundation-recovery carve-out for `PERMISSION_ADMIN`, suspended accounts not counting as coverage, super-admin holder reporting, non-zero exit on gaps).
- FOUNDATION-grant restriction covered by 5 `authorize` unit tests (allowlist member, FOUNDATION-flag holder, `PERMISSION_ADMIN`-only denied, suspended holder denied, no-account denied) and 3 integration tests: a `PERMISSION_ADMIN`-only caller is denied granting FOUNDATION on both create and update (with a passing control for non-FOUNDATION flags), while a FOUNDATION-flag holder succeeds.
- Verified no lockout regression: existing tests where a foundation payer grants FOUNDATION still pass (full `permission_test` suite: 17/17).


elitegreg (changes requested):
The FOUNDATION-grant guard (Part 1) looks solid — re-deriving the PDA from `payer_key` prevents passing someone else's FOUNDATION account, the foundation-allowlist short-circuit avoids lockout, and the iterator handling in create/update is behaviorally equivalent to the prior `authorize()` call. No concerns there.

**Blocking issue on `permission audit`: `MIGRATED_FLAGS` / `NON_MIGRATED_SUBSYSTEMS` are already stale on this branch, and they fail in the unsafe direction.**

The audit's correctness rests on those two lists mirroring which instructions actually route through `authorize()`. Grepping the processors on this branch (merge-base `1eed6441`, which already includes #3980 and #3984), `authorize()` is called with several flags the audit omits:

| Flag | Example call site | In `MIGRATED_FLAGS`? |
|---|---|---|
| `NETWORK_ADMIN` (15 sites) | `device/create`, `link/create`, `device/interface/create` | ❌ |
| `INFRA_ADMIN` (11) | `exchange/create`, `location/create` | ❌ |
| `TENANT_ADMIN` (6) | `tenant/create` | ❌ |
| `MULTICAST_ADMIN` (7) | `multicastgroup/create` | ❌ |
| `CONTRIBUTOR_ADMIN` (5) | `contributor/create` | ❌ |
| `GLOBALSTATE_ADMIN` (9) | `globalstate/*` | ❌ |

`NON_MIGRATED_SUBSYSTEMS` also still lists device, link, exchange, location, contributor, tenant, multicastgroup, and user as "NOT migrated … a Permission account grants nothing for them" — which is no longer true on this branch, so the printed warning actively misleads.

Impact: because these flags are omitted, the audit will not report a coverage gap for a foundation/authority key that today authorizes device/link/exchange/tenant/multicast administration but has no equivalent Permission account. An operator runs `permission audit`, sees "no coverage gaps," enables `require-permission-accounts`, and locks those subsystems out — the exact outage this command exists to prevent. It's a silent failure in the dangerous direction.

Requested changes:

1. Expand `MIGRATED_FLAGS` and `legacy_keys_for_flag` to cover every flag routed through `authorize()` (at minimum `NETWORK_ADMIN`, `INFRA_ADMIN`, `TENANT_ADMIN`, `MULTICAST_ADMIN`, `CONTRIBUTOR_ADMIN`, `GLOBALSTATE_ADMIN`), and drop the now-migrated subsystems from `NON_MIGRATED_SUBSYSTEMS`. Update the two unit tests that assert `gaps.len() == 5`.

2. (Strongly recommended, not strictly blocking) Reduce the drift risk. The hand-maintained mirror of `check_legacy_any` in the CLI is a standing liability — this PR ships stale and even adds a CLAUDE.md note acknowledging the hazard. Consider exporting the flag→legacy-key mapping from the serviceability crate so the program and the audit share one source, or add a test that fails when a processor calls `authorize()` with a flag the audit doesn't know about.

juan-malbeclabs:
Good catch — confirmed and fixed in fa82e1a1.

**On the blocking issue:** you're right that the stale lists fail in the unsafe direction. Rather than just topping up `MIGRATED_FLAGS`, I removed the hand-maintained mirror entirely and made the serviceability crate the single source of truth, which also addresses your point (2):

- `authorize::AUTHORIZE_GATED_FLAGS` — the canonical list of every flag routed through `authorize()`. I audited all call sites and it now covers `NETWORK_ADMIN`, `INFRA_ADMIN`, `TENANT_ADMIN`, `MULTICAST_ADMIN`, `CONTRIBUTOR_ADMIN`, `GLOBALSTATE_ADMIN`, and also `ACTIVATOR`/`HEALTH_ORACLE` (both are `authorize()`-gated and legacy-backed, so their authority keys are at risk too).
- `authorize::legacy_keys_for_flags` — the enumerating inverse of `check_legacy_any` (given a flag, which legacy keys authorize it). A `#[cfg(test)]` equivalence test asserts the two agree for every flag/key, so they can't drift. `check_legacy_any` itself is unchanged (stays allocation-free on the on-chain path).

The audit now derives its gap detection from these two, so there's nothing left to keep in sync by hand. `NON_MIGRATED_SUBSYSTEMS` is reduced to the only residual GlobalState-only privilege I could find: user create (custom owner override → foundation/sentinel; qa/foundation bypass of device status & seat limits). The CLAUDE.md note is updated to point at the shared lists.

Tests: updated the two `gaps.len() == 5` assertions to 11, and added a regression test asserting the previously-missing flags (network/infra/tenant/multicast/contributor/globalstate + activator/health-oracle) now surface gaps.

elitegreg (approved):

PR #3976, opened 2026-07-05
e2e: tolerate multiple tunnel statuses in all-devices unicast QA test

thijsvanemmerik:
## Summary

- `TestQA_AllDevices_UnicastConnectivity` called the deprecated `GetUserStatus`, which returns an error when a host reports more than one tunnel status. A host that holds both an IBRL and a Multicast tunnel therefore errored and was dropped from the batch, and a batch with fewer than two connected clients times out.
- Switch the two remaining call sites to `GetUserStatuses` + `FindIBRLStatus`, which select the IBRL status (and return the sole status on single-tunnel hosts). A host with a lingering Multicast tunnel is now kept via its IBRL status; a host with no IBRL status is excluded from routing.
- Log a warning when a host reports more than one tunnel status, so the multi-tunnel condition stays visible without failing the sweep.

## Testing Verification

- Added `TestFindIBRLStatus` covering: single status returned as-is, IBRL selected when an IBRL and a Multicast status are both present (including `IBRLWithAllocatedIP`), and nil when no IBRL status exists.
- `go test ./e2e/internal/qa/ -run TestFindIBRLStatus` passes; `go vet -tags qa ./e2e/...` clean.

### DIFF
--- a/e2e/qa_alldevices_unicast_test.go
+++ b/e2e/qa_alldevices_unicast_test.go
@@ -169,7 +169,7 @@ func TestQA_AllDevices_UnicastConnectivity(t *testing.T) {
 						return "", err
 					}
 					if len(statuses) > 1 {
-						log.Warn("Host reported multiple tunnel statuses; selecting IBRL", "client", c.Host, "count", len(statuses))
+						log.Warn("Host reported multiple tunnel statuses", "client", c.Host, "count", len(statuses))
 					}
 					// A host may report multiple tunnel statuses (an IBRL and
 					// a Multicast tunnel). Select the IBRL status, or the sole
@@ -406,7 +406,7 @@ func connectClientsAndWaitForRoutes(
 			continue
 		}
 		if len(userStatuses) > 1 {
-			log.Warn("Host reported multiple tunnel statuses; selecting IBRL", "client", c.Host, "count", len(userStatuses))
+			log.Warn("Host reported multiple tunnel statuses", "client", c.Host, "count", len(userStatuses))
 		}
 		// A host may report multiple tunnel statuses (an IBRL and a
 		// Multicast tunnel). Select the IBRL status, or the sole status
--- a/e2e/qa_alldevices_unicast_test.go
+++ b/e2e/qa_alldevices_unicast_test.go
@@ -164,11 +164,22 @@ func TestQA_AllDevices_UnicastConnectivity(t *testing.T) {
 		getStatus := func(hostname string) (string, error) {
 			for _, c := range clients {
 				if c.Host == hostname {
-					status, err := c.GetUserStatus(ctx)
+					statuses, err := c.GetUserStatuses(ctx)
 					if err != nil {
 						return "", err
 					}
-					return status.SessionStatus, nil
+					if len(statuses) > 1 {
+						log.Warn("Host reported multiple tunnel statuses; selecting IBRL", "client", c.Host, "count", len(statuses))
+					}
+					// A host may report multiple tunnel statuses (an IBRL and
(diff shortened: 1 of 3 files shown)


e2e/qa_alldevices_unicast_test.go
```diff
@@ -389,14 +400,26 @@ func connectClientsAndWaitForRoutes(
...
+		userStatuses, err := c.GetUserStatuses(ctx)
 		if err != nil {
 			log.Error("Failed to get user status", "client", c.Host, "error", err)
 			continue
 		}
-		statuses[c.Host] = status.SessionStatus
-		if !qa.IsStatusUp(status.SessionStatus) {
-			log.Warn("Client not up", "client", c.Host, "status", status.SessionStatus)
+		if len(userStatuses) > 1 {
+			log.Warn("Host reported multiple tunnel statuses; selecting IBRL", "client", c.Host, "count", len(userStatuses))
```
nikw9944: The 'Host reported multiple tunnel statuses; selecting IBRL' warning is logged before `FindIBRLStatus` runs, so on a multi-status host with no IBRL entry it is immediately followed by 'Client has no IBRL status' (line 417) — mildly contradictory. Consider logging just 'Host reported multiple tunnel statuses' and dropping the 'selecting IBRL' promise, or moving it after a successful selection. Same pattern in the getStatus closure at line 172. Control flow is correct; log-clarity nit only.

nikw9944 (commented):
Small, well-targeted fix migrating two call sites in `TestQA_AllDevices_UnicastConnectivity` from the deprecated `GetUserStatus` (which errors on multi-tunnel hosts and dropped them from the batch) to `GetUserStatuses` + `FindIBRLStatus`. Reuses existing tested helpers, error/empty semantics align with downstream consumers, CHANGELOG placed correctly. `TestFindIBRLStatus` passes; `go vet -tags qa ./e2e/...` and build are clean. No critical/high issues. Two minor points: (1) `WaitForStatusUp`/`waitForStatus` still require all tunnel statuses to be up, so a reconnecting host with a lingering non-up Multicast tunnel can still stall 90s and log a false DEVICE FAILURE before the new IBRL-selection code runs — pre-existing and likely out of scope, but the tolerance isn't fully end-to-end; (2) the 'selecting IBRL' warning is logged before selection and can be immediately followed by 'has no IBRL status', a minor log-clarity nit.

- `e2e/qa_alldevices_unicast_test.go:385` — medium: The multi-tunnel tolerance isn't fully end-to-end: `WaitForStatusUp` -> `waitForStatus` (e2e/internal/qa/client.go:753) iterates every entry in `resp.Status` and only returns true when *all* are up. For a reconnecting host that carries a lingering Multicast tunnel whose SessionStatus is not up, this call spins the full 90s timeout and logs a 'DEVICE FAILURE' (line 390) before the new IBRL-selecting block at line 403 is ever reached. Pre-existing behavior and arguably outside this PR's scope (the GetUserStatus *error* path), so not a blocker — but worth noting or a follow-up, since the PR's intent is to tolerate multi-tunnel hosts.

nikw9944 (approved):

e2e/qa_alldevices_unicast_test.go
thijsvanemmerik: Both call sites (the `getStatus` closure and `connectClientsAndWaitForRoutes`) now log "Host reported multiple tunnel statuses" and leave the selection outcome to the following lines, so the contradictory selecting-IBRL / no-IBRL-status pairing is gone.

The `WaitForStatusUp` end-to-end gap is a real pre-existing issue and is tracked as a follow-up in #3987 (swap to `WaitForUnicastStatusUp`, carrying the `doubleZeroIP` side effect).

PR #3978, opened 2026-07-05
serviceability: authorize contributor instructions via Permission accounts

juan-malbeclabs:
## Summary

Gate Contributor instructions (`create`, `update`, `suspend`, `resume`, `delete`) on the `CONTRIBUTOR_ADMIN` flag via `authorize()`. Behavior is preserved: the legacy foundation-allowlist check still authorizes while `RequirePermissionAccounts` is off, and the contributor owner retains the ops-manager-only update path.

One PR per domain in the incremental Permission-system rollout. See `smartcontract/programs/doublezero-serviceability/PERMISSION.md`.

## Testing Verification

- New CONTRIBUTOR_ADMIN permission-account test plus the existing `test_contributor` coverage (owner ops-manager update, non-owner rejection).
- `cargo test -p doublezero-serviceability`, `cargo test -p doublezero_sdk`, `make rust-lint` pass.

---

## Permission migration series

One of 8 per-domain PRs migrating serviceability instructions to the Permission (`authorize()`) system. The branches partition the change set with no overlap and can be reviewed and merged independently (only the CHANGELOG entry conflicts trivially).

| PR | Domain | Flag(s) |
|----|--------|---------|
| #3977 | Governance (globalstate/globalconfig/allowlists) | GLOBALSTATE_ADMIN |
| #3978 | Contributor | CONTRIBUTOR_ADMIN ← this PR |
| #3979 | Infra (location/exchange) | INFRA_ADMIN |
| #3980 | Devices + interfaces | NETWORK_ADMIN, HEALTH_ORACLE |
| #3981 | Links | NETWORK_ADMIN, HEALTH_ORACLE |
| #3982 | Multicast | MULTICAST_ADMIN, ACCESS_PASS_ADMIN |
| #3983 | Tenant | TENANT_ADMIN |
| #3984 | User (update / check_access_pass / check_status) | USER_ADMIN, ACTIVATOR |

PR #3977, opened 2026-07-05
serviceability: authorize governance instructions via Permission accounts

juan-malbeclabs:
## Summary

Migrate the governance admin instructions to the `authorize()` Permission-account system (see `smartcontract/programs/doublezero-serviceability/PERMISSION.md`), gated on the `GLOBALSTATE_ADMIN` flag:

- GlobalState: `SetAuthority`, `SetAirdrop`, `SetFeatureFlags`, `SetVersion`
- GlobalConfig: `SetGlobalConfig`
- Foundation and QA allowlist add/remove

Behavior is preserved: while `RequirePermissionAccounts` is off, the legacy foundation-allowlist check still authorizes; Permission-account holders with `GLOBALSTATE_ADMIN` are additionally accepted. Also corrects the stale legacy-mapping table in PERMISSION.md to match `check_legacy_any()`.

Part of the incremental rollout of the Permission system across serviceability (one PR per domain).

## Testing Verification

- New permission-account tests exercising the GLOBALSTATE_ADMIN path across the three account layouts (fixed, allowlist-mutating, and the 10-account SetGlobalConfig), plus the existing foundation/non-foundation regression tests.
- `cargo test -p doublezero-serviceability`, `cargo test -p doublezero_sdk`, and `make rust-lint` pass.

---

## Permission migration series

One of 8 per-domain PRs migrating serviceability instructions to the Permission (`authorize()`) system. The branches partition the change set with no overlap and can be reviewed and merged independently (only the CHANGELOG entry conflicts trivially).

| PR | Domain | Flag(s) |
|----|--------|---------|
| #3977 | Governance (globalstate/globalconfig/allowlists) | GLOBALSTATE_ADMIN ← this PR |
| #3978 | Contributor | CONTRIBUTOR_ADMIN |
| #3979 | Infra (location/exchange) | INFRA_ADMIN |
| #3980 | Devices + interfaces | NETWORK_ADMIN, HEALTH_ORACLE |
| #3981 | Links | NETWORK_ADMIN, HEALTH_ORACLE |
| #3982 | Multicast | MULTICAST_ADMIN, ACCESS_PASS_ADMIN |
| #3983 | Tenant | TENANT_ADMIN |
| #3984 | User (update / check_access_pass / check_status) | USER_ADMIN, ACTIVATOR |

PR #3980, opened 2026-07-05
serviceability: authorize device and interface instructions via Permission accounts

juan-malbeclabs:
## Summary

Gate Device and device-interface instructions on `NETWORK_ADMIN` (and `HEALTH_ORACLE` for `sethealth`) via `authorize()`, preserving the contributor-owner path:

- Device: `create`, `update`, `delete`, `resume`, `sethealth`
- Interface: `create`, `update`, `delete`

Owner-or-admin handlers become `owner OR authorize(NETWORK_ADMIN)`; `sethealth` composes `HEALTH_ORACLE | NETWORK_ADMIN` to keep both the oracle and foundation paths. Internal foundation-only sub-gates (contributor binding, privileged-field edits, status transitions) are extended to `NETWORK_ADMIN` holders. `interface/update` uses `split_trailing_permission` so the optional segment-routing/topology accounts stay unambiguous vs the SDK-appended Permission PDA.

Behavior is preserved via the legacy foundation/health-oracle fallback while `RequirePermissionAccounts` is off. One PR per domain; see PERMISSION.md.

## Testing Verification

- Full device/interface integration suites (create/update/delete/onchain-allocation/contributor-binding/mgmt-vrf/location-update) and SDK command tests pass.

---

| PR | Domain | Flag(s) |
|----|--------|---------|
| #3980 | Devices + interfaces | NETWORK_ADMIN, HEALTH_ORACLE ← this PR |
| #3981 | Links | NETWORK_ADMIN, HEALTH_ORACLE |
| #3984 | User (update / check_access_pass / check_status) | USER_ADMIN, ACTIVATOR |


elitegreg (changes requested):
Reviewed the authorization migration against the shared `authorize()` helper and the SDK's trailing-account handling. The `owner OR authorize(NETWORK_ADMIN)` rewrites are faithful (NETWORK_ADMIN legacy-maps exactly to `foundation_allowlist`), the `sethealth` `HEALTH_ORACLE | NETWORK_ADMIN` composition preserves the old two-key check, and I traced `split_trailing_permission` in `interface/update.rs` through all layouts (legacy / node_segment_idx+onchain / topologies, each ±permission) — it's correct. Behavior is preserved under the legacy fallback.

Two items below.

**1. Missing device/interface-level integration test for the Permission-account path**

The test changes in this PR are the mechanical `execute_transaction` → `execute_authorized_transaction` mock rename; the only permission-account coverage lives in the shared `authorize.rs` unit tests. Given how subtle the `split_trailing_permission` account-peeling in `interface/update.rs` is (variable-length middle + trailing `payer`/`system`/`permission`), could we add at least one integration test that drives `process_update_device_interface` with a real Permission (`NETWORK_ADMIN`) account appended in the tail, alongside `seg_ext`/topology accounts? That's exactly the case the count-heuristic → PDA-peel change is protecting, and it's the part most likely to regress silently.

**2. Nit (possibly pre-existing)**

In `smartcontract/sdk/rs/src/commands/device/sethealth.rs`, the switched mock's `.with(predicate::eq(...))` matches on `DoubleZeroInstruction::UpdateDevice(...)` rather than `SetDeviceHealth`. Since the PR touches this line anyway, worth confirming the mock asserts the right instruction.

juan-malbeclabs:
Thanks for the thorough trace, @elitegreg. Both items are addressed in b2d0f80 (rebased onto the latest branch as 1097f49):

**1. Integration test for the Permission-account path**

Added `test_update_interface_node_segment_idx_authorized_by_network_admin_permission` in `interface_onchain_allocation_test.rs`. It drives `process_update_device_interface` with a real `NETWORK_ADMIN` Permission account appended to the tail while the variable-length `SegmentRoutingIds` account sits in the middle — exercising `split_trailing_permission` peeling payer/system/permission with a mid account present, plus the `authorize(NETWORK_ADMIN)` path that gates `node_segment_idx`. The signer is neither the contributor owner nor a foundation key. The test also asserts the negative case (same update without a Permission account is rejected with `NotAllowed`), proving the Permission is what authorizes the update rather than some other trailing account.

**2. sethealth mock**

Good catch — the mock was matching the stale `UpdateDevice` instruction (the test module was cloned from the update command). Rewrote it to construct `SetDeviceHealthCommand` and assert `SetDeviceHealth` on `execute_authorized_transaction`, checking the `[device, globalstate]` account layout.

elitegreg (approved):

PR #3981, opened 2026-07-05
serviceability: authorize link instructions via Permission accounts

juan-malbeclabs:
## Summary

Gate Link instructions on `NETWORK_ADMIN` (and `HEALTH_ORACLE` for `sethealth`) via `authorize()`, preserving the contributor-owner path:

- `create`, `update`, `delete`, `suspend`, `resume`, `accept`, `sethealth`

Owner-or-admin handlers become `owner OR authorize(NETWORK_ADMIN)`; `accept` (previously owner-only) gains a NETWORK_ADMIN bypass; `sethealth` composes `HEALTH_ORACLE | NETWORK_ADMIN`. `delete` and `update` are variable-length (topology union parsed off the tail) and use `split_trailing_permission` so the trailing Permission PDA is disambiguated by PDA match. Internal foundation-only sub-gates (per-link contributor binding, tunnel-field / topology / unicast_drained edits) are extended to `NETWORK_ADMIN` holders.

Behavior preserved via the legacy fallback while `RequirePermissionAccounts` is off. One PR per domain; see PERMISSION.md.

## Testing Verification

- Link WAN / onchain-allocation / bandwidth-validation integration suites and SDK command tests pass.

---

| PR | Domain | Flag(s) |
|----|--------|---------|
| #3980 | Devices + interfaces | NETWORK_ADMIN, HEALTH_ORACLE |
| #3981 | Links | NETWORK_ADMIN, HEALTH_ORACLE ← this PR |
| #3984 | User (update / check_access_pass / check_status) | USER_ADMIN, ACTIVATOR |

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/processors/link/accept.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/link/accept.rs
@@ -58,7 +58,9 @@ pub fn process_accept_link(
     let accounts_iter = &mut accounts.iter();
 
     // Account layout: [link, contributor, side_z_dev, globalstate, side_a_dev,
-    //                  device_tunnel_block, link_ids, payer, system]
+    //                  device_tunnel_block, link_ids, payer, system, permission?]
+    // The trailing optional permission account (the payer's Permission PDA, appended
+    // by the SDK when it exists on-chain) is consumed by authorize() below.
     let link_account = next_account_info(accounts_iter)?;
     let contributor_account = next_account_info(accounts_iter)?;
     let side_z_account = next_account_info(accounts_iter)?;
--- a/smartcontract/programs/doublezero-serviceability/src/processors/link/accept.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/link/accept.rs
@@ -1,4 +1,5 @@
 use crate::{
+    authorize::authorize,
     error::DoubleZeroError,
     processors::{
         link::resource_onchain_helpers::validate_and_allocate_link_resources,
@@ -8,8 +9,10 @@ use crate::{
     state::{
         contributor::Contributor,
         device::Device,
+        globalstate::GlobalState,
         interface::{InterfaceCYOA, InterfaceDIA, InterfaceStatus},
         link::*,
+        permission::permission_flags,
     },
 };
 use borsh::BorshSerialize;
@@ -59,7 +62,7 @@ pub fn process_accept_link(
     let link_account = next_account_info(accounts_iter)?;
     let contributor_account = next_account_info(accounts_iter)?;
     let side_z_account = next_account_info(accounts_iter)?;
-    let _globalstate_account = next_account_info(accounts_iter)?;
+    let globalstate_account = next_account_info(accounts_iter)?;
 
     let side_a_device_account = next_account_info(accounts_iter)?;
     let device_tunnel_block_ext = next_account_info(accounts_iter)?;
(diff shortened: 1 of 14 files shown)


smartcontract/programs/doublezero-serviceability/src/processors/link/accept.rs
```diff
@@ -59,7 +62,7 @@ pub fn process_accept_link(
     let link_account = next_account_info(accounts_iter)?;
```
elitegreg, line 64: Nit: the account-layout comment at the top of this function (`[link, contributor, side_z_dev, globalstate, side_a_dev, device_tunnel_block, link_ids, payer, system]`) wasn't updated to note the trailing optional permission account that `authorize()` now consumes. `delete`/`update` document their `permission?` tail — worth matching that here for consistency.

elitegreg (changes requested):
Migration logic looks correct and behavior-preserving under the current feature flag — the `split_trailing_permission` tail-parsing for the variable-length `delete`/`update` paths holds up across the topology-union / permission-present-or-absent cases, and the fixed-length instructions consume payer/system before calling `authorize()`.

One blocking ask:

**Add integration test coverage for the link permission-account path.** The diff is 7 processors + 5 SDK files + CHANGELOG with no test changes. `permission_test.rs` exercises Permission CRUD but nothing drives a link instruction *through* an appended permission account. The subtlest logic in this PR is `delete`/`update` peeling the payer Permission PDA off the tail past the caller-controlled topology union (`split_trailing_permission`), and that path is currently untested end-to-end. Please add at least one integration test proving a `NETWORK_ADMIN` permission holder can `delete` and `update` a link with a non-empty topology union present (i.e. permission account + topology accounts in the same tail), so a future change to the tail layout can't silently break disambiguation.

Minor nit inline on `accept.rs`.

smartcontract/programs/doublezero-serviceability/src/processors/link/accept.rs
juan-malbeclabs: Done in 36d27d2a — the layout comment now lists the trailing `permission?` account and notes that `authorize()` consumes it, matching `delete`/`update`.

juan-malbeclabs:
Both review items addressed in 36d27d2a (branch also rebased onto `main`).

**Blocking ask — integration coverage for the link permission-account path.** Added two tests to `link_onchain_allocation_test.rs` that drive a link instruction *through* an appended Permission account, so `split_trailing_permission` is exercised end-to-end:

- `test_delete_link_via_network_admin_permission_with_topology_union` — a `NETWORK_ADMIN` Permission holder (neither the contributor owner nor a foundation member) deletes a link with a non-empty topology union present, i.e. permission account + topology accounts in the same tail. It also asserts the disambiguation isn't a blanket allow: **without** the appended Permission account the same non-owner caller is denied with `InvalidOwnerPubkey`, since legacy `NETWORK_ADMIN` is foundation-only.
- `test_update_link_topologies_via_network_admin_permission` — the `update` counterpart: the same admin adds a second topology to `link_topologies` (a network-admin-only field) with the topology union riding in the same tail as the permission account.

Because the admin is not foundation, the only path to authorization is the Permission account itself, so a green test proves both the auth and the tail-parsing. A future change to the tail layout that broke disambiguation would fail these.

**Nit — `accept.rs` layout comment.** Fixed (replied inline).

Verification: all 8 tests in `link_onchain_allocation_test` pass (6 existing + 2 new); `authorize` unit tests green; clippy clean.

elitegreg (approved):

PR #3984, opened 2026-07-05
serviceability: authorize user instructions via Permission accounts

juan-malbeclabs:
## Summary

Gate the administrative user instructions via `authorize()`:

- `UpdateUser` -> `USER_ADMIN` (uses `split_trailing_permission` for its optional old/new tenant accounts)
- `CheckAccessPass` -> `ACTIVATOR`
- accesspass `CheckStatus` -> `ACTIVATOR | USER_ADMIN` (activator or foundation)

Behavior preserved via the legacy foundation/activator fallback while `RequirePermissionAccounts` is off.

`CreateUser`/`CreateSubscribeUser` and `SetBgpStatus` are intentionally **not** migrated: the account owner authorizes via access-pass ownership (create) or `device.metrics_publisher_pk` (set_bgp_status), so the owner must not need a Permission account — the Permission system is for administrators only.

Final PR in the incremental Permission-system rollout; see PERMISSION.md.

## Testing Verification

- User and accesspass integration suites (user_tests, user_old, user_migration, create_subscribe, delete_user_dynamic_accesspass, accesspass_test) and SDK command tests pass.

---

| PR | Domain | Flag(s) |
|----|--------|---------|
| #3980 | Devices + interfaces | NETWORK_ADMIN, HEALTH_ORACLE |
| #3981 | Links | NETWORK_ADMIN, HEALTH_ORACLE |
| #3984 | User (update / check_access_pass / check_status) | USER_ADMIN, ACTIVATOR ← this PR |


elitegreg (changes requested):
Nice, clean finale to the Permission-system series. The `authorize()` migration is correct across all three instructions: the helper is called only after every fixed account is consumed, the legacy fallbacks preserve behavior (`ACTIVATOR` → `activator_authority_pk`, `USER_ADMIN` → `foundation_allowlist`), and the `msg!`/import gating is right.

The one thing I'd like added before merge is **test coverage for the new permission-account path**, specifically for `UpdateUser`.

The existing suites (`user_tests`, `accesspass_test`, etc.) all pass through the *legacy* branch — they don't supply a permission account, so the new code path in `process_update_user` isn't exercised end to end. The novel logic here is the tenant-pair-plus-permission disambiguation:

```rust
let (payer_account, _system_program, leading, permission_account) =
    split_trailing_permission(program_id, &remaining)?;
let (old_tenant_account, new_tenant_account) = if leading.len() >= 2 { ... };
```

`split_trailing_permission` itself has no direct unit tests, and it's only exercised elsewhere by the topology processors, which don't have update's optional-pair *leading* accounts. So the highest-risk combination — **`UpdateUser` with both tenant accounts AND an appended permission account** (tail of length 5) — is untested. That's exactly the case a future regression in the split logic would land on.

Requesting one of:
- an integration test for `UpdateUser` that supplies a valid `USER_ADMIN` permission account, ideally including the tenant-update variant so the length-5 tail is covered; and/or
- direct unit tests for `split_trailing_permission` covering the 2/3/4/5-length tails.

If the earlier PRs in the series already established a per-instruction permission-path test pattern, matching it here would be ideal — let me know if that's the case and I'm happy to reconsider whether the generic `authorize()` unit tests are deemed sufficient.

Everything else LGTM.

juan-malbeclabs:
Thanks for the thorough review. Addressed both requests in ac7dcada.

**Integration test for `UpdateUser` via the Permission-account path** (`tests/user_tests.rs`, step 10c in `test_user`): a non-foundation key holding `USER_ADMIN` reassigns the user's tenant. It uses `create_authorized_transaction` to reproduce the exact on-wire layout the SDK emits — `[old_tenant, new_tenant, payer, system, permission]`, i.e. the **length-5 tail** — so the tenant-pair-plus-permission disambiguation in `split_trailing_permission` is exercised end to end. Because the key is not in the foundation allowlist, the legacy `USER_ADMIN → foundation_allowlist` fallback would deny it, so a passing test confirms authorization went through the Permission account rather than the legacy branch. Reference counts on both tenants are asserted.

**Direct unit tests for `split_trailing_permission`** (`src/authorize.rs`): cover the 2/3/4/5-length tails plus the too-short error case, including both disambiguation branches — trailing account matches the payer's Permission PDA (peeled) vs. doesn't (treated as system, leaving the leading accounts intact).

On the established pattern: the sibling migrations (`contributor_test`, `location_test`, etc.) do have per-instruction Permission-path tests, but none carries `UpdateUser`'s optional *leading* accounts before the tail — so this fills the specific gap you flagged rather than just matching the generic pattern.

elitegreg (approved):

PR #3988, opened 2026-07-06
client/doublezero connect: surface underlying CreateUser error

nikw9944:
## Summary of Changes

- `doublezero connect` swallowed the real error on `CreateUser` failure: both arms returned a bare `eyre::eyre!("Error creating user")` and dropped `e`, so the underlying cause (e.g. `unable to confirm transaction`, an access-pass/allowlist rejection, etc.) only went to the suppressed spinner `println` and never reached the returned error / caller logs.
- Include `{e:?}` in the returned error at both sites (`connect.rs:586`, `connect.rs:832`) so the cause is visible to callers (the QA e2e harness surfaces it directly; interactive users still get the spinner detail too).

## Testing Verification

- `cargo build` of the client with this change was exercised earlier as a snapshot (`0.29.0-SNAPSHOT`) deployed to the QA hosts; it's what made the underlying `CreateUser` failure legible ("Error creating user: unable to confirm transaction …") that led to the DZ-ledger cluster + oracle root causes.

### DIFF
--- a/client/doublezero/src/command/connect.rs
+++ b/client/doublezero/src/command/connect.rs
@@ -746,7 +746,9 @@ impl ProvisioningCliCommand {
                             Err(e) => {
                                 spinner.println("❌ Error adding publisher subscription");
                                 spinner.println(format!("\n{}: {:?}\n", "Error", e));
-                                eyre::bail!("Error adding publisher subscription to existing user");
+                                eyre::bail!(
+                                    "Error adding publisher subscription to existing user: {e:?}"
+                                );
                             }
                         }
                     }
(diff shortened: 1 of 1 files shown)


thijsvanemmerik (approved):
Reviewed the full diff and traced the error types this touches. Safe and does what it says.

- Confirmed `{e:?}` cannot leak sensitive material: the payer keypair is only used to sign transactions, and the returned error variants carry only program error codes and onchain program logs. Nothing secret flows into these errors. The same `{e:?}` was already printed by the spinner, so there's no new disclosure surface either.
- The e2e substring match still holds: `"Error creating user: {e:?}"` still begins with `"Error creating user"`, so any existing matcher keeps working.
- Coverage looks complete for the goal. The arm at `connect.rs:696` already surfaced the cause, and the remaining `map_err(|_| ...)` arms (pubkey parse, device fetch) are intentionally generic.

Two optional nits, neither blocking:

1. `connect.rs:585-586` (and the sibling arms): interactive users now see the cause printed twice, once from `spinner.println` and once from the returned error. Harmless, just slightly noisy on a TTY.
2. `connect.rs:696` uses `"...: {:?}", e` while the changed arms use inline `{e:?}`. Worth aligning for style, functionally identical.

LGTM.

PR #3989, opened 2026-07-06
cli: block device delete while enabled in shred-subscription

martinsander00:
## Summary

- `doublezero device delete` now refuses to delete a serviceability device that is still **enabled** in the shred-subscription program. Deleting such a device orphans it: the shred oracle only settles `Activated` serviceability devices, so it can never settle the orphan, yet the orphan still counts toward `total_enabled_devices` and permanently deadlocks epoch advance. This implements step N3 ("prevent recurrence") of malbeclabs/infra#1769.
- The guard derives the device's `DeviceHistory` PDA and checks its `enabled` flag on the correct Solana L1 cluster before submitting the delete: DZ mainnet-beta → Solana mainnet-beta, DZ testnet → Solana **devnet** (testnet's shred-subscription runs on Solana devnet, not Solana testnet). On devnet/local, where the program isn't deployed, the check is skipped.
- Fail-closed: any RPC error aborts the delete (no bypass flag). A device whose `DeviceHistory` exists but is already disabled is still allowed to be deleted.

## Design notes

- This is a CLI-side guard (advisory): it prevents the operator accident that caused the incident without coupling the serviceability program to shred-subscription onchain or requiring a program redeploy. The issue also lists an onchain guard/CPI and oracle-side reconciliation as alternatives; those remain open as separate hardening options.
- Scope is `device delete` only (not drain / `device update`→Drained), per discussion.
- The shred-subscription program has no Rust SDK in-repo; the account layout (PDA seeds `["device_history", device_pubkey]`, discriminator, and `enabled` = flag bit 1 at offset 40) is mirrored from the Go SDK at `sdk/shreds/go/`.

## Testing Verification

- Added a config mapping test asserting DZ testnet resolves the shred check to Solana devnet (and that it differs from the testnet `solana_l1_rpc_url`, which is Solana testnet), and that devnet/local return no config.
- Added unit tests for the `DeviceHistory` flag parser: enabled bit set/clear, wrong discriminator, and too-short account.
- Confirmed the guard short-circuits with no network call on devnet/local — the existing `device delete` CLI test runs under `Environment::Local` and stays green.

Implements step N3 of malbeclabs/infra#1769.

### DIFF
--- a/smartcontract/cli/src/device/shred_guard.rs
+++ b/smartcontract/cli/src/device/shred_guard.rs
@@ -1,0 +1,378 @@
+//! Guard that prevents deleting a serviceability device while it is still
+//! enabled in the shred-subscription program.
+
+use borsh::BorshDeserialize;
+use doublezero_config::{Environment, ShredSubscriptionConfig};
+use solana_client::nonblocking::rpc_client::RpcClient;
+use solana_sdk::{
+    account::Account, commitment_config::CommitmentConfig, hash::hash, pubkey::Pubkey,
+};
+use std::{future::Future, sync::LazyLock, time::Duration};
+
+/// PDA seed for DeviceHistory accounts (see `sdk/shreds/go/pda.go`).
+const DEVICE_HISTORY_SEED: &[u8] = b"device_history";
+
+/// Length of a shred-subscription account discriminator.
+const DISCRIMINATOR_LEN: usize = 8;
+
+/// `enabled` is bit 1 of `DeviceHistory::flags` (see `DeviceHistory::IsEnabled`
+/// in `sdk/shreds/go/state.go`).
+const ENABLED_FLAG: u64 = 1 << 1;
+
+/// Bounds each Solana RPC call so a hung endpoint can't hang the delete.
+const SHRED_RPC_TIMEOUT: Duration = Duration::from_secs(15);
+
+/// Number of attempts for the DeviceHistory lookup before failing closed. The
+/// default RPC is the public, rate-limited Solana endpoint, so a transient 429
+/// or timeout should retry rather than abort a legitimate delete.
+const SHRED_RPC_ATTEMPTS: usize = 3;
+
+/// Delay between DeviceHistory RPC retry attempts.
+const SHRED_RPC_RETRY_DELAY: Duration = Duration::from_millis(500);
+
+/// Account discriminator: first 8 bytes of `sha256("dz::account::device_history")`
+/// (Solana's `hash` is SHA-256), matching `sdk/shreds/go/discriminator.go`.
+static DEVICE_HISTORY_DISCRIMINATOR: LazyLock<[u8; DISCRIMINATOR_LEN]> = LazyLock::new(|| {
+    hash(b"dz::account::device_history").to_bytes()[..DISCRIMINATOR_LEN]
+        .try_into()
+        .expect("sha256 digest is 32 bytes")
+});
+
+/// Leading fields of the shred-subscription `DeviceHistory` account, up to the
+/// `flags` we read. Mirrors `DeviceHistory` in `sdk/shreds/go/state.go`.
+///
+/// Do NOT add fields past `flags`: the onchain account is a fixed C-style layout
+/// (read via `binary.Read` in the Go SDK) with padding after `flags`. Borsh
+/// coincides with that layout only for these leading, unpadded fields; anything
+/// added after would decode at the wrong offset. `deserialize` (not
+/// `try_from_slice`) is used so the trailing account bytes are ignored.
+#[derive(BorshDeserialize)]
+struct DeviceHistoryHeader {
+    #[allow(dead_code)]
+    device_key: [u8; 32],
+    flags: u64,
+}
+
+/// Errors out if `device_pubkey` still has an enabled DeviceHistory account in
+/// the shred-subscription program.
+///
+/// Fail-closed: any RPC failure aborts (returns an error) rather than allowing
+/// a potentially unsafe delete. Environments without a deployed
+/// shred-subscription program (devnet, local) are skipped.
+pub async fn ensure_device_not_enabled_in_shred_subscription(
+    env: Environment,
+    device_pubkey: &Pubkey,
+) -> eyre::Result<()> {
+    let Some(cfg) = env.shred_subscription_config() else {
+        // Shred-subscription is not deployed for this environment.
+        return Ok(());
+    };
+
+    let rpc = RpcClient::new_with_timeout(cfg.solana_rpc_url.clone(), SHRED_RPC_TIMEOUT);
+    let rpc_url = cfg.solana_rpc_url.clone();
+
+    check_device_history(&cfg, device_pubkey, move |history_pda| async move {
+        fetch_with_retry(SHRED_RPC_ATTEMPTS, SHRED_RPC_RETRY_DELAY, || async {
+            rpc.get_account_with_commitment(&history_pda, CommitmentConfig::confirmed())
+                .await
+                .map(|response| response.value)
+        })
+        .await
+        .map_err(|e| {
+            eyre::eyre!(
+                "failed to query shred-subscription DeviceHistory {history_pda} on {rpc_url} \
+                 after {SHRED_RPC_ATTEMPTS} attempts: {e}"
+            )
+        })
+    })
+    .await
+}
+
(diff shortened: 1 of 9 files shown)


smartcontract/cli/src/device/shred_guard.rs
```diff
@@ -0,0 +1,178 @@
...
+                "failed to query shred-subscription DeviceHistory {history_pda} on {}: {e}",
+                cfg.solana_rpc_url
+            )
+        })?
+        .value;
+
+    let Some(account) = account else {
+        // No DeviceHistory account: the device was never enrolled in
+        // shred-subscription, so it is safe to delete.
+        return Ok(());
```
nikw9944, line 151: The guard fails *open* here: 'no DeviceHistory account found' → delete allowed. This is correct behavior, but it makes the guard's entire protective value depend on `find_program_address(["device_history", device_pubkey], program_id)` deriving the *actual* PDA. If shred-subscription keys `DeviceHistory` by anything other than the serviceability device account pubkey returned by `get_device` (e.g. a device index, metrics_publisher_pk, or public IP), the derived PDA never exists and the guard silently becomes a no-op — no error, no protection. I verified seed/discriminator/program-id/flag-bit/offset all match the Go SDK, but could not confirm from the repo that shred's `DeviceKey` == the serviceability device pubkey. That assumption is load-bearing for the whole fix; recommend confirming against a known-enabled mainnet-beta device.

smartcontract/cli/src/device/shred_guard.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    if device_history_enabled(&account.data)? {
+        eyre::bail!(
+            "Device {device_pubkey} is still enabled in the shred-subscription program \
+             (DeviceHistory {history_pda} on {}). Disable it there before deleting the device, \
+             otherwise the shred oracle will deadlock on epoch settlement. See malbeclabs/infra#1769.",
+            cfg.solana_rpc_url
+        );
+    }
```
nikw9944, line 165: No test exercises this bail path (enabled device → delete refused) or the 'account exists but disabled → allowed' path end to end. The parser (`device_history_enabled`) is well tested, but the `ensure_...` wiring is not. For a safety guard written in response to a production incident, an RPC-mocked test asserting an enabled account blocks the delete would meaningfully raise confidence and protect against future wiring regressions (see the fail-open concern above).

smartcontract/cli/src/device/shred_guard.rs
```diff
@@ -0,0 +1,178 @@
...
+        return Ok(());
+    };
+
+    let (history_pda, _bump) = Pubkey::find_program_address(
+        &[DEVICE_HISTORY_SEED, device_pubkey.as_ref()],
+        &cfg.program_id,
+    );
+
+    let rpc = RpcClient::new_with_timeout(cfg.solana_rpc_url.clone(), SHRED_RPC_TIMEOUT);
+    let account = rpc
```
nikw9944: Fail-closed against `https://api.mainnet-beta.solana.com` (public, aggressively rate-limited) with a single call, 15s timeout, no retry, and no bypass flag. A transient 429/timeout aborts a legitimate delete with no operator escape hatch. Deliberate tradeoff per the PR and consistent with the L1 URL also being public — but consider honoring an env override for the shred RPC URL (as `config()` honors `DZ_SOLANA_RPC_URL`) and/or a small retry.

smartcontract/cli/src/device/shred_guard.rs
```diff
@@ -0,0 +1,178 @@
...
+/// (Solana's `hash` is SHA-256), matching `sdk/shreds/go/discriminator.go`.
+static DEVICE_HISTORY_DISCRIMINATOR: LazyLock<[u8; DISCRIMINATOR_LEN]> = LazyLock::new(|| {
+    hash(b"dz::account::device_history").to_bytes()[..DISCRIMINATOR_LEN]
+        .try_into()
+        .expect("sha256 digest is 32 bytes")
+});
+
+/// Leading fields of the shred-subscription `DeviceHistory` account, up to the
+/// `flags` we read. Mirrors `DeviceHistory` in `sdk/shreds/go/state.go`; borsh
+/// decodes only these fields and ignores the rest of the account.
```
nikw9944: The onchain/Go layout is a fixed C-style struct read via `binary.Read` with explicit padding (BumpSeed, Padding0 [6]byte, etc. after `flags`), not Borsh. Borsh decoding works here *only* because the two fields read (`device_key`, `flags`) sit at the front with no padding, where Borsh coincides with the C layout. If anyone extends `DeviceHistoryHeader` past `flags`, Borsh will silently diverge from the C-aligned layout. Worth a one-line warning: 'do not add fields past `flags` — the rest of the account is a fixed C layout, not Borsh.' (Using `deserialize` rather than `try_from_slice` to tolerate trailing bytes is correct.)

smartcontract/cli/src/device/shred_guard.rs
```diff
@@ -0,0 +1,178 @@
...
+    // Only trust accounts actually owned by the shred-subscription program.
+    if account.owner != cfg.program_id {
+        return Ok(());
+    }
+
+    if device_history_enabled(&account.data)? {
+        eyre::bail!(
+            "Device {device_pubkey} is still enabled in the shred-subscription program \
+             (DeviceHistory {history_pda} on {}). Disable it there before deleting the device, \
+             otherwise the shred oracle will deadlock on epoch settlement. See malbeclabs/infra#1769.",
```
nikw9944: The user-facing bail message references the private infra issue `malbeclabs/infra#1769`. Operators are internal so this is fine, but consider whether a private-repo reference belongs in CLI output vs. the code comment that already carries it (line 16).

nikw9944 (commented):
CLI-side guard blocking `doublezero device delete` while a device is still enabled in the shred-subscription program (incident #1769, step N3). Verified every mirrored constant against the Go SDK (`sdk/shreds/go/`): PDA seed, discriminator, program ID, enabled flag bit, and offset 40 all match. Config mapping (DZ testnet → Solana devnet) is correct and tested. All new tests pass locally, plus the existing `test_cli_device_delete` (runs under `Local`, guard is a no-op).

No critical or high-severity issues — well-scoped and well-documented. Medium findings: (M1) the guard fails *open* if no DeviceHistory account is found, so its protection hinges on the unverifiable-in-repo assumption that shred-subscription keys `DeviceHistory` by the serviceability device pubkey — worth confirming against a known-enabled mainnet device since the failure mode is silent; (M2) no positive test that an enabled device is actually blocked; (M3) fail-closed against the public Solana RPC with no override/retry could block legitimate deletes during RPC hiccups. Low: (L1) 'borsh' decode works only because the read fields precede the C-layout padding — a maintenance trap worth a warning comment; (L2) private infra issue reference in user-facing error; (L3) guard is CLI-only (documented).

- L3 (low, informational): The guard lives only in the CLI verb; the SDK `DeleteDeviceCommand` and the onchain processor have no equivalent check, so any non-CLI delete path still orphans a device. The PR explicitly scopes this as an advisory CLI guard with onchain/CPI hardening as separate follow-ups — expected, noted only so reviewers are aware the protection is not enforced at the program level.

smartcontract/cli/src/device/shred_guard.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/3989/changes/b50edf1a4bb0effaecfe6ddb836195dea2799563

smartcontract/cli/src/device/shred_guard.rs
martinsander00: added both — a `DZ_SHRED_SOLANA_RPC_URL` override (mirrors how `config()` honors `DZ_SOLANA_RPC_URL`) so we can point at a non-public rpc, plus a 3-attempt/500ms retry before failing closed. turned out the backward-compat e2e was actually hitting the public rpc under `--env mainnet-beta` and 429ing, so the e2e manager now sets that override to the local ledger. 321a0462

smartcontract/cli/src/device/shred_guard.rs
martinsander00: good call — added a comment on `DeviceHistoryHeader` spelling out that borsh only lines up with the fixed C layout for the leading unpadded fields, so nothing should be added past `flags`. 321a0462

nikw9944 (approved):

PR #3990, opened 2026-07-06
ci(release.docker.client): auto-publish mainnet-beta base + notify edge-connect

juan-malbeclabs:
## What

Two changes to `release.docker.client.yml`:

1. **Auto-publish the mainnet-beta base image.** `publish-mainnet-beta` was `workflow_dispatch`-only. This adds a daily `schedule` plus an **idempotency guard**: it resolves the latest version from the mainnet-beta stable Cloudsmith channel and compares it to the `org.opencontainers.image.version` label on the currently-published `:mainnet-beta` image, publishing **only when they differ**. Manual dispatch still forces a publish. (Version resolution stays tied to the stable channel — this is *not* chained off every `client/v*` tag.)
2. **Notify edge-connect.** After each successful base publish (both testnet and mainnet-beta), fire a `repository_dispatch` (`doublezero-base-published`, payload `{env, version}`) at `malbeclabs/doublezero-edge-connect` so it rebuilds the matching variant immediately.

## Why

The `doublezero:mainnet-beta` image was stuck at **0.27.1** while **0.28.0** was already in the mainnet-beta stable channel, because the image only rebuilt on a manual dispatch that nobody ran. edge-connect layers on this base and tracks it, so it was stuck too. This makes the base self-heal (schedule + guard) and propagate to edge-connect event-driven instead of via its ~24h poll.

## Methodology notes (matches existing workflows)

- Cross-repo trigger uses **`actions/github-script@v7`** with `github.rest.repos.createDispatchEvent`, same idiom as `trusted-fork-e2e.yml` (the repo does not use the `gh` CLI in workflows).
- New scoped secret **`EDGE_CONNECT_DISPATCH_TOKEN`** (fine-grained PAT with `actions: write` on edge-connect), consistent with the repo's purpose-scoped secrets (`MALBEC_INFRA_RO`, `GA_ANSIBLE_REPO_RO`, …). Alternative: reuse the existing `DOUBLEZERO_PAT` **iff** its scope already includes edge-connect `actions: write` — say the word and I'll switch it.
- Docker actions kept at the repo's `@vN` pinning style.

## Sequence & prerequisites (important)

`repository_dispatch` only triggers workflows on the target repo's **default branch**, so order matters:

- [ ] 1. Merge the receiver PR first: **malbeclabs/doublezero-edge-connect#74**
- [ ] 2. Create the `EDGE_CONNECT_DISPATCH_TOKEN` secret in this repo
- [ ] 3. Merge **this** PR
- [ ] 4. (Runbook) add a step after "Promote client packages → doublezero" in the release checklist to run `publish-mainnet-beta` for an immediate publish; the daily schedule is the backstop.

## Validation

- `actionlint` clean.
- Guard verified against the live registry: currently `have=0.27.1-1` ≠ `want=0.28.0-1` → first run publishes 0.28.0; a second run is a no-op.

### DIFF
--- a/.github/workflows/release.docker.client.yml
+++ b/.github/workflows/release.docker.client.yml
@@ -8,6 +8,11 @@ on:
   workflow_run:
     workflows: ["releaser.client"]
     types: [completed]
+  # Daily auto-publish of the mainnet-beta image when the stable Cloudsmith channel
+  # advances (guarded below so a run with no new version is a no-op). This removes
+  # the manual-only gate that previously left :mainnet-beta lagging the release.
+  schedule:
+    - cron: "37 5 * * *" # daily 05:37 UTC
   pull_request:
     paths:
       - "client/Dockerfile"
@@ -103,10 +108,26 @@ jobs:
             ${{ env.IMAGE }}:testnet
             ${{ env.IMAGE }}:testnet-${{ steps.v.outputs.version }}
 
-  # Manual dispatch path: publish mainnet-beta tags (:mainnet-beta, :version, :latest).
-  # The version can be supplied directly or resolved from the public Cloudsmith repo.
+      # Tell doublezero-edge-connect to rebuild its :testnet variant against the
+      # base image we just published (event-driven; its daily poll is the fallback).
+      - name: Notify edge-connect to rebuild (testnet)
+        if: steps.v.outputs.publish == 'true'
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.EDGE_CONNECT_DISPATCH_TOKEN }}
+          script: |
+            await github.rest.repos.createDispatchEvent({
+              owner: context.repo.owner,
+              repo: 'doublezero-edge-connect',
+              event_type: 'doublezero-base-published',
+              client_payload: { env: 'testnet', version: '${{ steps.v.outputs.version }}' },
+            });
+
+  # mainnet-beta image publish: manual (workflow_dispatch) or automatic (schedule).
+  # The version is resolved from the mainnet-beta stable Cloudsmith channel (the
+  # deliberate promotion gate), so this is NOT chained off every client/v* tag.
   publish-mainnet-beta:
-    if: github.event_name == 'workflow_dispatch'
+    if: github.event_name == 'workflow_dispatch' || github.event_name == 'schedule'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -136,7 +157,29 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      # Idempotency guard: skip the (re)publish when the already-published
+      # :mainnet-beta image is at the same version the stable channel resolves to.
+      # A manual workflow_dispatch always publishes (override).
+      - name: Decide whether a publish is needed
+        id: gate
+        shell: bash
+        run: |
+          want='${{ steps.v.outputs.version }}'
+          have="$(docker buildx imagetools inspect \
+            --format '{{ index .Image.Config.Labels "org.opencontainers.image.version" }}' \
+            '${{ env.IMAGE }}:mainnet-beta' 2>/dev/null || true)"
+          echo "published mainnet-beta version: ${have:-<none>} ; stable channel latest: $want"
+          if [ '${{ github.event_name }}' = 'workflow_dispatch' ]; then
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+          elif [ "$want" != "$have" ]; then
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "changed=false" >> "$GITHUB_OUTPUT"
+            echo "mainnet-beta image already at $have; nothing to do"
+          fi
+
       - name: Build & push (:mainnet-beta + version + :latest)
+        if: steps.gate.outputs.changed == 'true'
         uses: docker/build-push-action@v6
         with:
           context: ./client
@@ -156,6 +199,21 @@ jobs:
             ${{ env.IMAGE }}:mainnet-beta
             ${{ env.IMAGE }}:mainnet-beta-${{ steps.v.outputs.tag }}
             ${{ env.IMAGE }}:latest
+
+      # Tell doublezero-edge-connect to rebuild its :mainnet-beta variant against
+      # the base image we just published (event-driven; poll is the fallback).
+      - name: Notify edge-connect to rebuild (mainnet-beta)
+        if: steps.gate.outputs.changed == 'true'
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.EDGE_CONNECT_DISPATCH_TOKEN }}
+          script: |
+            await github.rest.repos.createDispatchEvent({
+              owner: context.repo.owner,
+              repo: 'doublezero-edge-connect',
+              event_type: 'doublezero-base-published',
+              client_payload: { env: 'mainnet-beta', version: '${{ steps.v.outputs.version }}' },
+            });
 
   # Validation path: on PRs touching the image, build all three variants for
   # amd64 but do NOT push, to catch Dockerfile/entrypoint regressions before
--- a/.github/workflows/release.docker.client.yml
+++ b/.github/workflows/release.docker.client.yml
@@ -39,6 +39,11 @@ jobs:
   publish:
     if: github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success'
     runs-on: ubuntu-latest
+    # Serialize testnet publishes so two closely-spaced releases don't race the
+    # mutable :testnet tag.
+    concurrency:
+      group: release-docker-client-testnet
+      cancel-in-progress: false
     steps:
       - uses: actions/checkout@v4
         with:
@@ -112,7 +117,12 @@ jobs:
       # base image we just published (event-driven; its daily poll is the fallback).
       - name: Notify edge-connect to rebuild (testnet)
         if: steps.v.outputs.publish == 'true'
+        # A failed cross-repo dispatch must not fail an already-successful publish;
+        # edge-connect's daily poll is the fallback.
+        continue-on-error: true
         uses: actions/github-script@v7
+        env:
+          DZ_VERSION: ${{ steps.v.outputs.version }}
         with:
           github-token: ${{ secrets.EDGE_CONNECT_DISPATCH_TOKEN }}
           script: |
@@ -120,7 +130,7 @@ jobs:
               owner: context.repo.owner,
               repo: 'doublezero-edge-connect',
               event_type: 'doublezero-base-published',
-              client_payload: { env: 'testnet', version: '${{ steps.v.outputs.version }}' },
+              client_payload: { env: 'testnet', version: process.env.DZ_VERSION },
             });
 
   # mainnet-beta image publish: manual (workflow_dispatch) or automatic (schedule).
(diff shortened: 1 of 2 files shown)


ben-dz (commented):
Auto-publish of the mainnet-beta base image plus cross-repo edge-connect notifications. Well-structured and internally consistent — the idempotency guard's label comparison is sound (BUILD_VERSION = raw deb version = `want`, and the Dockerfile label is `${BUILD_VERSION}`), and the fork/PR trust boundary is correct (secrets confined to event-gated publish jobs; `pull_request` reaches only the push-less `build-only` job). No critical or high-confidence blocking defects. A set of Medium robustness/hardening items worth addressing: (M1) no `concurrency:` group so a scheduled run and a manual dispatch can race the mutable `:latest`/`:mainnet-beta` tags; (M2) the notify step lacks `continue-on-error`, so a failed cross-repo dispatch marks an already-successful publish as failed and a re-run force-republishes `:latest`; (M3) the dispatch `version` payload is semver for testnet but raw deb version for mainnet-beta; (M4) mainnet-beta uses raw `apt-cache madison | head -1` with no stability filter, relying entirely on Cloudsmith channel hygiene. Low: operator-only expression-injection hardening, fail-open guard, and a cosmetic tag/label format mismatch. Note: no sibling release workflow uses `concurrency`/`continue-on-error`, so M1/M2 are net-new improvements, not convention violations. Verify before merge: the `EDGE_CONNECT_DISPATCH_TOKEN` secret exists and the edge-connect receiver is on its default branch (both in the PR's stated sequence).

- Pre-merge (operator): the `EDGE_CONNECT_DISPATCH_TOKEN` secret must exist before merge, or every scheduled publish will push the image then fail the job on the notify step. Listed in the PR's stated sequence.
- Pre-merge (operator): the edge-connect receiver workflow (event_type `doublezero-base-published`) must be on that repo's default branch — `repository_dispatch` receivers only fire from the default branch. Listed in the PR's stated sequence.
- Note: the daily `schedule` only fires once this file is on `main`; it will not run from the PR branch. Expected.

- `.github/workflows/release.docker.client.yml:146` — medium: mainnet-beta version resolution uses raw `apt-cache madison doublezero | head -1` with no stability filter, unlike the testnet path which pins to the released semver (`grep -E '^${SEMVER}-'`, line 85). `apt-cache madison` ordering isn't a hard version-sort guarantee, and if the stable `doublezero` channel ever contains a pre-release/RC/epoch entry that lists first, the daily schedule would auto-publish it to `:mainnet-beta` and `:latest` unattended. If channel hygiene (only promoted stable versions land here) is guaranteed by the release process, document that invariant in the comment; otherwise add a semver/stability filter mirroring the testnet path.
- `.github/workflows/release.docker.client.yml:139` — low: `VERSION="${{ github.event.inputs.version }}"` interpolates operator-controlled free text directly into `run:` bash, and `steps.v.outputs.version` later reaches a single-quoted github-script JS literal at line 215. Not an exploitable injection — the input is only settable via `workflow_dispatch`, which requires write access (attacker == victim) — but it's fragile: a benign operator typo containing `'` or `;` breaks the run. Optional hardening: pass the input via `env:` and reference `"$VERSION"`, and pass `version` to github-script via env rather than a raw literal.
- `.github/workflows/release.docker.client.yml:150` — low: The version-suffixed immutable tag sanitizes `~`/`:` → `-` (`tr '~:' '--'`), while `:mainnet-beta`/`:latest`, the `org.opencontainers.image.version` label, and the dispatch payload all carry the raw version. No functional bug (Docker tags can't contain `~`/`:`), but anyone reverse-mapping the immutable `:mainnet-beta-<tag>` to a deb version must reverse the `tr`. Only observable if a version ever contains `~` or an epoch `:`. Worth a one-line comment.

.github/workflows/release.docker.client.yml
```diff
@@ -136,7 +157,29 @@ jobs:
...
+            '${{ env.IMAGE }}:mainnet-beta' 2>/dev/null || true)"
+          echo "published mainnet-beta version: ${have:-<none>} ; stable channel latest: $want"
+          if [ '${{ github.event_name }}' = 'workflow_dispatch' ]; then
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+          elif [ "$want" != "$have" ]; then
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "changed=false" >> "$GITHUB_OUTPUT"
+            echo "mainnet-beta image already at $have; nothing to do"
+          fi
```
ben-dz, line 204: No `concurrency:` group on this job. The gate reads `have` at job start (line 168), then build+push takes minutes (line 181). A daily `schedule` run and a manual `workflow_dispatch` (which always sets `changed=true`) can overlap and both push the mutable `:mainnet-beta`/`:latest` tags — last-writer-wins, and the two edge-connect dispatches can arrive out of order. The guard is a best-effort optimization, not a lock. Add a job-level `concurrency: { group: publish-mainnet-beta, cancel-in-progress: false }` — not `true`, since cancelling mid-push can leave a partially-updated multi-tag set.

.github/workflows/release.docker.client.yml
```diff
@@ -157,6 +200,21 @@ jobs:
...
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.EDGE_CONNECT_DISPATCH_TOKEN }}
+          script: |
+            await github.rest.repos.createDispatchEvent({
+              owner: context.repo.owner,
+              repo: 'doublezero-edge-connect',
+              event_type: 'doublezero-base-published',
+              client_payload: { env: 'mainnet-beta', version: '${{ steps.v.outputs.version }}' },
+            });
```
ben-dz, line 246: This notify step (and the testnet one at line 113) has no `continue-on-error`. If `createDispatchEvent` fails *after* the image is already pushed (token unset/expired, receiver not yet on edge-connect's default branch, API 5xx, rate limit), the job goes red even though the image published correctly. A re-run then hits the always-publish dispatch path (line 172-173) and force re-pushes `:latest`/`:mainnet-beta` plus fires a second dispatch. Notify is an optimization with a documented daily-poll fallback — add `continue-on-error: true` to both notify steps so it can't fail the publish.

.github/workflows/release.docker.client.yml
```diff
@@ -157,6 +200,21 @@ jobs:
...
+        if: steps.gate.outputs.changed == 'true'
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.EDGE_CONNECT_DISPATCH_TOKEN }}
+          script: |
+            await github.rest.repos.createDispatchEvent({
+              owner: context.repo.owner,
+              repo: 'doublezero-edge-connect',
+              event_type: 'doublezero-base-published',
+              client_payload: { env: 'mainnet-beta', version: '${{ steps.v.outputs.version }}' },
```
ben-dz: The `client_payload.version` here is the raw deb version (e.g. `0.28.0-1`, from `apt-cache madison … $3` on line 149), whereas the testnet dispatch at line 123 sends the semver (e.g. `0.28.0`, from `${TAG#client/v}`). The same payload field carries two different formats depending on env. Confirm the edge-connect receiver normalizes both, or make them consistent.

.github/workflows/release.docker.client.yml
```diff
@@ -136,7 +157,29 @@ jobs:
...
+      # :mainnet-beta image is at the same version the stable channel resolves to.
+      # A manual workflow_dispatch always publishes (override).
+      - name: Decide whether a publish is needed
+        id: gate
+        shell: bash
+        run: |
+          want='${{ steps.v.outputs.version }}'
+          have="$(docker buildx imagetools inspect \
+            --format '{{ index .Image.Config.Labels "org.opencontainers.image.version" }}' \
+            '${{ env.IMAGE }}:mainnet-beta' 2>/dev/null || true)"
```
ben-dz, line 195: `docker buildx imagetools inspect … 2>/dev/null || true` collapses all failures (registry 5xx, auth blip, missing label, first-ever run) into an empty `have`, so `want != ""` → republish. Correct for the genuine first-run case, but it defeats the idempotency guard exactly when the registry is flaky (spurious rebuild + spurious edge-connect dispatch of identical content). Optional: distinguish 'image/label absent' from 'inspect errored', or note the fail-open intent in the comment.

juan-malbeclabs:
Thanks for the thorough review — pushed 1f4ee9a4 addressing the actionable items.

- **M1 (concurrency race):** Added per-job `concurrency` groups (`cancel-in-progress: false`) to both `publish` and `publish-mainnet-beta`, so a scheduled run and a manual dispatch serialize instead of racing the mutable `:mainnet-beta`/`:latest` (and `:testnet`) tags.
- **M2 (notify fails a successful publish):** Both "Notify edge-connect" steps are now `continue-on-error: true`. A failed cross-repo dispatch no longer marks an already-successful publish as failed, which also removes the re-run-force-republishes footgun (edge-connect's daily poll is the fallback).
- **M4 (no stability filter):** `apt-cache madison doublezero | head -1` now pipes through `grep -v '~'` to defensively exclude Debian pre-release markers so an RC can never be auto-promoted to `:mainnet-beta`/`:latest`. Added a comment documenting the reliance on stable-channel hygiene.
- **Low (expression injection / fragility):** Operator input now flows through `env: INPUT_VERSION` instead of being interpolated into bash, and the dispatch `version` is passed to github-script via `env: DZ_VERSION` (`process.env.DZ_VERSION`) rather than a raw literal.
- **Low (tag/label format mismatch):** Added a comment clarifying that the `tr '~:' '--'` sanitization applies only to the immutable version-suffixed tag; the label and dispatch payload keep the raw version.

**M3 (payload version format differs — semver for testnet, raw deb version for mainnet-beta):** Left as-is intentionally. The two paths carry genuinely different version identifiers — testnet is triggered by a `client/v*` release tag (semver), while mainnet-beta resolves the deb version from the stable channel. Normalizing would change the semantics the edge-connect receiver consumes; happy to align both to the deb version in a follow-up if that's preferred.

Pre-merge operator prerequisites (`EDGE_CONNECT_DISPATCH_TOKEN` secret and the edge-connect receiver on its default branch) remain as noted in the PR description.

.github/workflows/release.docker.client.yml
juan-malbeclabs: Done in 1f4ee9a4 — added job-level `concurrency: { group: release-docker-client-mainnet-beta, cancel-in-progress: false }` (and an equivalent group on the testnet `publish` job). Used `false` as you noted, to avoid cancelling mid-push and leaving a partially-updated multi-tag set.

.github/workflows/release.docker.client.yml
juan-malbeclabs: Done in 1f4ee9a4 — added `continue-on-error: true` to both notify steps (testnet and mainnet-beta). A failed dispatch after a successful push no longer reds the job, so the re-run force-republish path can't be triggered by a notify failure; the daily poll remains the fallback.

.github/workflows/release.docker.client.yml
juan-malbeclabs: Left as-is intentionally. The two paths carry genuinely different version identifiers — testnet is triggered by a `client/v*` release tag (semver), while mainnet-beta resolves the deb version from the stable channel — so normalizing here would change the semantics the edge-connect receiver consumes. Happy to align both to the deb version in a follow-up if that's preferred.

.github/workflows/release.docker.client.yml
juan-malbeclabs: This is fail-open by design — on any inspect failure we'd rather republish identical content than silently skip a real update. The mainnet-beta version resolution now also filters pre-releases (`grep -v '~'`), and republishing the same content is harmless beyond a spurious dispatch. Kept the fail-open behavior; can add a clarifying comment on the `|| true` if you'd like it spelled out.

ben-dz (approved):

PR #3991, opened 2026-07-06
Add dz-ledger-program-review skill for on-chain program reviews

nikw9944:
## What

Adds a repo-local Claude Code skill at `.claude/skills/dz-ledger-program-review/SKILL.md` that captures this repo's review conventions for its on-chain programs (serviceability, geolocation). Claude Code auto-discovers it and applies it when reviewing on-chain program changes in this repo.

## Where it comes from

The guidance is distilled from this repo's own PR review history (the serviceability/geolocation program PRs) — 15 feedback classes (account & signer validation, PDA seeds+bump, access control, instruction validation, checked arithmetic, rent/space, (de)serialization, testing, error handling, etc.), each with a concrete "what to look for in a diff" checklist and real verbatim review quotes linked to their PRs.

## Why it's scoped to this repo

These programs use the DoubleZero **Ledger** serialization model — **borsh serde + a 1-byte `AccountType` enum**. The skill deliberately **excludes Solana zero-copy conventions** (bytemuck, 8-byte precomputed discriminators, `ZeroCopyAccount` loaders), which belong to the separate Solana program repos and would be actively wrong here. Keeping the skill in-repo means a reviewer only ever sees this repo's conventions — there's no blended, cross-network rubric to mis-apply.

Sibling repos (`doublezero-shreds`, `doublezero-solana`) get their own repo-local skills scoped to their Solana model.

## Notes

- No behavior/code change — documentation/tooling only.
- Optional follow-up: a one-line pointer in `CLAUDE.md` for discoverability (skills are auto-discovered regardless).

### DIFF
--- a/.claude/skills/dz-ledger-program-review/SKILL.md
+++ b/.claude/skills/dz-ledger-program-review/SKILL.md
@@ -1,0 +1,361 @@
+---
+name: dz-ledger-program-review
+description: Use when reviewing on-chain program (Rust) code or PRs in this repo — the DoubleZero Ledger serviceability and geolocation programs (borsh serde + AccountType enum). Complements smartcontract/programs/CLAUDE.md (the canonical rules) with diff-scan checklists, real review examples, severity guidance, and a review process. Covers sibling-processor parity, account/signer validation, PDA seeds+bump, access control, instruction validation, checked arithmetic, rent/space/heap caps, error handling, off-chain RPC hygiene, and testing.
+---
+
+# DoubleZero Ledger program review
+
+This skill distills the review conventions actually applied to on-chain program PRs in `malbeclabs/doublezero` — specifically the DoubleZero **Ledger** programs under `smartcontract/programs/` (the serviceability and geolocation programs). Apply it when reviewing Rust that touches these programs or their instruction processors, state accounts, and client/off-chain helpers.
+
+**Source of truth & precedence.** `smartcontract/programs/CLAUDE.md` is the canonical statement of these programs' conventions and is always loaded in this repo. This skill does not restate those rules — it adds what CLAUDE.md doesn't: concrete "what to look for in a diff" checklists, real review examples with PR links, severity guidance, and a review process. Where a rule is codified in CLAUDE.md, this skill cites it (e.g. "CLAUDE.md Security §1") instead of repeating it. **On any conflict, CLAUDE.md wins** — treat it as correct and flag the drift. A few checks below are curated from reviewer guidance not yet quote-backed by a mined PR; they are marked *(reviewer-curated)*.
+
+**Serialization model — read this first.** These programs use **borsh serde for account (de)serialization** with a **custom 1-byte `AccountType` enum discriminator**. Accounts are read via `T::try_from(&AccountInfo)`, which borsh-deserializes the data and checks the **discriminant only** (`account_type == AccountType::X`). It does **not** check `account.owner` — every processor must verify ownership separately (CLAUDE.md Security §1), with the singleton exception (ProgramConfig, GlobalState) where the discriminant check alone suffices. Do not assume `try_from` pins identity. State creation/mutation/close goes through the `try_acc_create` / `try_acc_write` / `try_acc_close` helpers. Authority is enforced with `GlobalState` + `authorize(...)` + `permission_flags`. PDAs are derived with `find_program_address`, with the bump cached on the account only when it must sign. Errors use the `DoubleZeroError` custom enum, and validation is done with `assert!` / `assert_eq!` and `return Err(DoubleZeroError::…)`.
+
+**This is NOT a bytemuck / zero-copy program.** Solana zero-copy conventions (fixed Pod layouts, 8-byte precomputed discriminators, `try_next_accounts` account loaders, etc.) do **not** apply here and should never be suggested — the discriminator is a plain 1-byte `AccountType` enum and accounts are borsh-serialized. If a suggestion only makes sense for a zero-copy program, it is wrong for this repo.
+
+Review window for these conventions: **2025-07 through 2026-03**.
+
+## How to apply — severity and voice
+
+Report findings at their real weight; a review that flags everything equally is noise.
+
+- **Blocking** (fix before merge): real correctness bugs; a missing owner check (CLAUDE.md Security §1); arithmetic over/underflow on counts, seats, or lamports; a collection cap not enforced at the mutating instruction (heap-brick risk); a missing changelog entry (CI-enforced).
+- **Nit** (raise, non-blocking): naming/idiom, debug/`Display` representations, other style items — label them as nits.
+- **Follow-up** (defer to a separate PR): broad cosmetic sweeps unrelated to the change.
+
+Match the voice the examples model:
+- Frame judgment calls as questions ("when would this ever be true?", "should this be `checked_sub`?"), not commands.
+- Give the exact replacement code, not just a complaint.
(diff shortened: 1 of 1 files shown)


ben-dz (commented):
## 1. State explicitly that `try_from` is not an owner check
The preamble says accounts are read via `T::try_from(&AccountInfo)`, "which borsh-deserializes the data and checks that `account_type == AccountType::X`". True (`state/device.rs:657`) — but the `AccountInfo`-level impl never looks at `account.owner` (`state/device.rs:665-676` just borrows data and deserializes). As written, the preamble invites the inference "`try_from` pins identity," which quietly defeats the skill's own owner-check class. Add one sentence: `try_from` checks the discriminant only; every processor must verify `account.owner` separately, with the singleton exception already codified in `smartcontract/programs/CLAUDE.md` (Security §1).

## 2. Foreign program ids are hardcoded, never passed in
The account-validation class says "verify the owner is the serviceability program ID" but not where that id comes from. The in-repo convention is a hardcoded const — geolocation checks `account.owner` against `crate::serviceability_program_id()` (`doublezero-geolocation/src/processors/mod.rs:21-22`, `geo_probe/create.rs:56-57`). Add a check: flag a diff that takes the expected program id from an instruction arg or a passed-in account instead of the hardcoded const — a caller-supplied "expected owner" is no check at all.

## 3. Sibling-processor parity
Serviceability is a family of parallel processors (create/update/suspend/resume/delete × device/link/user/exchange/…). A recurrent review theme: a new instruction or entity must match its established sibling — same `authorize()`/`permission_flags` gating, same status transitions, same `reference_count` handling, same validation. None of the 15 classes catches this, because each checks the diff in isolation; drift-from-sibling is its own bug class. Suggested check: for a new processor, diff it against the closest sibling processor; unexplained divergence in authorization, status handling, or counters is a finding.

## 4. Unbounded account collections vs the 32KB heap
The testing class asks for "an allowlist growing near the heap max" test, but no class flags the hazard itself: a `Vec` on an account (allowlists, interfaces, parent devices, samples) that a diff grows with no cap enforced at the mutating instruction. Deserializing an over-grown vec blows the 32KB program heap and panics every instruction that reads the account — the account is effectively bricked. Check: newly grown collection → where is the cap at the mutating processor, and where is the near-cap test.

## 5. A severity & framing section
The skill lists classes but never says what blocks versus what's a nit, so a skill-driven review will emit everything at equal weight — the opposite of how these reviews actually read. Suggest a short section: blocking = real bugs, missing owner checks, arithmetic hazards, caps not enforced at the mutating instruction, missing changelog; naming/idiom/debug-representation classes = nits; cosmetic sweeps = follow-up PR. Plus the style the verbatim quotes model but the skill never states: frame judgment calls as questions, give the exact replacement code, prove serialization claims with a runnable `#[test]`, and distinguish "more than just a suggestion" from "up to you".

## 6. The converse arithmetic checks
The checked-arithmetic class only pushes toward `checked_*`. The same review history also prunes in the other direction: question a `checked_*`/guard that can never fail ("seems impossible for this `checked_sub` to be `None`, right?"), question `wrapping_*` where wraparound isn't actually intended, and accept bare `+= 1` where overflow is physically impossible (u64 epoch/index counters). Without the converse, the class contradicts the skill's own redundant-validation philosophy and generates noise findings.

## 7. Off-chain transaction/RPC hygiene
The skill's stated scope includes "client/off-chain helpers," but its one RPC class is entirely derived from #729. Recurring asks it misses:

- No `.unwrap()`/`.expect()` on RPC-derived data — a network blip becomes a process panic; `?`-propagate in `Result`-returning functions.
- Fetch the recent blockhash inside the send method, not as a staleable argument.
- `get_multiple_accounts` over N single fetches.
- Pre-flight existence checks (account/ATA) and only add the create instruction if needed — a bare revert is poor UX.
- Don't send zero-value/no-op transactions.
- Don't set a CU price on uncontended DZ-ledger transactions; build the CU limit from real costs with headroom.

## 8. Minor
- Test-setup realism: reach target state via real instruction sequences, not `set_*`/poke helpers; flag feature-gated tests that never actually run under the default CI build.
- Naming: `_info` suffix for `AccountInfo` bindings; constants named generically, not after one incidental use.

## Note: overlap with `smartcontract/programs/CLAUDE.md`
Several classes are already codified there (assert-specific-errors, owner verification including the singleton exception, `repr(u32)` errors, no stored bumps, no redundant args, `BorshDeserializeIncremental`). Two documents will drift; suggest the skill cite `programs/CLAUDE.md` as the source of truth for shared items, or state which wins on conflict.

nikw9944:
Thanks — this was really useful, especially the pointer to `programs/CLAUDE.md`. Pushed a revision that addresses all of it. Point-by-point:

**Structural (the CLAUDE.md overlap note):** Added a **"Source of truth & precedence"** note up top — `programs/CLAUDE.md` is canonical for the rules it codifies, this skill defers to it and cites it (e.g. "CLAUDE.md Security §1") instead of restating, and **on conflict CLAUDE.md wins**. Trimmed the overlapping guidance (owner checks, system-program, `repr(u32)`, no-stored-bumps, no-redundant-args, borsh derives, assert-specific-errors, full-struct-eq, integration tests) down to citations while keeping the diff-scan checklists + examples.

1. **`try_from` is not an owner check** — fixed the preamble: it checks the discriminant only, *not* `account.owner`; every processor must verify ownership separately, with the singleton exception per CLAUDE.md §1. Added "do not assume `try_from` pins identity."
2. **Foreign program ids hardcoded** — added to account-validation: the *expected* owner must be a hardcoded const (`serviceability_program_id()`), never an instruction arg / passed account; a caller-supplied "expected owner" is no check. New diff-scan bullet.
3. **Sibling-processor parity** — new class (placed first): diff a new processor against its closest sibling; unexplained divergence in `authorize()`/`permission_flags`, status transitions, or `reference_count` handling is a finding.
4. **Unbounded collections vs 32KB heap** — added to the rent/space class (renamed "…/ heap caps"): a diff-grown account `Vec` must have its cap enforced *at the mutating instruction* (+ near-cap test), or an over-grown vec bricks the account by blowing the heap on every read.
5. **Severity & framing** — new **"How to apply — severity and voice"** section: blocking vs nit vs follow-up, plus the voice the examples model (frame as questions, give exact replacement code, prove serialization claims with a runnable `#[test]`, distinguish "more than a suggestion" from "up to you").
6. **Converse arithmetic** — added: question a `checked_*`/guard that can't fail, question `wrapping_*` where wrap isn't intended, accept bare `+= 1` where overflow is physically impossible (u64). Softened the "plain `+` is always a bug" framing.
7. **Off-chain RPC/tx hygiene** — expanded the RPC class with all six: no `unwrap`/`expect` on RPC data (`?`-propagate), blockhash inside `send`, `get_multiple_accounts`, pre-flight existence checks, no zero-value tx, no CU price on uncontended ledger + build the CU limit from real costs.
8. **Minor** — testing: reach state via real instruction sequences (not `set_*`/poke), flag feature-gated tests that never run under default CI. Naming: `_info` suffix on `AccountInfo` bindings. Also added pointers to CLAUDE.md **Permissions** (`AUTHORIZE_GATED_FLAGS` sync) and **Resource Allocation** (`resource verify` sync).

Two notes:
- A few additions (sibling-parity, the arithmetic converse, the extra RPC items) come from your review rather than a mined PR quote, so I marked them *(reviewer-curated)* rather than fabricate an example.
- The cross-cutting items (severity, sibling-parity, heap caps, hardcoded program id, RPC hygiene, test realism) apply to the two sibling PRs too (`doublezero-shreds#558`, `doublezero-solana#122`) — I'll propagate them there once this shape looks right to you.

ben-dz (approved):

PR #3995, opened 2026-07-08
daemon-cli: move latency and routes verbs into daemon-cli crate

ben-dz:
## Summary of Changes
* Move the `latency` and `routes` daemon-control verbs, plus their device-selection/latency-polling utilities (`best_latency`, `retrieve_latencies`, `select_tunnel_endpoint`, `retrieve_routes`) and a copy of `resolve_client_ip`, into the `doublezero-daemon-cli` crate per RFC-20 ([rfcs/rfc20-cli-standardization.md](../blob/main/rfcs/rfc20-cli-standardization.md)).
* Verbs now use the RFC-20 `execute<D: DaemonClient, L: LedgerClient, W: Write>` signature; output routes through the shared `writeln!(out)` helper and diagnostics through `tracing`. `LedgerClient` gains `list_device()`; the binary's `LedgerAdapter` is now generic to back it.
* Behavior is unchanged: `doublezero latency` / `doublezero routes` keep their flags, `--json`, output, and version-check semantics (only `enable`/`disable`/`status` skip the version check).
* PR 4 of 8 in the daemon-cli extraction stack (parent malbeclabs/infra#1458); prerequisite for PR 6 (connect).
* Fixes malbeclabs/infra#1490

### Deviation from the plan
The plan/issue listed `client/doublezero/src/dzd_latency.rs` for deletion, but the not-yet-migrated `connect.rs` (moves in PR 6) still imports those utilities against the binary's `ServiceController` trait — deleting now would break the build. So `dzd_latency.rs` is **kept** in the binary and the utilities are **copied** into the crate against `DaemonClient` (mirroring the plan's explicit copy-not-move decision for `resolve_client_ip`). PR 6 removes the binary copy. `routes.rs` (utility) had no other users and was deleted. Orphaned dead code left by the move (`command/util.rs`, and `RouteRecord`/`routes()` in `servicecontroller.rs`) was removed to keep `-Dwarnings` clean.

### Size note
~950 new production+test lines added; the diff is large but low in novelty — almost entirely mechanical moves with signature changes. The parent issue explicitly blessed this split exceeding the ~500-line guideline.

## Testing Verification
* Migrated the full `dzd_latency` device-selection and latency-polling test suite into the crate (converted `#[tokio::test]` → `block_on`), plus new per-verb tests for `latency`/`routes` and crate-side `resolve_client_ip` tests, all against `MockDaemonClient` + `MockLedgerClient` + `Vec<u8>` writer — 72 crate tests pass.
* `cargo test -p doublezero` passes (122 tests) — existing binary behavior unchanged after removing the standalone verb dispatch and the orphaned `RouteRecord` JSON-contract tests (equivalent coverage already lives in the crate's `client.rs`).
* Verified the version-check skip list in `main.rs` still contains only `Enable`/`Disable`/`Status`, so `latency`/`routes` remain version-checked exactly as before.


juan-malbeclabs (approved):
Approve. Faithful mechanical move verified: crate + binary compile cleanly with tests (`cargo check --tests` exit 0), `best_latency` body is identical to the retained binary copy modulo the `ServiceController→DaemonClient` signature swap, and behavior-preservation claims check out — `latency`/`routes` remain version-checked (not in the `skip_version_check` list in main.rs), and `check_daemon` is logically equivalent to the old `check_doublezero`. Good test migration and min-vs-avg ranking coverage.

Minor non-blocking nits:
- Temporary duplication of latency utils between `dzd_latency.rs` and the crate until PR 6 removes the binary copy — keep that window short so the copies don't diverge.
- Error output on pre-flight failure now routes through `tracing` rather than `spinner.println`/`eprintln!`, so the CHANGELOG's 'output unchanged' is slightly overstated for the error path (intentional RFC-20 change, cosmetic only).

juan-malbeclabs (approved):
Faithful RFC-20 relocation of `latency`/`routes` plus their utilities. Verified locally on the PR head: 72 crate tests pass, binary builds, clippy `-Dwarnings` clean on crate and binary. The migrated latency/device-selection logic is byte-identical to the retained `dzd_latency.rs` (only trait bound + param name differ), routes error semantics preserved, and the version-check skip list still excludes latency/routes so they stay version-checked. The plan deviation (keep the binary copy until connect migrates in PR 6, copy-not-move) is the correct call and clearly documented. LGTM.

PR #4000, opened 2026-07-08
serviceability: guard permission suspend/update against self-lockout

### LINKED ISSUE #3996: permission: review follow-ups (self-lockout guard, device/update account heuristic, SDK cache, CLI error handling)

## Cleanup (lower priority)

## Test gaps

juan-malbeclabs:
Resolves: #3996 (item 1)

## Summary of Changes
- Block a `PERMISSION_ADMIN` caller from suspending or modifying **their own** Permission account, mirroring the existing self-removal guard in `delete`. Since foundation lockout recovery is foundation-only, a sole non-foundation admin could otherwise permanently lock itself out.
- Reject a post-state `permissions == 0` in `update`, preserving the create-time invariant that an Activated Permission account always grants at least one flag (an all-zero Activated account is meaningless and misleads audits).
- Add integration tests for self-suspend, self-update, and clear-last-flag rejection.

## Diff Breakdown
| Core logic   |     2 | +22 / -0    |  +22 |
| Tests        |     1 | +151 / -0   | +151 |
| **Total**    |     3 | +173 / -0   | +173 |

Small, behavior-preserving authorization hardening; the bulk of the diff is test coverage.

<details>

- `smartcontract/programs/doublezero-serviceability/src/processors/permission/update.rs` — self-modification guard (`user_payer == payer`) plus post-state `permissions == 0` rejection.
- `smartcontract/programs/doublezero-serviceability/src/processors/permission/suspend.rs` — self-suspension guard mirroring `delete`.
- `smartcontract/programs/doublezero-serviceability/tests/permission_test.rs` — `test_suspend_self_rejected`, `test_update_self_rejected`, `test_update_permission_to_zero_rejected`.

</details>

## Testing Verification
- `cargo test -p doublezero-serviceability --test permission_test` — 20/20 pass, including the 3 new cases.
- New tests assert the guards specifically: a foundation key that self-suspends / self-updates is denied, and removing the last flag of a third-party permission is rejected while the sibling enable/disable paths still succeed.


smartcontract/programs/doublezero-serviceability/tests/permission_test.rs
```diff
@@ -698,6 +698,157 @@ async fn test_delete_self_removal_rejected() {
...
+        }),
+        vec![
+            AccountMeta::new(permission_pda, false),
+            AccountMeta::new_readonly(globalstate_pubkey, false),
+        ],
+        &payer,
+    )
+    .await;
+
+    assert!(
```
ben-dz: H1: This test cannot detect removal of the self-modification guard. The permission is created with only PERMISSION_ADMIN and the update removes exactly that flag, so without the guard at update.rs:72 the transaction still fails at the new permissions==0 check (update.rs:117) with the identical DoubleZeroError::InvalidArgument (Custom(65)) — even a specific-error assertion couldn't distinguish them. Fix: create with PERMISSION_ADMIN | USER_ADMIN and remove only USER_ADMIN, or use an add-only self-update (add: NETWORK_ADMIN, remove: 0); either input reaches the write path unless the self-guard exists.

smartcontract/programs/doublezero-serviceability/tests/permission_test.rs
```diff
@@ -698,6 +698,157 @@ async fn test_delete_self_removal_rejected() {
...
+        DoubleZeroInstruction::SuspendPermission(PermissionSuspendArgs {}),
+        vec![
+            AccountMeta::new(permission_pda, false),
+            AccountMeta::new_readonly(globalstate_pubkey, false),
+        ],
+        &payer,
+    )
+    .await;
+
+    assert!(
```
ben-dz: M1: All three new tests assert only is_err(), but smartcontract/programs/CLAUDE.md (Testing #1) requires asserting specific error codes — these assertions stay green if the tx fails for an unrelated reason (missing signer, NotAllowed from authorize, account-count regression). Match on TransactionError::InstructionError(0, InstructionError::Custom(65)) (DoubleZeroError::InvalidArgument). For test_update_permission_to_zero_rejected, also re-read the account and assert permissions is unchanged. Note this alone does not fix H1 (both guards share Custom(65)); the test input must change too.

smartcontract/programs/doublezero-serviceability/src/processors/permission/update.rs
```diff
@@ -66,6 +66,13 @@ pub fn process_update_permission(
         return Err(ProgramError::InvalidArgument);
     }
 
+    // Prevent self-modification: a caller with PERMISSION_ADMIN cannot modify their own
+    // permission, since removing their own flags would lock them out (recovery is
+    // foundation-only). Mirrors the self-removal guard in delete.
+    if &permission.user_payer == payer_account.key {
```
ben-dz, line 77: M2: The guard is unconditional, but foundation_allowlist members cannot be locked out by definition (foundation_permission_recovery always grants them PERMISSION_ADMIN), yet they can no longer update/suspend their own Permission account — under strict mode a sole foundation key must bootstrap a second admin key (create has no self-guard) to fix its own flags, a runbook documented nowhere. Also, the guard blocks self-*adding* flags, broader than the lockout rationale in the comment. Decide deliberately: exempt foundation_allowlist members (uniformly, delete included) or document the "keep at least two admin keys" requirement, and extend the comment to cover the add-only case if the breadth is intentional.

smartcontract/programs/doublezero-serviceability/src/processors/permission/update.rs
```diff
@@ -103,6 +110,14 @@ pub fn process_update_permission(
     }
 
     permission.permissions = (permission.permissions | value.add) & !value.remove;
+
+    // A Permission account must always grant at least one flag. `create` enforces this
+    // with `!= 0`; `update` must preserve it, as an Activated permission that grants
+    // nothing is meaningless (and would masquerade as a privileged account in audits).
+    if permission.permissions == 0 {
```
ben-dz: L1: Defined flags occupy bits 0–17, but this check doesn't mask against them: add = 1<<127 with remove = <all real flags> yields a permission granting nothing (every authorize() test is permissions & flags) that still passes — exactly the "Activated permission that grants nothing" state the comment says this prevents. Same gap pre-exists in create.rs:110 (whose comment says "at least one known flag" while checking != 0). Consider ALL_FLAGS in permission_flags and enforcing permissions & ALL_FLAGS != 0 in both create and update.

smartcontract/programs/doublezero-serviceability/src/processors/permission/suspend.rs
```diff
@@ -61,6 +61,13 @@ pub fn process_suspend_permission(
         return Err(ProgramError::InvalidArgument);
     }
 
+    // Prevent self-suspension: a caller with PERMISSION_ADMIN cannot suspend their own
```
ben-dz: L2: The comment says "a caller with PERMISSION_ADMIN cannot suspend their own permission", but the guard fires for any signer whose key equals user_payer, before authorize() runs, regardless of flags held. Reword to "no caller may suspend their own Permission account" (same imprecision in update.rs:69 and pre-existing in delete.rs:64). Consider also a one-line comment in resume.rs noting it deliberately has no self-guard (restoring access can't lock out).

ben-dz (commented):
The self-lockout guards are correct and not bypassable (PDA re-derivation + ownership + discriminator pin the target; guards run before authorize(); no wedge path — a sole admin can bootstrap a second key via create, and foundation keys always retain PERMISSION_ADMIN via foundation_permission_recovery). Requesting changes on test effectiveness, not logic: `test_update_self_rejected` cannot detect removal of the guard it tests — removing the only flag also trips the new permissions==0 check with the identical Custom(65). Also: `is_err()`-only assertions against the repo's assert-specific-errors convention, and a deliberate-decision ask on whether foundation-allowlist members (who cannot be locked out) should be exempt from the self-guard. Pre-existing follow-up for #3996: resource/closeaccount.rs closes any program account with no discriminator/PDA check, bypassing all three self-lockout guards.

- L3: Missing test for the realistic #3996 scenario — a non-foundation PERMISSION_ADMIN self-suspending with its own Permission PDA passed as the SDK's trailing account. The existing grant_permission_admin helper (permission_test.rs:853) makes this ~15 lines and pins that the guard's placement before authorize() is load-bearing. (severity: low)
- L4: CLI `doublezero permission set` on one's own key now succeeds when no PDA exists (create path, no self-guard) but fails when one does (update path) — defensible, but surprising; a CLI help note or payer==user_payer pre-check with a clear message resolves it. (severity: low)
- Out-of-scope context for #3996 follow-up (pre-existing, not a finding on this PR): resource/closeaccount.rs:33-68 closes any writable program-owned account with no AccountType discriminator or PDA check — a RESOURCE_ADMIN can close their own or anyone's Permission PDA, bypassing all three self-lockout guards. Deserves its own issue.

smartcontract/programs/doublezero-serviceability/tests/permission_test.rs
juan-malbeclabs: Fixed in 257ba670. `test_update_self_rejected` now creates the permission with `PERMISSION_ADMIN | USER_ADMIN` and removes only `USER_ADMIN`, so `permissions` stays non-zero and the write path is reachable — the self-modification guard is now the *only* thing that can reject the tx, and removing it would let the update succeed (so the test would fail, as intended).

Also added `test_suspend_self_rejected_non_foundation_admin`: a non-foundation `PERMISSION_ADMIN` whose authorization comes purely from its own Permission PDA (passed as the trailing account) tries to suspend itself. `authorize()` would accept it, so this pins that the self-guard runs *before* `authorize()` and its placement is load-bearing.

smartcontract/programs/doublezero-serviceability/tests/permission_test.rs
juan-malbeclabs: Fixed in 257ba670. Added an `assert_custom_error` helper + `INVALID_ARGUMENT` const (Custom(65)); all three new tests now assert the specific code via `TransactionError::InstructionError(_, Custom(65))` instead of `is_err()`. `test_update_permission_to_zero_rejected` also re-reads the account afterward and asserts `permissions` is unchanged.

As you noted, this alone doesn't fix H1 — that's handled separately by changing the test input so the two guards no longer share the failure path.

smartcontract/programs/doublezero-serviceability/src/processors/permission/update.rs
juan-malbeclabs: Decided deliberately: keep the guard **unconditional** and document the requirement, rather than exempt foundation-allowlist members. Reasoning: exempting them uniformly (update/suspend/delete) adds branching + test surface for a case that a two-key operating posture already covers, and the guard being unconditional is the simpler invariant to reason about.

Addressed in 16796368:
- The comment now states the guard fires for any signer whose key equals `user_payer`, before `authorize()` and regardless of flags held — so it intentionally also blocks self-*adding* flags (the breadth you flagged).
- Added an operational note: a lone foundation key cannot edit its own Permission this way and must bootstrap a second admin key via `create` (which has no self-guard); deployments should keep at least two `PERMISSION_ADMIN` keys.
- Same wording applied to suspend/delete; resume.rs now documents that it deliberately has no self-guard (restoring access can't lock out).

smartcontract/programs/doublezero-serviceability/src/processors/permission/update.rs
juan-malbeclabs: Fixed in 16796368. Added `permission_flags::ALL_FLAGS` (OR of every defined flag) and changed **both** `create` and `update` to enforce `permissions & ALL_FLAGS != 0` instead of `!= 0`, so a value built only from undefined bits (e.g. `1 << 127`) is rejected since no `authorize()` check (`permissions & flag`) could ever match it.

Added `test_update_permission_undefined_bits_only_rejected`: `add: 1 << 127, remove: USER_ADMIN` leaves a non-zero `permissions` that a bare `!= 0` would accept, but the `ALL_FLAGS` mask correctly rejects it (Custom(65)), and the account is left unchanged.

smartcontract/programs/doublezero-serviceability/src/processors/permission/suspend.rs
juan-malbeclabs: Fixed in 16796368. Reworded to "No caller may suspend/modify/delete their own Permission account" in suspend/update/delete, noting the guard fires for any signer whose key equals `user_payer`, before `authorize()` and regardless of flags held. Added a one-line comment in resume.rs explaining it deliberately has no self-guard (restoring one's own access can never cause a lockout).

ben-dz (approved):
All review findings verified as addressed at the updated head. H1: test_update_self_rejected now uses two flags and removes one, so the self-guard is the only rejection path and the test genuinely pins it. M1: all new tests assert Custom(65) via a shared helper, with post-state assertions where asked. M2: the unconditional guard (including self-adds) is documented as intentional with the two-admin-key runbook note in update.rs. L1: ALL_FLAGS mask in create and update plus an undefined-bits test. L2: guard comments corrected in all three processors and resume's deliberate lack of a guard documented. L3: new non-foundation-admin self-suspend test with the trailing PDA pins guard-before-authorize placement. L4: CLI permission set pre-checks self-modification with an actionable message and a unit test. The added scope (foundation allowlist remove: panics converted to typed InvalidFoundationAllowlist errors with isolated tests) was reviewed as fresh code and is clean. Verified locally: 26/26 program tests and 7/7 CLI set tests pass; CI green (e2e shards in progress).

- Optional nits, non-blocking: the grant_permission_admin doc comment still says "plain PERMISSION_ADMIN" though the helper now takes a flags parameter, and assert_custom_error is duplicated in permission_test.rs and foundation_allowlist_test.rs instead of living in test_helpers.rs.

PR #4001, opened 2026-07-08
serviceability: fix device update account parsing vs trailing permission

### LINKED ISSUE #3996: permission: review follow-ups (self-lockout guard, device/update account heuristic, SDK cache, CLI error handling)

juan-malbeclabs:
Resolves: #3996 (item 2)

## Summary of Changes
- Fix `process_update_device`'s optional-location detection. It used `accounts.len() > 5`, which counts the entire account slice — including the Permission PDA the SDK appends as the trailing account. A minimal update (no location change, no resources) by a caller who holds a Permission account produces 6 accounts, so `6 > 5` was true and `globalstate`/`payer` were misparsed as the two location accounts, reverting the instruction.
- The parser now peels `[payer, system, permission]` off the tail with `split_trailing_permission` (the same helper the other variable-length instructions use), then derives location presence from the leading length — which no longer includes the Permission account — and validates it against the expected count, erroring clearly on any other shape.
- Add a regression test exercising the previously-broken path: a NETWORK_ADMIN Permission holder updating a device it doesn't own in the minimal no-location shape, with its Permission PDA appended.

Latent on `origin/main` (the SDK always sends both location accounts, so the count never landed on the broken value), but the heuristic was unsound for any hand-built instruction or future SDK change.

## Diff Breakdown
| Core logic   |     1 | +47 / -17   |  +30 |
| Tests        |     1 | +63 / -0    |  +63 |
| **Total**    |     2 | +110 / -17  |  +93 |

Account-parsing correctness fix plus a targeted regression test.

<details>

- `smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs` — replace the `accounts.len() > 5` location heuristic with `split_trailing_permission` + leading-length detection.
- `smartcontract/programs/doublezero-serviceability/tests/device_contributor_binding_test.rs` — `test_update_device_network_admin_minimal_shape_with_permission_account`.

</details>

## Testing Verification
- `cargo test -p doublezero-serviceability` across `device_contributor_binding_test` (6/6, incl. the new case), `device_update_location_test`, `device_test` (9/9), and `device_onchain_allocation_test` (2/2) — covering the no-location, location-present, and resource-bearing shapes.
- The new test fails on `origin/main`'s parser (the trailing Permission account is misread as a location) and passes with this change.


ben-dz (commented):
The fix is correct and well-scoped: it replaces the unsound `accounts.len() > 5` location heuristic with the tail-anchored `split_trailing_permission` idiom used by the other variable-length instructions, and the exact leading-count check is fail-closed. All reachable account-list shapes (including adversarial ones with attacker-controlled `resource_count`) were traced — no silent-misparse or privilege-escalation path exists; permission→payer binding is enforced inside `authorize()`. The SDK always sends both location accounts and appends the permission PDA last, so no in-repo caller produces a rejected shape. Verified: 18 tests across 4 device suites pass on the branch (incl. the new regression test, which fails on main by trace), clippy and rustfmt clean.

No critical or high findings. Asks: (1) add a negative test asserting `InvalidArgument` for the new malformed-count branch per programs/CLAUDE.md ("assert specific errors"); (2) add an integration test for the SDK's most common production shape with a Permission account (locations+permission, and ideally the resource-bearing tail). Low-severity notes inline.

**Follow-ups (subsequent PR, not blocking this one):**
- Pre-existing duplicated system-program assert at `update.rs:235-239` and `242-246` (the identical `assert_eq!` twice). Per `programs/CLAUDE.md`, system-program checks are unnecessary anyway — a follow-up should delete both rather than dedupe. Same assert exists in `link/update.rs`.
- `user/update.rs:109` uses lenient `leading.len() >= 2` and silently ignores extra leading accounts; the exact-match approach in this PR is the better pattern — consider migrating the sibling.
- Helper-level inefficiencies shared by all `split_trailing_permission` call sites: the `Vec<&AccountInfo>` collect exists only to fit the `&[&AccountInfo]` signature (this call site passes the whole slice), and `get_permission_pda` (a `find_program_address` loop) is derived once in the split and again inside `authorize`. Worth a helper-level cleanup across all callers, not this PR.

- Medium — The SDK's most common production shapes with a Permission account are untested at integration level. Covered: minimal+permission (new test), locations no-permission (device_update_location_test), resources+locations no-permission (resource_extension_test). Uncovered: locations+permission (8 accounts — what the SDK emits for every update by a Permission holder, since UpdateDeviceCommand always sends both location metas) and the resource-bearing tail with permission ([dev, ctr, loc, loc, gs, gc, r0..rn, payer, sys, perm]), where the count formula must include both the usize::from(rc>0) and +2 terms. One test mirroring the resource_extension test with the payer's Permission PDA appended would pin both.

smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs
```diff
@@ -132,28 +132,60 @@ pub fn process_update_device(
...
+    } else if leading.len() == leading_without_locations + 2 {
+        true
+    } else {
+        msg!(
+            "Unexpected account count: {} leading accounts, expected {} or {}",
+            leading.len(),
+            leading_without_locations,
+            leading_without_locations + 2
+        );
+        return Err(DoubleZeroError::InvalidArgument.into());
```
ben-dz, line 176: New error branch has no test: no test exercises the leading-count-mismatch path (`DoubleZeroError::InvalidArgument`). Repo convention (smartcontract/programs/CLAUDE.md: "All error cases", "Assert specific errors") expects a negative test asserting the specific custom error for a malformed shape (e.g. a single location account = 4 leading, or a stray trailing account), locking in the "error clearly on any other shape" contract.

smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs
```diff
@@ -132,28 +132,60 @@ pub fn process_update_device(
...
 
-    let device_account = next_account_info(accounts_iter)?;
-    let contributor_account = next_account_info(accounts_iter)?;
-    // Update location accounts (old and new)
+    // Whether the optional (location_old, location_new) pair is present, derived from the
+    // leading length — now independent of the Permission account. The non-location prefix
+    // is [device, contributor, globalstate, globalconfig?, resources..]; the location pair
+    // adds exactly two accounts.
+    let leading_without_locations =
+        3 + usize::from(value.resource_count > 0) + value.resource_count;
```
ben-dz: Unchecked arithmetic on attacker-controlled input: `value.resource_count` is a borsh-decoded usize (u64) and the workspace sets no `overflow-checks` profile, so `3 + usize::from(..) + value.resource_count` wraps in release SBF builds (e.g. `resource_count = u64::MAX` wraps the expected count to 3). All wrap paths end in reverts (the globalconfig/resource reads immediately exhaust the iterator), so there is no exploit — but a `checked_add`, an upfront bound (`resource_count > 64` → InvalidArgument), or workspace-level `overflow-checks = true` would make the safety explicit and remove the debug/release divergence (test builds panic where deployed builds wrap).

smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs
```diff
@@ -132,28 +132,60 @@ pub fn process_update_device(
...
+    let has_locations = if leading.len() == leading_without_locations {
+        false
+    } else if leading.len() == leading_without_locations + 2 {
+        true
+    } else {
+        msg!(
```
ben-dz, line 163: Error message counts "leading" accounts while callers count total metas: someone debugging from a transaction log sees "6 leading accounts, expected 3 or 5" against a 9-meta instruction and must know about the payer/system/permission peel to reconcile. Consider including the total meta count (and/or whether a trailing Permission was detected) and naming the optional piece — "expected N (no locations) or N+2 (with location_old/location_new)".

smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs
```diff
@@ -132,28 +132,60 @@ pub fn process_update_device(
     accounts: &[AccountInfo],
     value: &DeviceUpdateArgs,
 ) -> ProgramResult {
-    let accounts_iter = &mut accounts.iter();
+    // Account layout:
+    //   [device, contributor, (location_old, location_new)?, globalstate,
+    //    (globalconfig)?, resource_0..resource_n, payer, system, (permission)?]
```
ben-dz: Nit: `resource_0..resource_n` in the layout comment reads as n+1 resources; with `resource_count = n` it's `resource_0..resource_{n-1}`.

smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs
juan-malbeclabs: Addressed in 2cfb9aa3. Added `test_update_device_rejects_malformed_account_count`: a no-location update with a single stray leading account ([device, contributor, <stray>, globalstate] = 4 leading) hits the new mismatch branch. It asserts the specific error `InstructionError::Custom(65)` (`DoubleZeroError::InvalidArgument`), not `is_err()`, and checks the device is unchanged — locking in the "error clearly on any other shape" contract per programs/CLAUDE.md.

smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs
juan-malbeclabs: Agreed on the analysis: `resource_count` is a borsh-decoded `usize` here and every wrap path reverts (the globalconfig/resource reads exhaust the iterator), so there's no exploit — but the debug-panics / release-wraps divergence is real. Out of scope for this account-count correctness fix; tracking it as a follow-up. An upfront bound (`resource_count > MAX` -> InvalidArgument) is likely the cheapest fix and would also tighten the shape check. Noting too that create/delete use `u8` for `resource_count` while update uses `usize` — worth reconciling in the same follow-up.

smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs
juan-malbeclabs: Valid readability point — "6 leading accounts" doesn't reconcile with a 9-meta tx log without knowing about the payer/system/permission peel. Will fold into the follow-up cleanup, including the total meta count and naming the optional piece (location_old/location_new).

smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs
juan-malbeclabs: Correct — with `resource_count = n` it's `resource_0..resource_{n-1}`. Will fix in the follow-up cleanup.

juan-malbeclabs:
Thanks for the thorough review. Disposition:

**Addressed in this PR (2cfb9aa3):**
- ✅ **Negative test** — added `test_update_device_rejects_malformed_account_count`, which drives a malformed 4-leading-account shape into the new mismatch branch and asserts the specific `Custom(65)` (`InvalidArgument`), not `is_err()`. Full `device_contributor_binding_test` suite green (7/7).

**Follow-up (separate PR, not blocking):**
- Unchecked `resource_count` arithmetic (non-exploitable — all wrap paths revert; likely an upfront bound + reconciling the `usize` vs `u8` mismatch with create/delete).
- Error-message wording ("leading" vs total meta count) and the `resource_0..resource_n` comment nit.
- The pre-existing duplicated system-program assert at `update.rs:235-239`/`242-246` (delete both per programs/CLAUDE.md) and the sibling in `link/update.rs`.
- Migrating `user/update.rs`'s lenient `leading.len() >= 2` to the exact-match pattern used here.
- Helper-level cleanup shared by all `split_trailing_permission` callers (redundant `Vec` collect + double `get_permission_pda` derivation).

The integration coverage for the SDK's most-common production shapes (locations+permission, resource-bearing tail+permission) is a fair ask — I'd prefer to land those alongside the follow-up cleanup rather than expand this scope, unless you'd like them here before merge.

smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs
juan-malbeclabs: Done in aa2b765d — folded into this PR rather than deferring. The message now reads e.g. "Unexpected account count: 6 total metas (4 leading + payer/system), expected 3 leading (no locations) or 5 (with location_old/location_new)", and flags `/permission` when a trailing Permission was peeled.

smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs
juan-malbeclabs: Fixed in aa2b765d: comment now reads `resource_0..resource_{n-1}`.

juan-malbeclabs:
Update — folded more of the review into this PR (aa2b765d), on top of the negative test:

- ✅ **Integration coverage (Medium finding)** — added `test_update_device_network_admin_locations_shape_with_permission_account`: a NETWORK_ADMIN Permission holder updating a device it doesn't own in the **locations+permission** shape (8 accounts: [device, contributor, location_old, location_new, globalstate] + payer/system/permission). Asserts the location change actually applies. This is what the SDK emits for every location-changing update by a Permission holder.
- ✅ **Error-message wording (#3)** — now reports total meta count, the leading count, whether a trailing Permission was peeled, and names location_old/location_new.
- ✅ **Comment nit (#4)** — `resource_0..resource_{n-1}`.

Full `device_contributor_binding_test` suite green (8/8); `device_update_location_test`, `device_test` (9/9), `device_onchain_allocation_test` unaffected.

**Still deferred to a follow-up** (behavior change / cross-file, out of scope here): the unchecked `resource_count` arithmetic + `usize`/`u8` reconciliation, the duplicated system-program assert cleanup, the `user/update.rs` lenient-check migration, and the shared `split_trailing_permission` helper cleanup.

ben-dz (approved):
All review feedback has been addressed at head aa2b765d. The new `test_update_device_rejects_malformed_account_count` pins the malformed-count branch with the specific `InvalidArgument` (Custom(65)) error and asserts the device is unchanged, and `test_update_device_network_admin_locations_shape_with_permission_account` pins the SDK's production 8-account locations+permission shape, asserting the location change actually applies. The error message now reports total metas, leading count, and permission detection, and the layout comment notation is fixed. The remaining low note (checked arithmetic on `resource_count`) was optional hardening — all overflow paths were previously traced to clean reverts.

Verified at aa2b765d: device_contributor_binding_test 8/8, device_test 9/9, device_update_location_test 1/1, device_onchain_allocation_test 2/2, resource_extension_test 21/21; rustfmt and clippy clean.

smartcontract/programs/doublezero-serviceability/src/processors/device/update.rs
juan-malbeclabs: Addressed in 4d22f7a1 (backward-compatible variant). Switched the count computation to `saturating_add`, so an out-of-range `resource_count` clamps to usize::MAX and falls through to the mismatch branch — identical behavior in debug and release SBF builds, no more wrap/panic divergence. Deliberately kept the field as `usize`: changing it to `u8` to match create/delete would alter the borsh wire format of a production instruction (and shift BorshDeserializeIncremental offsets), breaking existing clients — that type reconciliation stays a separate, coordinated follow-up.

juan-malbeclabs:
Update — folded the backward-compatible follow-ups into this PR (4d22f7a1):

- ✅ **Overflow hardening (#2)** — count computation now uses `saturating_add`, removing the release-wrap / debug-panic divergence without any wire-format change.
- ✅ **Redundant system-program asserts** — deleted from both `device/update.rs` (the duplicated pair) and `link/update.rs` (the single sibling), per programs/CLAUDE.md.
- ✅ **user/update.rs lenient check** — migrated `leading.len() >= 2` to an exact match (0 or 2 → else `InvalidArgument`), matching this PR's pattern.

**Verified backward-compatible:** the user-update SDK only ever emits 0 or 2 tenant metas (other-language SDKs are read-only), so the exact-match rejects no shape any real caller produces; `user_tests` already covers the 0- and 2-tenant paths. Full crate suite green: 271 passed, 0 failed; clippy + rustfmt clean.

**Kept as follow-up (out of scope / would break compat):**
- `resource_count` `usize`→`u8` reconciliation with create/delete — borsh wire-format change, needs coordinated SDK/fixture update.
- `split_trailing_permission` helper cleanup (redundant `Vec` collect + double `get_permission_pda`) — broad refactor across all 7 callers + `authorize()` signature, compute-only benefit.

juan-malbeclabs:
Follow-up — added the missing negative test for the new user/update.rs error branch (3a0dc8ad4). `test_user` now drives an `UpdateUser` with a single tenant account (leading length 1, neither 0 nor 2) and asserts it's rejected with `Custom(65)` (`InvalidArgument`) and leaves the tenant assignment unchanged — so every exact-match branch this PR introduces (device and user) now has explicit error coverage. Rebased onto the latest main merge; `user_tests`, `device_contributor_binding_test`, and `link_wan_test` green.

PR #4002, opened 2026-07-08
sdk: invalidate permission-account cache on create/delete

### LINKED ISSUE #3996: permission: review follow-ups (self-lockout guard, device/update account heuristic, SDK cache, CLI error handling)

juan-malbeclabs:
Resolves: #3996 (item 3)

## Summary of Changes
- Fix `DZClient::resolve_permission_account`, which memoized the "no Permission account exists" result for the client's entire lifetime with no invalidation. A long-lived client (daemon, controlplane, test harness) that bootstraps its own Permission account and then performs another gated action in the same session kept serving the stale `None` — so the newly created PDA was never injected as the trailing account and `authorize()` silently fell back to the legacy allowlist (or failed).
- Drop the memoized lookup after a `CreatePermission`/`DeletePermission` this client lands, so the next authorized transaction re-resolves the PDA. `UpdatePermission` does not invalidate (it changes the flags bitmask, not whether the account exists; the cached value is only the PDA address).

The positive-cache optimization is preserved (legacy-key operators that never hold a Permission account still resolve once), so there is no per-transaction RPC regression.

## Diff Breakdown
| Core logic   |     1 | +52 / -2    |  +50 |

Single-file client fix; the added lines are the invalidation helper, the post-success hook, and a unit test.

<details>

- `smartcontract/sdk/rs/src/client.rs` — add `instruction_invalidates_permission_cache`, clear the cache on a successful create/delete in `execute_transaction_inner`, and a unit test asserting which instructions invalidate.

</details>

## Testing Verification
- `cargo test -p doublezero_sdk --lib` — 156/156 pass, including the new `permission_cache_invalidated_only_by_create_and_delete`, which asserts create/delete invalidate while update and unrelated instructions do not.

### DIFF
--- a/smartcontract/sdk/rs/src/client.rs
+++ b/smartcontract/sdk/rs/src/client.rs
@@ -238,8 +238,9 @@ impl DZClient {
 
     /// Resolve the payer's Permission PDA as a read-only [`AccountMeta`], or
     /// `None` when no Permission account exists on-chain. The lookup is retried
-    /// on transient RPC errors and memoized until a `CreatePermission` /
-    /// `DeletePermission` this client lands invalidates it (see
+    /// on transient RPC errors and memoized until this client sends a
+    /// `CreatePermission` / `DeletePermission` (regardless of the reported send
+    /// result), which invalidates it (see
     /// [`Self::instruction_invalidates_permission_cache`]).
     fn resolve_permission_account(&self, payer: &Pubkey) -> Option<AccountMeta> {
         let mut cache = self.permission_account_cache.lock().unwrap();
@@ -294,24 +295,29 @@ impl DZClient {
             ..RpcSendTransactionConfig::default()
         };
 
-        match self
+        let send_result = self
             .client
             .send_and_confirm_transaction_with_spinner_and_config(
                 &transaction,
                 self.client.commitment(),
                 send_config,
-            ) {
-            Ok(sig) => {
-                // A CreatePermission/DeletePermission this client just landed can flip
-                // whether the payer's Permission PDA exists on-chain. Drop the memoized
-                // lookup so the next authorized transaction re-resolves it — covering the
-                // bootstrap case where a long-lived client creates its own Permission
-                // account and then performs another gated action in the same session.
-                if Self::instruction_invalidates_permission_cache(&instruction) {
-                    *self.permission_account_cache.lock().unwrap() = None;
-                }
-                Ok(sig)
-            }
+            );
+
+        // A CreatePermission/DeletePermission this client just landed can flip whether the
+        // payer's Permission PDA exists on-chain. Drop the memoized lookup so the next
+        // authorized transaction re-resolves it — covering the bootstrap case where a
+        // long-lived client creates its own Permission account and then performs another
+        // gated action in the same session. This runs regardless of the send result:
+        // skip_preflight=true means a tx can land on-chain even when the client sees an
+        // Err (confirmation timeout, connection drop), so gating invalidation on Ok would
+        // leave the stale entry — the exact silent legacy-fallback this fix closes. A
+        // spurious invalidation (tx never landed) only costs one re-resolution.
+        if Self::instruction_invalidates_permission_cache(&instruction) {
+            *self.permission_account_cache.lock().unwrap() = None;
+        }
+
+        match send_result {
+            Ok(sig) => Ok(sig),
             Err(client_err) => {
                 let tx_err = match client_err.kind.as_ref() {
                     ClientErrorKind::TransactionError(e) => Some(e.clone()),
--- a/smartcontract/sdk/rs/src/client.rs
+++ b/smartcontract/sdk/rs/src/client.rs
@@ -225,9 +225,22 @@ impl DZClient {
         ]
     }
 
+    /// Whether landing `instruction` can change whether the payer's Permission PDA
+    /// exists on-chain, so the memoized lookup must be dropped. `CreatePermission` and
+    /// `DeletePermission` flip existence; `UpdatePermission` only changes the flags
+    /// bitmask (the cached value is just the PDA address, so it is unaffected).
+    fn instruction_invalidates_permission_cache(instruction: &DoubleZeroInstruction) -> bool {
+        matches!(
+            instruction,
+            DoubleZeroInstruction::CreatePermission(_) | DoubleZeroInstruction::DeletePermission(_)
+        )
+    }
+
     /// Resolve the payer's Permission PDA as a read-only [`AccountMeta`], or
     /// `None` when no Permission account exists on-chain. The lookup is retried
-    /// on transient RPC errors and memoized for the client's lifetime.
+    /// on transient RPC errors and memoized until a `CreatePermission` /
+    /// `DeletePermission` this client lands invalidates it (see
+    /// [`Self::instruction_invalidates_permission_cache`]).
     fn resolve_permission_account(&self, payer: &Pubkey) -> Option<AccountMeta> {
         let mut cache = self.permission_account_cache.lock().unwrap();
         if let Some(cached) = cache.as_ref() {
@@ -288,7 +301,17 @@ impl DZClient {
                 self.client.commitment(),
                 send_config,
             ) {
-            Ok(sig) => Ok(sig),
+            Ok(sig) => {
+                // A CreatePermission/DeletePermission this client just landed can flip
+                // whether the payer's Permission PDA exists on-chain. Drop the memoized
+                // lookup so the next authorized transaction re-resolves it — covering the
+                // bootstrap case where a long-lived client creates its own Permission
+                // account and then performs another gated action in the same session.
+                if Self::instruction_invalidates_permission_cache(&instruction) {
+                    *self.permission_account_cache.lock().unwrap() = None;
+                }
+                Ok(sig)
+            }
             Err(client_err) => {
                 let tx_err = match client_err.kind.as_ref() {
                     ClientErrorKind::TransactionError(e) => Some(e.clone()),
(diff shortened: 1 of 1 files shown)


ben-dz (commented):
The fix is correct, minimal, and safe for what it covers: the invalidation predicate scope (create/delete yes, update no) is exactly right, both real senders go through the hooked path, the mutex prevents re-poisoning races, and the onchain authorize() is fail-closed in both stale-cache directions (verified: 156/156 lib tests, clippy and rustfmt clean). Requesting changes for one cheap item: the cache is invalidated only on the Ok arm, but with skip_preflight=true a client-side send error (confirmation timeout, connection drop) can leave a landed CreatePermission unobserved — the stale Some(None) then reintroduces the exact silent legacy-allowlist fallback this PR fixes (hard NotAllowed outage under RequirePermissionAccounts strict mode); invalidate unconditionally for create/delete. Follow-up (non-blocking): the pre-existing negative cache latches any non-retryable RPC error as "no account" — each invalidation now re-opens that window; and cross-process create/delete still never invalidates (restart-only recovery), worth a TTL or re-resolve-on-auth-failure before strict mode is enabled for long-lived Rust clients.

- (medium, follow-up — pre-existing at client.rs:250-256, introduced with the cache in #3206) The negative cache latches any non-retryable RPC failure as "no account" via .is_ok(), and this PR widens its blast radius: every create/delete now opens a fresh re-resolve window, and if that single lookup fails non-retryably — or hits a lagging load-balanced RPC node right after the create — the client re-latches the stale legacy path until the next create/delete or restart, undercutting the guarantee this PR adds. Distinguish AccountNotFound (cache the negative) from other errors (don't cache).
- (medium, follow-up — documented scope limitation at client.rs:239-258) Cross-process create/delete never invalidates and recovery is restart-only. Cross-process create of this payer's Permission account leaves a lifetime stale negative (silent legacy fallback today, NotAllowed under strict mode); cross-process delete is worse — the cached Some(meta) keeps appending a now-closed PDA and authorize() fails the owner check with InvalidAccountData on every gated action until restart. Worth deciding before RequirePermissionAccounts is enabled for long-lived Rust clients: a TTL on the negative entry only, or invalidate-and-retry-once on an authorization failure, closes both flavors.

smartcontract/sdk/rs/src/client.rs
```diff
@@ -288,7 +301,17 @@ impl DZClient {
...
                 send_config,
             ) {
-            Ok(sig) => Ok(sig),
+            Ok(sig) => {
+                // A CreatePermission/DeletePermission this client just landed can flip
+                // whether the payer's Permission PDA exists on-chain. Drop the memoized
+                // lookup so the next authorized transaction re-resolves it — covering the
+                // bootstrap case where a long-lived client creates its own Permission
+                // account and then performs another gated action in the same session.
+                if Self::instruction_invalidates_permission_cache(&instruction) {
```
ben-dz: Cache invalidation runs only on the Ok arm, but this send uses skip_preflight=true and send_and_confirm can return Err for reasons that don't mean the tx failed (confirmation timeout, blockhash-expiry ambiguity, connection drop after send) — the Err arm's own comment at line 334 says "The tx may have landed onchain" and even fetches the confirmed tx's logs by signature. A landed-but-Err CreatePermission leaves the stale Some(None) memo in place, so every subsequent gated action omits the PDA and silently falls back to the legacy allowlist — the exact staleness class this PR fixes; under RequirePermissionAccounts strict mode it becomes a hard NotAllowed outage until restart. Fix: invalidate unconditionally for create/delete (hoist this if-block above the match, or duplicate it into the Err arm). Over-invalidation costs at most one extra get_account on the next gated tx. If adopted, update the doc comment at lines 241-243 from "this client lands" to "this client sends".

smartcontract/sdk/rs/src/client.rs
```diff
@@ -906,6 +929,33 @@ mod assemble_instructions_tests {
         assert_eq!(perm.pubkey, permission_pda);
         assert!(!perm.is_signer && !perm.is_writable);
     }
+
+    /// The permission-account cache must be dropped exactly when an instruction can
+    /// change whether the payer's Permission PDA exists — otherwise a client that
+    /// bootstraps its own Permission account keeps serving the stale "does not exist"
+    /// result for the rest of its lifetime.
+    #[test]
+    fn permission_cache_invalidated_only_by_create_and_delete() {
```
ben-dz, line 961: This test only exercises the pure predicate — it re-states the matches! it tests. Nothing asserts that execute_transaction_inner actually clears permission_account_cache after a landed create/delete, so a regression that drops the if-block at lines 310-312 (or leaves it on the wrong arm) still passes the suite. Options: drive the send arms with RpcClient::new_mock/new_mock_with_mocks (no validator needed), or extract the cache-maintenance step into a small helper (e.g. note_instruction_sent(&self, &DoubleZeroInstruction)) and assert cache state around it — the field is already reachable from this in-file test module. If the Err-arm fix lands, the test worth writing is "cache cleared even when send errors". Optional: add SuspendPermission/ResumePermission asserts so a future "suspend closes the account" refactor trips this test.

smartcontract/sdk/rs/src/client.rs
```diff
@@ -225,9 +225,22 @@ impl DZClient {
         ]
     }
 
+    /// Whether landing `instruction` can change whether the payer's Permission PDA
+    /// exists on-chain, so the memoized lookup must be dropped. `CreatePermission` and
+    /// `DeletePermission` flip existence; `UpdatePermission` only changes the flags
+    /// bitmask (the cached value is just the PDA address, so it is unaffected).
+    fn instruction_invalidates_permission_cache(instruction: &DoubleZeroInstruction) -> bool {
```
ben-dz, line 240: Two small doc-comment precision points: (a) invalidation is deliberately not scoped to the client's own payer — PermissionCreateArgs.user_payer may target someone else, and PermissionDeleteArgs carries no pubkey at all (the target PDA rides in the accounts list), so an admin doing bulk grants drops its own cache each time. That's harmless, conservative over-invalidation (one extra RPC), but a one-line note that it's intentional would prevent someone "optimizing" it to payer-scoped later and introducing create/delete asymmetry. (b) "the cached value is just the PDA address" — it's an Option<AccountMeta> (address plus signer/writable flags).

juan-malbeclabs:
Thanks for the thorough review — all three points are valid.

**Blocking item (Ok-only invalidation): fixed in 8e00230c.** You're right that with `skip_preflight=true` a `CreatePermission`/`DeletePermission` can land on-chain even when the client observes an `Err` (confirmation timeout, connection drop), and gating invalidation on the `Ok` arm left the stale entry — reintroducing the exact silent legacy-fallback this PR closes. Invalidation now runs before matching the send result, so it fires regardless of the reported outcome. A spurious invalidation (tx never landed) only costs one re-resolution, so this is strictly the safe direction.

**Follow-ups (non-blocking):** filed both so they don't get lost, keeping them out of this PR since they're pre-existing:
- Negative cache latching non-retryable RPC errors as "no account" via `.is_ok()` (widened by this PR's re-resolve windows): #4011
- Cross-process create/delete never invalidating (restart-only recovery; cross-process delete being the worse `InvalidAccountData` case): #4012

Both are worth resolving before `RequirePermissionAccounts` strict mode is enabled for long-lived Rust clients, as you note.

Verification after the fix: 156/156 lib tests, clippy and rustfmt clean.

smartcontract/sdk/rs/src/client.rs
juan-malbeclabs: Fixed in 0e2adda8. Invalidation is now hoisted above the match into note_transaction_sent(&instruction), so it fires regardless of the send result — a landed-but-Err create/delete no longer leaves the stale memo. Over-invalidation (tx never landed) costs at most one extra get_account on the next gated tx, as you note. Also updated the doc comment from "lands" to "sends".

smartcontract/sdk/rs/src/client.rs
juan-malbeclabs: Addressed in 6c88593d. Took the helper route: extracted the cache-maintenance step into note_transaction_sent(&self, &DoubleZeroInstruction) and added note_transaction_sent_clears_cache_only_for_create_and_delete, which asserts the actual permission_account_cache state (Some(None)/Some(Some(meta)) -> None on create/delete, untouched on update/unrelated), using RpcClient::new_mock so no validator is needed. This is the "cache cleared even when send errors" guard — the helper is what execute_transaction_inner calls on both arms, so dropping or misplacing the call now trips the test. Left SuspendPermission/ResumePermission out for now since they don't close the account today; can add if that changes.

smartcontract/sdk/rs/src/client.rs
juan-malbeclabs: Both applied in 6c88593d. (a) Added a note that the invalidation is intentionally not payer-scoped — CreatePermission may target another user_payer and DeletePermission carries no pubkey, so bulk admin grants drop the caller's own memo each time; harmless over-invalidation, and the comment warns against "optimizing" it to payer-scoped. (b) Reworded to say the cached AccountMeta records the PDA address plus its read-only flags, not a bare address.

ben-dz (approved):
All feedback from the prior review has been addressed at head 6c88593d. H1 (High): the cache is now invalidated unconditionally after the send via the new note_transaction_sent helper, called between the send and the result inspection, so a CreatePermission that lands despite a client-side Err (skip_preflight=true confirmation timeout / dropped connection) no longer leaves the stale Some(None) memo; doc comments updated from "lands" to "sends... regardless of the reported send result" with the rationale recorded. M1 (Medium): the cache maintenance was extracted into note_transaction_sent and a behavioral test now asserts the actual cache-clear transitions from both seeded states (stale-negative cleared by create, resolved-present cleared by delete, memo intact across update/unrelated) using RpcClient::new_mock. L1 (Low): the predicate doc comment now accurately describes the cached value and documents the intentionally payer-unscoped, conservative over-invalidation. Verified: 164/164 lib tests pass (both cache tests by name), clippy -Dwarnings clean, nightly rustfmt clean, PR head matches the reviewed checkout, and the rebased diff vs merge-base remains a clean single-file change (+119/−3) with no new issues introduced by the delta. The two previously agreed follow-ups remain open by design and are not findings against this PR: the pre-existing negative cache latching non-retryable RPC errors as "account does not exist" (client.rs:280-286), and cross-process create/delete never invalidating (restart-only recovery) — worth resolving before RequirePermissionAccounts strict mode is enabled for long-lived Rust clients.

PR #4027, opened 2026-07-09
serviceability/cli: accept exchange and multicast group codes in feed verbs

nikw9944:
## Summary of Changes
* The `feed` create/update/get/delete verbs accept an exchange code for `--exchange` and multicast group codes for `--group`, in addition to pubkeys, matching the rest of the CLI (e.g. `device create`).
* Resolution is parse-first via two new `helpers.rs` functions (`resolve_exchange_arg`, `resolve_multicastgroup_arg`): a valid pubkey passes through with no onchain lookup, so existing pubkey-based scripts are unaffected; only codes hit the ledger.
* Fixes malbeclabs/infra#1918

## Testing Verification
* New mockall unit tests per verb: code inputs resolve through `GetExchangeCommand`/`GetMulticastGroupCommand` into the SDK command; pubkey inputs (including a 42-char short-encoding pubkey) pass through with no `get_*` calls; unknown code errors as `Exchange not found: <input>`.
* `e2e/feed_crud_test.go` now creates the feed with `--exchange xlax` (code) and asserts the stored `feed.Exchange` equals the resolved pubkey (runs in CI; needs testcontainers/cEOS).


smartcontract/cli/src/feed/get.rs
```diff
@@ -64,3 +63,79 @@ impl GetFeedCliCommand {
...
+            GetFeedCliCommand {
+                pubkey: "feed01".to_string(),
+                exchange: Some("xchi".to_string()),
+                json: false,
+            }
+            .execute(&ctx, &client, &mut output),
+        );
+        assert!(res.is_ok(), "{res:?}");
+        let output_str = String::from_utf8(output).unwrap();
+        assert!(output_str.contains("feed01"), "{output_str}");
```
martinsander00: can we assert_eq! the exact rendered output here instead of contains("feed01")?

smartcontract/cli/src/feed/create.rs
```diff
@@ -54,3 +53,175 @@ impl CreateFeedCliCommand {
...
+        let res = block_on(
+            CreateFeedCliCommand {
+                code: "feed01".to_string(),
+                name: "Feed".to_string(),
+                exchange: "xchi".to_string(),
+                groups: vec!["mg01".to_string()],
+            }
+            .execute(&ctx, &client, &mut output),
+        );
+        assert!(res.is_ok(), "{res:?}");
```
martinsander00, line 198: same here, this only checks res.is_ok() while the pubkeys test asserts the exact signature line. nice to be consistent.

smartcontract/cli/src/helpers.rs
```diff
@@ -66,6 +68,43 @@ pub fn resolve_exchange_pk<C: CliCommand>(
...
+    }
+}
+
+/// Resolve a multicast group argument that is either a base58 pubkey or a group code.
+///
+/// A pubkey input is used as-is with no onchain lookup or validation; only a
+/// code queries the backend for the account. Classification uses a full base58
+/// decode rather than [`parse_pubkey`]'s 43-44 char window so pubkeys with
+/// leading zero bytes (shorter encodings) still pass through.
+pub fn resolve_multicastgroup_arg<C: CliCommand>(client: &C, input: &str) -> eyre::Result<Pubkey> {
```
martinsander00: resolve_multicastgroup_arg's not-found path has no test; only the exchange unknown-code case is covered. can we mirror the exchange test for this one?

smartcontract/cli/src/helpers.rs
```diff
@@ -66,6 +68,43 @@ pub fn resolve_exchange_pk<C: CliCommand>(
...
 }
 
+/// Resolve an exchange argument that is either a base58 pubkey or an exchange code.
+///
+/// Unlike [`resolve_exchange_pk`], a pubkey input is used as-is with no onchain
+/// lookup or validation; only a code queries the backend for the account.
+/// Classification uses a full base58 decode rather than [`parse_pubkey`]'s
+/// 43-44 char window so pubkeys with leading zero bytes (shorter encodings)
+/// still pass through.
+pub fn resolve_exchange_arg<C: CliCommand>(client: &C, input: &str) -> eyre::Result<Pubkey> {
```
martinsander00: can we rename these? resolve_exchange_arg and resolve_exchange_pk sit right next to each other but behave differently: _pk always queries the ledger, _arg passes valid pubkeys straight through. the suffixes do not signal that.

smartcontract/cli/src/feed/get.rs
nikw9944: Done in f805804e: the test now asserts the exact rendered table (all six rows, including account/exchange/owner pubkeys).

smartcontract/cli/src/feed/create.rs
nikw9944: Done in f805804e: the codes test now asserts the exact `Signature: ...` line like the pubkeys test.

smartcontract/cli/src/helpers.rs
nikw9944: Added test_cli_feed_create_unknown_group_code in f805804e, mirroring the exchange case (asserts "Multicast group not found: nope").

smartcontract/cli/src/helpers.rs
nikw9944: Renamed in f805804e to parse_or_resolve_exchange / parse_or_resolve_multicastgroup (parse as pubkey, else resolve the code onchain), and added a doc note on resolve_exchange_pk pointing to the pass-through variant.

nikw9944:
Addressed all four review comments in f805804e:

* Renamed the helpers to `parse_or_resolve_exchange` / `parse_or_resolve_multicastgroup` and cross-referenced them from `resolve_exchange_pk`, so the pass-through vs always-lookup split is visible in the names.
* The `get` test asserts the exact rendered table and the create-with-codes test asserts the exact signature line.
* Added an unknown-group-code test mirroring the exchange one.

No behavior changes outside the rename.

martinsander00 (approved):

PR #4028, opened 2026-07-09
serviceability: verify account type in closeaccount resource extension

### LINKED ISSUE #4009: serviceability: closeaccount closes any program-owned account without a type/PDA check (bypasses Permission self-lockout guards)
## Summary

`process_closeaccount_resource_extension` (`smartcontract/programs/doublezero-serviceability/src/processors/resource/closeaccount.rs`) closes **any** writable, program-owned account after only a `RESOURCE_ADMIN` authorization check. It never verifies that the target account is actually a `ResourceExtension` — there is no `AccountType` discriminator check and no PDA re-derivation. A `RESOURCE_ADMIN` can therefore pass an arbitrary serviceability account (e.g. a `Permission` PDA) as `resource_account` and close it.

This bypasses all three `Permission` self-lockout guards (in `update`/`suspend`/`delete`), since closing the account destroys it outright without ever reaching those code paths.

## Where

`smartcontract/programs/doublezero-serviceability/src/processors/resource/closeaccount.rs`

The handler only checks:
- L47: `resource_account.owner == program_id` (owned by the program)
- L55: `resource_account.is_writable`
- L60: `authorize(RESOURCE_ADMIN)`
- L68: `try_acc_close(resource_account, owner_account)` — closes unconditionally

Missing: an `AccountType::ResourceExtension` discriminator check and/or `get_*_pda` re-derivation to pin the account to a real ResourceExtension before closing.

## Impact

- A `RESOURCE_ADMIN` (a lower-privileged role than `PERMISSION_ADMIN`) can close any program-owned account, including `Permission` PDAs — their own or anyone else's — defeating the self-lockout protections and enabling privilege disruption.
- More broadly, any account type owned by the serviceability program is closable through this path.

## Suggested fix

Before `try_acc_close`, deserialize/verify the account is a `ResourceExtension` (via `try_from` discriminator check) and re-derive its expected PDA, mirroring the ownership + discriminator + PDA pattern used elsewhere. Add a negative test that closing a non-ResourceExtension account (e.g. a `Permission` PDA) is rejected.

## Context

Pre-existing (not introduced by #4000). Surfaced during review of #4000 (permission self-lockout guards) — see the review discussion there. Related follow-up context: #3996.

elitegreg:
## Summary

- `process_closeaccount_resource_extension` closed **any** writable, program-owned account after only a `RESOURCE_ADMIN` authorization check, never confirming the target is actually a `ResourceExtension`. A `RESOURCE_ADMIN` could pass a `Permission` PDA (or any other serviceability account) as `resource_account` and close it outright, bypassing the `Permission` self-lockout guards in `update`/`suspend`/`delete`.
- Adds a discriminator check that rejects any account whose type is not `AccountType::ResourceExtension` with `InvalidAccountType` before `try_acc_close`.

## Why a first-byte discriminator check

- **Not PDA re-derivation:** the by-pubkey close path (`CloseResourceByPubkeyCommand`, used by `doublezero resource verify --fix`) intentionally closes *orphaned* extensions whose `ResourceType` is no longer derivable from onchain state, and the account doesn't store its `ResourceType`. A type check closes the hole while preserving orphan cleanup, with no instruction/SDK/CLI signature change.
- **Not `ResourceExtensionOwned::try_from`:** it does `data.split_at(88)`, but a `Permission` account is only 83 bytes, so `try_from` would panic instead of returning a clean, assertable error. Reading the first discriminator byte is panic-safe and yields `ProgramError::Custom(9)`.

This is pre-existing (not introduced by #4000); it surfaced during that review.

Fixes #4009

## Testing Verification

- Added `test_close_resource_rejects_non_resource_extension`: creates a `Permission` PDA, attempts to close it via `CloseResource`, asserts `Custom(9)` (`InvalidAccountType`), and confirms the account survives.
- The existing happy-path close in `test_create_device_tunnel_block_resource` now also exercises the new check on a real `ResourceExtension` and continues to pass.
- `cargo test -p doublezero-serviceability --test resource_extension_test --test permission_test` — all green.

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/processors/resource/closeaccount.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/resource/closeaccount.rs
@@ -1,7 +1,8 @@
 use crate::{
     authorize::authorize,
+    error::DoubleZeroError,
     serializer::try_acc_close,
-    state::{globalstate::GlobalState, permission::permission_flags},
+    state::{accounttype::AccountType, globalstate::GlobalState, permission::permission_flags},
 };
 use borsh::BorshSerialize;
 use borsh_incremental::BorshDeserializeIncremental;
@@ -64,6 +65,21 @@ pub fn process_closeaccount_resource_extension(
         &globalstate,
         permission_flags::RESOURCE_ADMIN,
     )?;
+
+    // Verify the target really is a ResourceExtension before closing it. Without
+    // this, a RESOURCE_ADMIN could pass any writable, program-owned account (e.g.
+    // a Permission PDA) and close it outright, bypassing the Permission
+    // self-lockout guards in update/suspend/delete. See issue #4009.
+    let account_type = resource_account
+        .data
+        .borrow()
+        .first()
+        .copied()
+        .map(AccountType::from)
+        .unwrap_or(AccountType::None);
+    if account_type != AccountType::ResourceExtension {
+        return Err(DoubleZeroError::InvalidAccountType.into());
+    }
 
     try_acc_close(resource_account, owner_account)?;
 
--- a/smartcontract/programs/doublezero-serviceability/tests/resource_extension_test.rs
+++ b/smartcontract/programs/doublezero-serviceability/tests/resource_extension_test.rs
@@ -13,7 +13,7 @@ use doublezero_serviceability::{
     instructions::DoubleZeroInstruction,
     pda::{
         get_contributor_pda, get_device_pda, get_exchange_pda, get_location_pda,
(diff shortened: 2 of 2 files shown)


juan-malbeclabs (approved):
Correct, well-scoped fix. The first-byte discriminator check is panic-safe (handles zero-length accounts via `unwrap_or(AccountType::None)`) and correctly identifies the account type, closing the privilege-escalation hole while preserving orphan cleanup in the by-pubkey path (orphaned extensions still carry discriminator 12). Good targeted regression test asserting `Custom(9)` and account survival. LGTM.

PR #4029, opened 2026-07-09
sdk: don't cache RPC failures as absent permission account

### LINKED ISSUE #4011: sdk: permission negative cache latches non-retryable RPC errors as "no account"
## Context
Follow-up from PR #4002 (review).

`DZClient::resolve_permission_account` (`smartcontract/sdk/rs/src/client.rs`, ~L250-256, introduced with the cache in #3206) determines existence via:

```rust
let exists = (|| self.client.get_account(&permission_pda))
    .retry(Self::rpc_retry_builder())
    .when(Self::is_retryable_rpc_error)
    .call()
    .is_ok();
```

`.is_ok()` treats **any** non-retryable failure (not just `AccountNotFound`) as "the account does not exist" and caches that negative result.

## Why it matters now
PR #4002 makes every `CreatePermission`/`DeletePermission` invalidate the cache, opening a fresh re-resolution window each time. If that single lookup fails non-retryably — or hits a lagging load-balanced RPC node right after the create — the client re-latches the stale legacy path until the next create/delete or a restart, undercutting the guarantee #4002 adds. Under `RequirePermissionAccounts` strict mode this surfaces as a hard `NotAllowed`.

## Proposed fix
Distinguish `AccountNotFound` (cache the negative) from other errors (do not cache; let the next call retry). Only latch the negative result on a definitive "account does not exist" response.

Severity: medium. Pre-existing, but widened by #4002.

elitegreg:
## Summary

- Fix `DZClient::resolve_permission_account` latching a non-retryable RPC failure as "no Permission account" and caching it. The lookup used `get_account(...).is_ok()`, so any non-`AccountNotFound` failure was recorded as a definitive absence.
- Switch to `get_account_with_commitment`: a definitive absence comes back as `Ok(value: None)` and is cached, while a genuine RPC error returns `None` for the attempt **without** caching, so the next authorized transaction re-resolves.

## Why

Follow-up from #4002 (review) — issue #4011. #4002 makes every `CreatePermission`/`DeletePermission` invalidate the cache, opening a fresh re-resolution window each time. If that single lookup failed non-retryably — or hit a lagging load-balanced RPC node right after the create — the client re-latched the stale legacy path until the next create/delete or a restart, undercutting the guarantee #4002 adds. Under `RequirePermissionAccounts` strict mode this surfaced as a hard `NotAllowed`.

Only a definitive present/absent result is now memoized; a failure is a single legacy fallback, not a latch.

Closes #4011.

## Testing Verification

- `resolve_permission_account_caches_definitive_absence` — a successful "not found" response caches `Some(None)` so the next call is served from cache.
- `resolve_permission_account_does_not_cache_rpc_failure` — a genuine RPC error leaves the cache unresolved (`None`), proving it is no longer latched and the next call re-queries.

### DIFF
--- a/smartcontract/sdk/rs/src/client.rs
+++ b/smartcontract/sdk/rs/src/client.rs
@@ -274,20 +274,42 @@ impl DZClient {
     /// transaction failing with an authorization-shaped error that indicates the
     /// memo went stale cross-process (see [`Self::is_stale_permission_error`] and
     /// the retry loop in `execute_transaction_inner`).
+    ///
+    /// Only a *definitive* answer is cached: a successful RPC response — whether
+    /// the account is present or absent — is memoized, but a genuine RPC failure
+    /// is not. Caching a failure as "no account" would latch the legacy allowlist
+    /// path until the next invalidation or restart, which under
+    /// `RequirePermissionAccounts` strict mode surfaces as a spurious `NotAllowed`
+    /// (issue #4011).
     fn resolve_permission_account(&self, payer: &Pubkey) -> Option<AccountMeta> {
         let mut cache = self.permission_account_cache.lock().unwrap();
         if let Some(cached) = cache.as_ref() {
             return cached.clone();
         }
         let (permission_pda, _) = get_permission_pda(&self.program_id, payer);
-        let exists = (|| self.client.get_account(&permission_pda))
-            .retry(Self::rpc_retry_builder())
-            .when(Self::is_retryable_rpc_error)
-            .call()
-            .is_ok();
-        let meta = exists.then(|| AccountMeta::new_readonly(permission_pda, false));
-        *cache = Some(meta.clone());
-        meta
+        // Use `get_account_with_commitment` so a definitive "account does not exist"
+        // comes back as `Ok(value: None)` rather than an `Err`. Any `Err` here is a
+        // genuine RPC failure (retries exhausted or non-retryable), which must NOT be
+        // cached as "no account": doing so would latch the legacy allowlist path until
+        // the next create/delete invalidation or a restart (see issue #4011). Leave the
+        // cache empty so the next authorized transaction re-resolves.
+        match (|| {
+            self.client
+                .get_account_with_commitment(&permission_pda, self.client.commitment())
+        })
+        .retry(Self::rpc_retry_builder())
+        .when(Self::is_retryable_rpc_error)
+        .call()
+        {
+            Ok(response) => {
+                let meta = response
+                    .value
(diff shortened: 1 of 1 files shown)


smartcontract/sdk/rs/src/client.rs
```diff
@@ -268,20 +268,42 @@ impl DZClient {
...
         let (permission_pda, _) = get_permission_pda(&self.program_id, payer);
-        let exists = (|| self.client.get_account(&permission_pda))
-            .retry(Self::rpc_retry_builder())
-            .when(Self::is_retryable_rpc_error)
-            .call()
-            .is_ok();
-        let meta = exists.then(|| AccountMeta::new_readonly(permission_pda, false));
-        *cache = Some(meta.clone());
-        meta
+        // Use `get_account_with_commitment` so a definitive "account does not exist"
```
juan-malbeclabs, line 290: Minor (non-blocking): the rationale for not caching failures is now stated in both this inline comment and the doc comment above the function. Slightly redundant — the inline version is well-placed at the decision point, so consider trimming the doc-comment repetition. Fine to leave as-is.

smartcontract/sdk/rs/src/client.rs
```diff
@@ -1022,6 +1044,53 @@ mod assemble_instructions_tests {
...
+
+    /// Regression for #4011: a genuine RPC failure must NOT be cached as "no
+    /// account". Caching it would latch the legacy path until the next
+    /// create/delete invalidation or a restart. The cache stays unresolved so the
+    /// next call re-queries.
+    #[test]
+    fn resolve_permission_account_does_not_cache_rpc_failure() {
+        // The "fails" mock returns a malformed body that fails to deserialize into
+        // the expected account response — a non-retryable RPC error.
+        let client = client_with_mock_rpc("fails");
```
juan-malbeclabs, line 1188: Minor (non-blocking): this relies on the internal behavior of `RpcClient::new_mock("fails")` (malformed body -> deserialize error) to synthesize a non-retryable RPC failure. Reasonable and well-documented, but it couples the test to Solana mock-sender internals that could shift across SDK bumps. Acceptable given there's no cleaner injection point.

juan-malbeclabs (approved):
Clean, tightly scoped bug fix. The switch to `get_account_with_commitment` correctly separates a definitive absence (`Ok(None)`, cached) from a genuine RPC failure (`Err`, not cached), closing the negative-caching latch from #4011. Ran both new regression tests locally — they pass. Approving with two minor, non-blocking notes inline.

PR #4030, opened 2026-07-09
serviceability: per-feed billing state on the EdgeSeat FeedSeat

martinsander00:
Resolves malbeclabs/infra#1907 (part of malbeclabs/infra#1906, Phase 1 negotiated-price subscriptions).

## Summary of Changes

- Widen `FeedSeat` so each seat on an EdgeSeat access pass carries the feed's whole billing lifecycle: the current cap (`max_users`), a future cap (`max_future_users`), the `window_end` boundary where the cap flips, the `terminates_at` date where the feed is removed, and the `anniversary_day` used to recompute renewals without drift. The oracle reads these off the pass so its expiry loop needs no database.
- Thread the same fields through the provisioning inputs (`FeedSeatConfig` on the instruction, `FeedSeatProvision` in the Rust SDK).
- `SetAccessPassFeeds` validates each provisioned seat, and every reject has its own error variant so tests pin the exact branch: `max_users > 0`, `max_users >= current_users`, `max_future_users >= max_users`, `anniversary_day ∈ 1..=31`, and a future, ordered billing window (`now < window_end <= terminates_at`, where `now` is the Clock sysvar). The future cap is restricted to growth for now because shrinking it would require deciding which live users to drop when the cap flips, which is unspecified. (The `max_future_users >= max_users` rule is stricter than #1907, which said to leave `max_future_users` unconstrained; revisit if attrition-based reduction is wanted.)
- Update the Go/Python/TypeScript read SDKs and regenerate the checked-in EdgeSeat fixture to the new 52-byte `FeedSeat` layout.

**Layout change, not breaking:** this changes the `AccessPass` borsh layout for EdgeSeat passes, but `EdgeSeat` is new and not yet deployed (no on-chain passes carry the old layout), so no migration is required.

**Out of scope (separate issues):** the oracle promoting `max_users = max_future_users` at `window_end` and removing the feed at `terminates_at` (malbeclabs/infra#1911); connect-time admission still enforces against `max_users` (already wired).

## Diff Breakdown

| Core logic   |     5 | +145 / -29   | +116 |
| Scaffolding  |     5 | +65 / -9     | +56  |
| Tests        |     5 | +439 / -14   | +425 |
| Fixtures     |     3 | +34 / -5     | +29  |
| Docs         |     1 | +4 / -10     | -6   |
| **Total**    |    19 | +687 / -67   | +620 |

A small core layout + validation change; the bulk of the diff is test coverage (negative-path cases per error variant) and mechanical SDK/fixture parity across the four languages.

<details>

- [`processors/accesspass/set_feeds.rs`](https://github.com/malbeclabs/doublezero/pull/4030/files#diff-63dd5a501d3467404e1dd61da66222df861bfde6a200da3db5403b467aad51f8) — widen `FeedSeatConfig`, build each `FeedSeat` from it (preserving `current_users`), read the Clock sysvar, and add the per-seat validation (non-zero/ordered caps, `max_future_users >= max_users`, `anniversary_day` range, future+ordered window), each with a distinct error variant.
- [`sdk/serviceability/python/serviceability/state.py`](https://github.com/malbeclabs/doublezero/pull/4030/files#diff-62de0c93cd427ad79b62e8ebc78f91b6c3dd54432c55fe31e5b53e862c24b4c7) — deserialize the new `FeedSeat` layout; add a signed-i64 helper for the timestamp fields.
- [`src/state/accesspass.rs`](https://github.com/malbeclabs/doublezero/pull/4030/files#diff-80e0bb5297b02b4e5f38a82aa45cb97bdeb0a7c4b07248e6acdf53dceda15802) — widen the `FeedSeat` struct (u16→u8 caps + the four new fields) and its size test.
- [`sdk/rs/src/commands/accesspass/set_feeds.rs`](https://github.com/malbeclabs/doublezero/pull/4030/files#diff-e38b79df68fc7461ab1f7f0f3af1be9c740fed61bdc2fd520924996aa580fec8) — widen `FeedSeatProvision` and its map to `FeedSeatConfig`.
- [`sdk/serviceability/typescript/serviceability/state.ts`](https://github.com/malbeclabs/doublezero/pull/4030/files#diff-5b4745b5b07c2d9c8be441ab85aa558f94242792e3760df72a96f84c95263ebf) — deserialize the new layout (`BigInt.asIntN` for the i64 timestamps).
- [`sdk/go/serviceability/deserialize.go`](https://github.com/malbeclabs/doublezero/pull/4030/files#diff-b71f90b1b3e00de9bd820920e63f5615edb9b7d110c2b39f38895847509a7844) / [`state.go`](https://github.com/malbeclabs/doublezero/pull/4030/files#diff-057f627e5f2629b35f1976fba6eaa1b3c507c511b0f47d2f80a3a1d0c684c4a1) — widen the `FeedSeat` struct and read the 52-byte layout.

</details>

## Testing Verification

- Serviceability integration tests pass. Each validation branch now has its own error variant and its negative test asserts that exact `Custom` code (no longer a shared `InvalidArgument`): `max_users == 0`, `max_users < current_users`, `max_future_users < max_users`, `anniversary_day = 0` / `= 32`, an inverted window, an unset (zero) window, and a positive-but-already-elapsed window. The past-window case reads the bank Clock so it exercises the future check directly; other tests use far-future bounds to avoid a time-dependent flake.
- `FeedSeat` unit tests updated for the new 52-byte serialized size and the per-feed cap behavior.
- Go, Python, and TypeScript EdgeSeat fixture-parity tests pass against the regenerated `access_pass_edge_seat` fixture, confirming all four languages agree on the new borsh layout (including the signed i64 timestamp fields).


elitegreg (approved):

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_feeds.rs
```diff
@@ -153,10 +159,62 @@ pub fn process_set_access_pass_feeds(
...
+                "max_future_users {} below max_users {} for feed {}",
+                config.max_future_users,
+                config.max_users,
+                feed_key
+            );
+            return Err(DoubleZeroError::InvalidArgument.into());
+        }
+
+        // anniversary_day is a calendar day-of-month; the granter clamps to the shortest month.
+        if !(1..=31).contains(&config.anniversary_day) {
```
bgm-malbeclabs, line 190: Non-blocking: all five reject paths here (`max_users < current_users`, `max_future_users < max_users`, this `anniversary_day` range, `window_end/terminates_at <= 0`, and `window_end > terminates_at`) return the same `DoubleZeroError::InvalidArgument`, and every negative test asserts only that code. So e.g. the `anniversary_day = 0` case in `test_cannot_set_feeds_with_invalid_billing_window` would pass just as well if the `window_end` check fired instead — the tests confirm *a* rejection, not the *intended* one. The `msg!` lines disambiguate in logs, so this is a known tradeoff rather than a defect, but the PR body's claim that each case asserts "the specific `Custom` code" slightly overstates it (they're all the same code). Distinct error variants would let these tests actually pin the branch.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_feeds.rs
```diff
@@ -153,10 +159,62 @@ pub fn process_set_access_pass_feeds(
...
+                "anniversary_day {} out of range 1..=31 for feed {}",
+                config.anniversary_day,
+                feed_key
+            );
+            return Err(DoubleZeroError::InvalidArgument.into());
+        }
+
+        // window_end and terminates_at are absolute unix seconds; a non-positive value is unset or
+        // in the distant past, i.e. a feed the oracle would treat as already terminated on arrival.
+        if config.window_end <= 0 || config.terminates_at <= 0 {
```
bgm-malbeclabs: Non-blocking, confirm intent: `window_end`/`terminates_at` are only checked for positivity + ordering, not for being in the future. A *positive but already-elapsed* value slips through — which is exactly the "a feed the oracle would treat as already terminated on arrival" case the comment says it's guarding against (that rationale only holds for non-positive values). If accepting a past-but-positive timestamp is fine because the oracle is trusted and you're deliberately avoiding a `Clock` dependency, all good — just want to confirm it's intentional rather than a gap in the positivity reasoning.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_feeds.rs
```diff
@@ -141,8 +147,8 @@ pub fn process_set_access_pass_feeds(
             .map(|s| s.current_users)
             .unwrap_or(0);
 
-        // A re-provision must not set a cap below the live count carried over from the prior seat,
-        // or the seat would start over its cap (enforced once connect-time ticking lands).
+        // A re-provision must not set the current cap below the live count carried over from the
+        // prior seat, or the seat would start over its cap.
         if config.max_users < current_users {
```
bgm-malbeclabs, line 165: Non-blocking, confirm intent: for a fresh seat `current_users == 0`, so `max_users == 0` passes this check (and `max_future_users == 0` then passes the next). That provisions a zero-cap seat that admits no users. If that's an intended placeholder / "present but capped to zero" state, great — otherwise it might be worth rejecting as nonsensical input at the boundary, or dropping a comment that a zero cap is supported.

smartcontract/programs/doublezero-serviceability/src/state/accesspass.rs
```diff
@@ -26,8 +29,12 @@ pub struct FeedSeat {
         )
     )]
     pub feed_key: Pubkey, // 32
-    pub max_users: u16,     // 2
-    pub current_users: u16, // 2
+    pub max_users: u8,        // 1   cap until window_end
```
bgm-malbeclabs, line 32: Non-blocking, confirm ceiling: the `u16 -> u8` narrowing caps per-feed `current`/`max`/`future` users at 255. Fine to do here since EdgeSeat is new and undeployed (no migration needed) and it tightens the struct nicely — just confirming 255 concurrent users per SKU-feed is an acceptable product ceiling.

bgm-malbeclabs (commented):
A few non-blocking observations on the new per-feed billing validation — all surfaced as inline comments below. Nice, well-tested change overall (cap enforced at the mutating instruction, current_users preserved, clean four-language SDK parity).

smartcontract/programs/doublezero-serviceability/src/state/accesspass.rs
martinsander00: what do you think about this one? I think 255 considering this is for each feed.

juan-malbeclabs (approved):
LGTM. The three points from bgm's earlier review (distinct error variants, future-window `Clock` check, and zero-cap rejection) are all addressed in the current revision, each with its own negative test.

On the open question: agree that u8 (255) is a fine ceiling — it's a per-feed, per-access-pass concurrent-user cap, so realistic fan-out is well under that. Only caveat worth noting: since EdgeSeat is undeployed, u8 is cheap now, but it becomes effectively permanent once deployed (widening later = layout change + migration). Acceptable trade-off.

Non-blocking nit (pre-existing, not this PR): the Go read-SDK bound check `count*52 <= Remaining()` does uint32 arithmetic that could overflow for a huge `count`; only theoretical since the program caps at MAX_ACCESS_PASS_FEEDS=64.

bgm-malbeclabs (approved):

martinsander00:
@juan-malbeclabs addrssed this guy:

> Non-blocking nit (pre-existing, not this PR): the Go read-SDK bound check count*52 <= Remaining() does uint32 arithmetic that could overflow for a huge count; only theoretical since the program caps at MAX_ACCESS_PASS_FEEDS=64.

PR #4031, opened 2026-07-09
sdk: recover permission cache from cross-process create/delete

### LINKED ISSUE #4012: sdk: cross-process permission create/delete never invalidates client cache (restart-only recovery)
## Context
Follow-up from PR #4002 (review). Documented scope limitation at `smartcontract/sdk/rs/src/client.rs` ~L239-258.

The permission-account cache is only invalidated by `CreatePermission`/`DeletePermission` that **this** client sends. A create or delete of the payer's Permission PDA performed by another process is never observed, so recovery is restart-only.

## Failure modes
- **Cross-process create:** the client keeps serving the stale negative (`Some(None)`) for its lifetime → silent legacy fallback today, hard `NotAllowed` under `RequirePermissionAccounts` strict mode.
- **Cross-process delete (worse):** the cached `Some(meta)` keeps appending a now-closed PDA as the trailing account, so `authorize()` fails the owner check with `InvalidAccountData` on every gated action until the process restarts.

## Options to consider
- A TTL on the negative cache entry only (bounded staleness without re-adding a per-tx RPC on the hot legacy path), and/or
- Invalidate-and-retry-once when a gated transaction fails with an authorization / `InvalidAccountData` error.

Worth deciding before `RequirePermissionAccounts` is enabled for long-lived Rust clients (daemon, controlplane).

Severity: medium, follow-up.

elitegreg:
## Summary of Changes
- Recover the Rust client's Permission-PDA cache from cross-process create/delete, which previously required a process restart.
- Add invalidate-and-retry-once: when a gated transaction fails with an authorization-shaped error, drop the memo, re-resolve the PDA, and resend once — only if the re-resolved account actually changed, so genuine denials fail fast with no wasted resend.
- Cover both failure modes: a cross-process delete (stale entry appends a now-closed PDA → `InvalidAccountData` on every gated action) and a cross-process create (stale negative → `NotAllowed` under strict mode).

Follow-up to #4002. Resolves #4012. No change to the happy path — the extra `get_account` and resend happen only after an authorization failure.

## Diff Breakdown
| Core logic  |     1 | +198 / -94  | +104 |

Single file: `execute_transaction_inner` restructured into an at-most-two-attempts loop, two small helpers extracted, plus unit tests (~55 lines of the additions are tests).

<details>

- `smartcontract/sdk/rs/src/client.rs` — retry loop in `execute_transaction_inner`; new `parse_transaction_error` (factored out of the old error path) and `is_stale_permission_error` classifier (`InvalidAccountData` / `NotAllowed`); updated doc comments; classifier + parse round-trip tests.

</details>

## Testing Verification
- `is_stale_permission_error` unit test: fires for `InvalidAccountData` and `NotAllowed` (code derived via `ProgramError::from` so an error-enum renumber is caught), rejects other custom and non-instruction errors.
- Cross-checked against `authorize.rs`: delete path yields `InvalidAccountData`, strict-mode create yields `NotAllowed` (both retried); present-but-insufficient permission yields `NotAllowed` with unchanged state (fails fast, no resend).
- Full loop is not driven end-to-end because `RpcClient::new_mock` can't return fail-then-succeed on successive identical sends; the retry decision is covered at the unit level (classifier + the `Option<AccountMeta>` inequality).

### DIFF
--- a/smartcontract/sdk/rs/src/client.rs
+++ b/smartcontract/sdk/rs/src/client.rs
@@ -256,6 +256,10 @@ impl DZClient {
     /// connection, so gating on `Ok` would leave the stale entry — reintroducing the
     /// silent legacy-allowlist fallback this fix closes. A spurious invalidation (the
     /// tx never landed) only costs one re-resolution on the next gated transaction.
+    ///
+    /// This handles only create/delete this client itself sends. A cross-process
+    /// create/delete is recovered reactively by the retry loop in
+    /// `execute_transaction_inner` (see [`Self::is_stale_permission_error`]).
     fn note_transaction_sent(&self, instruction: &DoubleZeroInstruction) {
         if Self::instruction_invalidates_permission_cache(instruction) {
             *self.permission_account_cache.lock().unwrap() = None;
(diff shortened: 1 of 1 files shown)


juan-malbeclabs (approved):
Approved. The retry logic is correct: the `refreshed != permission` guard makes the broad `InvalidAccountData`/`NotAllowed` classification safe (resend only when PDA existence actually flipped), and resending is idempotent because we only retry on a definite failure that reverted before any state change. Cross-process delete → `InvalidAccountData` (owner check) and cross-process create → `NotAllowed` (strict-mode None path) both check out against authorize.rs. Nice touch deriving NotAllowed's code via `ProgramError::from` in the test.

Minor, non-blocking: the happy path now clones `accounts`/`permission` on every send (was a move); negligible for CLI use.

PR #4035, opened 2026-07-09
rfcs: add RFC-26 Rust instruction-builder library

juan-malbeclabs:
## Summary

Adds **RFC-26**, proposing a new Rust crate `doublezero_serviceability_instruction` that provides one pure builder function per `doublezero-serviceability` instruction — the SPL pattern (`spl-token::instruction::transfer(...)`): each builder takes resolved arguments and returns a single unsigned `Instruction`, with no RPC. The caller composes the transaction, prepends the compute-budget prelude, and signs.

This PR adds the RFC document only; no implementation.

## Why

Today the assembly logic for a serviceability instruction (PDA derivation, `AccountMeta` ordering, borsh packing) is coupled to execution — every `commands/<domain>/<verb>.rs` has an `execute()` that builds *and* signs+sends. That creates four problems the RFC sets out to fix:

- **No pure builders.** A caller that wants an unsigned `Instruction` — to batch, simulate, inspect, or sign with a different signer — has to go through the command layer, which owns the RPC client and the send path.
- **Heavy dependency to build one instruction.** Any consumer that needs instruction bytes must depend on `doublezero_sdk`, pulling `solana-client`, `tokio`, `backon`, and the rest of the RPC tree — just to lay out accounts and borsh-pack args.
- **Account ordering is duplicated and drift-prone.** The processor (`next_account_info` order) and each command (`AccountMeta` vec) independently encode the same layout; when they disagree the processor rejects the transaction at runtime. The trailing `[payer, system_program, permission?]` convention lives in `client.rs::assemble_instructions`, so every command must line its accounts up with that tail.
- **Implicit length-detection hazard.** `CreateUser`/`DeleteUser`/`CreateSubscribeUser` detect an optional trailing account via `accounts.len()`; a stray appended account corrupts that parsing, and nothing in the current API prevents it.

The fix is the split SPL uses: a pure, RPC-free instruction library beneath the RPC-bearing SDK, with the account-order convention centralized in one reviewable place (`common::build`) and backed by golden fixtures and `solana-program-test` tests.

## Key design decisions

- **Rust only** for now; Go/Python/TS parity is a later phase.
- **Pure `Instruction` return**, SPL-style — not a `Vec` or a `Transaction`. The compute-budget prelude is a separate helper.
- **Permission account is derived from the payer** (`get_permission_pda(program_id, payer)`), so it is never a caller-supplied argument. The append is deferred and re-enabled per instruction as the `authorize()` migration rolls out (shipped commented in `common.rs`), keeping the hazard out of the type.
- **Anti-drift** via a centralized `common::build`, verbatim account-layout doc-comments copied from each processor, and a `solana-program-test` suite that runs every builder against the real program.
- **Incremental rollout** (R0–R10 + fixtures), each PR ~500 lines of new code, ending with the migration of `commands/*` to delegate to the builders.

## Testing Verification

Documentation-only change. Reviewed for internal consistency and validated the RFC's technical claims against the codebase: `get_permission_pda` is payer-derived and `authorize()` recomputes the expected PDA from the payer; instruction tag numbers (CreateDevice=20, CreateLink=28, CreateSubscribeUser=59, ClearTopology=109, AssignTopologyNodeSegments=110), the `resource_count` field, the compute-budget constants (1_400_000 CU / 256 KiB heap), and the `assemble_instructions` / permission-cache behavior all match the current implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,8 @@ All notable changes to this project will be documented in this file.
   - Fix dry runs dead-ending after `stage-programs`: the job-level skip of `push-tags` transitively skipped every downstream default-condition job (a skipped ancestor poisons `success()` for the whole graph, past `verify-cloudsmith`'s own override), so `gate-programs` through `announce` never ran in dry-run mode. The tag workflow gains a `dry_run` input and the tag jobs now run as validated no-ops instead of skipping — which also means dry runs exercise the `testnet` tag-approval prompt — and `verify-cloudsmith`'s special-case condition is deleted.
 - Dependencies
   - Pin the workspace `solana-system-interface` requirement to `"3"` and `solana-loader-v3-interface` to `"6"` (were multi-major ranges `>=1,<=3` and `>=5,<=6`). The wide ranges let `cargo update --workspace` silently rebind our crates' edges onto the older major the agave tree also pulls in, making lock resolution non-deterministic (poisoned the v0.30.0 release bump). No resolved versions change. (#4043)
+- RFCs
+  - Add RFC-26 proposing a pure, RPC-free Rust instruction-builder library (`doublezero_serviceability_instruction`) with one SPL-style builder per serviceability instruction, decoupling instruction construction from signing/sending and centralizing account-order conventions.
 
 ## [v0.29.0](https://github.com/malbeclabs/doublezero/compare/client/v0.28.0...client/v0.29.0) - 2026-07-02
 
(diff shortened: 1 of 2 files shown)


elitegreg (approved):
Personally, I think new crates should go under `./crates/`. Besides that, it looks good.

juan-malbeclabs:
Thanks @elitegreg, I agree with your suggestion. We'll also need to gradually reorganize other libraries into the `/crates/` directory.

martinsander00 (approved):

martinsander00:
i like this, I think this will come in handy for the shred oracle in the near future

PR #4036, opened 2026-07-09
ci: thread testnet release Slack posts under a per-run parent in #bots

ben-dz:
## Summary of Changes
* All testnet release orchestrator Slack posts now land in one `#bots` thread per run: `preflight` posts a parent message and every subsequent post (PR links, approval pings, program-deploy call, success, failure) replies to it — mirroring what the team already does by hand when starting a deploy.
* Webhooks can't start threads (no `ts` in the response), so posts move from the two incoming webhooks to Slack Web API `chat.postMessage` via a new `slack-thread-post` composite action and a bot token. `dispatch-and-wait`'s `slack-webhook` input is replaced by `slack-bot-token`/`slack-channel`/`slack-thread-ts` (manual infra dispatches are unaffected — Slack only activates when the orchestrator passes a token).
* New: a threaded post after `open-prs` with both PR links and the gate-1 nudge, covering the two human steps that previously had no Slack signal.
* Slack delivery is best-effort: any failure (outage, missing token) emits a `::warning::` and never fails or blocks the release (`--max-time`/`--retry` bound the curl); when the parent post failed, downstream posts degrade to flat top-level messages.
* Fixes malbeclabs/infra#1925

## Pre-merge checklist
- [ ] Replace the `SLACK_CHANNEL_ID: C0853CEAUR4 (#bots) — filled.
- [ ] `SLACK_BOT_TOKEN` repo secret exists (being handled in malbeclabs/infra#1636 setup) and the bot is invited to `#bots`.

## Testing Verification
* actionlint clean on the changed workflow; YAML parse + `bash -n` clean on both composite actions.
* Exercised the jq payload construction for all four shapes — `{channel, text}`, `{channel, text, thread_ts}`, `{channel, attachments}`, `{channel, attachments, thread_ts}` — each exactly as intended, with `thread_ts` present only when a parent ts is passed and attachments only when `color` is set.
* Exercised the `slack-thread-post` script end-to-end with stubbed/blocked curl: HTTP-200-but-`ok:false` (e.g. `invalid_auth`, `channel_not_found`), curl failure, malformed `ts`, and success — all failure paths warn and exit 0 without setting the `ts` output; success emits `ts`.
* Verified every job reading `needs.preflight.outputs.thread_ts` lists `preflight` as a direct need, every job using the local composite has a prior checkout, and no reference to `SLACK_TESTNET_ALERTS_WEBHOOK`/`SLACK_INT_TECH_WEBHOOK` or the removed `slack-webhook` input remains in `.github/`.
* No workflow was dispatched and nothing was posted to Slack; live thread behavior gets validated by the next dry run after merge.

## Review follow-up candidates (not blocking)
Independent architecture/security reviews found no critical/high issues. Applied from review: curl timeout/retry, `mrkdwn_in`+`fallback` on colored attachments, Slack `ts` shape validation, `.error` sanitization, sparse checkouts. Deliberately deferred: `reply_broadcast` on action-required pings (threading is the point of this change; revisit if thread replies prove too quiet), moving the channel ID to a repo variable, and de-duplicating the inline `chat.postMessage` in `dispatch-and-wait` into the composite.


.github/actions/slack-thread-post/action.yml
```diff
@@ -0,0 +1,66 @@
...
+             then {attachments: [{color: $color, text: $text, mrkdwn_in: ["text"], fallback: ($text | split("\n")[0])}]}
+             else {text: $text} end)
+          + (if $ts != "" then {thread_ts: $ts} else {} end)')
+        RESP=$(curl -sf --max-time 15 --retry 3 --retry-max-time 60 --retry-all-errors \
+          https://slack.com/api/chat.postMessage \
+          -H "Authorization: Bearer $TOKEN" \
+          -H 'Content-type: application/json; charset=utf-8' \
+          --data "$PAYLOAD") || { echo "::warning::slack post failed (curl)"; exit 0; }
+        # chat.postMessage returns HTTP 200 with ok=false on errors; check .ok, not the curl exit code.
+        echo "$RESP" | jq -e '.ok' >/dev/null || {
```
nikw9944: When Slack (or an intermediary) returns HTTP 200 with a non-JSON body, both jq invocations print raw parse errors to the step log and the warning ends with an empty error suffix (`slack post failed: `). Redirecting jq's stderr (`2>/dev/null`) in the `.ok` check and the `.error` extraction would keep the log clean. Same pattern in `.github/actions/dispatch-and-wait/action.yml:85-87`. Behavior is still correct — it degrades to a warning and exits 0.

.github/workflows/release.testnet.yml
```diff
@@ -171,6 +203,18 @@ jobs:
...
+      # Skipped in dry runs, where push-tags doesn't run at all.
+      - name: Post tag-approval nudge to Slack thread
+        if: ${{ !inputs.dry_run }}
+        uses: ./.github/actions/slack-thread-post
+        with:
+          token: ${{ secrets.SLACK_BOT_TOKEN }}
+          channel: ${{ env.SLACK_CHANNEL_ID }}
+          thread-ts: ${{ needs.preflight.outputs.thread_ts }}
+          text: |-
+            Version PRs confirmed merged. The 9 tag jobs are now waiting on the `testnet` environment prompt: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
```
nikw9944: This nudge says the 9 tag jobs "are now waiting on the `testnet` environment prompt", but it posts while `gate-tags` is still running; the prompt only appears once `gate-tags` completes and `push-tags` starts. The gap is seconds, but a fast clicker may land on a run with no pending approval yet. "are about to wait" (or accepting the race) both work.

.github/actions/slack-thread-post/action.yml
```diff
@@ -0,0 +1,66 @@
...
+        THREAD_TS: ${{ inputs.thread-ts }}
+        COLOR: ${{ inputs.color }}
+      run: |
+        set -euo pipefail
+        # fallback is the plain-text push-notification preview; mrkdwn_in makes
+        # the attachment text render *bold* etc. like a plain message would.
+        PAYLOAD=$(jq -n --arg ch "$CHANNEL" --arg text "$TEXT" --arg ts "$THREAD_TS" --arg color "$COLOR" '
+          {channel: $ch}
+          + (if $color != ""
+             then {attachments: [{color: $color, text: $text, mrkdwn_in: ["text"], fallback: ($text | split("\n")[0])}]}
```
nikw9944: The notification `fallback` is the raw first line of the mrkdwn text, so push-notification previews for colored posts show literal asterisks (`*Testnet release v1.2.3 failed*`). Stripping `*` in the fallback expression would polish this. Optional.

nikw9944 (approved):
Well-built change; no critical or high issues. Verified: thread_ts plumbing (all seven consumers list preflight as a direct need), prior checkout before every local-composite use, no leftover webhook/slack-webhook references in .github/, all four jq payload shapes correct, and all slack-thread-post failure paths (ok:false, curl failure, malformed ts, non-JSON body, newline-in-error) warn and exit 0 without emitting ts. Confirmed C0853CEAUR4 resolves to the public #bots channel in the malbeclabs workspace. Two medium operational items: (1) merging before SLACK_BOT_TOKEN is provisioned and the bot is invited to #bots silently drops all release Slack signal including failure alerts, since the working webhooks are removed and delivery is best-effort — treat the pre-merge checklist as a hard ordering constraint; (2) the PR description is stale — it says #int-tech and asks to replace a C0XXXXXXX placeholder, but the branch already hardcodes the #bots channel ID; confirm the channel choice and update the description. Plus three low nits on log noise and message polish.

- Medium — merge sequencing: until SLACK_BOT_TOKEN is provisioned and the bot is invited to #bots, releases have no Slack signal at all, including failure alerts. The removed webhooks presumably work today; after merge every post (notify-failure included) degrades to a step warning nobody is watching, and best-effort delivery means nothing fails loudly. Treat the PR's pre-merge checklist as a hard ordering constraint: land the secret and channel invite before merging.
- Medium — the PR description contradicts the committed code on the destination channel: the body says #int-tech and the checklist asks to replace a C0XXXXXXX placeholder, but the branch already hardcodes C0853CEAUR4, which resolves to the public #bots channel (verified against the Slack workspace). Code, docs, and CHANGELOG are internally consistent on #bots. Confirm #bots is the intended destination, update the description, and point the bot-invite checklist item at #bots.

.github/actions/slack-thread-post/action.yml
ben-dz: Fixed in fc9185ab — both jq calls (the `.ok` check and the `.error` extraction) now send stderr to /dev/null in both actions, and an unparseable body warns `slack post failed: unparseable response` instead of trailing off empty.

.github/workflows/release.testnet.yml
ben-dz: Fixed in fc9185ab — reworded to "will request the `testnet` environment prompt momentarily", keeping the post at gate completion and making the wording honest about the seconds-level gap.

.github/actions/slack-thread-post/action.yml
ben-dz: Fixed in fc9185ab — fallback now strips `*` via `gsub`, so push-notification previews read clean.

PR #4037, opened 2026-07-10
daemon-cli: move multicast transport verbs into daemon-cli crate

ben-dz:
## Summary of Changes
* Move the multicast transport verbs (`multicast subscribe`/`unsubscribe`/`publish`/`unpublish`) from the `doublezero` binary into the `doublezero-daemon-cli` crate per [RFC-20](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc20-cli-standardization.md) — PR 7 of the daemon-cli extraction stack (parent: malbeclabs/infra#1458).
* The verbs use the RFC-20 execute signature (`CliContext` + generic `DaemonClient` + `LedgerClient` + writer) but are **not** `DaemonCommand` variants — the binary's `MulticastCommands` enum nests them so `doublezero multicast <verb>` invocations are unchanged. Onchain `multicast group` CRUD still routes to the serviceability crate.
* Delete the binary's `command/helpers.rs` and the `servicecontroller.rs` remnant (these verbs were their last users; the crate's `resolve_client_ip`/`DaemonClient` survive), and drop the now-orphaned deps (`hyper` stack, `chrono`, `indicatif`) from the binary. Dedupe `init_spinner` into the crate's `helpers.rs` (was copied verbatim in connect/disconnect/multicast).
* Behavior parity: message text, skip semantics, role-carry flags, warning conditions, per-group failure aggregation, and exit codes are unchanged. Informational lines now route through the stdout writer instead of the stderr spinner, matching the connect/disconnect migrations.
* Fixes malbeclabs/infra#1493

## Testing Verification
* All 21 binary unit tests ported to the crate and strengthened: they now exercise the full `execute` path through `MockDaemonClient` (the old tests bypassed IP resolution via `execute_inner`) and assert rendered writer output. Added `unsubscribe_errors_when_daemon_has_no_client_ip` and a publish partial-failure aggregation test (one group fails onchain → remaining groups still attempted, aggregated non-zero exit).
* CLI surface verified with the built binary: `doublezero multicast --help` shows the same five subcommands, `multicast subscribe --help` unchanged, and no transport verb leaks to the top level.
* Checked e2e usages (`e2e/main_test.go`, `multicast_register_test.go`): they assert exit codes only, so the stderr→stdout message move is safe.
* Four independent reviews (architecture, security, and two style/domain reviewers) found no critical/high issues; the `init_spinner` dedup and publish failure test came out of them.

## Follow-ups (pre-existing behavior deliberately preserved for parity)
* The four verbs skip `check_requirements()`/`check_daemon` preflights that `connect`/`disconnect` run — a missing keypair or env mismatch surfaces mid-loop instead of upfront.
* `finish_update` prints the "✅ Updated." line even when every per-group call failed (exit code is still non-zero).
* Duplicate group args (`subscribe g g`) issue the onchain update twice; multiple multicast users on one IP are picked in HashMap order.


client/doublezero/src/main.rs
```diff
@@ -447,18 +443,38 @@ async fn main() -> eyre::Result<()> {
...
-        // they depend on `ServiceControllerImpl` and `resolve_client_ip`.
-        Command::Multicast(args) => match args.command {
-            MulticastCommands::Group(args) => {
-                args.command.execute(&ctx, &client, &mut handle).await
+        // `MulticastGroupCommands::execute`; the daemon-coupled transport verbs
+        // (Subscribe/Unsubscribe/Publish/Unpublish) route to
+        // `doublezero-daemon-cli`, nested here to keep the
+        // `doublezero multicast <verb>` invocation.
+        Command::Multicast(args) => {
+            let daemon = DaemonClientImpl::new(
```
juan-malbeclabs, line 451: Minor (non-blocking): `daemon` and `ledger` are built unconditionally here, but the `MulticastCommands::Group` arm below uses neither. Both are cheap (a socket-path string and a borrow), so it's negligible — moving the construction into the four transport arms would make the daemon coupling more explicit. Also, this block is now byte-identical to the one in the `Command::Daemon` arm above; a small shared helper could dedupe the two.

juan-malbeclabs (approved):
Clean, faithful RFC-20 migration. Verified no dangling references to the deleted `servicecontroller`/`command` modules, the dropped `hyper`/`chrono`/`indicatif` deps are truly unused in the binary, dispatch/role-carry/warning/failure-aggregation logic matches the originals, and the only intended change is info lines moving from the stderr spinner to the stdout writer (consistent with the connect/disconnect migrations). Test coverage is stronger than before — the ported tests now exercise the full async `execute` path and assert rendered output, plus two good new cases. Approving; one minor, non-blocking note inline.

client/doublezero/src/main.rs
ben-dz: I think this is okay- I'm going to leave it as is until it annoys someone (or myself) in the future.

PR #3879, opened 2026-06-11
ci: add testnet release orchestrator

ben-dz:
## Summary of Changes

* Add `release.testnet.yml`, a `workflow_dispatch` orchestrator that drives a full testnet release from one run: preflight checks (version format, devnet daily green) → automated version-bump PRs in doublezero and infra → human gate (`testnet-release-gate`) that verifies both PRs merged → 9-component tag push → CloudSmith package polling → Solana program build (serviceability default, telemetry/geolocation `--features testnet`) and staging via infra → human gate (`testnet-program-deploy`) around the manual program deploy → onchain version verification → infra core/client deploys and QA → Slack announce, with a failure notifier on every job. Supports `dry_run` (draft PRs, no tags, check-mode deploys) and `skip_devnet_check` inputs.
* Add `scripts/release/bump-version.sh`: bumps the single workspace version in `Cargo.toml`, runs `cargo update --workspace`, and promotes the CHANGELOG `## Unreleased` section into a dated, compare-linked version section with a fresh empty skeleton above it.
* Add `.github/actions/dispatch-and-wait`, a composite action that dispatches a `workflow_dispatch` workflow in another repo with a generated `correlation_id`, locates the run by its run-name, and watches it to completion.
* Add an optional `skip_existing` input (default `false`, today's behavior unchanged) to `release.testnet.push.tags.yml` so orchestrator re-runs succeed past already-pushed tags.
* Add `docs/testnet-release.md`, the stage-by-stage runbook (what is automated vs. what the human does, dry-run mode and cleanup, recovery semantics).
* This replaces the manual multi-step testnet release process with a single resumable run that pauses at the two points that genuinely need a human: merging/approving the version PRs and deploying the Solana programs.
* The infra-side pieces this references (`scripts/bump-testnet-versions.sh`, `stage-programs.testnet.yml`, `correlation_id` inputs on the deploy/QA workflows) are runtime-only dependencies built under the sibling sub-issue of malbeclabs/infra#1636. The `RELEASE_BOT_*` / `SLACK_TESTNET_ALERTS_WEBHOOK` secrets and the `testnet-release-gate` / `testnet-program-deploy` environments will be created before first run.
* Fixes malbeclabs/infra#1637

## Testing Verification

[...]

PR #4041, opened 2026-07-10
ci: testnet release post-dry-run fixes (single tag approval, pipeline-complete guard)

ben-dz:
## Summary of Changes

Fixes from the v0.29.2 dry runs. Draft — collecting further findings as the current run progresses.

* **One approval for the tag phase instead of two, with no new environments.** Approving gate 1 already meant "both version PRs are merged" (verified immediately after approval, failing fast if clicked early); the tag jobs then raised a second, redundant `testnet` prompt seconds later. The reusable tag workflow now carries no environment at all — it is pure mechanism — and the approval prompt lives with its callers: the orchestrator'''s gate 1, and a new `approve` job (on `testnet`, one click covering all components, same as today) in the manual `release.testnet.push.tags.components` dispatcher. Net environment count unchanged; no admin setup. (The env prompt on the reusable workflow was never a boundary against repo-write users anyway — `DOUBLEZERO_PAT` is a repo-level secret — its job is operator confirmation, which both entry paths keep.) A `github.workflow_ref` caller allowlist in the tag workflow makes the binding structural: it is workflow_call-only, and any caller other than the orchestrator or the components dispatcher fails closed until added — with its own approval story — in a reviewed change.
* **Make every approval self-describing.** Gate jobs get display names that appear verbatim in the review dialog ("gate 1: approve only after BOTH version PRs are merged — this pushes the tags" / "gate 2: approve only after the programs are deployed on nyc-tn-bm2"); the Slack PR-links post says approval must wait for both merges and states what approving does; the old tag-approval nudge becomes a tag-push notice (nothing left to nudge).
* **Add a `pipeline-complete` guard job** (`if: always()`, needs every stage): fails unless every stage concluded `success`, so a run that silently skipped stages (the transitive-skip hollow-success from #4040's investigation) can never end green. Wired into `notify-failure`'s needs so hollow successes also alert Slack.
* **Guard the version-bump script against resolver side effects** (found during the v0.30.0 wet run): `cargo update --workspace` silently rebound the six onchain/sdk crates' `solana-system-interface` edges from 3.2.0 to 2.0.0 — the workspace requirement `">=1,<=3"` admits both locked majors, and incremental re-resolution folds onto the copy the agave tree pins (`^2`). The script now fails if the lock diff contains any added line beyond the members' new version lines (pure removals are logged as pruning). Tested against both the poisoned and clean bump diffs. A systemic fix for the requirement itself is being built separately.
* Runbook updated throughout: approval count (now ~6 interactions), gate semantics, dry-run prompt list, and the guard.

## Testing Verification

* Guard's jq assertion exercised locally against hollow (`qa=skipped announce=skipped` → fail) and clean (→ pass) needs payloads.
* actionlint on both workflows: only the pre-existing SC2086 (agave install idiom, untouched).
* Live validation: next dry run after merge — expect exactly one prompt at the tag phase and a `pipeline-complete` success as the final job.

Related: #4040 (dry-run graph fix), #4036 (Slack threading), #3879 (orchestrator).


.github/workflows/release.testnet.yml
```diff
@@ -515,15 +519,38 @@ jobs:
...
             ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
 
+  # Guard against hollow successes: a skipped job transitively skips everything
+  # below it with a default status condition, and the run would still conclude
+  # "success" (see #4040 — dry runs silently lost gate-programs through
+  # announce this way). This job always runs and fails unless every stage
+  # actually succeeded, so an incomplete pipeline is loud, not green.
+  pipeline-complete:
+    runs-on: ubuntu-latest
+    needs: [preflight, open-prs, gate-tags, push-tags, verify-cloudsmith, build-programs, stage-programs, gate-programs, verify-onchain, deploy-core, deploy-clients, qa, announce]
```
nikw9944: **`announce` here quietly defeats `notify-failure`'s deliberate announce exclusion.** Job-level `failure()` looks at the whole transitive dependency chain, so announce fails → `pipeline-complete` fails (`announce=failure`) → `notify-failure` fires — exactly the path the comment on `notify-failure` (lines 549–551) says is deliberately excluded; that comment is now inaccurate. Practical bad case: a release deploys successfully, `announce` fails for a non-Slack-outage reason (the `slack-thread-post` action never fails the job, but announce's checkout step can), and Slack gets a *"Testnet release vX.Y.Z failed"* alert for a release that succeeded. Dropping `announce` from this needs list loses no hollow-success coverage: announce has no `if` of its own, so it can only be skipped when `qa`/`preflight` didn't succeed — both already in the list — and its own failure still reddens the run. Alternatively keep it and update the `notify-failure` comment, accepting the misleading-alert edge case.

.github/workflows/release.testnet.yml
```diff
@@ -515,15 +519,38 @@ jobs:
...
+    if: always()
```
nikw9944: **`if: always()` makes this job fail on every cancelled run.** Cancelling a run (a normal way to abandon a release or dry run while a gate waits for approval) leaves the stages `cancelled`/`skipped`; `always()` runs the guard anyway and the assertion fails, so every deliberately-cancelled run ends with a red failed job — and depending on how `failure()` evaluates for `notify-failure` after a cancel with a genuinely failed ancestor, possibly a spurious "release failed" Slack alert too. `if: ${{ !cancelled() }}` keeps the guard's entire purpose (skipped stages in a live run still trigger it, since `skipped` ≠ `cancelled`) while staying quiet on operator cancels; GitHub's docs recommend `!cancelled()` over `always()` for exactly this.

.github/workflows/release.testnet.push.tags.yml
```diff
@@ -26,12 +26,30 @@ on:
...
   create-and-push-tag:
     name: Push ${{ inputs.version }} tag for ${{ inputs.component }}
     runs-on: ubuntu-latest
-    environment: testnet
     steps:
+      - name: Require an approval-carrying caller
+        env:
+          WORKFLOW_REF: ${{ github.workflow_ref }}
+        run: |
+          case "$WORKFLOW_REF" in
```
nikw9944, line 43: The allowlist binds to workflow *path*, not reviewed content: the patterns match `@*` (any ref), so a modified copy of `release.testnet.yml` with `gate-tags` deleted, dispatched from any branch, passes and pushes tags with no prompt at all. Before this PR the `environment: testnet` on this job prompted even for that rogue-branch caller. The PR description already accepts this threat model (`DOUBLEZERO_PAT` is a repo-level secret usable directly by any branch workflow, so the environment never protected the PAT), and I agree — flagging only so the loss of that incidental backstop is a recorded decision. If a harder line is ever wanted, it's PAT scoping (environment secret or gate-time app token), not this allowlist.

scripts/release/bump-version.sh
```diff
@@ -14,6 +14,22 @@ COUNT=$(grep -c "^version = \"$PREV\"\$" Cargo.toml)
...
 cargo update --workspace
 
+# cargo update -w re-resolves the members' dependency edges too, and a
+# requirement that admits multiple locked majors can silently rebind them
+# (observed: solana-system-interface 3.2.0 -> 2.0.0 under ">=1,<=3"). A bump
+# may only ADD the members' new version lines — any other added line is a
+# rebind/upgrade/downgrade and means a poisoned bump PR, so fail loudly
+# instead of opening it. Pure removals are cargo pruning stale lock entries
+# (legitimate normalization); log them for the PR reviewer.
+BAD_LINES=$(git diff -U0 -- Cargo.lock | grep -E '^\+[^+]' | grep -vE "^\+version = \"$NEW\"\$" || true)
```
nikw9944: `$NEW` (and `$PREV` on line 30) are interpolated into `grep -vE` patterns with unescaped dots, so the *exclusion* is slightly wider than intended (with `NEW=0.30.0`, an added `version = "0x30y0"` line would be excluded from `BAD_LINES`). Not reachable with real semver lock entries, and genuine rebinds are still caught via their `checksum` line change regardless — but `NEW_RE=${NEW//./\\.}` would make the filter exact.

nikw9944 (approved):
Solid PR that does what it says: the tag phase collapses to one approval with the prompt moved to the two entry points, the caller allowlist fails closed, the `pipeline-complete` guard correctly catches the transitive-skip hollow success from #4040, and the bump-version guard correctly distinguishes member bumps from resolver rebinds. Load-bearing mechanics verified independently: the jq assertion fails hollow needs payloads and passes clean ones; the bump-guard greps fail a synthetic rebind (version and checksum lines both caught), pass a member-only bump, and log-only on prunes; `verify-onchain` short-circuits at step level in dry runs so `pipeline-complete` won't false-positive; the needs list covers exactly the 13 upstream jobs; and the reusable tag workflow has exactly the two allowlisted callers (`github.workflow_ref` reflects the top-level caller, so the check binds correctly for both entry paths).

No critical or high issues. Two medium findings, both about `pipeline-complete`'s interaction with its neighbors (the announce exclusion in `notify-failure` is now defeated transitively, and `if: always()` makes the guard fail on deliberately-cancelled runs), plus two low notes.

- Verified in review: the jq guard fails a hollow needs payload (`qa=skipped announce=skipped`) and passes an all-success one; the bump-guard greps fail a synthetic rebind diff, pass a clean member-only bump, and log-only on prune-style removals; `verify-onchain` short-circuits at step level in dry runs so `pipeline-complete` is dry-run-safe; `pipeline-complete`'s needs cover exactly the 13 upstream jobs; the reusable tag workflow has exactly the two allowlisted callers, and `github.workflow_ref` in a reusable workflow reflects the top-level caller, so the check binds correctly for both entry paths.

.github/workflows/release.testnet.yml
ben-dz: Fixed in fbdf233a — `announce` dropped from the guard's needs (your coverage argument is right: it can only skip when `qa`/`preflight` didn't succeed, both still checked, and its own failure still reddens the run). The `notify-failure` exclusion comment is accurate again, and the guard's comment now records why announce is excluded.

.github/workflows/release.testnet.yml
ben-dz: Fixed in fbdf233a — `if: ${{ !cancelled() }}`, with the comment noting why (deliberate cancels stay quiet; skipped ≠ cancelled keeps the hollow-success coverage).

.github/workflows/release.testnet.push.tags.yml
ben-dz: Agreed and recorded as accepted: the allowlist binds path, not content, and the old `environment:` was an incidental backstop for a rogue-branch copy — not real protection while `DOUBLEZERO_PAT` is repo-level and directly usable from any branch workflow. If we ever want the harder line, PAT scoping (environment secret or gate-time app token) is the mechanism, as you say. Leaving the code as is.

scripts/release/bump-version.sh
ben-dz: Fixed in fbdf233a — `NEW_RE=${NEW//./\\.}` / `PREV_RE=${PREV//./\\.}` and the grep exclusions use the escaped forms.

PR #4046, opened 2026-07-11
ci: fire edge-connect rebuild on base publish (release-bot App token + devnet notify)

juan-malbeclabs:
## Problem

The event-driven `doublezero-edge-connect` rebuild has **never fired**. On every base-image publish the "Notify edge-connect to rebuild" step failed with:

```
Error: Input required and not supplied: github-token
```

because it referenced `secrets.EDGE_CONNECT_DISPATCH_TOKEN`, a secret that was **never created** in this repo. The steps are `continue-on-error: true`, so the failure was silent and the publishes stayed green.

Confirmed on recent runs:
- testnet — run `29119708072` (2026-07-10)
- mainnet-beta — run `28999867250` (2026-07-09)

On the receiver side, `doublezero-edge-connect` has **zero** `repository_dispatch` events / dispatch-workflow runs. The only thing keeping the images in sync has been edge-connect's daily poll (`05:23 UTC`) — up to ~24h late, and one poll run (2026-07-09, devnet base moved) was lost to a hosted-runner acquisition flake before self-healing the next day.

Separately, **devnet had no notify step at all** (`release.docker.client.devnet.yml`), so devnet was never event-driven even in intent.

## Change

- **Auth via the release-bot GitHub App** instead of the missing secret. Each publisher mints a short-lived, least-privilege token scoped to `doublezero-edge-connect` (`actions/create-github-app-token@v2` with `RELEASE_BOT_APP_ID` / `RELEASE_BOT_PRIVATE_KEY`) and uses it for the cross-repo `repository_dispatch`. This mirrors the app-token pattern already used in `release.testnet.yml`.
- **Add the missing devnet notify step.** All three variants — testnet, mainnet-beta, devnet — now trigger an event-driven rebuild on base publish.
- **Non-fatal throughout.** Both the mint and the notify steps are `continue-on-error`, and the notify is skipped cleanly when the token is empty. edge-connect's daily poll remains the fallback (and still backstops digest drift from any cause).

The receiver in edge-connect (`release.docker.edge-connect.dispatch.yml`) already validates and handles `testnet | mainnet-beta | devnet`, so no change is needed there.

## ⚠️ Action required before this works end-to-end

The **release-bot GitHub App must be installed on `doublezero-edge-connect`** with **Contents: write** (the permission `repository_dispatch` requires). If it is not yet installed there, an org admin needs to add the installation / grant the repo. Until then the mint step fails non-fatally and the daily poll continues to backstop — no regression.

## Verification
- `python3 -c "import yaml; ..."` parses both workflows; repo `actionlint` runs on this PR.
- After merge, the next base publish should produce a `doublezero-base-published` `repository_dispatch` in `doublezero-edge-connect` and a `release.docker.edge-connect.dispatch` run within ~1 min.

### DIFF
--- a/.github/workflows/release.docker.client.yml
+++ b/.github/workflows/release.docker.client.yml
@@ -113,10 +113,26 @@ jobs:
             ${{ env.IMAGE }}:testnet
             ${{ env.IMAGE }}:testnet-${{ steps.v.outputs.version }}
 
+      # Mint a short-lived, least-privilege token scoped to doublezero-edge-connect
+      # from the release-bot GitHub App, used only to fire the cross-repo dispatch
+      # below. Non-fatal: if the app isn't installed on edge-connect the mint fails,
+      # the dispatch is skipped, and edge-connect's daily poll backstops the rebuild.
+      - name: Mint edge-connect dispatch token (testnet)
+        id: ec_token
+        if: steps.v.outputs.publish == 'true'
+        continue-on-error: true
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ secrets.RELEASE_BOT_APP_ID }}
+          private-key: ${{ secrets.RELEASE_BOT_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: doublezero-edge-connect
+
       # Tell doublezero-edge-connect to rebuild its :testnet variant against the
       # base image we just published (event-driven; its daily poll is the fallback).
       - name: Notify edge-connect to rebuild (testnet)
-        if: steps.v.outputs.publish == 'true'
+        # Skip cleanly if the token mint failed (empty token) — the poll backstops it.
+        if: steps.v.outputs.publish == 'true' && steps.ec_token.outputs.token != ''
         # A failed cross-repo dispatch must not fail an already-successful publish;
         # edge-connect's daily poll is the fallback.
         continue-on-error: true
@@ -124,7 +140,7 @@ jobs:
         env:
           DZ_VERSION: ${{ steps.v.outputs.version }}
         with:
-          github-token: ${{ secrets.EDGE_CONNECT_DISPATCH_TOKEN }}
+          github-token: ${{ steps.ec_token.outputs.token }}
           script: |
             await github.rest.repos.createDispatchEvent({
               owner: context.repo.owner,
@@ -225,10 +241,26 @@ jobs:
             ${{ env.IMAGE }}:mainnet-beta-${{ steps.v.outputs.tag }}
             ${{ env.IMAGE }}:latest
 
+      # Mint a short-lived, least-privilege token scoped to doublezero-edge-connect
+      # from the release-bot GitHub App, used only to fire the cross-repo dispatch
+      # below. Non-fatal: if the app isn't installed on edge-connect the mint fails,
+      # the dispatch is skipped, and edge-connect's daily poll backstops the rebuild.
+      - name: Mint edge-connect dispatch token (mainnet-beta)
+        id: ec_token
+        if: steps.gate.outputs.changed == 'true'
+        continue-on-error: true
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ secrets.RELEASE_BOT_APP_ID }}
+          private-key: ${{ secrets.RELEASE_BOT_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: doublezero-edge-connect
+
       # Tell doublezero-edge-connect to rebuild its :mainnet-beta variant against
       # the base image we just published (event-driven; poll is the fallback).
       - name: Notify edge-connect to rebuild (mainnet-beta)
-        if: steps.gate.outputs.changed == 'true'
+        # Skip cleanly if the token mint failed (empty token) — the poll backstops it.
+        if: steps.gate.outputs.changed == 'true' && steps.ec_token.outputs.token != ''
         # A failed cross-repo dispatch must not fail an already-successful publish;
         # edge-connect's daily poll is the fallback.
         continue-on-error: true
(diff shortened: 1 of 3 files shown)


.github/workflows/release.docker.client.yml
```diff
@@ -113,18 +113,34 @@ jobs:
             ${{ env.IMAGE }}:testnet
             ${{ env.IMAGE }}:testnet-${{ steps.v.outputs.version }}
 
+      # Mint a short-lived, least-privilege token scoped to doublezero-edge-connect
```
ben-dz, line 116: The comment says "least-privilege," but `create-github-app-token@v2` with only `owner`/`repositories` mints a token carrying ALL permissions the release-bot App holds on the repo, not just the `contents: write` that `createDispatchEvent` needs — it is repo-scoped, not permission-scoped. Optional hardening: add `permission-contents: write` (and nothing else) to genuinely restrict the token and match the comment. Same applies to the mainnet-beta block (line 244) and the devnet workflow (release.docker.client.devnet.yml:79).

.github/workflows/release.docker.client.yml
```diff
@@ -113,18 +113,34 @@ jobs:
...
         # edge-connect's daily poll is the fallback.
         continue-on-error: true
         uses: actions/github-script@v7
         env:
           DZ_VERSION: ${{ steps.v.outputs.version }}
         with:
-          github-token: ${{ secrets.EDGE_CONNECT_DISPATCH_TOKEN }}
+          github-token: ${{ steps.ec_token.outputs.token }}
           script: |
             await github.rest.repos.createDispatchEvent({
```
ben-dz, line 148: The cross-repo receiver contract (repo `doublezero-edge-connect`, event_type `doublezero-base-published`, `client_payload` shape `{env, version}`) is duplicated in three places (here, mainnet-beta at ~273, and the devnet workflow). `createDispatchEvent` returns 204 whether or not the receiver listens, and `continue-on-error` further masks a typo, so any drift becomes a silent no-op. The daily poll backstops it (no regression), but these three payloads must stay in lockstep with the receiver in doublezero-edge-connect.

ben-dz (approved):
CI-only change wiring the edge-connect cross-repo rebuild dispatch to a release-bot GitHub App token (replacing the never-created `EDGE_CONNECT_DISPATCH_TOKEN` secret) and adding the missing devnet notify step. The core correctness holds: a failed or skipped token mint cleanly skips the dispatch (`continue-on-error` + `steps.ec_token.outputs.token != ''` guard) and never fails the publish, notify can never fire against an unpublished image (implicit `success()` gating on the non-continue-on-error Build & push), gating is correct per job, and owner/repository scoping matches the dispatch target. The old secret is fully removed and the pattern mirrors the existing `release.testnet.yml` usage. Findings are all Low/optional. Nothing blocks merge. Note: end-to-end success still requires the release-bot App to be installed on `doublezero-edge-connect` with Contents: write — an operational prerequisite the PR description already calls out, not a code defect.

- Nit (maintainability): the mint+notify block is copy-pasted across all three jobs (testnet, mainnet-beta, devnet), differing only by env label and gate. A local composite action (e.g. .github/actions/notify-edge-connect taking env/version) would collapse the drift surface — relevant given the 3x-duplicated receiver contract.
- Informational: the devnet workflow dispatches on every nightly run even when the resolved version is unchanged (no digest gate; :latest always re-pushed). Documented as intentional (release.docker.client.devnet.yml:93-96) and absorbed by edge-connect's poll — flagging the per-night churn only.

juan-malbeclabs:
Thanks for the thorough review! Agreed on both points — here's my thinking on leaving them as-is:

**Composite action for the mint+notify block:** Good call on the duplication. I looked at extracting a `.github/actions/notify-edge-connect` composite, but the gate differs per job (`publish == 'true'` for testnet, `gate.changed == 'true'` for mainnet-beta, no gate for devnet), so the `if` conditions would stay in the callers regardless — the composite would only fold in ~10 lines of CI glue while adding a layer of indirection to a workflow that changes rarely. Given it's CI-only and the receiver contract is stable, I'd rather keep it inline and readable for now, and revisit if a fourth variant or a contract change makes the drift surface real.

**Per-night devnet churn:** Correct that devnet dispatches on every nightly run since there's no digest gate (`:latest` is always re-pushed). This is intentional and matches the testnet notify-on-every-publish pattern — edge-connect's poll absorbs the churn, and the dispatch is cheap. Documented inline at `release.docker.client.devnet.yml:96-99` so it's not a surprise later.

One follow-up from the inline thread: I did apply the token-scoping suggestion in 839abe95 — all three mint blocks now set `permission-contents: write` so the token is genuinely least-privilege rather than inheriting every App permission on the repo. Ready to merge once that lands.

.github/workflows/release.docker.client.yml
juan-malbeclabs: Good catch — you're right that repo-scoping alone still inherits every permission the App holds on the repo, so "least-privilege" was aspirational rather than accurate. Fixed in 839abe95: added `permission-contents: write` to all three mint blocks (testnet, mainnet-beta, and the devnet workflow), which scopes the token to exactly what `createDispatchEvent` needs and makes the comment truthful. Thanks for pushing on this.

.github/workflows/release.docker.client.yml
juan-malbeclabs: Agreed, and the silent-no-op risk is the sharp edge here: 204 regardless of a listener plus `continue-on-error` means a drifted `event_type` or payload shape fails invisibly. I'm leaving the three copies inline for now (the per-job gates would keep the callers from fully collapsing into a composite anyway), but the lockstep concern is real. The daily poll backstops any drift so there's no regression, and the contract (`doublezero-base-published` + `{env, version}`) is intentionally simple so it's stable to maintain by hand. If a fourth variant or a payload change shows up, that's the trigger to extract a shared action and cut the drift surface.

PR #4049, opened 2026-07-13
serviceability-instruction: scaffold pure instruction-builder crate (RFC-26 R0)

### LINKED ISSUE #4015: serviceability: scaffold Rust instruction-builder crate + common helpers + exemplar builders
Part of #4014.

Scaffold the new **`doublezero_serviceability_instruction`** crate at `crates/doublezero-serviceability-instruction/` and establish the pattern every later PR copies.

### Scope
- New crate + `Cargo.toml`; register it as a workspace member. Dependencies limited to `doublezero-serviceability`, `solana-program`, `doublezero-program-common` (no `solana-client`/`tokio`/`backon`).
- `src/common.rs`: the `build(program_id, instruction, accounts, payer)` helper (the single place that appends the trailing `[payer(signer,writable), system_program]`). The read-only Permission PDA is derived from the payer inside `build` and appended last, but ships **commented out** until each instruction's `authorize()` migration is activated — it is never a caller-supplied parameter. Also `compute_budget_prelude() -> [Instruction; 2]` (CU `1_400_000`, heap `256 KiB`).
- `src/lib.rs`: module wiring, re-exports, and a documented list of excluded deprecated variants.
- 4 exemplar builders, one per family, to prove the shape end to end:
  - `create_device` — variable `dz_prefix` blocks (builder fixes `resource_count`).
  - `create_link` — fixed accounts.
  - `delete_device` — variable-account builder with a legacy/atomic-close split and RPC-read resource-owner accounts (replaces the originally-planned `suspend_device`, which is a deprecated variant — `SuspendDevice`/`ResumeDevice` return `DoubleZeroError::Deprecated`; suspend/resume now go through `UpdateDevice{desired_status}`).
  - `create_subscribe_user` — length-detected feed (no `permission` parameter; never appends a Permission account).

### Acceptance
- Builders return a single unsigned `Instruction` with accounts in exact processor order incl. trailing accounts; no RPC.
- Unit tests assert the account list (ported from the existing `mockall` command expectations) and `data[0] == <tag>`.
- First `ix_*.bin/.json` golden fixtures emitted for the exemplars (see RF).
- `make rust-build`, `make rust-test`, `make rust-fmt` clean.

juan-malbeclabs:
## Summary

RFC-26 **R0**: scaffold the pure, RPC-free instruction-builder crate `doublezero-serviceability-instruction` — SPL-style builders that return a single unsigned `Instruction` per serviceability instruction, no signing/sending.

- `common::build` (no-permission trailing `[payer, system]`) + `common::build_with_permission` (the deferred, activate-in-one-place Permission append) + `compute_budget_prelude` (1.4M CU / 256 KiB).
- Four exemplar builders establishing the pattern every later PR copies: `create_device`, `delete_device` (legacy/atomic), `create_link`, `create_subscribe_user`.
- Deps limited to `doublezero-serviceability` + `solana-program` + `solana-system-interface` + `solana-compute-budget-interface` — no RPC tree.

The `suspend_device` exemplar from the RFC was replaced with `delete_device` (`SuspendDevice` is a deprecated variant); the RFC and #4015 were updated, and the "length-detected family" classification was corrected (only `CreateUser` is length-detected).

## Why

Instruction assembly is currently coupled to signing+sending inside `commands/*::execute()`; there is no reusable `build_xxx(args) -> Instruction`. See [rfcs/rfc26-rust-instruction-builder-library.md](rfcs/rfc26-rust-instruction-builder-library.md).

## Testing Verification

- Unit tests assert the exact `AccountMeta` list (incl. trailing `[payer, system]`) and the borsh tag byte for each exemplar, including the `delete_device` legacy vs atomic layouts and the `create_subscribe_user` optional-feed placement.

Closes #4015. Part of RFC-26.

---

1. #4049 — R0 scaffold + exemplars  ← **this PR**
2. #4050 — R1 device
3. #4051 — R2 link
4. #4052 — R3 user
8. #4056 — R7 topology + feed

### DIFF
--- a/crates/doublezero-serviceability-instruction/src/common.rs
+++ b/crates/doublezero-serviceability-instruction/src/common.rs
@@ -35,6 +35,11 @@ pub(crate) fn build(
     payer: &Pubkey,
 ) -> Instruction {
     accounts.push(AccountMeta::new(*payer, true));
+    // The system program is marked **writable** for byte-parity with today's
+    // `client.rs::assemble_instructions` (`AccountMeta::new`); the runtime demotes
+    // reserved keys, so this is harmless. Byte-parity is deliberate — the golden
+    // fixtures freeze this flag, and diverging from the SDK would defeat the
+    // drop-in-replacement goal. The RFC glossary is aligned to match.
     accounts.push(AccountMeta::new(
         solana_system_interface::program::ID,
         false,
@@ -68,6 +73,17 @@ pub(crate) fn build(
 /// accounts.push(AccountMeta::new_readonly(permission, false)); // MUST be last
 /// Instruction::new_with_bytes(*program_id, &instruction.pack(), accounts)
 /// ```
+///
+/// **Activation precondition.** This append is a pure, offline operation: it
+/// cannot check whether the payer's Permission PDA actually exists onchain. When
+/// the account does not exist, `authorize()` fails **hard** with
+/// `InvalidAccountData` — the program-ownership check runs before the
+/// foundation-recovery/legacy fallback (see `authorize.rs`), so a missing
+/// Permission account is never routed to the legacy allowlist path. Today's SDK
+/// sidesteps this by appending the PDA only after an RPC existence check, which a
+/// pure builder cannot replicate. The append MUST therefore stay deferred until
+/// every payer is guaranteed a Permission account (e.g. `RequirePermissionAccounts`
+/// enforced), or activating it would break every payer without one.
 pub(crate) fn build_with_permission(
     program_id: &Pubkey,
     instruction: DoubleZeroInstruction,
--- a/crates/doublezero-serviceability-instruction/src/common.rs
+++ b/crates/doublezero-serviceability-instruction/src/common.rs
@@ -1,0 +1,88 @@
+//! The anti-drift keystone.
+//!
+//! [`build`] is the single place that appends the trailing accounts every
+//! serviceability instruction shares, so the payer/system convention lives in
+//! exactly one reviewable location instead of being re-encoded per builder.
+
+use doublezero_serviceability::instructions::DoubleZeroInstruction;
+use solana_compute_budget_interface::ComputeBudgetInstruction;
+use solana_program::{
+    instruction::{AccountMeta, Instruction},
+    pubkey::Pubkey,
+};
+
+/// Protocol-max compute-unit limit for a serviceability transaction. Mirrors
+/// `sdk/rs/src/client.rs::MAX_COMPUTE_UNIT_LIMIT`.
+pub const MAX_COMPUTE_UNIT_LIMIT: u32 = 1_400_000;
+/// Protocol-max heap-frame request (256 KiB). Mirrors
+/// `sdk/rs/src/client.rs::MAX_HEAP_FRAME_BYTES`.
+pub const MAX_HEAP_FRAME_BYTES: u32 = 256 * 1024;
+
+/// Assemble a single serviceability [`Instruction`] from its instruction-specific
+/// account metas (in processor order, WITHOUT payer/system) plus the shared
+/// trailing `[payer, system]`.
+///
+/// This is the permanent **no-permission** path: instructions that never route
+/// through `authorize()` (e.g. `CreateUser`) call this. Instructions that do
+/// route through `authorize()` call [`build_with_permission`] instead.
+///
+/// The wire encoding is `1 tag byte + borsh(args)`, obtained for free from
+/// [`DoubleZeroInstruction::pack`] — builders never hand-write tag bytes.
+pub(crate) fn build(
+    program_id: &Pubkey,
+    instruction: DoubleZeroInstruction,
+    mut accounts: Vec<AccountMeta>,
+    payer: &Pubkey,
+) -> Instruction {
+    accounts.push(AccountMeta::new(*payer, true));
+    accounts.push(AccountMeta::new(
+        solana_system_interface::program::ID,
+        false,
+    ));
+
+    Instruction::new_with_bytes(*program_id, &instruction.pack(), accounts)
+}
+
+/// Assemble an instruction whose processor routes through `authorize()` and will
+/// therefore carry a trailing, read-only Permission PDA (derived from the payer,
+/// never a caller-supplied argument) once the permission model is rolled out.
+///
+/// Builders are **assigned** to this method per-instruction as they are
+/// implemented; the append itself is **deferred** (RFC-26 "Permission account").
+/// Today it simply delegates to [`build`], so assigning a builder here is
+/// behavior-preserving. At the permission rollout, enable the append **here, in
+/// this one place** — it then activates for every builder already assigned to
+/// this method at once, while builders on [`build`] stay untouched.
+///
+/// The Permission account MUST be appended **last**: `authorize()` reads it as
+/// the final account and identifies it by PDA match (`get_permission_pda(payer)`)
+/// — so the variable-length / length-detected families tolerate it (the feed /
+/// tenant slots sit ahead of the payer and are never confused with it). To
+/// activate, assemble here instead of delegating:
+///
+/// ```ignore
+/// accounts.push(AccountMeta::new(*payer, true));
+/// accounts.push(AccountMeta::new(solana_system_interface::program::ID, false));
+/// let (permission, _) =
+///     doublezero_serviceability::pda::get_permission_pda(program_id, payer);
+/// accounts.push(AccountMeta::new_readonly(permission, false)); // MUST be last
+/// Instruction::new_with_bytes(*program_id, &instruction.pack(), accounts)
+/// ```
+pub(crate) fn build_with_permission(
+    program_id: &Pubkey,
+    instruction: DoubleZeroInstruction,
+    accounts: Vec<AccountMeta>,
+    payer: &Pubkey,
+) -> Instruction {
+    build(program_id, instruction, accounts, payer)
+}
+
+/// The transaction-level compute-budget prelude: set the compute-unit limit and
+/// request the heap frame. This is NOT inside each builder — the caller prepends
+/// it once per transaction over the built instruction(s).
+pub fn compute_budget_prelude() -> [Instruction; 2] {
+    [
+        ComputeBudgetInstruction::set_compute_unit_limit(MAX_COMPUTE_UNIT_LIMIT),
+        ComputeBudgetInstruction::request_heap_frame(MAX_HEAP_FRAME_BYTES),
+    ]
+}
--- a/crates/doublezero-serviceability-instruction/src/device.rs
+++ b/crates/doublezero-serviceability-instruction/src/device.rs
@@ -1,0 +1,371 @@
+//! Device-domain instruction builders.
+
+use crate::common;
+use doublezero_serviceability::{
+    instructions::DoubleZeroInstruction,
+    pda::{get_device_pda, get_globalconfig_pda, get_globalstate_pda, get_resource_extension_pda},
+    processors::device::{create::DeviceCreateArgs, delete::DeviceDeleteArgs},
+    resource::ResourceType,
+};
+use solana_program::{
+    instruction::{AccountMeta, Instruction},
+    pubkey::Pubkey,
+};
+
+/// `CreateDevice` (variant 20).
+///
+/// Account layout (processor `next_account_info` order), before the trailing
+/// `[payer, system]` appended by [`common::build`]:
+///
+/// ```text
+/// device                (writable)  — PDA get_device_pda(account_index)
+/// contributor           (writable)
+/// location              (writable)
+/// exchange              (writable)
+/// globalstate           (writable)
+/// globalconfig          (writable)
+/// tunnel_ids resource   (writable)  — ResourceType::TunnelIds(device, 0)
+/// dz_prefix_block[i]    (writable)  — one per args.dz_prefixes entry
+/// ```
+///
+/// The `dz_prefix` blocks and `args.resource_count` are produced from the same
+/// loop, so the declared count can never disagree with the account list.
+///
+/// `account_index` is the **new** device's index: the caller passes
+/// `globalstate.account_index + 1`, not the raw current value.
+pub fn create_device(
+    program_id: &Pubkey,
+    payer: &Pubkey,
+    contributor: &Pubkey,
+    location: &Pubkey,
+    exchange: &Pubkey,
+    account_index: u128,
+    mut args: DeviceCreateArgs,
+) -> Instruction {
+    let (device, _) = get_device_pda(program_id, account_index);
+    let (globalstate, _) = get_globalstate_pda(program_id);
+    let (globalconfig, _) = get_globalconfig_pda(program_id);
+    let (tunnel_ids, _, _) =
+        get_resource_extension_pda(program_id, ResourceType::TunnelIds(device, 0));
+
+    let mut accounts = vec![
+        AccountMeta::new(device, false),
+        AccountMeta::new(*contributor, false),
+        AccountMeta::new(*location, false),
+        AccountMeta::new(*exchange, false),
+        AccountMeta::new(globalstate, false),
+        AccountMeta::new(globalconfig, false),
+        AccountMeta::new(tunnel_ids, false),
+    ];
+
+    let dz_prefix_count = args.dz_prefixes.len();
+    for idx in 0..dz_prefix_count {
+        let (dz_prefix, _, _) =
+            get_resource_extension_pda(program_id, ResourceType::DzPrefixBlock(device, idx));
+        accounts.push(AccountMeta::new(dz_prefix, false));
+    }
+
+    // One TunnelIds account plus one DzPrefixBlock per advertised prefix, derived
+    // from the same loop that produced the accounts above. The count is bounded
+    // by the transaction's account budget, so overflow is unreachable in
+    // practice; the debug_assert keeps the "count == accounts" invariant honest.
+    let resource_total = 1 + dz_prefix_count;
+    debug_assert!(
+        resource_total <= u8::MAX as usize,
+        "device resource_count {resource_total} exceeds u8::MAX"
+    );
+    args.resource_count = u8::try_from(resource_total).unwrap_or(u8::MAX);
+
+    common::build_with_permission(
+        program_id,
+        DoubleZeroInstruction::CreateDevice(args),
+        accounts,
+        payer,
+    )
+}
+
+/// How `delete_device` closes the device's resource accounts. Selecting the
+/// path explicitly (rather than inferring it from an empty slice) prevents an
+/// activated device from silently getting the legacy layout.
+pub enum DeviceDeleteResources<'a> {
+    /// Never-activated device with no live resource accounts.
+    Legacy,
+    /// Activated device: atomically close its resource accounts. `owners[i]` is
+    /// the onchain-read owner of resource `i` (idx 0 = `TunnelIds(device, 0)`,
+    /// idx 1.. = `DzPrefixBlock(device, idx - 1)`), not offline-derivable.
+    Atomic {
+        location: &'a Pubkey,
+        exchange: &'a Pubkey,
+        owners: &'a [Pubkey],
+        device_owner: &'a Pubkey,
+    },
+}
+
+/// `DeleteDevice` (variant 26).
+///
+/// Two layouts, selected by [`DeviceDeleteResources`]:
+///
+/// - [`DeviceDeleteResources::Legacy`]:
+///   ```text
+///   device       (writable)
+///   contributor  (writable)
+///   globalstate  (writable)
+///   ```
+/// - [`DeviceDeleteResources::Atomic`]:
+///   ```text
+///   device                (writable)
+///   contributor           (writable)
+///   globalstate           (writable)
+///   location              (writable)
+///   exchange              (writable)
+///   resource[i]           (writable)  — idx 0: TunnelIds(device, 0);
+///                                        idx 1..: DzPrefixBlock(device, idx-1)
+///   resource_owner[i]     (writable)  — the onchain owner of resource[i]
+///   device_owner          (writable)
+///   ```
+///
+/// `owners.len()` drives both the resource-PDA loop and `args.resource_count`.
+/// This builder never appends a Permission account.
+pub fn delete_device(
+    program_id: &Pubkey,
+    payer: &Pubkey,
+    device: &Pubkey,
+    contributor: &Pubkey,
+    resources: DeviceDeleteResources,
+) -> Instruction {
+    let (globalstate, _) = get_globalstate_pda(program_id);
+
+    let mut accounts = vec![
+        AccountMeta::new(*device, false),
+        AccountMeta::new(*contributor, false),
+        AccountMeta::new(globalstate, false),
+    ];
+
+    let (location, exchange, owners, device_owner) = match resources {
+        DeviceDeleteResources::Legacy => {
+            return common::build_with_permission(
+                program_id,
+                DoubleZeroInstruction::DeleteDevice(DeviceDeleteArgs::default()),
+                accounts,
+                payer,
+            );
+        }
+        DeviceDeleteResources::Atomic {
+            location,
+            exchange,
+            owners,
+            device_owner,
+        } => (location, exchange, owners, device_owner),
+    };
+
+    accounts.push(AccountMeta::new(*location, false));
+    accounts.push(AccountMeta::new(*exchange, false));
+    // Resource PDAs, in the order the processor consumes them.
+    for idx in 0..owners.len() {
+        let resource_type = if idx == 0 {
+            ResourceType::TunnelIds(*device, 0)
+        } else {
+            ResourceType::DzPrefixBlock(*device, idx - 1)
+        };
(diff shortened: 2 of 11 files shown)


crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -0,0 +1,371 @@
...
+///   location              (writable)
+///   exchange              (writable)
+///   resource[i]           (writable)  — idx 0: TunnelIds(device, 0);
+///                                        idx 1..: DzPrefixBlock(device, idx-1)
+///   resource_owner[i]     (writable)  — the onchain owner of resource[i]
+///   device_owner          (writable)
+///   ```
+///
+/// `owners.len()` drives both the resource-PDA loop and `args.resource_count`.
+/// This builder never appends a Permission account.
```
nikw9944: **[Architecture] Doc contradiction:** this says the builder never appends a Permission account, but both paths call `common::build_with_permission`, and `process_delete_device` calls `authorize()` (NETWORK_ADMIN, `processors/device/delete.rs:110-121`) — so at Permission rollout this builder WILL carry one. In a crate whose doc comments are the anti-drift spec, this sentence instructs a future maintainer to reassign the builder to `common::build`, silently dropping the NETWORK_ADMIN path for non-contributor payers. Replace with the accurate statement used in `create_subscribe_user`'s doc: the processor routes through `authorize()`, so the builder is assigned to `build_with_permission` and carries a trailing Permission PDA once activated.

rfcs/rfc26-rust-instruction-builder-library.md
```diff
@@ -186,10 +186,10 @@ Respecting the ~500-lines-of-new-code-per-PR norm (tests excluded):
 
 | PR | Scope |
 |----|-------|
-| R0 | Scaffold crate + `common` + `compute_budget_prelude` + 4 exemplar builders (`create_device`, `create_link`, `suspend_device`, `create_subscribe_user`) + first fixtures |
+| R0 | Scaffold crate + `common` + `compute_budget_prelude` + 4 exemplar builders (`create_device`, `create_link`, `delete_device`, `create_subscribe_user`) + first fixtures |
```
nikw9944: **[Architecture] Plan/PR mismatch:** this R0 row (edited by this PR) still promises "first fixtures," but the PR ships neither fixtures nor `solana-program-test` coverage — and the RFC names account-order drift "the biggest risk." The unit tests are self-referential: they recompute the expected `Vec<AccountMeta>` with the same `pda.rs` helpers and the same human-transcribed ordering as the builders, so they cannot detect drift from the processor; only tag bytes are independently asserted. Land the fixture generator alongside R0, add one `solana-program-test` exemplar for the four builders, or amend this row (and the PR description) so plan and PR agree.

crates/doublezero-serviceability-instruction/src/common.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/// Assemble an instruction whose processor routes through `authorize()` and will
+/// therefore carry a trailing, read-only Permission PDA (derived from the payer,
+/// never a caller-supplied argument) once the permission model is rolled out.
+///
+/// Builders are **assigned** to this method per-instruction as they are
+/// implemented; the append itself is **deferred** (RFC-26 "Permission account").
+/// Today it simply delegates to [`build`], so assigning a builder here is
+/// behavior-preserving. At the permission rollout, enable the append **here, in
```
nikw9944, line 58: **[Architecture] Activation recipe overstates safety:** `authorize()` returns `InvalidAccountData` when the appended Permission PDA does not exist onchain (the owner check runs before the foundation-recovery/legacy fallback), and today's SDK appends the PDA only after an RPC existence check — which a pure builder cannot do. Flipping this append on unconditionally would break every payer without an onchain Permission account. Add a sentence that the append must not be enabled until the RFC's open existence question is resolved (e.g. `RequirePermissionAccounts` enforced or all payers guaranteed a Permission account).

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -0,0 +1,371 @@
...
+    // One TunnelIds account plus one DzPrefixBlock per advertised prefix, derived
+    // from the same loop that produced the accounts above. The count is bounded
+    // by the transaction's account budget, so overflow is unreachable in
+    // practice; the debug_assert keeps the "count == accounts" invariant honest.
+    let resource_total = 1 + dz_prefix_count;
+    debug_assert!(
+        resource_total <= u8::MAX as usize,
+        "device resource_count {resource_total} exceeds u8::MAX"
+    );
+    args.resource_count = u8::try_from(resource_total).unwrap_or(u8::MAX);
```
nikw9944: **[Architecture + Security] Silent invariant break in release:** `unwrap_or(u8::MAX)` (here and in `delete_device`, lines 181-186) emits a `resource_count` that disagrees with the account list — the exact invariant this crate exists to make impossible — guarded only by `debug_assert!`. The condition is unreachable (>255 resources exceeds any transaction's account budget; security review confirmed no exploit path), and the existing SDK returns a hard error here. Replace with `u8::try_from(...).expect("resource count exceeds u8::MAX")` — a panic in the unreachable case is strictly better than an instruction whose declared count disagrees with its accounts.

crates/doublezero-serviceability-instruction/src/common.rs
```diff
@@ -0,0 +1,88 @@
...
+/// The wire encoding is `1 tag byte + borsh(args)`, obtained for free from
+/// [`DoubleZeroInstruction::pack`] — builders never hand-write tag bytes.
+pub(crate) fn build(
+    program_id: &Pubkey,
+    instruction: DoubleZeroInstruction,
+    mut accounts: Vec<AccountMeta>,
+    payer: &Pubkey,
+) -> Instruction {
+    accounts.push(AccountMeta::new(*payer, true));
+    accounts.push(AccountMeta::new(
```
nikw9944, line 43: **[All reviewers] System program marked writable vs RFC glossary:** `AccountMeta::new` here matches today's `assemble_instructions` (`sdk/rs/src/client.rs:215`; harmless — the runtime demotes reserved keys), but the RFC glossary defines the trailing convention as "system_program (readonly)". Golden fixtures will freeze whichever flag ships. Decide now: switch to `new_readonly`, or keep writable for byte parity and correct the RFC glossary; record the choice in a comment.

crates/doublezero-serviceability-instruction/src/user.rs
```diff
@@ -0,0 +1,201 @@
...
+    ];
+
+    for idx in 0..dz_prefix_count as usize {
+        let (dz_prefix, _, _) =
+            get_resource_extension_pda(program_id, ResourceType::DzPrefixBlock(*device, idx));
+        accounts.push(AccountMeta::new(dz_prefix, false));
+    }
+
+    // Optional trailing Feed account (EdgeSeat metro gate), appended BEFORE
+    // payer/system and never followed by a Permission account.
```
nikw9944: **[Architecture] Comment contradicts the function's own doc:** "never followed by a Permission account" is wrong — the doc above (and the processor's `split_trailing_permission`) correctly says a Permission PDA, once activated, follows payer/system. Reword to "the Permission PDA, once activated, follows payer/system, not the feed," or delete the clause.

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -0,0 +1,371 @@
...
+    let (tunnel_ids, _, _) =
+        get_resource_extension_pda(program_id, ResourceType::TunnelIds(device, 0));
+
+    let mut accounts = vec![
+        AccountMeta::new(device, false),
+        AccountMeta::new(*contributor, false),
+        AccountMeta::new(*location, false),
+        AccountMeta::new(*exchange, false),
+        AccountMeta::new(globalstate, false),
+        AccountMeta::new(globalconfig, false),
```
nikw9944, line 62: **[Security] Over-broad writable flags:** `globalconfig` is only read by `create_resource`, and `delete_device`'s `globalstate` (line 141) is validated `writable = false` and never written. Not exploitable and matches the existing SDK, but it costs spurious write-lock contention on hot shared accounts. Use `AccountMeta::new_readonly`, or document byte-parity as the deliberate choice before fixtures freeze it (same decision as the system-program flag in common.rs).

crates/doublezero-serviceability-instruction/src/user.rs
```diff
@@ -0,0 +1,201 @@
...
+/// so it always matches the number of `dz_prefix_block` accounts produced.
+#[allow(clippy::too_many_arguments)]
+pub fn create_subscribe_user(
+    program_id: &Pubkey,
+    payer: &Pubkey,
+    device: &Pubkey,
+    mgroup: &Pubkey,
+    accesspass: &Pubkey,
+    dz_prefix_count: u8,
+    feed: Option<Pubkey>,
```
nikw9944: **[Architecture]** `feed: Option<Pubkey>` is owned while every other account parameter is `&Pubkey`, breaking the crate's own "non-derivable accounts are passed as `&Pubkey`" rule (it does match the RFC exemplar). Use `Option<&Pubkey>` or amend the RFC rule to allow owned Copy scalars/optionals.

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -0,0 +1,371 @@
...
+///
+/// `account_index` is the **new** device's index: the caller passes
+/// `globalstate.account_index + 1`, not the raw current value.
+pub fn create_device(
+    program_id: &Pubkey,
+    payer: &Pubkey,
+    contributor: &Pubkey,
+    location: &Pubkey,
+    exchange: &Pubkey,
+    account_index: u128,
```
nikw9944: **[Architecture]** `account_index` here vs `link_index` in `create_link` both mean "globalstate.account_index + 1 for the new entity," with the +1 convention living only in prose. Unify the name (e.g. `device_index`/`link_index`) before R1/R2 replicate the split across ~90 builders.

crates/doublezero-serviceability-instruction/src/link.rs
```diff
@@ -0,0 +1,124 @@
...
+    resource::ResourceType,
+};
+use solana_program::{
+    instruction::{AccountMeta, Instruction},
+    pubkey::Pubkey,
+};
+
+/// The unicast-default topology account is required; `CreateLink` auto-tags the
+/// link into it.
+const UNICAST_DEFAULT_TOPOLOGY: &str = "unicast-default";
```
nikw9944: **[Architecture]** Fourth copy of the `"unicast-default"` literal (processor `link/create.rs:254`, SDK `commands/link/create.rs`, CLI). Export a `pub const` from `doublezero-serviceability` next to `get_topology_pda` and consume it everywhere.

crates/doublezero-serviceability-instruction/src/common.rs
```diff
@@ -0,0 +1,88 @@
...
+use doublezero_serviceability::instructions::DoubleZeroInstruction;
+use solana_compute_budget_interface::ComputeBudgetInstruction;
+use solana_program::{
+    instruction::{AccountMeta, Instruction},
+    pubkey::Pubkey,
+};
+
+/// Protocol-max compute-unit limit for a serviceability transaction. Mirrors
+/// `sdk/rs/src/client.rs::MAX_COMPUTE_UNIT_LIMIT`.
+pub const MAX_COMPUTE_UNIT_LIMIT: u32 = 1_400_000;
```
nikw9944, line 16: **[Architecture]** `MAX_COMPUTE_UNIT_LIMIT`/`MAX_HEAP_FRAME_BYTES` duplicate private consts in `sdk/rs/src/client.rs:64-65` with only reciprocal "Mirrors" comments as a tether. Fine for R0; at the SDK-migration PR, make the SDK consume these exports. Tracking only.

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -0,0 +1,371 @@
...
+};
+use solana_program::{
+    instruction::{AccountMeta, Instruction},
+    pubkey::Pubkey,
+};
+
+/// `CreateDevice` (variant 20).
+///
+/// Account layout (processor `next_account_info` order), before the trailing
+/// `[payer, system]` appended by [`common::build`]:
```
nikw9944: **[Architecture] Rustdoc warnings (verified with `cargo doc`):** public builder docs link the private `common::build` (also `link.rs:22`, `user.rs:18`), emitting `private_intra_doc_links` warnings — and these builders actually call `build_with_permission`, not `build`. Also `lib.rs:8`'s `[Transaction](solana_program)` points at the crate root, not a `Transaction` type. Reference public items (or `build_with_permission` in plain text) and fix the lib.rs link.

crates/doublezero-serviceability-instruction/src/user.rs
```diff
@@ -0,0 +1,201 @@
...
+    mut args: UserCreateSubscribeArgs,
+) -> Instruction {
+    args.dz_prefix_count = dz_prefix_count;
```
nikw9944, line 59: **[Architecture]** The write-back silently discards any caller-set `dz_prefix_count` (same for `resource_count` in `create_device`). Documented, but `debug_assert!(args.dz_prefix_count == 0 || args.dz_prefix_count == dz_prefix_count)` would catch caller confusion cheaply.

nikw9944 (changes requested):
RFC-26 R0 scaffold is fundamentally sound: all four exemplar builders were verified line-by-line against their processors and the existing SDK commands — account ordering, writability, PDA derivations, tag bytes (20/26/28/59), and args write-back all match exactly, and tests/clippy/fmt pass. However, the crate's anti-drift value proposition has holes that should be fixed before merge: (1) `delete_device`'s doc falsely claims it never appends a Permission account while it routes through `build_with_permission` (and `process_delete_device` does call `authorize()`), inviting a future rollout regression; (2) R0 ships without the RFC-promised golden fixtures / `solana-program-test` coverage, and the unit tests are self-referential (they recompute the expected accounts with the same helpers and ordering as the builders, so they can't catch processor drift); (3) the RFC is left internally contradictory about the Permission activation mechanism, and `build_with_permission`'s activation recipe overstates safety — `authorize()` hard-fails on a non-existent Permission PDA before any legacy fallback. Plus: release-mode `unwrap_or(u8::MAX)` can silently break the count==accounts invariant, and the system program is marked writable while the RFC glossary says readonly (fixtures will freeze whichever ships). All fixes are small — doc corrections, an RFC section, an assert swap, and one writability decision.

- **[Architecture — Medium] RFC left internally contradictory about Permission activation:** rfcs/rfc26-rust-instruction-builder-library.md:111 (unchanged by this PR) still says the Permission append ships "commented out in `common.rs` … enabled per-builder by the activating PR," while the implementation and the PR-updated risk section describe a separate `build_with_permission` enabled centrally, activating every assigned builder at once. The two designs have different blast radii — update the RFC's "Permission account" section and `common.rs` sketch to the `build`/`build_with_permission` split and state which activation model applies.
- **[Architecture — Low] Dead code until R1/RF:** nothing consumes the crate yet; only its self-referential unit tests exercise it in CI. Acceptable per the rollout plan, but worth noting in the PR description; prioritize the first consumer.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Fixed. The doc now states that `process_delete_device` routes through `authorize()` (NETWORK_ADMIN), so the builder is assigned to `build_with_permission` and will carry a trailing Permission PDA once activated — no more misleading "never appends a Permission account".

rfcs/rfc26-rust-instruction-builder-library.md
juan-malbeclabs: Amended. The R0 row now says the four exemplars ship with unit tests (tag bytes + account layout), and the golden fixtures + `solana-program-test` coverage land in a follow-up PR alongside the fixture generator, so the plan and this PR agree. Also noted in the PR description.

crates/doublezero-serviceability-instruction/src/common.rs
juan-malbeclabs: Added an explicit activation precondition to the `build_with_permission` doc: `authorize()` hard-fails with `InvalidAccountData` on a non-existent Permission PDA (the program-ownership check runs before the foundation-recovery/legacy fallback), and a pure builder cannot replicate the SDK's RPC existence check. The append must therefore stay deferred until every payer is guaranteed a Permission account (e.g. `RequirePermissionAccounts` enforced).

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Done. Both sites (`create_device` and `delete_device`) now use `u8::try_from(...).expect(...)` and the redundant `debug_assert` was removed — a panic in the unreachable case is strictly better than emitting a `resource_count` that disagrees with the account list.

crates/doublezero-serviceability-instruction/src/common.rs
juan-malbeclabs: Decided on byte-parity: kept `AccountMeta::new` (writable) to match `assemble_instructions`, since diverging would defeat the drop-in-replacement goal and the fixtures freeze this flag. Documented the choice in a comment here and aligned the RFC glossary (the trailing convention now states `system_program` is emitted writable, runtime-demoted).

crates/doublezero-serviceability-instruction/src/user.rs
juan-malbeclabs: Reworded. The comment now says the Permission PDA, once activated, is appended after payer/system (not after the feed) and is peeled by PDA match, so the feed and the Permission account never collide.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Kept for byte-parity — these mirror the SDK exactly (`globalstate` is sent writable in device delete too, even though the processor validates it `writable = false`). Documented this as the deliberate choice in the `create_device`/`delete_device` doc comments rather than tightening the flags before the fixtures freeze the layout.

crates/doublezero-serviceability-instruction/src/user.rs
juan-malbeclabs: Changed `feed` to `Option<&Pubkey>` so it follows the crate's own `&Pubkey` rule; updated the RFC exemplar and clarified the rule (optional accounts are `Option<&Pubkey>`).

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Unified. Renamed `create_device`'s `account_index` to `device_index` for parity with `create_link`'s `link_index`; updated the RFC exemplar to match. The +1 convention stays in the doc.

crates/doublezero-serviceability-instruction/src/link.rs
juan-malbeclabs: Added a canonical `UNICAST_DEFAULT_TOPOLOGY_NAME` const in `doublezero-serviceability` next to `get_topology_pda`; this crate and the `link/create` processor now consume it (removing two copies). The remaining test/SDK/CLI literals are a mechanical follow-up sweep.

crates/doublezero-serviceability-instruction/src/common.rs
juan-malbeclabs: Acknowledged — leaving these as reciprocal mirrors for R0. The SDK will consume these exports at the SDK-migration PR, as the comments note. Tracking only, no change here.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Fixed. Public docs no longer link the private `common::build`/`build_with_permission` (referenced in plain text, and they now correctly name `build_with_permission`), and the `lib.rs` `Transaction` link is gone. `cargo doc` is warning-free.

crates/doublezero-serviceability-instruction/src/user.rs
juan-malbeclabs: Added `debug_assert!(args.dz_prefix_count == 0 || args.dz_prefix_count == dz_prefix_count)` before the write-back to catch caller confusion cheaply.

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -0,0 +1,380 @@
...
+            &payer,
+            &Pubkey::new_unique(),
+            &Pubkey::new_unique(),
+            &Pubkey::new_unique(),
+            1,
+            args,
+        );
+        let decoded = DoubleZeroInstruction::unpack(&ix.data).unwrap();
+        match decoded {
+            DoubleZeroInstruction::CreateDevice(a) => assert_eq!(a.resource_count, 3),
```
nikw9944, line 290: **[dz-ledger-program-review: Testing; CLAUDE.md Testing §3]** Nit: the args structs derive `PartialEq + Clone`, but this test (and `device.rs:353`, `user.rs:147`) asserts only the count field after decoding. Clone the input args, set the expected count, and `assert_eq!(decoded, expected)` — that additionally pins that the builder's `mut args` write-back touches *only* the count field, and future args fields get asserted automatically.

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -0,0 +1,380 @@
...
+        AccountMeta::new(*device, false),
+        AccountMeta::new(*contributor, false),
+        AccountMeta::new(globalstate, false),
+    ];
+
+    let (location, exchange, owners, device_owner) = match resources {
+        DeviceDeleteResources::Legacy => {
+            return common::build_with_permission(
+                program_id,
+                DoubleZeroInstruction::DeleteDevice(DeviceDeleteArgs::default()),
```
nikw9944, line 160: **[dz-ledger-program-review: State & invariants]** Nit: if `DeviceDeleteArgs` gains a field, `default()` silently fills it with no compile error here — and on the serializing side a silently-defaulted new field may be exactly wrong. Use `DeviceDeleteArgs { resource_count: 0 }` so a future field addition is a compile error, forcing the builder to take a position (same guidance as the #2332 review on `Default`-derived args structs).

nikw9944 (approved):
Round-2 verification at head d55a4e59: all 13 findings from the posted review are resolved — 11 fixed and 2 resolved by documented decision (system-program and globalconfig/globalstate writability kept for byte-parity with the SDK, with the RFC glossary aligned) — and every fix was verified in code. The RFC now scopes golden fixtures and `solana-program-test` coverage to a follow-up PR, resolving the plan/PR mismatch; note drift protection stays deferred until that follow-up lands. Re-verified on the new head: 7 crate tests + 292 serviceability program tests pass, clippy/fmt clean on both touched crates, rustdoc warnings now zero. Remaining open items are three optional nits from the dz-ledger-program-review supplement plus one optional residual; nothing blocking.

- **[Low, informational] Changelog skipped via label:** the `skip-changelog` label passes the CI gate, and the CHANGELOG delta on the branch came in via the merge of main (#4027), not this PR. Defensible for a scaffold nothing consumes yet; the RFC-26 series should land a changelog entry no later than the first consuming or publishing PR.
- **[Residual, non-blocking] Drift protection deferred:** the RFC now honestly scopes golden fixtures and `solana-program-test` coverage to a follow-up PR, but until it lands the unit tests remain self-referential and cannot catch builder-vs-processor account-order drift. Please prioritize that follow-up.
- **[Optional residual on finding 15]:** the `dz_prefix_count` debug_assert was added in `create_subscribe_user`; the analogous caller-set `resource_count` in `create_device` is still silently overwritten without an assert. Optional, fine to leave.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Done in 3753d36c. Replaced `DeviceDeleteArgs::default()` with `DeviceDeleteArgs { resource_count: 0 }` so a future field addition on the struct becomes a compile error here, forcing the builder to take an explicit position.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Done in 3753d36c. The three builder tests (`create_device`, `delete_device` atomic, `create_subscribe_user`) now clone the input args, set the expected count, and `assert_eq!` the full decoded args struct — pinning that the `mut args` write-back touches only the count field, with future fields covered automatically.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Addressed in a separate follow-up PR so this PR stays at the approved state: https://github.com/malbeclabs/doublezero/pull/4062 — `delete_device` now uses `DeviceDeleteArgs { resource_count: 0 }` instead of `::default()`, so a future field addition is a compile error.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Addressed in the follow-up PR https://github.com/malbeclabs/doublezero/pull/4062 — the three builder tests now `assert_eq!` the full decoded args struct (clone input + set expected count), pinning that the `mut args` write-back touches only the count field and auto-covering future fields.

PR #4050, opened 2026-07-13
serviceability-instruction: device domain builders (RFC-26 R1)

### LINKED ISSUE #4016: serviceability: device instruction builders (Rust)
Part of #4014.

Add `src/device.rs` builders covering the full device domain (create/update/suspend/resume/delete + interface instructions). Includes the variable `dz_prefix` block loop and `resource_count` handling established in R0.

### Acceptance
- One pure `Instruction`-returning function per non-deprecated device instruction, accounts in processor order.
- Verbatim account-layout doc-comment copied from each processor.
- Unit tests + `ix_*` golden fixtures for each builder.

juan-malbeclabs:
## Summary

RFC-26 **R1** (stacked on the previous phase's branch). Complete the device domain on top of the R0 exemplars: update_device (max(old,new) dz_prefix block), set_device_health, and interface create/delete/update (Vpnv4 topology PDAs, segment-routing reconcile).

## Testing Verification

Closes #4016. Part of RFC-26 ([rfcs/rfc26-rust-instruction-builder-library.md](rfcs/rfc26-rust-instruction-builder-library.md)).

---

2. #4050 — R1 device  ← **this PR**
3. #4051 — R2 link
4. #4052 — R3 user
8. #4056 — R7 topology + feed

### DIFF
--- a/crates/doublezero-serviceability-instruction/src/device.rs
+++ b/crates/doublezero-serviceability-instruction/src/device.rs
@@ -3,9 +3,21 @@
 use crate::common;
 use doublezero_serviceability::{
     instructions::DoubleZeroInstruction,
-    pda::{get_device_pda, get_globalconfig_pda, get_globalstate_pda, get_resource_extension_pda},
-    processors::device::{create::DeviceCreateArgs, delete::DeviceDeleteArgs},
+    pda::{
+        get_device_pda, get_globalconfig_pda, get_globalstate_pda, get_resource_extension_pda,
+        get_topology_pda,
+    },
+    processors::device::{
+        create::DeviceCreateArgs,
+        delete::DeviceDeleteArgs,
+        interface::{
+            create::DeviceInterfaceCreateArgs, DeviceInterfaceDeleteArgs, DeviceInterfaceUpdateArgs,
+        },
+        sethealth::DeviceSetHealthArgs,
+        update::DeviceUpdateArgs,
+    },
     resource::ResourceType,
+    state::interface::LoopbackType,
 };
 use solana_program::{
     instruction::{AccountMeta, Instruction},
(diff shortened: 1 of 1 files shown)


crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -201,6 +213,257 @@ pub fn delete_device(
...
+/// contributor           (writable)  — device.contributor_pk
+/// globalstate           (writable)
+/// device_tunnel_block   (writable)  — ResourceType::DeviceTunnelBlock
+/// segment_routing_ids   (writable)  — ResourceType::SegmentRoutingIds
+/// topology[i]           (readonly)  — one per topology_name, ONLY for Vpnv4 loopbacks
+/// ```
+///
+/// Topology PDAs are appended (and `args.topology_count` set) only when
+/// `args.loopback_type == Vpnv4`; otherwise `topology_names` is ignored.
+pub fn create_device_interface(
```
nikw9944, line 325: `create_device_interface` leaves `args.use_onchain_allocation` caller-controlled, but `process_create_device_interface` hard-rejects `false` as its first statement (interface/create.rs:90-92) and `false` is the default — `DeviceInterfaceCreateArgs::default()` (as used by all three new interface tests) builds an instruction that always fails with `InvalidArgument`. The SDK command forces `use_onchain_allocation: true`, and this builder already owns layout-derived fields (it overwrites `args.topology_count`). Set `args.use_onchain_allocation = true` inside the builder and assert the forced flag in the unpacked-args tests.

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -201,6 +213,257 @@ pub fn delete_device(
...
+/// Account layout, before the trailing accounts:
+///
+/// ```text
+/// device                (writable)
+/// contributor           (writable)  — device.contributor_pk
+/// globalstate           (writable)
+/// device_tunnel_block   (writable)  — ResourceType::DeviceTunnelBlock
+/// segment_routing_ids   (writable)  — ResourceType::SegmentRoutingIds
+/// ```
+pub fn delete_device_interface(
```
nikw9944, line 384: Same as `create_device_interface`: `process_delete_device_interface` rejects `use_onchain_deallocation == false` unconditionally (interface/delete.rs:56-58) and the SDK command forces it to `true`. Set `args.use_onchain_deallocation = true` inside the builder so default args can't produce an always-revert instruction.

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -201,6 +213,257 @@ pub fn delete_device(
...
+            let (topology, _) = get_topology_pda(program_id, name);
+            accounts.push(AccountMeta::new_readonly(topology, false));
+        }
+    }
+    let topology_count = if is_vpnv4 { topology_names.len() } else { 0 };
+    debug_assert!(
+        topology_count <= u8::MAX as usize,
+        "device interface topology_count {topology_count} exceeds u8::MAX"
+    );
+    args.topology_count = u8::try_from(topology_count).unwrap_or(u8::MAX);
```
nikw9944: `debug_assert!` + `unwrap_or(u8::MAX)`: in a release build with >255 `topology_names`, N topology accounts are pushed but `topology_count` encodes 255 — a count that disagrees with the account list, which the R0 siblings in this same file explicitly handle by panicking (`create_device`: "panicking is strictly better than emitting a `resource_count` that disagrees with the account list") and the SDK handles with a hard error. Practically unreachable (transaction account limit), but replace with `u8::try_from(topology_count).expect("topology_count exceeds u8::MAX")` to match the R0 pattern.

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -201,6 +213,257 @@ pub fn delete_device(
...
+    mut args: DeviceInterfaceUpdateArgs,
+) -> Instruction {
+    let update_topologies = topology_names.is_some();
+    let topology_count = topology_names.map_or(0, <[String]>::len);
+    debug_assert!(
+        topology_count <= u8::MAX as usize,
+        "device interface topology_count {topology_count} exceeds u8::MAX"
+    );
+    args.update_topologies = update_topologies;
+    args.topology_count = u8::try_from(topology_count).unwrap_or(u8::MAX);
```
nikw9944: Same `debug_assert!` + `unwrap_or(u8::MAX)` clamp as in `create_device_interface` — replace with `u8::try_from(topology_count).expect(...)` to match the R0 builders' panic-on-overflow policy. (Here the processor ignores `topology_count` and slices the account tail, so the truncated field is inert, but the file shouldn't carry three behaviors for one condition.)

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -201,6 +213,257 @@ pub fn delete_device(
...
+/// `args.dz_prefixes`. `args.resource_count` is written from the same loop.
+#[allow(clippy::too_many_arguments)]
+pub fn update_device(
+    program_id: &Pubkey,
+    payer: &Pubkey,
+    device: &Pubkey,
+    current_contributor: &Pubkey,
+    current_location: &Pubkey,
+    new_location: &Pubkey,
+    old_dz_prefix_count: usize,
```
nikw9944, line 245: `old_dz_prefix_count` is only meaningful when `args.dz_prefixes.is_some()`; otherwise the caller supplies a value that is silently ignored, and nothing couples the RPC-read old count to the new prefix list — a stale count fails only on-chain via the opaque "Wrong number of resource accounts provided" assert. Consider taking the pair together, e.g. `dz_prefixes: Option<(NetworkV4List, usize /* old onchain count */)>` with the builder writing `args.dz_prefixes` itself, following the `DeviceDeleteResources` precedent from R0.

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -201,6 +213,257 @@ pub fn delete_device(
...
+        AccountMeta::new(*current_location, false),
+        AccountMeta::new(*new_location, false),
+        AccountMeta::new(globalstate, false),
+    ];
+
+    let mut resource_count = 0usize;
+    if let Some(dz_prefixes) = args.dz_prefixes.as_ref() {
+        let (globalconfig, _) = get_globalconfig_pda(program_id);
+        accounts.push(AccountMeta::new(globalconfig, false));
+        let max_count = old_dz_prefix_count.max(dz_prefixes.len());
```
nikw9944, line 261: `args.dz_prefixes = Some(empty list)` with `old_dz_prefix_count == 0` emits `resource_count = 1`, which trips the processor's `resource_count >= 2` assert (processors/device/update.rs:195-198) — a guaranteed-fail instruction. Parity with the SDK's identical loop, but a `debug_assert!(!dz_prefixes.is_empty())` (or a doc note) would catch it at build time.

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -201,6 +213,257 @@ pub fn delete_device(
...
+        AccountMeta::new(*contributor, false),
+        AccountMeta::new(globalstate, false),
+        AccountMeta::new(device_tunnel_block, false),
+        AccountMeta::new(segment_routing_ids, false),
+    ];
+
+    let is_vpnv4 = args.loopback_type == LoopbackType::Vpnv4;
+    if is_vpnv4 {
+        for name in topology_names {
+            let (topology, _) = get_topology_pda(program_id, name);
```
nikw9944, line 349: `get_topology_pda` uses the uppercased name directly as a PDA seed; a name longer than 32 bytes makes `find_program_address` panic. Shared with the SDK, but this is a pure library taking caller-supplied `topology_names` (here and at line 454) — validate the post-uppercase byte length or document the panic condition on both builders.

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -201,6 +213,257 @@ pub fn delete_device(
...
+/// `UpdateDevice` (variant 23).
+///
+/// Account layout (processor `next_account_info` order), before the trailing
+/// accounts appended by [`common::build_with_permission`]:
+///
+/// ```text
+/// device                (writable)
+/// current_contributor   (writable)  — device.contributor_pk
+/// current_location      (writable)  — device.location_pk
+/// new_location          (writable)  — the new location, or current if unchanged
```
nikw9944, line 225: The processor also accepts a no-location-pair layout (has_locations derived from leading length) that this builder deliberately never emits. Worth one explicit line in the doc-comment: "always emits the with-locations layout; pass the current location for both when unchanged."

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -201,6 +213,257 @@ pub fn delete_device(
...
+        get_resource_extension_pda(program_id, ResourceType::SegmentRoutingIds);
+    let mut accounts = vec![
+        AccountMeta::new(*device, false),
+        AccountMeta::new(*contributor, false),
+        AccountMeta::new(globalstate, false),
+        AccountMeta::new(device_tunnel_block, false),
+        AccountMeta::new(segment_routing_ids, false),
+    ];
+
+    let is_vpnv4 = args.loopback_type == LoopbackType::Vpnv4;
```
nikw9944, line 346: Non-Vpnv4 loopbacks silently drop a non-empty `topology_names`. Documented and parity with SDK semantics, but a `debug_assert!(is_vpnv4 || topology_names.is_empty())` would surface caller confusion at build time instead of quietly emitting a different account list.

crates/doublezero-serviceability-instruction/src/device.rs
```diff
@@ -377,4 +640,308 @@ mod tests {
...
+        let nl = Pubkey::new_unique();
+        let args = DeviceUpdateArgs {
+            dz_prefixes: Some("10.0.0.0/8".parse().unwrap()),
+            ..Default::default()
+        };
+        // old count 1, new count 1 -> max 1 -> TunnelIds + DzPrefix0, resource_count 2.
+        let ix = update_device(&pid, &payer, &device, &cc, &cl, &nl, 1, args);
+        assert_eq!(ix.data[0], 23);
+        match DoubleZeroInstruction::unpack(&ix.data).unwrap() {
+            DoubleZeroInstruction::UpdateDevice(a) => assert_eq!(a.resource_count, 2),
```
nikw9944, line 669: Per the repo testing standard (full-struct equality when `PartialEq` is derived): this and the other unpack asserts (lines 774, 817, 941) check a single derived field instead of comparing the whole args struct against the expected value (input args with the builder-written fields applied). Full-struct equality also catches silent drift in any future `#[incremental]` field the builder must pass through untouched, e.g. `assert_eq!(a, DeviceUpdateArgs { resource_count: 2, dz_prefixes: Some(...), ..Default::default() })`.

nikw9944 (changes requested):
RFC-26 R1 device builders ([RFC](rfcs/rfc26-rust-instruction-builder-library.md)). Account order, writable/readonly flags, conditional-account logic, and variant tags (23/83/73/74/76) verified line-by-line against all five processors and the existing SDK commands — they match for valid inputs, and a mis-stated `old_dz_prefix_count` can only fail the transaction, never corrupt state. All 16 crate tests pass; clippy and fmt clean. The dz-ledger-program-review checklist pass found the on-chain-only classes correctly untouched (no `AUTHORIZE_GATED_FLAGS` / `resource verify` changes needed).

One High finding: `create_device_interface`/`delete_device_interface` leave `use_onchain_allocation`/`use_onchain_deallocation` caller-controlled (default `false`), which the processors reject unconditionally — the SDK forces `true`, and the builders should too (they already own `topology_count`). Medium: the release-mode `u8` clamp (`debug_assert!` + `unwrap_or(u8::MAX)`) can emit a `topology_count` that disagrees with the account list, contradicting the R0 siblings' explicit panic-on-overflow policy in the same file; `old_dz_prefix_count` should be coupled to the dz_prefixes it qualifies; and the tri-state's `Some(&[])` clear-topologies case is untested. Lows are hardening/doc/test-coverage items, including full-struct `PartialEq` asserts per the repo's testing standard.

- Medium (test gap): update_device_interface's tri-state middle value `Some(&[])` (clear all topologies) is untested — the one combination exercising the processor's presence heuristic with update_topologies=true, topology_count=0, seg-ext appended, zero topology accounts. Add test_update_device_interface_empty_topologies_clears asserting [device, contributor, globalstate, sri, payer, system], update_topologies == true, topology_count == 0.
- Low (awareness only): contributor/globalstate/globalconfig are sent writable although the processors only read them — deliberate byte-parity with the SDK per common.rs; downgrade to new_readonly when the fixture-parity constraint is lifted.
- Low (test gap): update_device resource-loop tests only cover old == new (1,1) and the no-prefix path; the max() asymmetry — shrink (old=3,new=1) and grow (old=1,new=3) — is untested, and shrink is the branch where the processor closes orphaned DzPrefixBlock accounts.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Fixed in 22b2125b. `delete_device_interface` now takes `mut args` and forces `args.use_onchain_deallocation = true`; the delete-interface test asserts it on the unpacked args.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Fixed in 22b2125b. The builder now sets `args.use_onchain_allocation = true` before packing, so default args can no longer produce an always-reverting instruction, and the two create-interface tests assert the forced flag on the unpacked args.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Good point. Leaving the signature as-is in this round to keep the change behavior-preserving against the SDK command, but I agree that taking the pair together (`dz_prefixes: Option<(NetworkV4List, usize)>`, following the `DeviceDeleteResources` precedent) is the cleaner shape and removes the silently-ignored argument. Tracking it as a follow-up rather than folding an API change into this pass.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: This mirrors the SDK loop exactly, so it only fails in the same case the SDK already fails on-chain. I will add the `debug_assert!(!dz_prefixes.is_empty())` build-time guard as part of the coupling follow-up above.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Fixed in 22b2125b. Replaced the `debug_assert!` + `unwrap_or(u8::MAX)` with `u8::try_from(topology_count).expect(...)`, matching the panic-on-overflow policy of the R0 builders in this file.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Fixed in 22b2125b, same change as `create_device_interface`. Agreed the field is inert here since the processor slices the account tail, but the file no longer carries two behaviors for one condition.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Agreed full-struct equality is stronger. The create-interface tests now also assert the forced allocation flag, but I will convert these unpack asserts to full-struct `PartialEq` comparisons in the follow-up so any future pass-through field is covered.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Acknowledged. The drop is intentional and matches SDK semantics; I will add the `debug_assert!(is_vpnv4 || topology_names.is_empty())` guard in the follow-up so caller confusion surfaces at build time.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Will add the one-line note to the doc-comment ("always emits the with-locations layout; pass the current location for both when unchanged") in the follow-up.

crates/doublezero-serviceability-instruction/src/device.rs
juan-malbeclabs: Agreed this is a caller-supplied panic path in a pure library, and it is not validated here today (same as the SDK). Enforcing the post-uppercase 32-byte cap belongs with the shared validators so it lives in one place rather than inline in each builder; I will document the panic condition on both builders in the follow-up.

nikw9944 (approved):

PR #4051, opened 2026-07-13
serviceability-instruction: link domain builders (RFC-26 R2)

### LINKED ISSUE #4017: serviceability: link instruction builders (Rust)
Part of #4014.

Add `src/link.rs` builders for the full link domain (create/accept/update/suspend/resume/delete, link health). Covers the `split_trailing_permission` family.

### Notes
- `split_trailing_permission` builders derive the Permission PDA from the payer and append it read-only as the last account (never a caller-supplied parameter). This append ships **commented out** until the instruction's `authorize()` migration is activated. It remains safe once enabled because the processor peels the trailing Permission account by PDA match.

### Acceptance
- One pure builder per non-deprecated link instruction.

juan-malbeclabs:
## Summary

RFC-26 **R2** (stacked on the previous phase's branch). accept_link, update_link (LinkUpdateAuthority preamble + conditional tunnel_net / tunnel-resource / topology-union sections), delete_link (topology reference-count accounts), set_link_health.

## Testing Verification

Closes #4017. Part of RFC-26 ([rfcs/rfc26-rust-instruction-builder-library.md](rfcs/rfc26-rust-instruction-builder-library.md)).

---

2. #4050 — R1 device
3. #4051 — R2 link  ← **this PR**
4. #4052 — R3 user
8. #4056 — R7 topology + feed

### DIFF
--- a/crates/doublezero-serviceability-instruction/src/link.rs
+++ b/crates/doublezero-serviceability-instruction/src/link.rs
@@ -7,7 +7,10 @@ use doublezero_serviceability::{
         get_globalstate_pda, get_link_pda, get_resource_extension_pda, get_topology_pda,
         UNICAST_DEFAULT_TOPOLOGY_NAME,
     },
-    processors::link::create::LinkCreateArgs,
+    processors::link::{
+        accept::LinkAcceptArgs, create::LinkCreateArgs, delete::LinkDeleteArgs,
+        sethealth::LinkSetHealthArgs, update::LinkUpdateArgs,
+    },
     resource::ResourceType,
 };
 use solana_program::{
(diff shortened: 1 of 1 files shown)


crates/doublezero-serviceability-instruction/src/link.rs
```diff
@@ -65,6 +68,224 @@ pub fn create_link(
...
+            AccountMeta::new(globalstate, false),
+        ],
+    };
+
+    if args.tunnel_net.is_some() {
+        accounts.push(AccountMeta::new(*side_a, false));
+        accounts.push(AccountMeta::new(*side_z, false));
+    }
+
+    let updating_tunnel_resources = args.tunnel_id.is_some() || args.tunnel_net.is_some();
```
ben-dz, line 181: The `tunnel_id`-only conditional path is untested. `update_link` has three independent conditionals; the tests cover the bare preamble and the `tunnel_net + topologies` case, but not `tunnel_id: Some(_)` with `tunnel_net: None` — the only shape that appends `device_tunnel_block`/`link_ids` WITHOUT the `side_a`/`side_z` pair. A future edit that accidentally couples the side-device push to `tunnel_id` would pass every existing test, and the PR claims to cover the conditional/variable-account paths. Add a test with `LinkUpdateArgs { tunnel_id: Some(n), ..Default::default() }` pinning the full vector `[link, contributor, globalstate, device_tunnel_block, link_ids, payer, system]` and `use_onchain_allocation == true`. While there, pin the inverse in `test_update_link_side_z_preamble`: pass `use_onchain_allocation: true` and assert it comes back `false` (line 189 flips it).

crates/doublezero-serviceability-instruction/src/link.rs
```diff
@@ -120,4 +341,213 @@ mod tests {
...
+                AccountMeta::new(side_z, false),
+                AccountMeta::new(globalstate, false),
+                AccountMeta::new(side_a, false),
+                AccountMeta::new(device_tunnel_block, false),
+                AccountMeta::new(link_ids, false),
+                AccountMeta::new(payer, true),
+                AccountMeta::new(system_program::ID, false),
+            ]
+        );
+        match DoubleZeroInstruction::unpack(&ix.data).unwrap() {
```
ben-dz, line 382: The unpacked-args assertions check only the forced flag. Per the repo's testing guidance (full struct equality), assert the whole struct instead — e.g. `assert_eq!(a, LinkAcceptArgs { side_z_iface_name: String::new(), use_onchain_allocation: true })` here and the equivalent for delete at line 525 — so accidental mutation of other arg fields is caught. Minor.

ben-dz (approved):
Faithful, low-risk PR. The four new builders (accept_link, update_link, delete_link, set_link_health) match their on-chain processors' account order, writability, PDA derivations, and borsh variant tags (66/31/34/84) exactly, and a security pass found no exploitable divergence — every wrong instruction a caller could assemble is rejected by the processor before any state write. One actionable item: `update_link`'s `tunnel_id`-only path (resource accounts without the side_a/side_z pair) is the one conditional layout no test pins, despite the PR claiming to cover the conditional paths. Plus a minor test nit.

crates/doublezero-serviceability-instruction/src/link.rs
juan-malbeclabs: Done in ed43ef88. Added `test_update_link_tunnel_id_only` pinning the full `[link, contributor, globalstate, device_tunnel_block, link_ids, payer, system]` vector with `tunnel_id: Some(42), tunnel_net: None` and asserting `use_onchain_allocation == true`. Also pinned the inverse in `test_update_link_side_z_preamble`: it now passes `use_onchain_allocation: true` and asserts it comes back `false`.

crates/doublezero-serviceability-instruction/src/link.rs
juan-malbeclabs: Done in ed43ef88. Both `accept` and `delete` now assert full struct equality (`LinkAcceptArgs { side_z_iface_name: String::new(), use_onchain_allocation: true }` and `LinkDeleteArgs { use_onchain_deallocation: true }`), so accidental mutation of other arg fields is caught.

PR #4052, opened 2026-07-13
serviceability-instruction: user domain builders (RFC-26 R3)

### LINKED ISSUE #4018: serviceability: user instruction builders (Rust)
Part of #4014.

Add `src/user.rs` builders (create, create-subscribe, update, delete, ban, BGP status, etc.).

### Notes
- **Only `CreateUser` is length-detected**: it identifies its optional trailing tenant account via `accounts.len()` and **never calls `authorize()`**, so its builder must **never** append a Permission account (assign it to `common::build`). A trailing Permission would corrupt its `accounts.len()` parsing.
- `DeleteUser` detects its optional tenant from **onchain state** (`user.tenant_pk`) and **does call `authorize()`**; `CreateSubscribeUser` uses **`split_trailing_permission`** (peels the Permission by PDA match) and **does call `authorize()`** (USER_ADMIN owner-override). Both tolerate a trailing Permission account → assign them to `common::build_with_permission` (append deferred; see #4015 / RFC-26). No builder takes a `permission` parameter (the PDA is derived from the payer).
- Variable `dz_prefix` blocks on subscribe; the optional `feed` sits in the leading slice, before payer/system.

### Acceptance
- One pure builder per non-deprecated user instruction.
- A test pins the account count for `CreateUser` (the length-detected builder).

juan-malbeclabs:
## Summary

RFC-26 **R3** (stacked on the previous phase's branch). create_user (length-detected -> build, no permission), update_user, delete_user, request_ban_user, check_user_access_pass, and set_user_bgp_status (metrics-publisher check, no authorize -> build).

## Testing Verification

Closes #4018. Part of RFC-26 ([rfcs/rfc26-rust-instruction-builder-library.md](rfcs/rfc26-rust-instruction-builder-library.md)).

---

2. #4050 — R1 device
3. #4051 — R2 link
4. #4052 — R3 user  ← **this PR**
8. #4056 — R7 topology + feed

### DIFF
--- a/crates/doublezero-serviceability-instruction/src/user.rs
+++ b/crates/doublezero-serviceability-instruction/src/user.rs
@@ -54,6 +54,13 @@ pub fn create_subscribe_user(
     feed: Option<&Pubkey>,
     mut args: UserCreateSubscribeArgs,
 ) -> Instruction {
+    // The processor rejects `dz_prefix_count == 0` as its first statement
+    // (create_subscribe.rs) — CreateSubscribeUser requires on-chain allocation — so a
+    // zero here can only ever fail. Catch it cheaply in debug builds.
+    debug_assert!(
+        dz_prefix_count > 0,
+        "dz_prefix_count must be > 0; CreateSubscribeUser requires on-chain allocation"
+    );
     // The write-back overwrites any caller-set `dz_prefix_count`; assert they agree
     // (or the caller left it zero) to catch confusion cheaply in debug builds.
     debug_assert!(
@@ -135,6 +142,13 @@ pub fn create_user(
     tenant: Option<Pubkey>,
     mut args: UserCreateArgs,
 ) -> Instruction {
+    // The processor rejects `dz_prefix_count == 0` as its first statement (create.rs)
+    // — CreateUser requires on-chain allocation — so a zero here can only ever fail.
+    // Catch it cheaply in debug builds.
+    debug_assert!(
+        dz_prefix_count > 0,
+        "dz_prefix_count must be > 0; CreateUser requires on-chain allocation"
+    );
     args.dz_prefix_count = dz_prefix_count;
 
     let (user, _) = get_user_pda(program_id, &args.client_ip, args.user_type);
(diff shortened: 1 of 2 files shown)


crates/doublezero-serviceability-instruction/src/user.rs
```diff
@@ -100,10 +112,355 @@ pub fn create_subscribe_user(
...
+pub fn update_user(
+    program_id: &Pubkey,
+    payer: &Pubkey,
+    user: &Pubkey,
+    device: &Pubkey,
+    dz_prefix_count: u8,
+    old_tenant: &Pubkey,
+    mut args: UserUpdateArgs,
+) -> Instruction {
+    args.dz_prefix_count = dz_prefix_count;
```
ben-dz, line 225: `debug_assert!(dz_prefix_count > 0)` guards `create_user` and `create_subscribe_user` but not `update_user`/`delete_user`/`request_ban_user`, whose processors reject `dz_prefix_count == 0` identically (update.rs:70, delete.rs:56, requestban.rs:50). Add the same assert here and in the other two so the fail-fast guardrail is uniform. (The assert only catches the zero case; a nonzero-but-understated count still silently under-deallocates the dz_ip bit, but a pure builder can't detect that without onchain state, so no further fix is expected.)

ben-dz (approved):
Re-reviewed at head f5170e0c. One non-blocking consistency nit: `debug_assert!(dz_prefix_count > 0)` was added to `create_user`/`create_subscribe_user` but not to `update_user`/`delete_user`/`request_ban_user`, whose processors reject `dz_prefix_count == 0` identically — apply it to all five for a uniform build-time guardrail. Everything else raised in the first pass is either addressed by the new commits (tenant-branch tests) or triaged as pre-existing/intentional SDK-parity behavior.

crates/doublezero-serviceability-instruction/src/user.rs
juan-malbeclabs: Done in 873f63cf. Added the same `debug_assert!(dz_prefix_count > 0)` to `update_user`, `delete_user`, and `request_ban_user`, each pointing at its processor guard (update.rs:70, delete.rs:56, requestban.rs:50), so the fail-fast guardrail is uniform across all five user builders. Agreed on the nonzero-but-understated case — a pure builder can't detect that without onchain state, so no further change here.

PR #4053, opened 2026-07-13
serviceability-instruction: location/exchange/contributor builders (RFC-26 R4)

juan-malbeclabs:
## Summary

RFC-26 **R4** (stacked on the previous phase's branch). Three account_index-seeded CRUD domains (create/update/suspend/resume/delete). Exchange create/update carry globalconfig; set_device_exchange carries the device; create_contributor carries the owner.

All builders route through `authorize()` -> `build_with_permission` unless noted; each carries a verbatim account-layout doc-comment copied from its processor.

## Testing Verification

- Unit tests assert the exact `AccountMeta` list (incl. trailing `[payer, system]`) and the borsh tag byte for every builder, covering the conditional/variable-account paths.

Closes #4019. Part of RFC-26 ([rfcs/rfc26-rust-instruction-builder-library.md](rfcs/rfc26-rust-instruction-builder-library.md)).

---

### PR stack (RFC-26 builder library)

Stacked PRs, merge in order (each is based on the previous one's branch):

1. #4049 — R0 scaffold + exemplars
2. #4050 — R1 device
3. #4051 — R2 link
4. #4052 — R3 user
5. #4053 — R4 location/exchange/contributor  ← **this PR**
6. #4054 — R5 multicastgroup + allowlists
7. #4055 — R6 tenant + permission
8. #4056 — R7 topology + feed
9. #4057 — R8 accesspass + resource
10. #4058 — R9 globalstate/config/allowlist/index/migrate

R10 (commands/* migration + program-test) and RF (fixtures) follow as separate PRs.

PR #4054, opened 2026-07-13
serviceability-instruction: multicastgroup + allowlist builders (RFC-26 R5)

### LINKED ISSUE #4020: serviceability: multicastgroup instruction builders (Rust)
Part of #4014.

Add `src/multicastgroup.rs` builders (create/update/suspend/resume/delete/activate + publisher/subscriber allowlist add/remove).

### Acceptance
- One pure builder per non-deprecated multicastgroup instruction.

juan-malbeclabs:
## Summary

RFC-26 **R5** (stacked on the previous phase's branch). create/update (conditional multicast_group_block)/suspend/reactivate/delete, update_multicast_group_roles, and the four pub/sub allowlist add/remove builders.

## Testing Verification

Closes #4020. Part of RFC-26 ([rfcs/rfc26-rust-instruction-builder-library.md](rfcs/rfc26-rust-instruction-builder-library.md)).

---

2. #4050 — R1 device
3. #4051 — R2 link
4. #4052 — R3 user
6. #4054 — R5 multicastgroup + allowlists  ← **this PR**
8. #4056 — R7 topology + feed

### DIFF
--- a/crates/doublezero-serviceability-instruction/src/multicastgroup.rs
+++ b/crates/doublezero-serviceability-instruction/src/multicastgroup.rs
@@ -65,6 +65,10 @@ pub fn create_multicast_group(
 /// `UpdateMulticastGroup` (variant 49).
 /// Accounts: `[mgroup, globalstate]`, plus `multicast_group_block` when
 /// `args.multicast_ip.is_some()` (a multicast-IP reallocation).
+///
+/// `args.use_onchain_allocation` is DERIVED from `args.multicast_ip.is_some()`;
+/// any caller-supplied value is ignored (the flag must stay in lockstep with
+/// whether the block account is emitted).
 pub fn update_multicast_group(
     program_id: &Pubkey,
     payer: &Pubkey,
@@ -82,6 +86,9 @@ pub fn update_multicast_group(
             get_resource_extension_pda(program_id, ResourceType::MulticastGroupBlock);
         accounts.push(AccountMeta::new(multicast_group_block, false));
     }
+    // Derived, not caller-supplied: keep the flag in lockstep with block-account
+    // emission. A stray `true` with no block emitted would make the processor
+    // consume the trailing payer as the resource-extension account.
     args.use_onchain_allocation = updating_multicast_ip;
     common::build_with_permission(
         program_id,
@@ -198,10 +205,13 @@ pub fn add_multicast_group_pub_allowlist(
     payer: &Pubkey,
     mgroup: &Pubkey,
     accesspass: &Pubkey,
-    user_payer: &Pubkey,
     args: AddMulticastGroupPubAllowlistArgs,
 ) -> Instruction {
     let (globalstate, _) = get_globalstate_pda(program_id);
+    // Single source of truth: the processor derives both the accesspass PDA and
+    // the funded key from `args.user_payer`, so the account meta MUST come from
+    // the same field — never a separate parameter that could diverge.
+    let user_payer = args.user_payer;
     common::build_with_permission(
         program_id,
         DoubleZeroInstruction::AddMulticastGroupPubAllowlist(args),
@@ -209,7 +219,7 @@ pub fn add_multicast_group_pub_allowlist(
             AccountMeta::new(*mgroup, false),
             AccountMeta::new(*accesspass, false),
             AccountMeta::new(globalstate, false),
-            AccountMeta::new(*user_payer, false),
+            AccountMeta::new(user_payer, false),
         ],
         payer,
     )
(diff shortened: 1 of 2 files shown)


crates/doublezero-serviceability-instruction/src/multicastgroup.rs
```diff
@@ -0,0 +1,542 @@
...
+}
+
+/// `AddMulticastGroupPubAllowlist` (variant 54).
+/// Accounts: `[mgroup, accesspass, globalstate, user_payer]`.
+pub fn add_multicast_group_pub_allowlist(
+    program_id: &Pubkey,
+    payer: &Pubkey,
+    mgroup: &Pubkey,
+    accesspass: &Pubkey,
+    user_payer: &Pubkey,
```
ben-dz: `user_payer` is specified twice — once as this account parameter and again inside `args.user_payer` (:212 builds the meta from this one) — with nothing binding them. A caller can pass divergent values; it fails closed on-chain (the processor asserts `accesspass.user_payer == value.user_payer` and `user_payer_account.key == funded_user_payer` before `airdrop_user_credits`), so no fund/authority risk, but it's a fail-late footgun in a correct-by-construction crate and diverges from the SDK, which derives both from a single `self.user_payer`. Drop the account parameter and build the meta from `args.user_payer`, or overwrite `args.user_payer = *user_payer` the way this file forces the allocation flags. Same on the subscriber variant (:247/:258).

crates/doublezero-serviceability-instruction/src/multicastgroup.rs
```diff
@@ -0,0 +1,542 @@
...
+        AccountMeta::new(*mgroup, false),
+        AccountMeta::new(globalstate, false),
+    ];
+    let updating_multicast_ip = args.multicast_ip.is_some();
+    if updating_multicast_ip {
+        let (multicast_group_block, _, _) =
+            get_resource_extension_pda(program_id, ResourceType::MulticastGroupBlock);
+        accounts.push(AccountMeta::new(multicast_group_block, false));
+    }
+    args.use_onchain_allocation = updating_multicast_ip;
```
ben-dz, line 92: This silently overrides a caller-supplied `use_onchain_allocation`, flipping `true`→`false` when `multicast_ip` is `None`. The override is correct — it keeps the flag in lockstep with block-account emission; a stray `true` with no block would make the processor consume the payer as the resource extension — but every other forced flag in this file carries a why-comment and this one doesn't. Add the one-line rationale and note in the doc-comment (:60-63) that the flag is derived and any caller value is ignored.

crates/doublezero-serviceability-instruction/src/multicastgroup.rs
```diff
@@ -0,0 +1,542 @@
...
+        );
+        match DoubleZeroInstruction::unpack(&with_ip.data).unwrap() {
+            DoubleZeroInstruction::UpdateMulticastGroup(a) => assert!(a.use_onchain_allocation),
+            other => panic!("unexpected: {other:?}"),
+        }
+
+        // No multicast_ip -> no block, no onchain allocation.
+        let without_ip =
+            update_multicast_group(&pid, &payer, &mgroup, MulticastGroupUpdateArgs::default());
+        assert_eq!(without_ip.accounts.len(), 4);
```
ben-dz: This is the only length-only assertion in the file; every other path pins the exact `AccountMeta` vec. A regression swapping account order or writability on the no-IP path would pass. Assert the full `vec![mgroup, globalstate, payer(signer), system]`.

ben-dz (approved):
All ten builders are byte-parity-correct against the on-chain processors and the reference SDK — tags, account order/writability/signer flags, the conditional-block path, and the forced allocation flags all check out. No blocking issues; three Low items only.

- L1: `add_multicast_group_pub_allowlist`/`add_multicast_group_sub_allowlist` take `user_payer` both as an account parameter and inside `args.user_payer`, unbound — a caller can pass divergent values. Fails closed on-chain (no fund/authority risk) but is a fail-late footgun in a correct-by-construction crate and diverges from the SDK's single-source pattern. Derive the account meta from `args.user_payer`, or overwrite `args.user_payer = *user_payer` as the file already does for the allocation flags.
- L2: `update_multicast_group` overrides the caller's `use_onchain_allocation` flag (flipping `true`→`false` when `multicast_ip` is `None`) with no rationale comment, unlike every other forced flag in the file. Add the one-line why and note the derivation in the doc-comment.
- L3: the no-IP update test asserts only `accounts.len() == 4` instead of the full `AccountMeta` vec like every other path; pin the full `vec![mgroup, globalstate, payer, system]`.

crates/doublezero-serviceability-instruction/src/multicastgroup.rs
juan-malbeclabs: Fixed in d3cc9549. Dropped the `user_payer` account parameter from both `add_multicast_group_pub_allowlist` and `add_multicast_group_sub_allowlist` and build the account meta from `args.user_payer`, so the account and the packed field share a single source and can no longer diverge. This matches the processor, which derives both the accesspass PDA and the funded key from `value.user_payer`.

crates/doublezero-serviceability-instruction/src/multicastgroup.rs
juan-malbeclabs: Fixed in d3cc9549. Added the why-comment at the override — the flag is kept in lockstep with block-account emission; a stray `true` with no block emitted would make the processor consume the trailing payer as the resource-extension account — and the doc-comment now notes that `use_onchain_allocation` is derived from `multicast_ip.is_some()` and any caller-supplied value is ignored.

crates/doublezero-serviceability-instruction/src/multicastgroup.rs
juan-malbeclabs: Fixed in d3cc9549. The no-IP path now pins the full `AccountMeta` vec `[mgroup, globalstate, payer(signer), system]` instead of only asserting the length.

PR #4055, opened 2026-07-13
serviceability-instruction: tenant + permission builders (RFC-26 R6)

juan-malbeclabs:
## Summary

RFC-26 **R6** (stacked on the previous phase's branch). Tenant CRUD + add/remove administrator + update_payment_status (globalstate writable on create, read-only elsewhere), and permission create/update/suspend/resume/delete (target PDA derived from args.user_payer).

All builders route through `authorize()` -> `build_with_permission` unless noted; each carries a verbatim account-layout doc-comment copied from its processor.

## Testing Verification

- Unit tests assert the exact `AccountMeta` list (incl. trailing `[payer, system]`) and the borsh tag byte for every builder, covering the conditional/variable-account paths.

Closes #4021. Part of RFC-26 ([rfcs/rfc26-rust-instruction-builder-library.md](rfcs/rfc26-rust-instruction-builder-library.md)).

---

### PR stack (RFC-26 builder library)

Stacked PRs, merge in order (each is based on the previous one's branch):

1. #4049 — R0 scaffold + exemplars
2. #4050 — R1 device
3. #4051 — R2 link
4. #4052 — R3 user
5. #4053 — R4 location/exchange/contributor
6. #4054 — R5 multicastgroup + allowlists
7. #4055 — R6 tenant + permission  ← **this PR**
8. #4056 — R7 topology + feed
9. #4057 — R8 accesspass + resource
10. #4058 — R9 globalstate/config/allowlist/index/migrate

R10 (commands/* migration + program-test) and RF (fixtures) follow as separate PRs.

PR #4056, opened 2026-07-13
serviceability-instruction: topology + feed builders (RFC-26 R7)

### LINKED ISSUE #4022: serviceability: topology + feed instruction builders (Rust)
Part of #4014.

Add `src/topology.rs` and `src/feed.rs` builders.

### Notes
- Topology `clear_topology` / `assign_topology_node_segments` are batched: expose a single-chunk builder plus a `*_batched(...) -> Vec<Instruction>` convenience. Move the batch-size consts into this crate (the 32-account cap math includes the trailing accounts the builder now owns).

### Acceptance
- One pure builder per non-deprecated topology/feed instruction; batched convenience helpers where applicable.
- Unit tests (incl. a batching boundary test) + `ix_*` golden fixtures.

juan-malbeclabs:
## Summary

RFC-26 **R7** (stacked on the previous phase's branch). Topology create/delete plus batched clear_topology / assign_topology_node_segments (single-chunk + *_batched; CLEAR_BATCH_SIZE=16 / BACKFILL_BATCH_SIZE=4 moved into the crate). Feed create/update/delete.

## Testing Verification

Closes #4022. Part of RFC-26 ([rfcs/rfc26-rust-instruction-builder-library.md](rfcs/rfc26-rust-instruction-builder-library.md)).

---

2. #4050 — R1 device
3. #4051 — R2 link
4. #4052 — R3 user
8. #4056 — R7 topology + feed  ← **this PR**

### DIFF
--- a/crates/doublezero-serviceability-instruction/src/feed.rs
+++ b/crates/doublezero-serviceability-instruction/src/feed.rs
@@ -1,0 +1,122 @@
+//! Feed-domain instruction builders.
+//!
+//! All route through `authorize()` -> [`common::build_with_permission`].
+//! Accounts are `[feed, globalstate]`; create derives the feed PDA from
+//! `args.code` and `args.exchange`.
+
+use crate::common;
+use doublezero_serviceability::{
+    instructions::DoubleZeroInstruction,
+    pda::{get_feed_pda, get_globalstate_pda},
+    processors::feed::{create::FeedCreateArgs, delete::FeedDeleteArgs, update::FeedUpdateArgs},
+};
+use solana_program::{
+    instruction::{AccountMeta, Instruction},
+    pubkey::Pubkey,
+};
+
+/// `CreateFeed` (variant 112). Accounts: `[feed, globalstate]`.
+///
+/// The feed PDA is derived from `args.code` and `args.exchange`.
+pub fn create_feed(program_id: &Pubkey, payer: &Pubkey, args: FeedCreateArgs) -> Instruction {
+    let (feed, _) = get_feed_pda(program_id, &args.code, &args.exchange);
+    let (globalstate, _) = get_globalstate_pda(program_id);
+    common::build_with_permission(
+        program_id,
+        DoubleZeroInstruction::CreateFeed(args),
+        vec![
+            AccountMeta::new(feed, false),
+            AccountMeta::new(globalstate, false),
+        ],
+        payer,
+    )
+}
+
+/// `UpdateFeed` (variant 113). Accounts: `[feed, globalstate]`.
+pub fn update_feed(
+    program_id: &Pubkey,
+    payer: &Pubkey,
+    feed: &Pubkey,
+    args: FeedUpdateArgs,
+) -> Instruction {
+    let (globalstate, _) = get_globalstate_pda(program_id);
+    common::build_with_permission(
+        program_id,
+        DoubleZeroInstruction::UpdateFeed(args),
+        vec![
+            AccountMeta::new(*feed, false),
+            AccountMeta::new(globalstate, false),
+        ],
+        payer,
+    )
+}
+
+/// `DeleteFeed` (variant 114). Accounts: `[feed, globalstate]`.
+pub fn delete_feed(
+    program_id: &Pubkey,
+    payer: &Pubkey,
+    feed: &Pubkey,
+    args: FeedDeleteArgs,
+) -> Instruction {
+    let (globalstate, _) = get_globalstate_pda(program_id);
+    common::build_with_permission(
+        program_id,
+        DoubleZeroInstruction::DeleteFeed(args),
+        vec![
+            AccountMeta::new(*feed, false),
+            AccountMeta::new(globalstate, false),
+        ],
+        payer,
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use solana_system_interface::program as system_program;
+
+    #[test]
+    fn test_create_feed_derives_pda_from_code_and_exchange() {
+        let pid = Pubkey::new_unique();
+        let payer = Pubkey::new_unique();
+        let exchange = Pubkey::new_unique();
+        let args = FeedCreateArgs {
+            code: "feed".to_string(),
+            exchange,
+            ..Default::default()
+        };
+        let ix = create_feed(&pid, &payer, args);
+        assert_eq!(ix.data[0], 112);
+        let (feed, _) = get_feed_pda(&pid, "feed", &exchange);
+        let (globalstate, _) = get_globalstate_pda(&pid);
+        assert_eq!(
+            ix.accounts,
+            vec![
+                AccountMeta::new(feed, false),
+                AccountMeta::new(globalstate, false),
+                AccountMeta::new(payer, true),
+                AccountMeta::new(system_program::ID, false),
+            ]
+        );
+    }
+
+    #[test]
+    fn test_feed_pubkey_verbs() {
+        let pid = Pubkey::new_unique();
+        let payer = Pubkey::new_unique();
+        let feed = Pubkey::new_unique();
+        let (globalstate, _) = get_globalstate_pda(&pid);
+        let expected = vec![
+            AccountMeta::new(feed, false),
+            AccountMeta::new(globalstate, false),
+            AccountMeta::new(payer, true),
+            AccountMeta::new(system_program::ID, false),
+        ];
+        let update = update_feed(&pid, &payer, &feed, FeedUpdateArgs::default());
+        assert_eq!(update.data[0], 113);
+        assert_eq!(update.accounts, expected);
+        let delete = delete_feed(&pid, &payer, &feed, FeedDeleteArgs {});
+        assert_eq!(delete.data[0], 114);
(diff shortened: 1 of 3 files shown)


crates/doublezero-serviceability-instruction/src/topology.rs
```diff
@@ -0,0 +1,347 @@
...
+//! `assign_topology_node_segments` are batched: a single-chunk builder plus a
+//! `*_batched(...) -> Vec<Instruction>` convenience. The batch-size constants
+//! account for the trailing `[payer, system]` the builder now owns, keeping each
+//! transaction under Solana's 32-account cap.
+//!
+//! **Name canonicalization.** Every builder uppercases `args.name` before
+//! deriving the topology PDA and packing the instruction. Topology names are
+//! canonically uppercase onchain: the `create` processor uppercases the name and
+//! derives/asserts the PDA from that form, storing the uppercase name; `delete`,
+//! `clear`, and `assign` derive their PDA from the raw arg, so they only ever
```
ben-dz: This rationale is factually wrong: `get_topology_pda` canonicalizes internally (`smartcontract/programs/doublezero-serviceability/src/pda.rs:121-124`, since c60e14de) — `let upper = name.to_ascii_uppercase();` before `find_program_address`. Every processor derivation routes through it (`delete.rs:52`, `clear.rs:85`, `assign_node_segments.rs:76`), so a builder passing `"topo"` without canonicalization derives the same PDA the processor computes; nothing references a non-existent account and nothing reverts (create's PDA assert passes too). The SDK's PDA can't diverge either, so "deliberate, correctness-preserving divergence from the SDK (which derives from the raw name)" is misleading — the only real effect of `args.name = args.name.to_ascii_uppercase()` is packing the canonical name, which for non-canonical input is itself a byte-parity divergence from the SDK. Fix: either drop the canonicalization from all four builders and delete this paragraph (restores exact SDK byte-parity; PDA already canonical), or keep it and rewrite the rationale to what it actually does (normalize the packed name so onchain logs/stored args carry the canonical form). The test comments repeating this claim (e.g. line 200: "delete/clear/assign derive the processor-side PDA from the packed name, so it must be uppercase") need the same correction either way.

crates/doublezero-serviceability-instruction/src/topology.rs
```diff
@@ -0,0 +1,347 @@
+//! Topology-domain instruction builders.
+//!
+//! All route through `authorize()` -> [`common::build_with_permission`]. The
+//! topology PDA is derived from `args.name`. `clear_topology` and
+//! `assign_topology_node_segments` are batched: a single-chunk builder plus a
+//! `*_batched(...) -> Vec<Instruction>` convenience. The batch-size constants
+//! account for the trailing `[payer, system]` the builder now owns, keeping each
+//! transaction under Solana's 32-account cap.
```
ben-dz: "Solana's 32-account cap" is not a real limit — the per-transaction account-lock limit is 64; the binding constraint for large batches is the ~1232-byte packet size. Inherited from the SDK comment, but restated here as fact and used to frame the batch-size math. Reword to name the real constraints (the current constants are safe under all of them).

crates/doublezero-serviceability-instruction/src/feed.rs
```diff
@@ -0,0 +1,139 @@
...
+//! **Why `build_with_permission`.** The feed `create`/`update`/`delete`
+//! processors call `authorize(.., FEED_AUTHORITY | FOUNDATION)`, so feed is a
+//! permission-gated instruction like topology/tenant and belongs on the
+//! permission-appending path. This intentionally diverges from today's Rust SDK
+//! feed commands, which use the plain (no-permission) `execute_transaction`;
+//! among authorize()-gated instructions the SDK feed commands are the odd ones
+//! out, not these builders.
+//!
+//! One pre-existing program-side gap, unrelated to these builders: `FEED_AUTHORITY`
+//! is absent from `authorize.rs::AUTHORIZE_GATED_FLAGS`. That list drives
```
ben-dz, line 16: The `FEED_AUTHORITY`-missing-from-`AUTHORIZE_GATED_FLAGS` gap is verified real (`authorize.rs:293-308`) and is an activation precondition for the deferred permission append, but it lives only in prose in two files — no tracking issue exists and nothing fails if the gap and the activation drift apart. File a tracking issue, and optionally add a one-assert canary test in this crate (`!AUTHORIZE_GATED_FLAGS.contains(&permission_flags::FEED_AUTHORITY)` with a pointer to this doc) so closing the program-side gap forces this doc to be revisited.

crates/doublezero-serviceability-instruction/src/feed.rs
```diff
@@ -0,0 +1,139 @@
...
+        let payer = Pubkey::new_unique();
+        let feed = Pubkey::new_unique();
+        let (globalstate, _) = get_globalstate_pda(&pid);
+        let expected = vec![
+            AccountMeta::new(feed, false),
+            AccountMeta::new(globalstate, false),
+            AccountMeta::new(payer, true),
+            AccountMeta::new(system_program::ID, false),
+        ];
+        let update = update_feed(&pid, &payer, &feed, FeedUpdateArgs::default());
```
ben-dz: `FeedUpdateArgs::default()` builds an instruction the processor rejects outright as a no-op update (`processors/feed/update.rs:30-33`, `InvalidArgument`). Commit 8f03288 fixed exactly this for `create_feed`; use e.g. `FeedUpdateArgs { name: Some("Feed".into()), groups: None }` so the fixture could actually execute.

ben-dz (commented):
The topology module doc's name-canonicalization rationale is factually wrong: `get_topology_pda` (pda.rs:121-124) has uppercased internally since c60e14de, so delete/clear/assign deriving "from the raw arg" still canonicalize and a non-uppercase name can never reference a non-existent account or revert — the builders' `to_ascii_uppercase()` is a PDA no-op whose only effect is packing the canonical name (itself a byte-parity divergence from the SDK for non-canonical input). Either drop the canonicalization or rewrite the rationale, and fix the test comments repeating the claim. Separately, the live SDK `ClearTopologyCommand` readonly bug this PR correctly documents is user-facing (clearing a live topology is the mandatory precursor to delete) and has no tracking issue. Builder account layouts, writability flags, wire tags, batch math, and the `FEED_AUTHORITY`/`AUTHORIZE_GATED_FLAGS` gap claim all verified correct; remaining items are low-severity doc and test-fixture fixes.

- Medium: the live SDK bug this PR documents at topology.rs:101 ("the SDK command is the one that needs fixing") is real, user-facing, and untracked. `smartcontract/sdk/rs/src/commands/topology/clear.rs` passes the topology PDA `new_readonly`, while the processor (`clear.rs:124-127`) asserts `is_writable` whenever the topology exists and at least one link is cleared; since `delete_topology` requires `reference_count == 0` (`delete.rs:77`), clearing a live topology is the mandatory precursor to delete — the SDK/CLI `topology clear` reverts in exactly its primary use case. No GitHub issue tracks it. File a tracking issue, or fix the one-line `new_readonly` → `new` in a small follow-up, so this doc's pointer doesn't dangle indefinitely.

crates/doublezero-serviceability-instruction/src/topology.rs
juan-malbeclabs: Good catch — you're right that `get_topology_pda` canonicalizes internally, so the PDA was never at risk and the rationale was wrong. Went with option 1: dropped `to_ascii_uppercase()` from all four builders and deleted the paragraph, restoring exact SDK byte-parity. The test comments repeating the claim are gone too.

crates/doublezero-serviceability-instruction/src/topology.rs
juan-malbeclabs: Fixed in 941c96c5. Reworded the module doc and both batch-size constants to name the real limits — the 64-account lock limit (`MAX_TX_ACCOUNT_LOCKS`) and the ~1232-byte packet size, which is the binding constraint as batches grow. The current constants stay safe under both.

crates/doublezero-serviceability-instruction/src/feed.rs
juan-malbeclabs: Filed #4079 to track the program-side gap, and added a `feed_authority_gap_is_still_open` canary test in this crate asserting `!AUTHORIZE_GATED_FLAGS.contains(&FEED_AUTHORITY)` with a pointer to the doc/issue — so closing the gap fails the test and forces this note to be revisited. 941c96c5.

crates/doublezero-serviceability-instruction/src/feed.rs
juan-malbeclabs: Fixed in 941c96c5 — the fixture now uses `FeedUpdateArgs { name: Some("Feed".into()), groups: None }` so it isn't rejected as a no-op, mirroring what 8f03288 did for `create_feed`.

ben-dz (approved):
lgtm — all findings from previous rounds verified fixed on the current head (canonicalization removed, SDK clear_topology bug tracked in #4078, batch-limit docs corrected, FEED_AUTHORITY gap tracked in #4079 with a passing canary test, realistic FeedUpdateArgs fixture).

PR #4057, opened 2026-07-13
serviceability-instruction: accesspass + resource builders (RFC-26 R8)

### LINKED ISSUE #4023: serviceability: accesspass + resource instruction builders (Rust)
Part of #4014.

Add `src/accesspass.rs` and `src/resource.rs` builders. Resource args carry data-bearing enums (`ResourceType`, `IdOrIp`) — cover their borsh encoding explicitly.

### Acceptance
- One pure builder per non-deprecated accesspass/resource instruction.
- Golden fixtures exercise the data-bearing enum encodings.

juan-malbeclabs:
## Summary

RFC-26 **R8** (stacked on the previous phase's branch). Access-pass set (conditional tenant pair)/close/check-status/set-feeds, and resource-extension allocate/create/deallocate/close (resource PDA + associated account derived from the data-bearing args.resource_type).

## Testing Verification

Closes #4023. Part of RFC-26 ([rfcs/rfc26-rust-instruction-builder-library.md](rfcs/rfc26-rust-instruction-builder-library.md)).

---

2. #4050 — R1 device
3. #4051 — R2 link
4. #4052 — R3 user
8. #4056 — R7 topology + feed
9. #4057 — R8 accesspass + resource  ← **this PR**

### DIFF
--- a/crates/doublezero-serviceability-instruction/src/accesspass.rs
+++ b/crates/doublezero-serviceability-instruction/src/accesspass.rs
@@ -176,18 +176,32 @@ mod tests {
         let user_payer = Pubkey::new_unique();
         let new_tenant = Pubkey::new_unique();
         let client_ip = Ipv4Addr::new(10, 0, 0, 1);
+        let current_tenant = Pubkey::new_unique();
         let ix = set_access_pass(
             &pid,
             &payer,
             &user_payer,
-            &Pubkey::default(),
+            &current_tenant,
             &new_tenant,
             set_args(client_ip),
         );
-        // 3 base + current_tenant(default) + new_tenant + payer + system = 7.
-        assert_eq!(ix.accounts.len(), 7);
-        assert_eq!(ix.accounts[3].pubkey, Pubkey::default());
-        assert_eq!(ix.accounts[4].pubkey, new_tenant);
+        assert_eq!(ix.data[0], 67);
+        let (accesspass, _) = get_accesspass_pda(&pid, &client_ip, &user_payer);
+        let (globalstate, _) = get_globalstate_pda(&pid);
+        // 3 base + current_tenant + new_tenant appended (either non-default), then
+        // the trailing payer/system.
+        assert_eq!(
+            ix.accounts,
+            vec![
+                AccountMeta::new(accesspass, false),
+                AccountMeta::new_readonly(globalstate, false),
+                AccountMeta::new(user_payer, false),
+                AccountMeta::new(current_tenant, false),
+                AccountMeta::new(new_tenant, false),
+                AccountMeta::new(payer, true),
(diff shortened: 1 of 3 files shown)


crates/doublezero-serviceability-instruction/src/accesspass.rs
```diff
@@ -0,0 +1,294 @@
...
+/// Accounts: `[accesspass, globalstate(readonly), feed[i]...(readonly)]`.
+///
+/// The access-pass PDA is derived from `args.client_ip` and `args.user_payer`.
+/// `feed_keys` are the feed accounts, in the same order as `args.feeds`.
+pub fn set_access_pass_feeds(
+    program_id: &Pubkey,
+    payer: &Pubkey,
+    feed_keys: &[Pubkey],
+    args: SetAccessPassFeedsArgs,
+) -> Instruction {
```
ben-dz, line 108: The comment says "panic here rather than emit a broken instruction," but `debug_assert_eq!` compiles out in release, so a release-build caller with `feed_keys.len() != args.feeds.len()` still emits the shifted instruction. It fails closed on-chain (displaced payer trips the `is_signer` assert or the feed-owner check), so no corruption lands — a comment/guard-strength nit only. Either reword to "panics in debug builds; release relies on the on-chain asserts," or promote to a hard `assert_eq!` (cold path, always a caller bug) so the comment is true.

ben-dz (commented):
Clean, mechanical PR. All eight builders reproduce their processors and the existing SDK commands byte-for-byte (account order, writable/readonly/signer flags, trailing `[payer, system]`), variant tags (67/69/70/115, 80/81/82/85) are correct, the conditional tenant-pair path matches the processor's `accounts.len() >= 7` detection, and `associated_account` derivation matches `get_resource_extension_pda`. One non-blocking nit: the `set_access_pass_feeds` comment overstates its `debug_assert_eq!` guard.

crates/doublezero-serviceability-instruction/src/accesspass.rs
juan-malbeclabs: Good catch — fixed in 3683ea40.

I went with the reword (option 1) rather than promoting to a hard `assert_eq!` to stay consistent with the crate-wide convention: every caller-bug invariant in this crate uses `debug_assert`, including the 6 in `user.rs` and the one in `device.rs` whose comment literally calls it "the exact invariant this crate exists to protect" — yet it is still `debug_assert!`. Promoting only this one would diverge from that identical-severity sibling. The new comment states the debug-only behavior explicitly and folds in your fail-closed explanation (displaced payer trips `is_signer` / feed-owner check).

ben-dz (approved):

PR #4058, opened 2026-07-13
serviceability-instruction: globalstate/globalconfig/allowlist/index/migrate builders (RFC-26 R9)

### LINKED ISSUE #4024: serviceability: globalstate/globalconfig/allowlist/index/migrate instruction builders (Rust)
Part of #4014.

Add the admin/misc builders: `src/globalstate.rs`, `src/globalconfig.rs`, `src/allowlist.rs`, `src/index.rs`, `src/migrate.rs` (init/set-authority/set-config/feature-flags, allowlist toggles, index, migrate).

### Excluded variants (documented in `lib.rs`)
The enum has 116 variants (tags 0–115, contiguous); builders cover all buildable variants (~94). Excluded, with no builder emitted (no `Err`/panic stubs — builders are infallible):
- **Explicit placeholders** kept only for discriminant stability: `Deprecated95`, `Deprecated96`, `Deprecated102`, `Deprecated103`, `Deprecated111`.
- **Deprecated handlers** returning `DoubleZeroError::Deprecated` (e.g. `ActivateDevice`, `RejectDevice`, `CloseAccountDevice`, the corresponding link/user/multicast lifecycle variants, and several `*DeviceInterface` variants).

### Acceptance
- One pure builder per non-deprecated instruction in these domains; empty-arg variants pack to just the tag byte.
- Unit tests + `ix_*` golden fixtures. After this PR all buildable variants are covered.

juan-malbeclabs:
## Summary

RFC-26 **R9** (stacked on the previous phase's branch). init_global_state and migrate (no authorize -> build); setters, set_global_config (config PDA + all 8 resource pools), foundation/QA allowlist toggles, index create/delete. After this every buildable variant has a builder.

## Testing Verification

Closes #4024. Part of RFC-26 ([rfcs/rfc26-rust-instruction-builder-library.md](rfcs/rfc26-rust-instruction-builder-library.md)).

---

2. #4050 — R1 device
3. #4051 — R2 link
4. #4052 — R3 user
8. #4056 — R7 topology + feed
10. #4058 — R9 globalstate/config/allowlist/index/migrate  ← **this PR**

### DIFF
--- a/crates/doublezero-serviceability-instruction/src/allowlist.rs
+++ b/crates/doublezero-serviceability-instruction/src/allowlist.rs
@@ -1,0 +1,123 @@
+//! Allowlist-domain instruction builders (foundation + QA).
+//!
+//! Each toggles a single pubkey on the GlobalState allowlist; the only account is
+//! GlobalState. All route through `authorize()` -> [`common::build_with_permission`].
+
+use crate::common;
+use doublezero_serviceability::{
+    instructions::DoubleZeroInstruction,
+    pda::get_globalstate_pda,
+    processors::allowlist::{
+        foundation::{add::AddFoundationAllowlistArgs, remove::RemoveFoundationAllowlistArgs},
+        qa::{add::AddQaAllowlistArgs, remove::RemoveQaAllowlistArgs},
+    },
+};
+use solana_program::{
+    instruction::{AccountMeta, Instruction},
+    pubkey::Pubkey,
+};
+
+fn build_globalstate_only(
+    program_id: &Pubkey,
+    payer: &Pubkey,
+    instruction: DoubleZeroInstruction,
+) -> Instruction {
+    let (globalstate, _) = get_globalstate_pda(program_id);
+    common::build_with_permission(
+        program_id,
+        instruction,
+        vec![AccountMeta::new(globalstate, false)],
+        payer,
+    )
+}
+
+/// `AddFoundationAllowlist` (variant 4). Accounts: `[globalstate]`.
+pub fn add_foundation_allowlist(
+    program_id: &Pubkey,
+    payer: &Pubkey,
+    args: AddFoundationAllowlistArgs,
+) -> Instruction {
+    build_globalstate_only(
+        program_id,
+        payer,
+        DoubleZeroInstruction::AddFoundationAllowlist(args),
+    )
+}
+
+/// `RemoveFoundationAllowlist` (variant 5). Accounts: `[globalstate]`.
+pub fn remove_foundation_allowlist(
+    program_id: &Pubkey,
+    payer: &Pubkey,
+    args: RemoveFoundationAllowlistArgs,
+) -> Instruction {
+    build_globalstate_only(
+        program_id,
+        payer,
+        DoubleZeroInstruction::RemoveFoundationAllowlist(args),
+    )
+}
+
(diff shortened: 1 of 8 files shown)


crates/doublezero-serviceability-instruction/src/allowlist.rs
```diff
@@ -0,0 +1,123 @@
...
+        foundation::{add::AddFoundationAllowlistArgs, remove::RemoveFoundationAllowlistArgs},
+        qa::{add::AddQaAllowlistArgs, remove::RemoveQaAllowlistArgs},
+    },
+};
+use solana_program::{
+    instruction::{AccountMeta, Instruction},
+    pubkey::Pubkey,
+};
+
+fn build_globalstate_only(
```
ben-dz: `build_globalstate_only` is duplicated verbatim from `globalstate.rs` (extracted in d60c29d). Consider hoisting the shared helper into `common.rs`, or leave the duplication deliberately. Non-blocking.

ben-dz (approved):
lgtm — builders faithfully mirror the processors (all 13 tag bytes, account orders, writability/signer flags, PDA derivations, and the build vs build_with_permission split verified), byte-parity with the existing SDK commands holds, and tests pin the full AccountMeta vector + tag byte per builder. One trivial nit; one informational note about a pre-existing SDK `SetVersionCommand` bug (the new `set_min_version` builder is correct where the SDK is stale) worth a separate follow-up.

- Informational (pre-existing, not this PR): `smartcontract/sdk/rs/src/commands/globalstate/setversion.rs` sends only `[globalstate]`, but the processor reads `[program_config, globalstate, payer, system]` since #3977, so the SDK's `SetVersionCommand` fails onchain today. The new `set_min_version` builder is correct — an intentional, documented divergence from SDK byte-parity. Worth a separate follow-up to fix the SDK command.

crates/doublezero-serviceability-instruction/src/allowlist.rs
juan-malbeclabs: Hoisted into `common.rs` as `pub(crate) fn build_globalstate_only`; both `globalstate.rs` and `allowlist.rs` now call it, and the test-only imports moved into their test modules. Fixed in 36666295.

juan-malbeclabs:
> Informational (pre-existing): the SDK's `SetVersionCommand` sends only `[globalstate]` while the processor reads `[program_config, globalstate, payer, system]` since #3977.

Good catch — rather than defer this, I fixed it here in 5259c95c: `SetVersionCommand` now prepends `program_config` to match the processor's account order, and the test pins the two-account vector. Heads up that this widens the PR slightly beyond the RFC-26 R9 builders; happy to split it onto its own branch if you'd prefer to keep this one strictly scoped.

ben-dz:
> > Informational (pre-existing): the SDK's `SetVersionCommand` sends only `[globalstate]` while the processor reads `[program_config, globalstate, payer, system]` since #3977.
> 
> Good catch — rather than defer this, I fixed it here in [5259c95](https://github.com/malbeclabs/doublezero/commit/5259c95c5ae10c5b1b6126a3fdba22c877b89c52): `SetVersionCommand` now prepends `program_config` to match the processor's account order, and the test pins the two-account vector. Heads up that this widens the PR slightly beyond the RFC-26 R9 builders; happy to split it onto its own branch if you'd prefer to keep this one strictly scoped.

LGTM, no need for a new branch.

PR #4059, opened 2026-07-13
serviceability-instruction: golden fixtures + CI guard (RFC-26 RF)

### LINKED ISSUE #4026: serviceability: golden fixtures for instruction builders + CI diff guard (Rust)
Part of #4014.

Extend the shared fixture infrastructure so builder output is captured as golden files and guarded in CI.

### Scope
- Extend `sdk/serviceability/testdata/fixtures/generate-fixtures/src/main.rs` to depend on the builder crate and emit, per instruction (deterministic inputs, fixed `program_id`):
  - `ix_<name>.bin` = `instruction.data` (tag + borsh).
  - `ix_<name>.json` = `{ variant, data_hex, accounts: [{pubkey, is_signer, is_writable}] }` (account order + flags).
- Wire regeneration into `make generate-fixtures`.
- CI check that fails on uncommitted fixture diffs.

### Notes
- This can land early (alongside R0) to establish the format, then each domain PR adds its fixtures. Fixtures remain available for future Go/Python/TS parity checks (out of scope now).

### Acceptance
- `make generate-fixtures` produces stable, committed `ix_*` files; CI fails if they drift.

juan-malbeclabs:
## Summary

RFC-26 **RF**: golden fixtures for the instruction builders + a CI drift guard.

- Extend the serviceability fixture generator to depend on `doublezero-serviceability-instruction` and emit, per instruction, `ix_<name>.bin` (wire bytes = tag + borsh) and `ix_<name>.json` (`{ variant, data_hex, accounts: [{pubkey, is_signer, is_writable}] }`) from fixed, deterministic inputs.
- Representative set covering the trickiest layouts: `create_device` (variable dz_prefix), `delete_device` (atomic close), `create_link`, `create_subscribe_user` (optional feed), `create_user` (length-detected), `clear_topology` / `assign_topology_node_segments` (batched), `set_global_config` (config PDA + all 8 pools).
- `make generate-fixtures-check` (regenerate + fail on drift, scoped to the emitted `.bin`/`.json`) and a `fixtures-check` job in the `rust` workflow.

These fixtures capture byte-for-byte the current SDK trailing convention (they will drive Go/Python/TS parity in a later phase).

## Testing Verification

- `make generate-fixtures-check` passes (regenerated output is byte-identical to the committed fixtures); a hand-edited fixture makes it fail as expected.

Closes #4026. Part of RFC-26 ([rfcs/rfc26-rust-instruction-builder-library.md](rfcs/rfc26-rust-instruction-builder-library.md)).

---

Stacked PRs, merge in order (each based on the previous one's branch):

2. #4050 — R1 device
3. #4051 — R2 link
4. #4052 — R3 user
8. #4056 — R7 topology + feed
11. this PR — RF fixtures + CI guard

R10 (commands/* migration + program-test) is the final PR and will carry the global changelog entry.


.github/workflows/rust.yml
```diff
@@ -25,6 +25,14 @@ jobs:
           components: clippy, rustfmt
       - uses: Swatinem/rust-cache@v2
       - run: make rust-lint
+  fixtures-check:
```
ben-dz: This `fixtures-check` job runs `make check-fixtures`, byte-for-byte the same command as the pre-existing `sdk-fixture-drift` job in `.github/workflows/sdk.yml:55-63`. `make check-fixtures` regenerates all SDK generators (not just serviceability) and `git diff`s the goldens, so the two jobs are strict duplicates — both trigger on every PR and push to main/hotfix with no path filters, running the drift check twice per PR for zero added coverage. This copy is also worse-provisioned: a paid `ubuntu-24.04-16c-64gb` runner (the existing one fits on `ubuntu-latest`) and no `timeout-minutes: 15` guard, so a hung download runs to the 6-hour default on 16 cores. Pick one home: delete this job, or move the check into rust.yml and delete `sdk-fixture-drift` in the same PR. If it stays, drop to `ubuntu-latest` and add a timeout — the work is four small cargo builds plus a `git diff`, and `rust-cache@v2` won't even warm-cache the generator crates' separate `target/` dirs without a `workspaces:` setting.

ben-dz (commented):
Two things worth changing before merge; nothing blocking. (1) The new `fixtures-check` job in rust.yml is a strict duplicate of the pre-existing `sdk-fixture-drift` job in sdk.yml — same `make check-fixtures`, both triggered on every PR/push with no path filters — so the drift check runs twice per PR for zero added coverage, and the new copy runs on an oversized `ubuntu-24.04-16c-64gb` runner with no `timeout-minutes`. Delete one, or move the check and delete the other. (2) The `set_global_config` fixture is built from `SetGlobalConfigArgs::default()`, emitting an all-zero body that pins only the length — it can't catch field-order/endianness/signedness/Option-encoding regressions, which is the whole point of a wire-format golden; use distinct nonzero values. Low: the fixture JSON records neither `program_id` nor the input args, so a future parity harness must read the Rust source to reproduce inputs. Fixtures are deterministic (regenerated locally, byte-identical) and otherwise correctly capture the builder output for the representative set.

sdk/serviceability/testdata/fixtures/generate-fixtures/src/main.rs
```diff
@@ -160,6 +171,192 @@ fn generate_user_delete_args(dir: &Path) {
...
+        &[a],
+        AssignTopologyNodeSegmentsArgs {
+            name: "topo".into(),
+        },
+    );
+    write_ix_fixture(dir, "assign_topology_node_segments", &assign);
+
+    // set_global_config (3): the config PDA plus all eight resource pools.
+    let set_global_config =
+        dzi::globalconfig::set_global_config(&program_id, &payer, SetGlobalConfigArgs::default());
```
ben-dz: `SetGlobalConfigArgs::default()` produces an all-zero body — the emitted `data_hex` is the tag byte plus 29 zero bytes. A wire-format golden should pin field order, endianness, signedness, and Option Some/None encoding, but an all-zero payload catches none of these (a `None` option byte is indistinguishable from a zeroed field; reordering the `NetworkV4` blocks or swapping the two u32 ASNs is invisible). This is the builder the PR highlights as exercising all eight resource pools, yet its fixture is the least discriminating. Build the args with distinct nonzero values per field, as `generate_global_config` already does for the account fixture.

sdk/serviceability/testdata/fixtures/generate-fixtures/src/main.rs
```diff
@@ -160,6 +171,192 @@ fn generate_user_delete_args(dir: &Path) {
...
+
+#[derive(Serialize)]
+struct IxAccountMeta {
+    pubkey: String,
+    is_signer: bool,
+    is_writable: bool,
+}
+
+#[derive(Serialize)]
+struct IxFixtureMeta {
```
ben-dz, line 191: `IxFixtureMeta` records `variant`/`data_hex`/`accounts` but not `instruction.program_id`, and the fixed program id (`pubkey_from_byte(255)`) plus per-builder arg values live only in this Rust source. A later Go/Python/TS parity harness must read the generator to reproduce inputs, and a program-id mismatch surfaces only as opaque PDA-key diffs. Add a `program_id` field (and optionally an args/fields block, as the account fixtures carry) so the goldens are self-describing. Cheaper to fix now, before the parity PR locks in the fixture shape.

.github/workflows/rust.yml
juan-malbeclabs: Done — deleted the `fixtures-check` job. `sdk-fixture-drift` in sdk.yml already runs `make check-fixtures` on `ubuntu-latest` with `timeout-minutes: 15`, so this was a strict duplicate with no added coverage; keeping the pre-existing, better-provisioned one. (ebe5e544)

sdk/serviceability/testdata/fixtures/generate-fixtures/src/main.rs
juan-malbeclabs: Fixed — `set_global_config` now builds `SetGlobalConfigArgs` with distinct nonzero values per field (mirroring `generate_global_config`): `local_asn: 65000`, `remote_asn: 65001`, the three tunnel/multicast `NetworkV4` blocks, `next_bgp_community: Some(10042)`, and `multicast_publisher_block`. The body grew from 30 to 32 bytes and `data_hex` is now `03e8fd0000e9fd0000...013a27...`, so the golden pins field order, endianness, both u32 ASNs, and the `Some`-encoding of the optional community rather than only the length. (ebe5e544)

sdk/serviceability/testdata/fixtures/generate-fixtures/src/main.rs
juan-malbeclabs: Added — `IxFixtureMeta` now records `program_id` (read straight off `ix.program_id`), so every `ix_*.json` golden is self-describing on the program id and a parity harness no longer has to read the generator to catch a mismatch. Left the per-builder arg block out for now since no consumer reads these yet (the parity harness is still future work); happy to add a `fields` block in the parity PR if it's useful there. (ebe5e544)

ben-dz (approved):
All three findings from the prior review are addressed in ebe5e544 and verified in the current code: the duplicate `fixtures-check` CI job is deleted from rust.yml (sdk.yml's better-provisioned `sdk-fixture-drift` retained); `set_global_config` now builds `SetGlobalConfigArgs` with distinct nonzero values so the golden pins field order, both u32 ASNs, and the optional community's `Some`-encoding instead of an all-zero body; and `IxFixtureMeta` records `program_id` in every `ix_*.json`. The passing `sdk-fixture-drift` job confirms the goldens regenerate byte-identically. lgtm.

Issue #3612, opened 2026-04-29
activator removal phase 3.1: delete activator/ crate from workspace


elitegreg:
Part of the activator removal effort — see tracker malbeclabs/doublezero#3607.

**Phase / PR:** Phase 3, PR 3.1

**Gate:** Phases 1 and 2 must be merged.

## Goal

Delete the `activator/` crate and unbreak the workspace.

## Files

- `Cargo.toml` (workspace root) — remove `"activator"` from workspace members.
- `activator/` — delete entire directory (8.6k LoC, 23 files, plus README).
- `Makefile` — remove any activator-specific build/test targets.
- Any workspace-level lint allow-list entries scoped to the activator.

## Acceptance

- `cargo build --workspace` succeeds.
- `make rust-build`, `make rust-lint`, `make rust-test` all pass.
- `grep -r doublezero-activator .` (excluding CHANGELOG and rfcs) returns nothing.

## Notes

SDK command modules under `smartcontract/sdk/rs/src/commands/.../{activate,reject,closeaccount}.rs` **stay** at this point — old CLIs still call into the corresponding onchain instructions until the min-version gate. They are removed in phase 7.

PR #4060, opened 2026-07-14
sdk: migrate serviceability commands to RFC-26 instruction builders (R10)

### LINKED ISSUE #4078: sdk: ClearTopologyCommand passes topology PDA read-only but the processor requires it writable
## Summary

The Rust SDK `ClearTopologyCommand` passes the topology PDA as **read-only**, but the on-chain `process_topology_clear` processor asserts the topology account is **writable** whenever the topology still exists and at least one link reference is actually cleared. As a result, `topology clear` reverts in its primary use case: clearing a live topology.

## Details

- SDK: `smartcontract/sdk/rs/src/commands/topology/clear.rs:31`
  ```rust
  AccountMeta::new_readonly(topology_pda, false),
  ```
- Processor: `smartcontract/programs/doublezero-serviceability/src/processors/topology/clear.rs:121-125`
  ```rust
  if !topology_account.data_is_empty() && cleared_count > 0 {
      assert!(
          topology_account.is_writable,
          "Topology Account is not writable"
      );
      // ... decrements reference_count
  }
  ```
  The processor decrements `reference_count` by the number of links whose reference was removed, so it needs the account writable.

## Impact (user-facing)

Deleting a topology requires `reference_count == 0` (`processors/topology/delete.rs:77`). Clearing a live topology's link references is therefore the **mandatory precursor to delete**. Because the SDK passes the topology PDA read-only, `topology clear` panics with `"Topology Account is not writable"` in exactly that path — the command's primary use case. It only succeeds on the already-closed-topology path (where the processor skips the write).

## Fix

One line in the SDK command:
```rust
- AccountMeta::new_readonly(topology_pda, false),
+ AccountMeta::new(topology_pda, false),
```
Writable is a strict superset privilege, so no other execution path regresses.

## Context

The RFC-26 R7 instruction-builder crate already builds `clear_topology` with the topology PDA **writable** (matching the processor) and documents this as its one deliberate byte-parity break with the SDK — see `crates/doublezero-serviceability-instruction/src/topology.rs` (`clear_topology` doc comment). This issue tracks the corresponding SDK fix so that doc pointer does not dangle.

juan-malbeclabs:
## Summary of Changes
- Delegate every serviceability `commands/*` `execute()` to the pure `doublezero-serviceability-instruction` builders (RFC-26), so account layout has a single source of truth in the builder crate rather than being hand-assembled per command.
- Replace the four `DoubleZeroClient::execute_*(instruction, accounts)` trait methods and the client-side account assembly with a single `send_transaction(Instruction)` that just prepends the compute-budget prelude, signs, and sends. Command `execute()` signatures/returns are unchanged, so CLI/sentinel/daemon consumers are unaffected. The `_quiet` variant is not carried forward — its only caller was the `activator/` crate, deleted in #3647.
- **Fixes #4078:** `doublezero topology clear` reverted with `Topology Account is not writable` on a live topology — the SDK passed the topology PDA read-only while the processor asserts it writable whenever it actually clears a link reference (it decrements `reference_count`). The command therefore only worked on the already-closed-topology path, blocking `topology delete`, which requires `reference_count == 0`. The `clear_topology` builder passes the account writable; writable is a strict superset privilege, so no other path changes.
- **Breaking (see CHANGELOG `### Breaking`):** the Rust SDK no longer attaches the payer's Permission PDA to serviceability transactions — `authorize()` uses the legacy GlobalState allowlist path until the builder-side Permission append is re-enabled per instruction (must be sequenced with the permission-model rollout). A payer whose only grant is an activated Permission account would lose access to gated commands.
  - **No cluster is exposed — verified against live chain state.** `RequirePermissionAccounts` is off everywhere, so the legacy fallback is live. mainnet-beta's 8 Permission accounts all grant `FOUNDATION` to keys already in `foundation_allowlist`. The one grant on testnet/devnet that no legacy authority covers (`ACCESS_PASS_ADMIN` to `EWFXDTBWhNxsmj4mbahZL7HH1XpmC6Ao3LZUhJGCyyPh`) belongs to the shred oracle, which never uses `DoubleZeroClient` — it assembles its serviceability instructions itself and appends its own Permission account after its own RPC existence check (`doublezero-shreds`, `crates/shred-oracle/src/dz_ledger.rs`, `permission_account_meta`, 9 call sites). It is unaffected by this send path. No deploy blocker.
  - **The next grant is now self-documenting.** `permission set` emits a `tracing::warn!` whenever the resulting bitmask carries an `AUTHORIZE_GATED_FLAGS` bit, and `permission audit` renders a matching paragraph (plus a `sdk_attaches_permission_accounts` field in `--json`), stating that the CLI/Rust SDK do not attach Permission accounts yet so the key still needs a legacy allowlist or authority entry. The chain-state check above is point-in-time by nature; this covers grants made after it. Both are removed when the append is re-enabled.
- Add a `solana-program-test` safety net that runs the highest-cardinality builders (`create_device`, `create_link`, `create_subscribe_user`, atomic `delete_device`, `clear_topology`) against the real program to catch account-order drift.
- Remove the now-unreachable permission-resolve/cache/stale-retry machinery from `client.rs` (the append, when re-enabled, lands in the builder — derived offline — so the RPC-based path is not needed).

This is the final PR of the RFC-26 stack; it carries the global changelog entry. See [`rfcs/rfc26-rust-instruction-builder-library.md`](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc26-rust-instruction-builder-library.md).

> **Stack:** R0 → R1 → … → R9 → RF (#4059, merged) → **R10 (this PR)**. Now rebased onto `main`, so the diff below is R10 only.

## Diff Breakdown
| Core logic   |    13 | +868 / -1441  | -573  |
| Scaffolding  |    78 | +2396 / -3298 | -902  |
| Tests        |     1 | +763 / -0     | +763  |
| Config       |     3 | +6 / -0       | +6    |
| Docs         |     2 | +10 / -3      | +7    |
| Generated    |     1 | +2 / -0       | +2    |

Net **−697 lines**: a mostly-mechanical migration (77 command delegations) on top of a genuine send-path/trait rewrite and one new program-test suite. This exceeds doublezero's ~500-line PR norm, but the migration + its program-test safety net are one logical unit and the net change is a reduction; splitting would land the safety net separately from the code it guards.

<details>

- `smartcontract/programs/doublezero-serviceability/tests/rfc26_builders_test.rs` (+763) — new program-test safety net; builds via the RFC-26 builders and runs them against the real program.
- `smartcontract/sdk/rs/src/client.rs` (+70/-560) — rework `execute_transaction_inner` into `send_transaction_inner(ix)`; drop `assemble_instructions`, the permission-resolve/cache and stale-retry loop, and their tests.
- `smartcontract/sdk/rs/src/commands/user/delete.rs` (+153/-246) — composed unsubscribe + `delete_user` builder; keeps cascade orchestration.
- `smartcontract/sdk/rs/src/commands/exchange/setdevice.rs` (+117/-131) — two-send command migrated to `set_device_exchange` builder.
- `smartcontract/sdk/rs/src/commands/tenant/delete.rs` (+108/-116) — `delete_tenant` builder; keeps cascade + reference-count poll.
- `smartcontract/sdk/rs/src/commands/link/update.rs` (+77/-81) — `update_link` builder with `LinkUpdateAuthority`; builder recomputes `use_onchain_allocation`.
- `smartcontract/sdk/rs/src/commands/device/delete.rs` (+59/-87) — legacy/atomic paths via `delete_device` + `DeviceDeleteResources`.
- `smartcontract/sdk/rs/src/commands/topology/clear.rs` (+57/-85) — batched sends via `clear_topology`; carries the #4078 writable-PDA fix.
- `smartcontract/sdk/rs/src/doublezeroclient.rs` (+7/-35) — trait: four `execute_*` methods → a single `send_transaction`.
- `smartcontract/cli/src/permission/{set,audit}.rs` (+92/-1) — operator-facing warning that Permission grants are inert until the builder-side append returns.

</details>

## Testing Verification
- New `solana-program-test` suite (`rfc26_builders_test.rs`): 5 tests exercising the highest-cardinality builders against the real program — `create_device`, `create_link` + `clear_topology`, `create_subscribe_user` (atomic subscriber), and atomic `delete_device` (owners read on-chain in processor order). All pass.
- `clear_topology` covers both processor branches: `cleared_count == 0` (untagged topology, no write) and `cleared_count > 0` (the topology the link is tagged into), the latter exercising the `is_writable` assert and `reference_count` decrement that #4078 is about. Verified it is a real guard — flipping the builder to `new_readonly` fails with `Topology Account is not writable`.
- `cargo test -p doublezero_sdk`: 167 passed. Rewritten command tests assert `predicate::eq(builder(...))` on the exact instruction each command sends. The composed cascades in `user/delete` and `tenant/delete` pin every step — the resolved access-pass PDA, `dz_prefix_count`, the optional `tenant` account, the `[current_tenant, new_tenant]` reset pair, and the trailing `owner` (which differs from the payer on the foundation-key path).
- Assertion strength across `sdk/rs/src/commands/` went up, not down: `predicate::always()` appears in 0 files here, down from 8 on `main` (where it relaxed the *accounts* argument of the two-arg `execute_*` expectation). The four `multicastgroup/allowlist/*` tests now pin the built instruction end to end — on `main` they asserted the `DoubleZeroInstruction` but not the accounts — with `user_payer` deliberately distinct from the group pubkey so an account-meta swap fails. Confirmed discriminating: perturbing `client_ip` in the expected instruction fails with `No matching expectation found`.
- `permission set`'s new warning is gated on `AUTHORIZE_GATED_FLAGS`, unit-tested to fire for every gated flag and stay quiet for `SENTINEL` / `QA` / `FEED_AUTHORITY` (which no processor hands to `authorize()`), so it cannot become background noise.
- The `u8` bounds on `resource_count` / `topology_count` stay errors rather than builder panics: `device/create`, `device/interface/create` and `device/interface/update` guard before building, with tests asserting the exact messages for both interface guards.
- Confirmed the batched topology and multi-send commands (`exchange/setdevice`, `topology/clear`, `topology/assign_node_segments`) preserve their per-instruction sequencing.


CHANGELOG.md
```diff
@@ -23,6 +23,9 @@ All notable changes to this project will be documented in this file.
...
+  - Migrate every serviceability `commands/*` `execute()` to delegate to those builders and a new `DoubleZeroClient::send_transaction` (compute-budget prelude + sign + send), replacing the four `execute_*(instruction, accounts)` methods and the client-side account assembly. **Behavior change:** the Rust SDK no longer attaches the payer's Permission PDA to serviceability transactions — `authorize()` uses the legacy GlobalState allowlist path until the builder-side Permission append is re-enabled per instruction (must be sequenced with the permission-model rollout). Command `execute()` signatures/returns are unchanged, so CLI/sentinel/daemon consumers are unaffected.
```
ben-dz: This is an authorization behavior change filed under `### Changes`, and it ships without a recorded verification gate.

`build_with_permission` defers the append (`crates/doublezero-serviceability-instruction/src/common.rs:87`) and `client.rs` no longer resolves the payer's Permission PDA, so every migrated instruction now takes the `None` branch of `authorize()` (`smartcontract/programs/doublezero-serviceability/src/authorize.rs:108-124`). With `RequirePermissionAccounts` off that falls through to `check_legacy_any` — so **a payer whose only grant is an activated Permission account (not in `foundation_allowlist`/`qa_allowlist`, not an authority key) starts getting `NotAllowed` on every gated command as soon as this SDK ships.** Nothing distinguishes that from a genuine permission problem; the operator just loses access.

Two things needed before release:

1. **Confirm zero permission-only grants exist** on devnet/testnet/mainnet-beta. `doublezero permission audit` does *not* answer this: `build_report` walks `AUTHORIZE_GATED_FLAGS × legacy_keys_for_flags(...)` and reports legacy keys *lacking* a Permission account (`smartcontract/cli/src/permission/audit.rs:63-88`) — the strict-mode direction. The inverse gap this PR opens is not in `AuditReport` at all, so verifying it means diffing `permission list` against the GlobalState allowlists by hand. Please state the result here.
2. **Keep `RequirePermissionAccounts` off** on every cluster until the builder-side append activates — with it on, every gated instruction from this SDK fails except the foundation `PERMISSION_ADMIN` recovery at `authorize.rs:130`.

Fix: move this bullet (or split out the behavior-change clause) into the `### Breaking` section that already exists in this Unreleased block at line 8, alongside the `feed_pk → feed_pks` entry, and record the audit result there as a deploy note — that bullet is the precedent for exactly this. Filed under `### Changes`, an operator reading release notes has no reason to check their allowlist membership before upgrading.

While here: this bullet should also note that `ClearTopologyCommand` now sends the topology account writable. This PR is the vehicle that ships the #4078 fix to SDK consumers, and the entry currently reads as pure refactor.

smartcontract/sdk/rs/src/commands/user/delete.rs
```diff
@@ -304,30 +270,14 @@ mod tests {
...
-                predicate::eq(vec![
-                    AccountMeta::new(mgroup_pubkey, false),
-                    AccountMeta::new(accesspass_pubkey, false),
-                    AccountMeta::new(user_pubkey, false),
-                    AccountMeta::new(globalstate_pubkey, false),
-                    AccountMeta::new(multicast_publisher_block_ext_unsub, false),
-                ]),
-            )
+            .expect_send_transaction()
+            .with(predicate::always())
```
ben-dz: The Testing Verification section says this PR includes "tightened expectations for the composed `user/delete`, `tenant/delete`, `user/requestban` commands" — for the first two the opposite happened. `predicate::always()` count went 0 → 6 here (lines 277, 317, 463, 503, 691, 739) and 3 → 7 in `tenant/delete.rs`. On `main`, `user/delete.rs:311` asserted the exact `UpdateMulticastGroupRoles` args *and* the exact five-account vector; that is now `predicate::always()`.

This is the one path in the migration that is more than mechanical: `DeleteUserCommand` resolves `accesspass_pk` through `GetAccessPassCommand`'s dynamic-then-exact fallback, derives `dz_prefix_count` from the device, and gates the optional `tenant` account on `user.tenant_pk != Pubkey::default()` (lines 60-93). None of that is asserted any more — a wrong access-pass PDA, a dropped tenant account, or a swapped unsubscribe/delete ordering all pass, and this cascade is what users hit on `doublezero user delete`.

Fix: use the pattern the sibling tests already use — `device/delete.rs` and `exchange/setdevice.rs` build `let expected = <builder>(...)` from the test's own values and assert `predicate::eq(expected)`. The single-send test at line 860 shows it works here too; apply it to these six expectations.

smartcontract/sdk/rs/src/commands/tenant/delete.rs
```diff
@@ -316,19 +305,13 @@ mod tests {
...
-            .expect_execute_authorized_transaction()
-            .with(
-                predicate::eq(DoubleZeroInstruction::DeleteUser(UserDeleteArgs {
-                    dz_prefix_count: 1,
-                    multicast_publisher_count: 1,
-                })),
-                predicate::always(),
-            )
+            .expect_send_transaction()
+            .with(predicate::always())
```
ben-dz: Same coverage regression as `user/delete.rs:277`: `predicate::always()` count went 3 → 7 in this file (lines 311, 335, 352, 438, 446), so the cascade's exact `UserDeleteArgs` / `SetAccessPassArgs` and account vectors are no longer pinned. With every predicate `always()`, even a swapped `SetAccessPass`/`DeleteTenant` ordering passes despite the `Sequence`.

Fix: build `let expected = <builder>(...)` from the test's own values and assert `predicate::eq(expected)`, as `device/delete.rs` and `exchange/setdevice.rs` do.

smartcontract/sdk/rs/src/doublezeroclient.rs
```diff
@@ -37,36 +34,15 @@ pub trait DoubleZeroClient {
...
-    ) -> eyre::Result<Signature>;
-
-    /// Like `execute_authorized_transaction`, but suppresses program log output on simulation failure.
-    /// Use this for authorized transactions where simulation failures are expected (e.g., race conditions).
-    fn execute_authorized_transaction_quiet(
-        &self,
-        instruction: DoubleZeroInstruction,
-        accounts: Vec<AccountMeta>,
-    ) -> eyre::Result<Signature>;
+    fn send_transaction_quiet(&self, instruction: Instruction) -> eyre::Result<Signature>;
```
ben-dz: `send_transaction_quiet` has no consumer anywhere — zero callers outside this trait and its impl, on `main` and on this branch. Its only consumer was the activator, which introduced the quiet path in #3197 and was itself deleted from the workspace (`CHANGELOG.md:384`, #3612). `SimulationError` / `SimulationTransactionError` are likewise produced only inside `client.rs:244-249`'s quiet branch and consumed by nobody.

RFC-26 left the trait shape "to be settled in R10 review," so this is the moment: drop `send_transaction_quiet`, the `quiet` parameter on `send_transaction_inner`, and both error types, leaving `send_transaction(Instruction)` as the single method. If the shape is being kept for a future out-of-repo consumer, worth saying so in the RFC — otherwise it's dead surface carried into the new trait.

smartcontract/programs/doublezero-serviceability/tests/rfc26_builders_test.rs
```diff
@@ -0,0 +1,696 @@
...
+        .expect("create_link builder should be accepted by the program");
+
+    let link = get_account_data(&mut banks_client, link_pubkey)
+        .await
+        .expect("link should exist")
+        .get_tunnel()
+        .unwrap();
+    assert_eq!(link.status, LinkStatus::Activated);
+
+    // clear_topology (batched) drift check. The builder passes the topology
```
ben-dz, line 400: This comment says "The builder passes the topology read-only, matching the SDK command's long-standing contract." The builder passes it **writable** (`crates/doublezero-serviceability-instruction/src/topology.rs:103`), documented there as the crate's one deliberate parity break with the SDK, fixing #4078; `ix_clear_topology.json` pins `is_writable: true`. The comment describes the pre-migration SDK command, and it is the kind of wrong that invites someone to "fix" the flag back.

The test then targets a fresh topology the link was never tagged into so `cleared_count == 0`, which means the processor branch at `processors/topology/clear.rs:121` — the one that asserts `topology_account.is_writable` and applies the `saturating_sub` ref-count decrement — never executes. The changelog advertises this suite as running `clear_topology` "against the real program to catch account-order drift," but the branch that motivated the divergence is unguarded.

Fix: target the topology the link *is* tagged into so `cleared_count > 0` — that exercises both the writable requirement and the decrement. Correct the comment either way.

smartcontract/sdk/rs/src/commands/device/interface/update.rs
```diff
@@ -36,49 +31,21 @@ pub struct UpdateDeviceInterfaceCommand {
...
-                accounts.push(AccountMeta::new_readonly(topology_pda, false));
-            }
-        }
-
-        client.execute_authorized_transaction(
-            DoubleZeroInstruction::UpdateDeviceInterface(DeviceInterfaceUpdateArgs {
+        // The builder appends the SegmentRoutingIds resource (when node_segment_idx
+        // or topologies change) and the topology PDAs, and writes update_topologies /
+        // topology_count from the topology_names choice.
+        client.send_transaction(update_device_interface(
```
ben-dz, line 54: This command dropped its `u8::try_from(n).map_err(|_| eyre!("too many topologies for one UpdateDeviceInterface call: {n} > 255"))?` guard and now reaches the builder's `u8::try_from(topology_count).expect(...)` (`crates/doublezero-serviceability-instruction/src/device.rs:457`) — a release-build panic on a CLI-supplied list. `device/interface/create.rs:38` and `device/create.rs` (`resource_count`) lost the equivalent guards the same way.

The builders being infallible is fine; the command is the right place for the check, and `user/delete.rs:80-87` already keeps exactly that guard for `dz_prefix_count`. Restore the one-line `u8::try_from(...).map_err(...)` in the three commands so oversized input stays an error.

smartcontract/sdk/rs/src/commands/topology/clear.rs
```diff
@@ -1,15 +1,13 @@
...
+use crate::DoubleZeroClient;
+use doublezero_serviceability::processors::topology::clear::TopologyClearArgs;
+use doublezero_serviceability_instruction::topology::clear_topology_batched;
+use solana_sdk::{pubkey::Pubkey, signature::Signature};
 
 /// Max link accounts per clear transaction. Solana caps transactions at 32
 /// accounts; with 2 fixed accounts (topology PDA, globalstate) plus the payer
-/// and system_program appended by the client, we stay well under that limit at
+/// and system_program appended by the builder, we stay well under that limit at
 /// 16 (same constant as backfill).
```
ben-dz: Two factual errors in this cap comment: "Solana caps transactions at 32 accounts" (the builder's own note at `crates/doublezero-serviceability-instruction/src/topology.rs:27` says the 64-account lock limit), and "(same constant as backfill)" (`CLEAR_BATCH_SIZE` is 16, `BACKFILL_BATCH_SIZE` is 4). Both predate the PR but these lines were rewritten in it, and wrong headroom math is what a future reader would take from them.

ben-dz (changes requested):
Two things should change before this merges.

**The Permission-PDA drop needs release gating.** With the builder-side append deferred and `client.rs`'s resolve/cache/retry deleted, every migrated instruction takes the `None` branch of `authorize()`. A payer whose only grant is an activated Permission account — not on `foundation_allowlist`/`qa_allowlist`, not an authority key — starts getting `NotAllowed` on every gated command the moment this ships. Please confirm zero such grants exist on devnet/testnet/mainnet-beta, and note that `doublezero permission audit` does not answer this (it reports the strict-mode direction only). The entry also belongs in the `### Breaking` section this Unreleased block already has, following the `feed_pk → feed_pks` precedent.

**The composed cascade tests lost their exact-instruction assertions**, opposite to what the Testing Verification section says: `predicate::always()` went 0 → 6 in `user/delete.rs` and 3 → 7 in `tenant/delete.rs`. That is the one path in this migration that is more than mechanical, and it is now unasserted.

Also worth addressing: `send_transaction_quiet` has no consumer left (the activator was deleted from the workspace), the `clear_topology` program test skips the writable path the builder's divergence exists for and its comment states the opposite, and three device commands traded an eyre error for a release-build panic on over-long input.

The mechanical migration itself holds up — tracing the risky layouts found no swapped pubkeys, dropped accounts, or writability drift beyond the documented `clear_topology` divergence.

- `rfcs/rfc26-rust-instruction-builder-library.md:176` — low: "The permission cache (`note_transaction_sent`/invalidation) stays" describes a subsystem this PR deletes, and contradicts the Open Questions bullet updated at line 245. One-sentence fix.

CHANGELOG.md
juan-malbeclabs: Valid, and the verification you asked for is done. Moved to `### Breaking` in a691c802.

I queried all three clusters directly (`getProgramAccounts` filtered on `AccountType::Permission = 15`) and cross-referenced every `user_payer` against live GlobalState, replicating `check_legacy_any`'s mapping flag by flag:

| Cluster | Permission accounts | Not covered by legacy |
|---|---|---|
| mainnet-beta | 8 — all `FOUNDATION`, every payer in `foundation_allowlist` | **0** |
| testnet | 16 | **1** |
| devnet | 1 | **1** |

The exposed key is the same on both: `EWFXDTBWhNxsmj4mbahZL7HH1XpmC6Ao3LZUhJGCyyPh`, `ACCESS_PASS_ADMIN`, Activated — absent from `foundation_allowlist`, not the sentinel, and `feed_authority_pk` is unset (default) on both clusters. It would lose `SetAccessPass`, `CloseAccessPass`, `SetAccessPassFeeds`, `SetAccessPassFlags`, and both multicast subscriber-allowlist verbs.

On your point 2: `RequirePermissionAccounts` is currently **off** everywhere — `feature_flags = 1` on all three clusters, which is the reserved deprecated bit 0, not strict mode (bit 1, mask 2). So the legacy fallback is live and there is no total lockout.

One near-miss worth recording so nobody re-flags it: `BUtAWK4GaUV42YRp7jSHZhchspsshabn67HnBHnKxzsY` (testnet, `FEED_AUTHORITY`) looks exposed on a per-flag reading but is not — `feed/{create,update,delete}` gate on `FEED_AUTHORITY | FOUNDATION` and that key is in `foundation_allowlist`, so the OR semantics cover it.

What I cannot settle from this repo is whether `EWFXDTBW…` drives those instructions through the **Rust** SDK, which is the only surface this PR changes. `PERMISSION.md:121` describes exactly this grant shape as the feed oracle's ("holding only `ACCESS_PASS_ADMIN | USER_ADMIN`, rather than sitting in `foundation_allowlist`"), so if the oracle is Go or TS it is untouched here. That needs whoever owns the key. Either confirm it, or add it to `foundation_allowlist` on testnet and devnet before this ships — the CHANGELOG entry now states the gate explicitly.

And confirmed on `permission audit`: `build_report` only emits `gaps` = legacy keys *lacking* a Permission account. The inverse direction this PR opens is not in `AuditReport` at all.

smartcontract/sdk/rs/src/commands/user/delete.rs
juan-malbeclabs: Correct on both the counts and the misleading Testing Verification line. Fixed in d807db7e.

All six expectations in this file now assert `predicate::eq(<builder>(...))` built from the test's own values:

- the two `update_multicast_group_roles` unsubscribes, pinning the resolved access-pass PDA and `client_ip`;
- the three `delete_user` sends, pinning `accesspass_pk`, `device_pk`, `dz_prefix_count`, the optional `tenant`, and the trailing `owner`.

The foundation-key test is the one that earns the most from this: it now pins `payer = foundation_key` while `owner = user_owner` and the access pass resolves through the exact-IP fallback, so a swap between payer and owner in the trailing accounts fails the test rather than passing it.

The dead `let (_user_tunnel_block_ext, _, _)` / `let (_globalstate_pubkey, _)` bindings the loosening left behind are gone too — they were the giveaway that the expectations were still computable.

`predicate::always()` across `sdk/rs/src/commands/` is now 4, down from 17, and all 4 are pre-existing on `main` (the multicast allowlist commands, 1 each). 167 tests pass.

smartcontract/sdk/rs/src/commands/tenant/delete.rs
juan-malbeclabs: Same fix applied here in d807db7e — all seven are now `predicate::eq(<builder>(...))`:

- `delete_user` with `Some(tenant_pubkey)`, so the cascade's optional tenant account is pinned;
- both `set_access_pass` sends, pinning the `[current_tenant, new_tenant]` pair the builder appends — `current_tenant = tenant_pubkey`, `new_tenant = default` is exactly the reset this command exists to perform, and it was unasserted;
- all four `delete_tenant` sends, including the failure-path test.

Your point about `Sequence` not saving an all-`always()` file is the sharp one: ordering alone can't distinguish a `SetAccessPass` from a `DeleteTenant` when neither predicate inspects the instruction. Both are now distinguishable by predicate, not just by position.

I replaced the dead `let (_globalstate_pubkey, _)` / `let (_vrf_ids_pda, _, _)` pairs in all five tests with the `program_id` / `payer` bindings the assertions actually need.

smartcontract/sdk/rs/src/doublezeroclient.rs
juan-malbeclabs: Agreed — dropped in d807db7e. The trait is now `send_transaction(Instruction)` only, and with it went the `quiet` parameter on `send_transaction_inner`, its whole branch, and both `SimulationError` / `SimulationTransactionError`.

Confirmed the deadness before removing: the quiet path arrived for the activator in f26c1d09 (#3197), the crate was deleted in aa373d31 (#3647), and `execute_transaction_quiet` had zero callers on `main` too — so this was pre-existing dead code the migration renamed rather than newly orphaned. `clippy --all-targets` across the workspace is clean afterwards, which is the check that would have caught an in-repo consumer.

The two error types are `pub`, so removing them is technically an API break. Nothing in or out of the workspace references them, and keeping them would have left orphans behind a removed method, so I took the removal and noted it in the CHANGELOG's SDK bullet.

Also settled the RFC question rather than leaving it open — `rfc26:244` now records that the quiet variant is not carried forward and why, instead of still proposing it.

smartcontract/programs/doublezero-serviceability/tests/rfc26_builders_test.rs
juan-malbeclabs: Both halves correct. Fixed in d807db7e.

The comment now says the builder passes the topology **writable** and names it as the deliberate parity break with the SDK command (#4078), so the flag reads as intentional rather than as something to "fix" back.

For coverage I kept the `cleared_count == 0` case (the tolerant path is worth pinning too) and added a second case targeting `unicast-default`, the topology `CreateLink` auto-tags the link into. That one runs `clear.rs:121`: it asserts `link_topologies` ends up empty and that the topology's `reference_count` is decremented by exactly 1.

I checked the new case actually guards the divergence rather than merely passing — flipping the builder's `AccountMeta::new(topology, false)` to `new_readonly` makes it fail with:

```
Topology Account is not writable
clear_topology builder should be accepted on the writable path:
  TransactionError(InstructionError(2, ProgramFailedToComplete))
```

Reverted, and the test passes with the builder as shipped.

smartcontract/sdk/rs/src/commands/device/interface/update.rs
juan-malbeclabs: Valid, and I agree the command is the right place — restored in all three in d807db7e, leaving the builders infallible per the RFC-26 contract:

- `device/interface/update.rs` — guards `Some(names).len()`, so `None` (leave topologies alone) still short-circuits;
- `device/interface/create.rs` — guards `topology_names.len()`;
- `device/create.rs` — guards `1 + dz_prefixes.len()` for `resource_count`, matching the bound the builder writes.

Same error strings as before the migration, so the messages are unchanged for anyone parsing them.

Added coverage for both interface guards (`…_rejects_too_many_topologies`) asserting the exact message, with no `send_transaction` expectation registered so the test also proves the command bails before it reaches the builder. Neither guard had a test on `main`, which is how they went missing quietly in the first place.

Worth noting both are unreachable in practice — 256 topologies or 255 dz_prefixes blows the per-transaction account limit long before the `u8`. `Err` → panic is still the wrong direction for CLI-supplied input, so the guards are back regardless.

smartcontract/sdk/rs/src/commands/topology/clear.rs
juan-malbeclabs: Both errors confirmed, and the same stale doc was duplicated on the sibling re-export in `assign_node_segments.rs:9`. Fixed in a691c802.

Rather than correct the arithmetic in two places, I dropped the restatement: these are now plain re-exports pointing at the builder crate's constants, which own the real headroom math (the 64-account lock limit and the ~1232-byte packet size, per `topology.rs:27`). That leaves one home for the numbers instead of three, so the next change to a batch size can't leave a stale copy behind — which is how "32 accounts" and "same constant as backfill" survived this long.

juan-malbeclabs:
Thanks — all seven findings held up under checking, including the two blocking ones. Addressed in d807db7e and a691c802, with details inline on each thread.

## The Permission-PDA gate — verified against live chain state

You were right that `permission audit` can't answer this, so I queried the clusters directly (`getProgramAccounts` on `AccountType::Permission = 15`) and cross-referenced every `user_payer` against live GlobalState, replicating `check_legacy_any`'s mapping flag by flag:

| Cluster | Permission accounts | Not covered by legacy |
|---|---|---|
| mainnet-beta | 8 — all `FOUNDATION`, every payer in `foundation_allowlist` | **0** |
| testnet | 16 | **1** |
| devnet | 1 | **1** |

Same key on both: `EWFXDTBWhNxsmj4mbahZL7HH1XpmC6Ao3LZUhJGCyyPh`, `ACCESS_PASS_ADMIN`, Activated — not in `foundation_allowlist`, not the sentinel, and `feed_authority_pk` unset on both. It would lose `SetAccessPass`, `CloseAccessPass`, `SetAccessPassFeeds`, `SetAccessPassFlags`, and both multicast subscriber-allowlist verbs.

On your second point: `RequirePermissionAccounts` is off everywhere — `feature_flags = 1` is the reserved deprecated bit 0, not strict mode (bit 1). The legacy fallback is live, so there's no total lockout.

**What still needs an owner, not code:** whether `EWFXDTBW…` drives those instructions through the *Rust* SDK, the only surface this PR changes. `PERMISSION.md:121` describes precisely this grant shape as the feed oracle's, so a Go/TS oracle would be untouched. Either confirm that, or add the key to `foundation_allowlist` on testnet and devnet before this ships. The entry now lives in `### Breaking` and states the gate explicitly.

## Everything else

- **Cascade assertions** — `predicate::always()` across `sdk/rs/src/commands/` is 4, down from 17; the remaining 4 are pre-existing on `main`. Both cascade files now assert `predicate::eq(<builder>(...))` at every send, and the dead `let (_pda, _)` bindings are gone. The Testing Verification section is corrected — that claim was backwards.
- **`clear_topology`** — comment fixed (writable, deliberate, #4078) and a `cleared_count > 0` case added against the tagged topology. Verified it's a real guard: flipping the builder to read-only fails with `Topology Account is not writable`.
- **`send_transaction_quiet`** — dropped, with `SimulationError` / `SimulationTransactionError` and the `quiet` branch. Dead since #3647 deleted `activator/`, so pre-existing rather than newly orphaned. `rfc26:244` now records the decision instead of still proposing the variant.
- **`u8` guards** — restored in all three commands with the original messages, builders left infallible, plus tests for both interface guards (neither had one on `main`).
- **Batch-size cap comments** — the stale doc was on both topology re-exports; replaced with pointers to the builder crate so the numbers live in one place.
- **`rfc26:176`** — now says the permission cache is removed, matching the Open Questions bullet.

Green: `cargo test -p doublezero_sdk` 167 passed, full `doublezero-serviceability` suite passed (incl. the 5 RFC-26 builder program tests), `make rust-lint` and `make rust-fmt` clean.

Two judgement calls, flagged rather than buried:

1. **Guard placement** — in the commands, not the builders, since RFC-26 fixes builders as infallible and making them fallible would ripple through every call site. The `.expect(...)` stays as an unreachable backstop.
2. **Removing the two `pub` error types** — technically an API break. Nothing references them in or out of the workspace, and keeping them would leave orphans behind a removed method. Noted in the CHANGELOG.

ben-dz (approved):
lgtm — all seven earlier findings are addressed, and the one open question is answered.

On whether `EWFXDTBWhNxsmj4mbahZL7HH1XpmC6Ao3LZUhJGCyyPh` loses authorization through the Rust SDK: it does not, for the documented candidate. The shred oracle — which `PERMISSION.md:119-122` describes as the holder of that `ACCESS_PASS_ADMIN`-only grant shape — never touches the SDK send path. `crates/shred-oracle` depends on `doublezero_sdk` only for data types and PDA helpers (`Device`, `User`, `Exchange`, `get_accesspass_pda`, `get_permission_pda`, …); there is no `DoubleZeroClient`, no `commands/*` usage, and no `execute_*`/`send_transaction` call anywhere in that repo. It assembles its serviceability instructions itself in `crates/shred-oracle/src/dz_ledger.rs` and appends its own Permission `AccountMeta` per instruction via `permission_account_meta()` (`:264-286`, applied at nine sites covering `SetAccessPass`, `CloseAccessPass`, `SetAccessPassFeeds`, `AddMulticastGroupSubAllowlist`, `RemoveMulticastGroupSubAllowlist`, `CreateSubscribeUser`, `UpdateMulticastGroupRoles`, and `DeleteUser` — the exact set the deploy note names). Its `doublezero_sdk` dependency is pinned by git rev, so the type surface does not move on merge either.

So the oracle keeps authorizing via its own append, independent of the SDK. Residual exposure would need an unidentified Rust-SDK consumer holding that key on testnet or devnet; the in-repo consumers are all operator-driven tools and mainnet-beta has zero exposure. No deploy blocker.

One note for whenever the deferred builder-side Permission append resumes: the oracle solved the offline-existence problem locally, so the builder will face the same constraint documented at `crates/doublezero-serviceability-instruction/src/common.rs:80-86`.

juan-malbeclabs:
Thanks for tracing the oracle — that closes the one thing I couldn't settle from this repo.

I verified your findings before encoding them, since the deploy note now asserts them:

- No `DoubleZeroClient` anywhere in `doublezero-shreds` (zero occurrences); its `doublezero_sdk` imports are data types and status enums only — `Device`, `User`, `Exchange`, `AccountData`, `AccountType`, `DeviceStatus`, `ExchangeStatus`, `UserStatus`, `UserType`, `GlobalState`. No `commands/*`, no send path.
- `permission_account_meta()` is at `crates/shred-oracle/src/dz_ledger.rs:264-287` with 9 call sites, doing exactly the RPC existence check + per-process memoization the SDK just deleted. (My local checkout of that repo was 38 commits behind and predates the helper, so I read it off remote `main` rather than trusting the stale copy.)

Two follow-ups pushed in 769573c4:

1. **The CHANGELOG note now records the answer instead of asking for it.** It previously told the reader to confirm the key doesn't drive those instructions through the Rust SDK before shipping — stale, and worse than useless once resolved, since it implies a blocker that no longer exists. It now states that no cluster is exposed and why. Same correction applied to the PR description.
2. **Your forward-looking note is captured where the next person will hit it** — `build_with_permission`'s activation-precondition doc now points at the oracle's caller-side solution and spells out the consequence: activation here needs either the `RequirePermissionAccounts` guarantee or a caller-supplied "permission exists" input threaded down from the command layer, because a pure builder can't own that check. That's the constraint your last paragraph identifies, and it's better placed next to the deferred append than only in review history.

Still green after both: `cargo test -p doublezero_sdk` 167 passed, `make rust-lint` clean.

elitegreg (changes requested):
Reviewed the full stack tip. The migration itself holds up well: I hand-compared the builder call against the old hand-assembled account list for every non-mechanical command (`link/update` including the side-A/side-Z preamble split and the conditional `side_a`/`side_z` + resource-extension appends, `device/delete` on both the legacy and atomic paths, `user/create`, `user/delete`, `tenant/delete`, `exchange/setdevice` two-send sequencing, `topology/clear`, `topology/assign_node_segments`) and found no drift. The `_quiet` removal checks out: nothing in the workspace calls the removed trait methods, and `geolocation/`'s `execute_transaction` is a separate trait, untouched.

Verified locally:

- `cargo test -p doublezero_sdk --lib` — 167 passed
- `cargo test -p doublezero-serviceability --test rfc26_builders_test` — 5 passed
- `cargo check --workspace --all-targets` — clean, no new warnings (the `doublezero_geolocation::entrypoint` feature-unification errors reproduce identically on `main`)

Four things to address before merge.

## 1. This PR fixes #4078 but does not say so

The builder passes the `clear_topology` topology PDA writable, which is exactly the fix #4078 asks for. That issue is open and describes a user-facing break: the processor asserts `topology_account.is_writable` whenever `cleared_count > 0`, so on `main` `topology clear` reverts with "Topology Account is not writable" on a live topology, which is the mandatory precursor to `topology delete`. After this PR it works.

Right now the only mention is a comment inside `rfc26_builders_test.rs`. Please add a `### Fixed` CHANGELOG entry under SDK and a `Fixes #4078` line in the PR body so the issue closes on merge. As written, a real bug fix ships invisibly and #4078 stays open pointing at a file that no longer has the bug.

## 2. Four allowlist tests lost their instruction assertion

`multicastgroup/allowlist/publisher/add.rs`, `publisher/remove.rs`, `subscriber/add.rs`, `subscriber/remove.rs` now read:

```rust
client
    .expect_send_transaction()
    .with(predicate::always())
    .returning(|_| Ok(Signature::new_unique()));
```

On `main` the same tests pinned the instruction and relaxed only the accounts:

```rust
.with(
    predicate::eq(DoubleZeroInstruction::AddMulticastGroupPubAllowlist(
        AddMulticastGroupPubAllowlistArgs { client_ip: ..., user_payer: pubkey },
    )),
    predicate::always(),
)
```

So `client_ip` and `user_payer` were asserted before and are not now. The PR body's "`predicate::always()` across `sdk/rs/src/commands/` is 4, all pre-existing on `main`" is accurate as a file count but understates this: the assertion strength dropped in all four, and these are the only migrated tests that assert nothing about what gets sent. One-line fix each, using the `predicate::eq(add_multicast_group_pub_allowlist(...))` pattern the other 73 migrations already use.

## 3. A Permission grant becomes a silent no-op with nothing telling the operator

The exposure analysis in the CHANGELOG is thorough and I am not disputing the conclusion. The gap is forward-looking: once this lands, `doublezero permission set` produces a grant that cannot work through the CLI or Rust SDK, and nothing at runtime says so. The CHANGELOG itself notes `permission audit` answers only the inverse (strict-mode) direction, so the natural place an operator would look does not surface it either.

Suggest a `tracing::warn!` on `permission set` / `permission update`, or an added line in `permission audit`, stating that the Rust SDK does not attach Permission accounts yet and gated commands still require a legacy allowlist or authority entry. It can be removed when the builder-side append is re-enabled. The chain-state check is point-in-time by nature; this makes the next grant self-documenting instead of a debugging session.

## 4. Please rebase onto `origin/main`

RF (#4059) merged as `9c2ceb76`, but this branch still carries its three unsquashed commits, so the PR renders as 115 files / +4548 −4751 when only ~96 files are actually new. I diffed the branch's RF content against merged `main`: byte-identical apart from the two intended builder-crate additions (`set_access_pass_flags`, the `build_with_permission` doc note), so the rebase is content-neutral and makes the diff match what needs reviewing.

## Non-blocking notes

- Several commands pass args fields the builder immediately overwrites: `use_onchain_allocation: false` in `link/update`, `resource_count: 0` in `device/create`, `dz_prefix_count` / `multicast_publisher_count` in `user/delete`, `user/create`, `user/requestban`. Correct, and documented on the builder side, but at the call site it reads as if the caller controls them. A short comment at each site would help the next reader.
- Roughly 60 commands drop the `GetGlobalStateCommand` round-trip, which is a real latency win, but they also lose the fast-fail "Globalstate not initialized" message in favor of an onchain error on an uninitialized cluster. Probably the right trade; flagging it as deliberate rather than incidental.
- `device/delete` has no `u8` guard on `owners.len()` where `device/create` and the two interface verbs got one, so it relies on the builder's `.expect(...)`. Unreachable given the transaction account budget, and the builder comment says as much, so this is consistency rather than a bug.
- Worth stating in the RFC or the test module doc: the migrated command tests are structurally tautological with respect to account layout. `predicate::eq(builder(...))` catches wrong arguments but cannot catch a builder layout regression, since both sides call the same builder. That is by design, with the golden fixtures and the program test as the real net, but it means around 70 builders currently have no program-level coverage. A note keeps a future reader from over-trusting the 167 passing tests.

juan-malbeclabs:
Thanks — all four held up under checking. Addressed in `cb9fa51c`, on a branch now rebased onto `main`.

**1. `#4078`.** Added `Fixes #4078` to the PR body and a CHANGELOG entry covering the revert-on-live-topology break and the writable-PDA fix.

One deviation worth flagging: you asked for a `### Fixed` section, but this CHANGELOG has never had one — every past fix is a bullet under `### Changes`, including the closely related `ClearTopology` account-layout fix (line 691). I followed the file's convention rather than introducing a new section. Happy to switch if you'd rather establish `### Fixed` here.

**2. Allowlist assertions.** All four now pin the built instruction with `predicate::eq(add_multicast_group_pub_allowlist(...))` and friends.

Two things beyond the one-line fix. `user_payer` is now a distinct pubkey from the group pubkey — previously both were the same value, so the builder's `AccountMeta::new(user_payer, false)` and the mgroup meta were indistinguishable and a swap between them would have passed. And I confirmed the assertion actually bites: perturbing `client_ip` in the expected instruction fails with `No matching expectation found`, then reverted.

You were also right that the PR body understated this, and the real numbers cut harder than the file count suggested. `predicate::always()` under `sdk/rs/src/commands/` is 8 files on `main` and 0 here. The four allowlist files kept the count while inverting the meaning — on `main` the `always()` was the accounts argument of the two-arg `execute_*` expectation, paired with an `eq()` on the instruction; here it had become the only predicate. Body rewritten to say that.

**3. Permission grant no-op.** Did both hooks you offered rather than picking one, since the audit is where an operator looks but `set` is where the mistake is made:

- `permission set` emits a `tracing::warn!` when the resulting bitmask carries an `AUTHORIZE_GATED_FLAGS` bit. Extracted as `inert_grant_warning(u128) -> Option<&'static str>` so it is unit-testable; the test asserts it fires for every gated flag and stays quiet for `SENTINEL` / `QA` / `FEED_AUTHORITY`. Those three are absent from `AUTHORIZE_GATED_FLAGS`, so no processor ever hands them to `authorize()` and a Permission account never authorized them in the first place — warning there would be noise, not signal.
- `permission audit` renders a matching paragraph and exposes `sdk_attaches_permission_accounts: false` in `--json`, both keyed off a single `const` to flip when the append returns.

Small correction: there is no `permission update` verb. The subcommands are `set`, `delete`, `flags`, `get`, `list`, `audit`, `resume`, `suspend`, and `set` is the only grant-creating path, so that is where the warning lives.

**4. Rebase.** Done. A plain `git rebase origin/main` conflicts immediately, since the branch's three RF commits collide with the same content squash-merged as `9c2ceb76`. Confirmed they were safe to drop outright before doing so — `git diff ebe5e544 9c2ceb76` is empty, i.e. the branch's RF tip tree is byte-identical to merged `main`. Rebased with `git rebase --onto origin/main ebe5e544`, replaying only the four R10 commits with zero conflicts, and `git diff` between the old and new tips is empty, so the rebase is provably content-neutral.

The PR renders 98 files rather than your predicted ~96 — the extra two are the `permission/{set,audit}.rs` changes from point 3. I also refreshed the Diff Breakdown table in the body, which still described the pre-rebase 115-file diff.

**Non-blocking notes.** Taking the first two as follow-ups rather than growing this PR further: the call-site comments on builder-overwritten args (`use_onchain_allocation`, `resource_count: 0`, `dz_prefix_count`) are worth doing but touch a dozen files that are otherwise untouched by this round. On the `GetGlobalStateCommand` round-trip: deliberate, not incidental — the builders derive the globalstate PDA offline, so the fetch existed only to produce the fast-fail message, and paying an RPC round-trip on ~60 commands to pre-empt an error that only occurs on an uninitialized cluster is the wrong trade. On `device/delete`'s missing `u8` guard: agreed it is consistency rather than a bug, and left as-is for the reason the builder comment gives — `owners.len()` is bounded well below `u8::MAX` by the transaction account budget, whereas the interface verbs' counts are caller-supplied.

Verified: `cargo test -p doublezero_sdk --lib` 167 passed, `cargo test -p doublezero-serviceability-cli --lib` 367 passed, `cargo clippy --all-targets -Dwarnings` clean on both crates.

elitegreg (approved):

PR #4063, opened 2026-07-15
cli: expose per-feed seat details in access-pass get --json

martinsander00:
## Summary

- `access-pass get --json` collapses an EdgeSeat pass to `type: "edge_seat: N feed(s)"`, so a pass's per-feed seat details (feed_key, max/future/current users, window_end, terminates_at) aren't exposed anywhere in machine-readable output.
- Adds a `feed_seats` array to the JSON view, populated from `AccessPassType::EdgeSeat`. Empty for other pass types. The table view is unchanged: the field is `#[tabled(skip)]`, matching the existing skipped-field pattern in `access-pass list`, and `type` still shows the feed count.

## Testing

- Added a unit test covering the EdgeSeat -> `feed_seats` mapping (per-seat fields) and the empty non-EdgeSeat case.
- `cargo test -p doublezero-serviceability-cli --lib -- accesspass::get::tests` passes.

### DIFF
--- a/smartcontract/cli/src/accesspass/get.rs
+++ b/smartcontract/cli/src/accesspass/get.rs
@@ -5,7 +5,7 @@ use doublezero_sdk::commands::{
     accesspass::get::GetAccessPassCommand, multicastgroup::list::ListMulticastGroupCommand,
     tenant::list::ListTenantCommand,
 };
-use doublezero_serviceability::state::accesspass::AccessPassType;
+use doublezero_serviceability::state::accesspass::{AccessPassType, FeedSeat};
 use serde::Serialize;
 use solana_sdk::pubkey::Pubkey;
 use std::{io::Write, net::Ipv4Addr};
@@ -24,20 +24,6 @@ pub struct GetAccessPassCliCommand {
     pub json: bool,
 }
 
-/// Per-feed seat detail for the JSON view of an EdgeSeat pass. The table view and the
-/// `type` summary show only a feed count, so these per-feed user counts and billing
-/// windows are otherwise invisible.
-#[derive(Serialize)]
-struct FeedSeatDisplay {
-    pub feed_key: String,
-    pub max_users: u8,
-    pub max_future_users: u8,
-    pub current_users: u8,
-    pub anniversary_day: u8,
-    pub window_end: i64,
-    pub terminates_at: i64,
-}
-
 #[derive(Tabled, Serialize)]
 struct AccessPassDisplay {
     pub account: String,
@@ -45,7 +31,7 @@ struct AccessPassDisplay {
     #[serde(rename = "type")]
     pub accesspass_type: String,
     #[tabled(skip)]
-    pub feed_seats: Vec<FeedSeatDisplay>,
+    pub feed_seats: Vec<FeedSeat>,
     pub client_ip: String,
     pub user_payer: String,
     pub tenant: String,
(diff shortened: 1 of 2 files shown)


smartcontract/cli/src/accesspass/get.rs
```diff
@@ -23,12 +24,28 @@ pub struct GetAccessPassCliCommand {
...
+    pub max_users: u8,
+    pub max_future_users: u8,
+    pub current_users: u8,
+    pub anniversary_day: u8,
+    pub window_end: i64,
+    pub terminates_at: i64,
+}
+
 #[derive(Tabled, Serialize)]
 struct AccessPassDisplay {
```
juan-malbeclabs, line 28: `FeedSeatDisplay` may be redundant. `Cargo.toml` enables `features = ["serde"]` on `doublezero-serviceability`, and `FeedSeat` already derives `serde::Serialize` with `feed_key` serialized as a string. The JSON emitted here is byte-for-byte identical to serializing `Vec<FeedSeat>` directly, so you could drop this struct and `feed_seat_displays` and embed the on-chain type:

```rust
#[tabled(skip)]
pub feed_seats: Vec<FeedSeat>,
```

Keeping a dedicated display struct is defensible — it decouples the CLI's JSON contract from the on-chain struct's serde layout, so a future change to `FeedSeat` won't silently alter CLI output. If that's the intent, a one-line comment saying so would help; otherwise reusing `FeedSeat` is less code. Not blocking either way.

smartcontract/cli/src/accesspass/get.rs
```diff
@@ -309,4 +347,27 @@ mod tests {
...
+        }];
+        let displays = super::feed_seat_displays(&AccessPassType::EdgeSeat(seats));
+        assert_eq!(displays.len(), 1);
+        assert_eq!(displays[0].feed_key, feed_key.to_string());
+        assert_eq!(displays[0].max_users, 2);
+        assert_eq!(displays[0].current_users, 1);
+        assert_eq!(displays[0].terminates_at, 200);
+
+        // Non-EdgeSeat passes carry no per-feed seats.
+        assert!(super::feed_seat_displays(&AccessPassType::Prepaid).is_empty());
```
juan-malbeclabs: Minor test-coverage gap: this exercises the `feed_seat_displays` helper directly (good), but there's no end-to-end assertion that `get --json` actually renders `feed_seats` for an EdgeSeat pass — the existing `test_cli_accesspass_get` uses `Prepaid` + `json: false`. A `json: true` + EdgeSeat variant would lock in the serialized shape that consumers depend on. Optional.

juan-malbeclabs (commented):
Small, well-scoped, and correct. I verified the mapping covers all seven `FeedSeat` fields, and `#[tabled(skip)]` correctly keeps the table view unchanged (matching the `access-pass list` pattern). Two non-blocking notes inline.

juan-malbeclabs (approved):
Approving, conditional on the two notes above. The change is correct and low-risk. Before merging, please either reuse `Vec<FeedSeat>` directly or add a one-line comment explaining the intentional decoupling, and consider a `get --json` + EdgeSeat test to lock in the serialized shape. Neither blocks merge.

smartcontract/cli/src/accesspass/get.rs
martinsander00: went with reusing the on-chain `FeedSeat` directly (dropped `FeedSeatDisplay` + the helper). done in 932252de2.

smartcontract/cli/src/accesspass/get.rs
martinsander00: added a `get --json` + EdgeSeat test that asserts the `feed_seats` array shape (per-seat users and windows), so the serialized contract is locked in. 932252de2.

PR #4065, opened 2026-07-17
e2e/qa: retry seat withdraw while in-flight rejection persists instead of polling pending flag

thijsvanemmerik:
## Summary

The multicast settlement QA test failed persistently on seat teardown. Two problems in the old flow:

- `wait_for_seat_allocation_acked` polled the client seat's pending flag to detect the oracle ack. That flag cannot distinguish "the oracle has not acked yet" from "no allocation request was created" (a re-fund of an already-active seat creates none), so the wait could time out with `saw_pending=false` when there was simply nothing to ack.
- `withdraw_seat` did a single-shot withdraw. The CLI preflight rejects with an in-flight error while it reads the allocation request PDA as still present, and that read can lag an already-closed request, so a single attempt fails transiently.

Changes:

- Remove the pending-flag ack-wait and make the withdraw retry while the in-flight rejection persists (`WithdrawSeatWaitingForAck`). The retry absorbs both a slow oracle ack and a lagging RPC read; an in-flight withdraw bails atomically with no state change, so retrying is safe. Any non-in-flight error fails fast.
- Log the seat's tenure and pending flag right after pay (diagnostic only), so a "no request created" run is visible in the run log.

## Tradeoff

With the pending-flag ack-wait removed, a genuine oracle outage now surfaces as `validate_tunnel_up` timing out rather than a specific "not acked" error. The withdraw's timeout error calls out a persistent in-flight rejection (a stale RPC view) versus a slow ack.

## Scope

Self-heal of a seat left active by a prior failed teardown is handled separately and is intentionally not part of this PR.

## Testing Verification

Verify with a manual `workflow_dispatch` of `qa.mainnet-beta` with `doublezero_branch` set to this PR branch, confirming `TestQA_MulticastSettlement` passes end to end including `withdraw_seat` (baseline is green after the stuck seat was cleared).


thijsvanemmerik:
Closing unmerged. This was a QA-environment issue handled separately; no change to land here.

ben-dz:
Superseded by #4066

PR #4066, opened 2026-07-17
e2e/qa: settlement test self-heals stuck seats and retries withdraws with endpoint rotation

ben-dz:
Resolves: malbeclabs/infra#2002
Supersedes #4065 — carries over its ack-wait removal and retry-instead-of-flag-polling approach (credit @thijsvanemmerik for the diagnosis and design there; co-authored on the absorbing commit).

## Summary of Changes
* `TestQA_MulticastSettlement` now self-heals from a seat left **active onchain** by a previous run whose withdraw failed — the poisoned state that turned `qa.mainnet-beta` red for 11+ consecutive hourly runs.
* `ensure_multicast_disconnected` scans onchain client seats for the client's public IP (shreds SDK `FetchAllClientSeats`) and withdraws any with `TenureEpochs > 0`, in addition to the existing session check. A stuck-active seat with no session (the actual failure state) is now detected instead of passing untouched.
* Every withdraw — the `withdraw_seat` step, the self-heal, and the cleanup — retries over a bounded window (`WithdrawSeatWithRetry`) instead of a single shot; the same spurious "request in flight" preflight bail failed on every run, so the escrow grew ~$30/run and never healed.
* On the in-flight bail the retry **rotates to a different Solana RPC endpoint** (`solanaRPCPool.Failover()`) before retrying: a fixed endpoint serves the stale `getMultipleAccounts` read behind the bail for seconds to hours, so a same-endpoint retry cannot converge. Rotation fires only for that pre-submission bail, never for submission timeouts (which may have landed onchain). Completion is confirmed against fresh `getAccountInfo` onchain state, not the CLI's error text.
* The `wait_for_seat_allocation_acked` step and its `WaitForSeatAllocationAcked` helper are removed (from #4065): polling the seat's pending flag cannot distinguish a fast ack from a re-fund of an already-active seat that never creates a request, and the retrying withdraw checks the request state onchain directly. Tradeoff: a genuine oracle outage now surfaces as `validate_tunnel_up` timing out rather than a specific "not acked" error.

Context: `shreds pay` on an already-active seat only tops up the escrow and never creates a new `InstantSeatAllocationRequest`, so the pending flag is never set, the oracle never acks, and the multicast tunnel can't come up — every subsequent run was poisoned. Reference failing runs: [29554004596](https://github.com/malbeclabs/infra/actions/runs/29554004596) (first, in-flight bail), [29580032054](https://github.com/malbeclabs/infra/actions/runs/29580032054) (steady state, `saw_pending=false`), [29588245744](https://github.com/malbeclabs/infra/actions/runs/29588245744) (re-stuck immediately after a manual unstick — no pool failover was logged, confirming the CLI preflight and the harness read shared one stale primary endpoint).

Scope note: no changes to the pay/withdraw CLI or the QA workflow.

## Testing Verification
* Field-validated on mainnet-beta before the `withdraw_seat` absorption: dispatch [29590898632](https://github.com/malbeclabs/infra/actions/runs/29590898632) hit the in-flight bail on the primary endpoint (logged, redacted), rotated, and the cleanup withdraw **succeeded on attempt 2** via the fallback — leaving the seat clean onchain instead of poisoning the next run.
* New table-driven unit tests: `TestFilterActiveSeats` (active-seat-by-IP selection: our-IP active kept, tenure-0 skipped, other-IP skipped, multi-device all-kept), `TestIsSeatNotFound` (sentinel match — asserts a transient "Blockhash not found" is **not** treated as a missing seat), and `TestIsInFlightPreflightBail` (matches the in-flight bail forms; asserts a submission timeout does **not** trigger endpoint rotation).
* Withdraw completion is confirmed against authoritative onchain state (`TenureEpochs == 0` / account gone via `getAccountInfo`, which stays fresh even when the endpoint's `getMultipleAccounts` is stale) rather than pattern-matching the external CLI's error text, so a transient RPC failure can't be misread as "already withdrawn" and silently leave a seat stuck.
* End-to-end validation: dispatch [29592673971](https://github.com/malbeclabs/infra/actions/runs/29592673971) (`qa.mainnet-beta` #6566) — **fully green**, the first green run since the incident began. The self-heal detected and withdrew the seat left stuck by the immediately preceding scheduled failure (`Self-healed stuck-active seat(s) count=1`: attempt 1 bailed in-flight on the stale primary, rotated, succeeded on attempt 3). The absorbed `withdraw_seat` step then converged after 5 attempts (~63s) across both endpoints on a different device's fresh seat — alternating genuine in-flight windows and stale reads that a single-shot withdraw (or fixed-endpoint retry) would have failed.

## Review follow-ups (non-blocking)
The blocking finding (over-broad error-text matching) is fixed via onchain verification. Deferred Low/nit items: `FetchAllClientSeats` is a discriminator-only `getProgramAccounts` scan (fine at QA scale; add a `ClientIPBits` memcmp filter in the SDK if it grows); seats are selected by IP only per the approved plan (filtering additionally by funding authority would harden against a same-IP seat funded by a different wallet); `SetPublicIP` validates non-nil but not IPv4-only (pre-existing).


ben-dz:
Pushed a follow-up (ab7fd67d) for the post-build finding on run [29588245744](https://github.com/malbeclabs/infra/actions/runs/29588245744): the seat re-stuck immediately after a manual unstick, with no pool failover logged, confirming the CLI withdraw preflight and the harness read shared one primary endpoint that serves stale `getMultipleAccounts` (closed request PDA reported as existing 27s–8h after close) while `getAccountInfo` stays fresh.

`WithdrawSeatWithRetry` now calls `solanaRPCPool.Failover()` after an **in-flight preflight bail** so the next `FeedSeatWithdraw` preflights against a different `CurrentURL()`, and logs the redacted endpoint per attempt. This covers both the self-heal and cleanup paths (both route through `WithdrawSeatWithRetry`). Rotation is gated on the in-flight bail only — submission timeouts don't rotate, since a timed-out submission may have landed onchain. New unit test `TestIsInFlightPreflightBail` pins that distinction.

e2e/internal/qa/client_settlement.go
```diff
@@ -256,6 +271,214 @@ func (c *Client) FeedSeatWithdraw(ctx context.Context, devicePubkey string) erro
...
+// found" must never read as "already withdrawn").
+func (c *Client) seatIsWithdrawn(ctx context.Context, shredsClient *shreds.Client, deviceKey solana.PublicKey, clientIPBits uint32) (bool, error) {
+	seat, err := shredsClient.FetchClientSeat(ctx, deviceKey, clientIPBits)
+	if isSeatNotFound(err) {
+		return true, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	return seat.TenureEpochs == 0, nil
```
nikw9944: **H1.** This treats `TenureEpochs == 0` as proof the withdraw took effect, but the SDK doc (`sdk/shreds/go/state.go:153-155`) says withdrawal is blocked onchain while `HasPendingInstantRequest` is set — a pending seat is definitionally not withdrawn. With `wait_for_seat_allocation_acked` removed, `withdraw_seat` now routinely races the oracle, so the first attempt can fail with a *genuine* in-flight rejection. If the external program sets `TenureEpochs` only at oracle ack (not at pay), this reads 0 on the pending seat and `WithdrawSeatWithRetry` returns success without withdrawing — the seat activates on ack and recreates the poisoned state this PR exists to fix (recovered only by next run's self-heal). One-line hardening that is strictly safer under either tenure semantics:
```go
return seat.TenureEpochs == 0 && !seat.HasPendingInstantRequest(), nil
```

e2e/internal/qa/client_settlement.go
```diff
@@ -256,6 +271,214 @@ func (c *Client) FeedSeatWithdraw(ctx context.Context, devicePubkey string) erro
...
+}
+
+// seatIsWithdrawn reads authoritative onchain state to decide whether the seat
+// for deviceKey no longer holds an active tenure: the account is gone, or
+// TenureEpochs == 0. Used to confirm a withdraw actually took effect rather than
+// pattern-matching the external CLI's error text (a transient "Blockhash not
+// found" must never read as "already withdrawn").
+func (c *Client) seatIsWithdrawn(ctx context.Context, shredsClient *shreds.Client, deviceKey solana.PublicKey, clientIPBits uint32) (bool, error) {
+	seat, err := shredsClient.FetchClientSeat(ctx, deviceKey, clientIPBits)
+	if isSeatNotFound(err) {
```
nikw9944, line 314: **M1.** Withdraw appears to close the seat account (not-found ⇒ withdrawn), so the seat is only minutes old at withdraw time. The PR's own analysis documents fallback endpoints stale for "seconds to hours"; such an endpoint serves not-found for the young seat, and one read makes `WithdrawSeatWithRetry` return success without withdrawing. The comment that getAccountInfo "stays fresh even on an endpoint whose getMultipleAccounts is stale" is an empirical observation about one provider, not a guarantee — a node behind by slots is stale for both. Fix options: call `c.solanaRPC.SelectHealthiestEndpoint(ctx)` before the confirm read (as `GetUSDCBalance` does), or accept not-found only on two consecutive reads.

e2e/internal/qa/client_settlement.go
```diff
@@ -256,6 +271,214 @@ func (c *Client) FeedSeatWithdraw(ctx context.Context, devicePubkey string) erro
...
+	deadline := time.Now().Add(withdrawRetryTimeout)
+	attempt := 0
+	for {
+		attempt++
+		withdrawErr := c.FeedSeatWithdraw(ctx, devicePubkey)
+		if withdrawErr == nil {
+			c.log.Info("Seat withdraw succeeded", "host", c.Host, "device", devicePubkey, "attempt", attempt)
+			return nil
+		}
+		if withdrawn, checkErr := c.seatIsWithdrawn(ctx, shredsClient, deviceKey, clientIPBits); checkErr == nil && withdrawn {
```
nikw9944: **L1.** `checkErr` is silently discarded; the following warn logs only `withdrawErr`, so a persistently failing confirm read is invisible in the run log. Add `checkErr` (scrubbed) to the warn attributes.

e2e/internal/qa/client_settlement.go
```diff
@@ -256,6 +271,214 @@ func (c *Client) FeedSeatWithdraw(ctx context.Context, devicePubkey string) erro
...
+func (c *Client) SelfHealStuckSeats(ctx context.Context) (int, error) {
+	seats, err := c.ActiveClientSeats(ctx)
+	if err != nil {
+		return 0, err
+	}
+	if len(seats) == 0 {
+		return 0, nil
+	}
+
+	shredsClient, clientIPBits, err := c.shredsQuery()
```
nikw9944: **L2.** `SelfHealStuckSeats` calls `shredsQuery()` a second time after `ActiveClientSeats` already did. Reuse a shared preamble or pass the client/ipbits through. Harmless, pure duplication.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
+- E2E/QA
+  - `TestQA_MulticastSettlement` recovers from the failure modes that kept mainnet-beta QA red: `ensure_multicast_disconnected` self-heals seats left stuck-active onchain by a previous run's failed withdraw (scanning client seats for the client's public IP via the shreds SDK `FetchAllClientSeats` and withdrawing any with `TenureEpochs > 0`), and every withdraw — the `withdraw_seat` step, the self-heal, and the cleanup — retries over a bounded window instead of failing on a single spurious "request in flight" preflight bail. The retry rotates to a different Solana RPC endpoint on the in-flight bail (the stale `getMultipleAccounts` read behind it is per-endpoint) and confirms completion against fresh onchain state rather than the CLI's error text. The `wait_for_seat_allocation_acked` step is removed: polling the seat's pending flag cannot distinguish a fast ack from a re-fund of an active seat that never creates a request, and the retrying withdraw checks the request state onchain directly. (#4066, supersedes #4065)
```
nikw9944: **L3.** "the retrying withdraw checks the request state onchain directly" (also the comment at `e2e/qa_multicast_settlement_test.go:248-251`) overstates the code: it checks the seat's `TenureEpochs`, not the `InstantSeatAllocationRequest`/pending state. Reword — or make it true by adopting H1.

e2e/internal/qa/client_settlement.go
```diff
@@ -256,6 +271,214 @@ func (c *Client) FeedSeatWithdraw(ctx context.Context, devicePubkey string) erro
...
+		c.log.Warn("Seat withdraw attempt failed, will retry",
+			"host", c.Host, "device", devicePubkey, "attempt", attempt,
+			"endpoint", redactURL(c.currentSolanaRPCURL()), "error", c.scrubRPCErr(withdrawErr))
+		if time.Now().After(deadline) {
+			return fmt.Errorf("seat withdraw did not succeed within %s on host %s after %d attempts: %s",
+				withdrawRetryTimeout, c.Host, attempt, c.scrubRPCErr(withdrawErr))
+		}
+		// The stale getMultipleAccounts read behind the in-flight bail lives on
+		// this endpoint; rotate so the next submission's preflight reads fresh.
+		if isInFlightPreflightBail(withdrawErr) && c.solanaRPC != nil && c.solanaRPC.EndpointCount() > 1 {
```
nikw9944, line 410: **L4.** With the ack-wait removed, a *genuine* in-flight rejection (withdraw racing the oracle) also matches `isInFlightPreflightBail`, so the loop rotates the shared pool's endpoint on every attempt even though nothing is stale. Bounded and harmless, but worth a comment acknowledging this, since the pool's current endpoint is shared with subsequent reads.

nikw9944 (changes requested):
The design is sound: onchain-state confirmation instead of CLI error-text matching, sentinel-based not-found detection, bounded retries, and a self-heal pass that converges across runs. Build, go vet, golangci-lint, and the new unit tests all pass. One High finding: `seatIsWithdrawn` treats `TenureEpochs == 0` as proof of withdrawal but ignores `HasPendingInstantRequest` — with the ack-wait step removed, the withdraw retry can declare false success on a paid-but-not-yet-acked seat and recreate the exact poisoned state this PR eliminates. The fix is one line. A Medium on trusting a single not-found read from a possibly-stale endpoint, plus four Lows.

- The two items the PR description already defers (no funding-authority filter in filterActiveSeats; discriminator-only FetchAllClientSeats scan) look acceptable at QA scale; no need to re-litigate.
- Non-test additions are ~250 lines, within the repo's 500-line guidance. Build, go vet, golangci-lint, and the new unit tests all pass locally.

e2e/internal/qa/client_settlement.go
ben-dz: Adopted in 2aebd0b7: `seatIsWithdrawn` now requires `TenureEpochs == 0 && !seat.HasPendingInstantRequest()`.

One data point on the conditional ("if the external program sets TenureEpochs only at oracle ack"): it doesn't. On the deployed program, `try_request_instant_seat_allocation_core` sets `tenure_epochs = 1` (together with `subscription_start_slot`) in the pay transaction, and the onchain history agrees: the QA seat's `subscription_start_slot` equals the pay tx slot, not the ack slot. So a paid-but-unacked seat reads tenure=1 and the old check already returned not-withdrawn. Adopted regardless: the SDK's documented contract (withdrawal blocked while the pending flag is set) is the stable thing to encode, and the hardened check is correct under either semantics.

e2e/internal/qa/client_settlement.go
ben-dz: Adopted the two-consecutive-reads option in 2aebd0b7: a not-found counts as withdrawn only when a second read, taken after `Failover()`, agrees. The shreds client is pool-backed, so the rotation genuinely changes which endpoint answers the second read.

One premise correction: withdraw does not close the seat account. `try_request_instant_seat_withdrawal` zeroes tenure and the account persists (observed on the QA seat all day: every post-withdraw read returns the account with tenure=0). The reachable version of this risk enters through a different door: a first pay against a device creates the seat minutes before the withdraw (run 6566's second device did exactly this), and a stale endpoint can serve not-found for a young account. The double-read covers that case.

On the `SelectHealthiestEndpoint` option: it probes slot height, and today's stale endpoint answered `getSlot` and `getAccountInfo` fresh while serving stale `getMultipleAccounts`, so the probe would have passed it. The rotate-and-reread approach does not depend on the staleness being visible to a probe.

e2e/internal/qa/client_settlement.go
ben-dz: Fixed in 2aebd0b7: `checkErr` is hoisted out of the if and logged (scrubbed) as `confirm_read_error` on the retry warn whenever it is non-nil.

e2e/internal/qa/client_settlement.go
ben-dz: Fixed in 2aebd0b7. `ActiveClientSeats` had exactly one caller, so it is inlined into `SelfHealStuckSeats` (which also removes the duplicate `shredsQuery`); `filterActiveSeats` stays as the unit-tested pure function.

CHANGELOG.md
ben-dz: Reworded in 2aebd0b7, and made true by adopting H1: the changelog now says the retrying withdraw "confirms completion against the seat's onchain tenure and pending-request state". The test comment at 248-251 only claims retry-with-rotation (it makes no request-state claim), so it stands as written.

e2e/internal/qa/client_settlement.go
ben-dz: Comment added in 2aebd0b7: a genuine in-flight rejection (the withdraw racing the oracle's ack) matches the same marker and also rotates, which is harmless because every pool endpoint is valid for the shared reads that follow and rotation is bounded by the retry window. For what it's worth, run 29592673971 exercised exactly that path: five attempts alternating across both endpoints, converging in ~63s.

e2e/internal/qa/client_settlement.go
```diff
@@ -256,6 +271,227 @@ func (c *Client) FeedSeatWithdraw(ctx context.Context, devicePubkey string) erro
...
+			"host", c.Host, "seat", seat.Pubkey, "device", deviceKey, "tenure_epochs", seat.TenureEpochs)
+
+		if err := c.WithdrawSeatWithRetry(ctx, deviceKey.String()); err != nil {
+			return healed, fmt.Errorf("failed to withdraw stuck seat %s on device %s (host %s): %w", seat.Pubkey, deviceKey, c.Host, err)
+		}
+
+		if err := poll.Until(ctx, func() (bool, error) {
+			s, fetchErr := shredsClient.FetchClientSeat(ctx, deviceKey, clientIPBits)
+			// A closed/absent seat account is fully healed.
+			if isSeatNotFound(fetchErr) {
```
nikw9944: **L5.** This poll still trusts a *single* not-found read (`isSeatNotFound(fetchErr) → return true`), and the comment "A closed/absent seat account is fully healed" contradicts the fact now documented at `seatIsWithdrawn` (line 306): withdrawal does not close the seat account, so a not-found here can only be a stale endpoint. Impact is negligible (the withdraw already succeeded; early exit just skips waiting for the tenure=0 read), but for consistency either treat not-found as "keep polling" like the transient-error branch, or update the comment.

nikw9944 (approved):
All six findings from the previous round are addressed and verified in the current diff at head 5b9cc9bc: H1 (pending-flag guard adopted in seatIsWithdrawn), M1 (not-found trusted only after a second read on a rotated endpoint), L1 (confirm_read_error logged), L2 (ActiveClientSeats inlined, duplicate shredsQuery removed), L3 (CHANGELOG reworded), L4 (genuine-in-flight rotation comment added). Build, unit tests, gofmt, and golangci-lint pass on the new head. One new Low: the self-heal confirmation poll still trusts a single not-found read, and its 'closed/absent seat account' comment contradicts the fact the hardening commit itself documents (withdraw does not close the seat account).

- All previous findings (H1, M1, L1-L4) verified as fixed in the current code at head 5b9cc9bc — dropped based on the diff, not on the author's replies. External-program claims in the replies (tenure set at pay; withdraw keeps the account) remain unverifiable from this repo, but the adopted fixes are correct independent of them.

e2e/internal/qa/client_settlement.go
ben-dz: Fixed in 6ed35018, one step further than the two options: the poll now reuses `seatIsWithdrawn` instead of carrying a third bespoke confirmation branch. That gives it the rotate-and-reread not-found handling and the pending-flag check for free, and the contradictory "closed/absent seat account" comment is gone with the branch.

PR #4068, opened 2026-07-17
ci: Improvements to testnet release flow

ben-dz:
## Summary of Changes
- Create the testnet release version-bump commit through the GraphQL `createCommitOnBranch` mutation instead of `git commit` + `git push`. Commits created via the API are signed by GitHub and show as Verified, so the version PR can merge under `main`'s required-signatures rule. The release App has no signing key, which left #4067 unmergeable until a human amended and re-signed the commit by hand.
- The release branch is created (or force-reset, preserving the previous `git push -f` re-run behavior) at the checked-out `main` HEAD via the git refs API before the commit is added.
- The infra pinned-versions step is unchanged: infra's `main` does not require signed commits.
- Gate 1 belt and suspenders: approving before both version PRs are merged no longer fails the release outright (that killed the v0.31.0 run). The gate posts a warning to the Slack thread and polls the PRs for up to 30 minutes, then proceeds when both merge. A closed PR or the timeout still fails the gate, and the prompt text still says to approve only after both merges.
- The orchestrator now tags the version-bump merge commit instead of the sha the run is pinned to, which by construction predates the bump. v0.30.0's tags landed one commit shy of the bump (the tagged tree's `Cargo.toml` says 0.29.0; shipped artifacts were unaffected because goreleaser injects the version from the tag name). Gate 1 resolves the doublezero version PR's merge commit and passes it to `release.testnet.push.tags.yml` via a new optional `ref` input; the manual components dispatcher passes no `ref` and keeps its current behavior (main at dispatch time).
- Slack messages that need a human lead with an *Operator Steps Required* numbered list (version-PR merge steps, the early-approval wait, the program-deploy instructions — mirrored in the run summary), with the gate approval spelled out as the last step. Informational posts are unchanged.

## Diff Breakdown
| Config/build |     2 | +124 / -20   | +104 |

Workflow-only change: the "Open doublezero version-bump PR" step, the `gate-tags` job, the `push-tags` call, and the operator-facing Slack/summary texts in `release.testnet.yml`, plus a new optional `ref` input in `release.testnet.push.tags.yml`.

## Testing Verification
- Built the mutation payload with the step's exact commands against the real `Cargo.toml`/`Cargo.lock`/`CHANGELOG.md` (GNU userland): ~597 KB payload, well under API limits, and the base64 `contents` fields round-trip back to byte-identical files.
- Verified every input field name (`CreateCommitOnBranchInput`, `CommittableBranch`, `FileChanges`, `FileAddition`, `CommitMessage`) against GitHub's published GraphQL schema.
- Tag-target resolution verified against the real v0.31.0 PR: `gh pr view 4067 --json mergeCommit --jq .mergeCommit.oid` returns the bump merge commit `f2248b0b9`. Confirmed the v0.30.0 regression it fixes: all nine `*/v0.30.0` tags point at `354190e83`, the direct parent of the bump commit `23048e954`.
- Empty `ref` on `actions/checkout` falls back to the run's pinned sha (the action's documented default), so dispatcher-driven tag pushes are byte-for-byte unchanged.
- The 30-minute wait can't be exercised by a dry run (dry-run version PRs stay open on purpose, so the gate proceeds immediately); it only runs in a real release approved early. The poll loop mirrors the existing `verify-cloudsmith` idiom, including treating transient API errors as failed attempts.
- End-to-end check before the next release: trigger `release.testnet.yml` with `dry_run=true` — the draft version PR it opens should show a Verified commit authored by `dz-release-bot[bot]`, and the dry-run tag jobs should no-op with an empty `ref`.

### DIFF
--- a/.github/workflows/release.testnet.push.tags.yml
+++ b/.github/workflows/release.testnet.push.tags.yml
@@ -11,6 +11,11 @@ on:
         description: 'The component to tag'
         required: true
         type: string
+      ref:
+        description: 'Commit to tag. Empty tags the sha the run is pinned to (the manual dispatcher path: main at dispatch).'
+        required: false
+        type: string
+        default: ''
       skip_existing:
         description: 'Succeed without re-tagging if the tag already exists (for orchestrator re-runs)'
         required: false
@@ -55,6 +60,8 @@ jobs:
         with:
           repository: malbeclabs/doublezero
           token: ${{ secrets.DOUBLEZERO_PAT }}
+          # empty ref = the sha this run is pinned to (actions/checkout default)
+          ref: ${{ inputs.ref }}
           fetch-tags: true
 
       - name: Validate format and push tag
--- a/.github/workflows/release.testnet.yml
+++ b/.github/workflows/release.testnet.yml
@@ -100,14 +100,33 @@ jobs:
           DRY_RUN: ${{ inputs.dry_run }}
         run: |
           set -euo pipefail
-          git config user.name "dz-release-bot"
-          git config user.email "dz-release-bot@malbeclabs.com"
           BRANCH="release/v${VERSION}"
           ./scripts/release/bump-version.sh "$VERSION"
-          git checkout -b "$BRANCH"
-          git add Cargo.toml Cargo.lock CHANGELOG.md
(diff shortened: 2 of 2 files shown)


juan-malbeclabs (approved):
LGTM

PR #4070, opened 2026-07-20
e2e/qa: report the epoch-tail closed window distinctly when wait_for_open_phase fails

### LINKED ISSUE #4069: Mainnet QA: report the epoch-tail closed window distinctly when wait_for_open_phase fails
# Mainnet QA: report the epoch-tail closed window distinctly when wait_for_open_phase fails

## Problem

The hourly `qa.mainnet-beta` workflow (malbeclabs/infra) runs this repo's QA e2e suite against Solana mainnet. `TestQA_MulticastSettlement/wait_for_open_phase` polls the shred-subscription program's execution controller for up to 2 minutes and, on timeout, fails with:

```
shred-subscription program did not enter OpenForRequests phase within timeout
program in "updating prices" phase, not yet open for requests
```

That message is indistinguishable from a real outage (oracle down, program stuck). But for the last ~15 minutes of every Solana epoch, the failure is expected by design:

- The shred oracle advances the program `OpenForRequests → ClosedForRequests` at slot `first_slot_of_next_epoch − closed_for_requests_grace_period_slots`. The grace period lives onchain in the shred-subscription ProgramConfig; on mainnet it is currently 2250 slots (~15 min).
- The oracle then settles seats (`ClosedForRequests`), updates prices (`UpdatingPrices`), and advances back to `OpenForRequests` only just AFTER the epoch boundary slot. The program is therefore closed for the full tail window of every epoch.
- The test's 2-minute wait can never bridge that window, so any run whose settlement test lands inside it fails. With hourly runs and ~2.06-day mainnet epochs, that is roughly one spurious red run per week.

Verified example (2026-07-20, infra actions run 29772283452, epoch 1004→1005, all UTC):

- 19:28:12 — `ClosedForRequests` at slot 434,157,756 (oracle target 434,157,750 = 434,160,000 − 2250)
- 19:31:29–19:33:26 — QA polls, sees "updating prices", times out, run goes red
- 19:44:04 — epoch boundary (slot 434,160,000, first slot of epoch 1005)
- 19:44:06 — `OpenForRequests` at slot 434,160,003; next hourly run green

## Goal

When `wait_for_open_phase` times out AND the timing is consistent with this epoch-tail window, the QA run must report it distinctly, so on-call can tell benign-window from real-outage at a glance. The mechanism is the implementer's choice — e.g. `t.Skip` with an explanatory message (run stays green), or a failure carrying a distinct, explicit message (`expected epoch-tail closed window: epoch N boundary at <time>, close target slot <slot>, current slot <slot>`). Weigh the tradeoff: QA redness is an alerting signal (the infra workflow posts failures to Slack), and a known-benign window arguably should not page — but silently skipping real coverage is worse. Pick one and justify the choice in the PR description.

## Hard requirements

- The "benign window" classification must be verified against live chain state, not assumed: compute the close-target slot from the target cluster's epoch schedule (RPC) and the actual onchain grace period, and check the current slot falls inside the window. Do not hardcode 2250.
- The settlement tests run against both mainnet-beta and testnet (`-env` flag in the infra QA workflows), so derive everything — RPC endpoint, program ID, epoch schedule — from whatever the settlement client already targets for the active environment. No mainnet-only constants.
- A `wait_for_open_phase` timeout OUTSIDE the window must keep failing exactly as loudly as it does today.
- Unit-test the window-classification logic (table-driven; no live RPC in unit tests).
- Keep the change inside this repo's e2e code; do not require changes to the infra repo's workflow jobs. If a distinct output marker would help infra's "Analyze QA failure" job classify runs, note that in the PR description as a follow-up.

## Pointers (verified 2026-07-20)

- Test: `e2e/qa_multicast_settlement_test.go` — the `wait_for_open_phase` subtest calls `client.WaitForOpenForRequests(ctx)` (~line 156).
- Poll loop: `e2e/internal/qa/client_settlement.go` — `WaitForOpenForRequests`, backoff `MaxElapsedTime = 2 * time.Minute`. The client in that package already holds Solana RPC access (endpoint pool with failover) — reuse it.
- Phase cycle

ben-dz:
## Summary of Changes

* `TestQA_MulticastSettlement` now skips — instead of failing — when `wait_for_open_phase` times out during the by-design closed window at the tail of every Solana epoch, so hourly QA redness stays a real alerting signal. Fixes #4069
* The benign classification is verified against live chain state, nothing assumed or hardcoded: the grace period is read from the shred-subscription ProgramConfig, the close-target slot is computed from the target cluster's epoch schedule (`getEpochInfo`), and the current slot must fall inside `[close target, epoch boundary)`. Everything derives from the settlement client's per-environment config, so the same code works for mainnet-beta and testnet.
* The verdict additionally requires the execution controller phase to be observably not `OpenForRequests` (so two minutes of RPC read failures against an open program still fails), and the wait-start slot (best-effort read) to also be inside the window (so a premature close before the oracle's target fails loudly). The classification reads select the healthiest pool endpoint first, so a lagging endpoint's stale slot view cannot skip a real outage.
* A timeout outside the window — or any classification error — falls through to the exact same loud `require.NoError` failure as today.

### Why skip rather than a distinctly-worded failure

QA redness pages Slack via the infra workflow, and this window is deterministic, verified against chain state at failure time, and unavoidable by design (the 2-minute wait can never bridge a ~15-minute closed window). Paging on-call weekly for a condition they can only acknowledge trains them to ignore QA reds. The skip is not silent: the run logs and the test summary carry a distinct `expected epoch-tail closed window: epoch N boundary at slot ..., close target slot ..., current slot ..., program phase ...` message, and the classification refuses to guess (falls through to a loud failure) whenever chain state cannot confirm the window. This also matches existing precedent in the same test (`ensure_program_unpaused` skips during migrations).

**Follow-up for infra:** the skip message is a stable marker (`expected epoch-tail closed window:`) that the "Analyze QA failure" job could match on if it wants to annotate these runs; no infra workflow change is required for this PR to work.

### Motivating example (infra actions run 29772283452, epoch 1004→1005, 2026-07-20 UTC)

- 19:28:12 — `ClosedForRequests` at slot 434,157,756 (oracle target 434,157,750 = 434,160,000 − 2250)
- 19:31:29–19:33:26 — QA polls, sees "updating prices", times out, run goes red
- 19:44:04 — epoch boundary (slot 434,160,000, first slot of epoch 1005)
- 19:44:06 — `OpenForRequests` at slot 434,160,003; next hourly run green

With this change, that run would have skipped with the distinct message instead of paging; a run red for any other reason (oracle down, program stuck outside the window, phase open but reads failing) still fails exactly as before.

## Diff Breakdown

| Core logic   |     1 | +167 / -0   | +167 |
| Tests        |     2 | +358 / -0   | +358 |
| **Total**    |     4 | +526 / -0   | +526 |

Mostly tests; the core logic is a single new classifier file plus wiring in the QA test.

<details>

- `e2e/internal/qa/epoch_window.go` — new `EpochTailWindow` classifier: pure slot/phase window math (`classifyEpochTailWindow`) plus `EpochTailClosedWindow`, which reads ProgramConfig grace period, controller phase, and `getEpochInfo` through the existing failover pool
- `e2e/qa_multicast_settlement_test.go` — `wait_for_open_phase` records the wait-start slot, classifies on timeout, and skips (subtest + parent) only on a verified-benign window
- `e2e/internal/qa/epoch_window_test.go` — table-driven unit tests (no live RPC) pinning the full classification struct at every boundary, all three benign gates, and every error branch

</details>

## Testing Verification

* Table-driven unit tests (no live RPC) pin the window boundaries exactly at the verified epoch 1004→1005 mainnet numbers: benign at the close-target slot and the last slot before the boundary; not benign one slot before the target, at the first slot of the next epoch, when the wait started before the close target, and when the phase reads `OpenForRequests` despite an in-window slot.
* Error branches are pinned with distinct message fragments: zero slots-in-epoch, slot index inconsistent with the current slot or epoch length, and a grace period covering the entire epoch (refuses to classify rather than mask outages).
* Fail-loud paths verified by inspection and covered by the switch structure: any classification error, out-of-window timeout, or open-phase timeout falls through to the pre-existing `require.NoError`.
* Reviewed by three independent review agents (correctness, SDK/idiom, security); all Medium findings (phase gate, wait-start gate, healthiest-endpoint selection, nil `getEpochInfo` guard) were applied in the second commit.

Note: the "benign skip" path itself can only fire against a live cluster inside the ~15-minute window; it was not exercised end-to-end here. The next epoch-tail collision on the hourly mainnet run will be the live validation, and a misclassification degrades to today's behavior (a red run), never to a silent skip.


e2e/internal/qa/epoch_window.go
```diff
@@ -0,0 +1,167 @@
...
+		// benign and permanently mask real outages; refuse rather than guess.
+		return EpochTailWindow{}, fmt.Errorf("grace period (%d slots) covers the entire epoch (%d slots); refusing to classify", grace, slotsInEpoch)
+	}
+	nextEpochStart := currentSlot - slotIndex + slotsInEpoch
+	closeTarget := nextEpochStart - grace
+	inWindow := currentSlot >= closeTarget
+	startedInWindow := waitStartSlot == 0 || waitStartSlot >= closeTarget
+	closedNow := phase != shreds.ExecutionPhaseOpenForRequests
+	return EpochTailWindow{
+		Benign:             inWindow && startedInWindow && closedNow,
```
nikw9944: **M1:** `Benign` verifies "closed while inside the window", not "closed by this epoch's scheduled close". If the oracle dies mid-epoch and the program sits closed for hours, an hourly run whose entire 2-minute wait lands inside the tail window satisfies all three gates (`startedInWindow` holds because the wait genuinely started in the window) and skips — violating the PR's invariant that a misclassification degrades to a red run, never a silent skip. The fix is nearly free: `EpochTailClosedWindow` already fetches the `ExecutionController` and discards everything but `GetPhase()`; the account carries `LastClosedForRequestsSlot` (sdk/shreds/go/state.go:109), and live mainnet confirms the oracle stamps it on each close — for this PR's motivating example (epoch 1004) it reads 434,157,756, the exact observed close slot, 6 slots after the 434,157,750 target (`LastOpenForRequestsSlot` = 434,160,003 likewise matches the documented reopen). Thread it into `classifyEpochTailWindow` and add a fourth conjunct, `lastClosedSlot >= closeTarget && lastClosedSlot < nextEpochStart`, proving the close was this epoch's scheduled one (`>=` tolerates the oracle's few-slot landing lag; an early close fails loudly, the safe direction). This also restores a strong gate in the degraded `waitStartSlot == 0` mode. Add table cases for "closed long before the target → not benign".

e2e/internal/qa/epoch_window.go
```diff
@@ -0,0 +1,167 @@
...
+		return EpochTailWindow{}, fmt.Errorf("invalid epoch info: slots-in-epoch is zero")
+	}
+	if slotIndex >= slotsInEpoch {
+		return EpochTailWindow{}, fmt.Errorf("inconsistent epoch info: slot index %d >= slots in epoch %d", slotIndex, slotsInEpoch)
+	}
+	if slotIndex > currentSlot {
+		return EpochTailWindow{}, fmt.Errorf("inconsistent epoch info: slot index %d > current slot %d", slotIndex, currentSlot)
+	}
+	grace := uint64(gracePeriodSlots)
+	if grace >= slotsInEpoch {
```
nikw9944: **M2:** This guard refuses only when `grace >= slotsInEpoch`. A fat-fingered or malicious onchain config update to, say, `slotsInEpoch − 1` passes the guard and classifies essentially every slot of every epoch as in-window — the exact "permanently mask real outages" failure the comment says this guard exists to prevent, enforced only at the boundary. The operational value is 2,250 slots (~15 min); tighten the refusal to a sanity bound, e.g. `grace > slotsInEpoch/4` (or an absolute cap), with the same "refusing to classify" error.

e2e/internal/qa/epoch_window.go
```diff
@@ -0,0 +1,167 @@
...
+		// A grace period spanning the whole epoch would classify every slot as
+		// benign and permanently mask real outages; refuse rather than guess.
+		return EpochTailWindow{}, fmt.Errorf("grace period (%d slots) covers the entire epoch (%d slots); refusing to classify", grace, slotsInEpoch)
+	}
+	nextEpochStart := currentSlot - slotIndex + slotsInEpoch
+	closeTarget := nextEpochStart - grace
+	inWindow := currentSlot >= closeTarget
+	startedInWindow := waitStartSlot == 0 || waitStartSlot >= closeTarget
+	closedNow := phase != shreds.ExecutionPhaseOpenForRequests
```
nikw9944: **L1:** `closedNow := phase != ExecutionPhaseOpenForRequests` is a denylist: an unknown future `Phase` byte (3–255 after a program upgrade, or deserialization drift) counts as "closed" and can gate a benign skip, inconsistent with this file's refuse-on-anomaly stance. Allowlist `ClosedForRequests`/`UpdatingPrices` and return an error for out-of-range phases.

e2e/internal/qa/epoch_window.go
```diff
@@ -0,0 +1,167 @@
...
+// Three conditions must all hold for Benign:
+//   - currentSlot is inside [close target, epoch boundary);
+//   - waitStartSlot (when known) is at or after the close target: a timeout
+//     means the program was closed for the entire wait, so a wait that began
+//     before the close target observed the program closed before the oracle
+//     was due to close it — an anomaly, not the benign window;
+//   - the program is observably not in OpenForRequests phase: a timeout can
+//     also be minutes of RPC read failures against a program that is in fact
+//     open, which must keep failing loudly.
+func classifyEpochTailWindow(epoch, currentSlot, slotIndex, slotsInEpoch uint64, gracePeriodSlots uint32, phase shreds.ExecutionPhase, waitStartSlot uint64) (EpochTailWindow, error) {
```
nikw9944: **L2:** Four adjacent positional `uint64`s all sourced from one `getEpochInfo` response plus a trailing `uint64 waitStartSlot`; the single call site at line 166 compiles under any transposition, and the unit tests exercise the pure function, not the wiring. Pass `*rpc.GetEpochInfoResult` (or a small named struct) plus the named `gracePeriodSlots`/`phase`/`waitStartSlot`.

e2e/internal/qa/epoch_window.go
```diff
@@ -0,0 +1,167 @@
...
+	programID, err := solana.PublicKeyFromBase58(c.ShredSubscriptionProgramID)
+	if err != nil {
+		return EpochTailWindow{}, fmt.Errorf("failed to parse shred subscription program ID %q: %w", c.ShredSubscriptionProgramID, err)
+	}
+
+	// Rotate off a lagging endpoint before reading: the pool fails over on RPC
+	// errors, not on stale-but-valid replies, and a lagging slot view here is
+	// what decides whether a real outage pages or is skipped as benign.
+	if c.solanaRPC != nil {
+		c.solanaRPC.SelectHealthiestEndpoint(ctx)
```
nikw9944, line 182: **L3:** `SelectHealthiestEndpoint` tolerates up to the 150-slot lag threshold (and is a no-op with one endpoint or when all probes fail), so a stale view taken just past the epoch boundary can report the previous epoch's tail and skip a still-closed real outage for one run (the next hourly run pages). If worth closing: expose the max probed slot from `SelectHealthiestEndpoint` and refuse `Benign` when `info.AbsoluteSlot` trails it by more than the threshold.

nikw9944 (commented):
The classifier is well-factored (pure, table-tested, refuses to guess on inconsistent inputs) and every ambiguity resolves toward failing loudly rather than skipping; the skip plumbing through subtest + parent `t.Skipf` is correct and the new error paths consistently scrub endpoint URLs. Three Medium findings harden the benign gates: (M1) `Benign` proves the timeout landed in the window but not that the closure was this epoch's scheduled close — the already-fetched `ExecutionController.LastClosedForRequestsSlot` closes that gap (verified against live mainnet: it reads 434,157,756, the exact observed close slot from the PR's motivating example); (M2) the grace-period refusal guard only catches the 100%-of-epoch case; (M3) a pre-existing unscrubbed error in `WaitForOpenForRequests` is exactly the message this PR's fail-loud path prints into CI logs. Low findings are ergonomics and defense-in-depth.

- **M3 (pre-existing, on the exact path this PR instruments):** `WaitForOpenForRequests` (e2e/internal/qa/client_settlement.go:203-205) wraps the raw fetch error (`fmt.Errorf("failed to fetch execution controller: %w", err)`). solana-go embeds the full request URL — including any API key — in its connectivity/HTTP error strings (the repo's own `scrubRPCErr` comment documents this), and that error is exactly what `require.NoError` at e2e/qa_multicast_settlement_test.go:213 prints into CI logs when classification is not benign. Every error path new in this PR scrubs correctly; fix this one (here or in an immediate follow-up) with `fmt.Errorf("failed to fetch execution controller: %s", c.scrubRPCErr(err))` inside the retry closure, as `EpochTailClosedWindow` already does.

packethog (approved):

e2e/internal/qa/epoch_window.go
ben-dz: Fixed in 13d668be. `classifyEpochTailWindow` now takes the already-fetched `*shreds.ExecutionController` and requires `LastClosedForRequestsSlot` in `[close target, epoch boundary)` as a fourth conjunct (`>=` tolerates the oracle's landing lag; an early close fails loudly). Added table cases for a stale close from an oracle dead mid-epoch (with wait start unknown, pinning the gate in degraded mode), a close one slot before the target, and a close at the boundary. The slot is also included in the window's `String()` for the run log.

e2e/internal/qa/epoch_window.go
ben-dz: Deliberately leaving this at the degenerate-case boundary for now: a near-epoch-length grace period requires a drastic onchain config change we judge unlikely, and the run log now prints the computed window (including the grace period) on every classification, so a misconfiguration that extreme would be visible immediately. Happy to add an absolute sanity cap in a follow-up if the config ever starts moving.

e2e/internal/qa/epoch_window.go
ben-dz: Fixed in 13d668be. The phase check is now an allowlist (`ClosedForRequests`/`UpdatingPrices` gate benign, `OpenForRequests` fails loudly) and any other phase byte returns a refuse-to-classify error, with a table case for phase 3.

e2e/internal/qa/epoch_window.go
ben-dz: Fixed in 13d668be using existing structs — the classifier now takes `*rpc.GetEpochInfoResult` and `*shreds.ExecutionController` (which also carries the last-closed slot needed for M1), leaving only `gracePeriodSlots` and `waitStartSlot` positional. Nil inputs error, with unit tests.

e2e/internal/qa/epoch_window.go
ben-dz: Deliberately not closing this one: the worst case is a single skipped run whose still-closed outage pages on the next hourly run, which doesn't justify plumbing the max probed slot out of `SelectHealthiestEndpoint`.

ben-dz:
Addressed the review in 13d668be and d7090f6f:

- **M1** — benign now additionally requires `LastClosedForRequestsSlot` in `[close target, epoch boundary)`, so an oracle dead mid-epoch fails loudly even when the whole wait lands inside the tail window. Table cases added for a stale close (including in degraded wait-start-unknown mode), an early close, and a close at the boundary.
- **M3** — the fetch error inside `WaitForOpenForRequests` is now scrubbed via `scrubRPCErr` before it reaches `require.NoError` and CI logs.
- **L1** — phase check is an allowlist (`ClosedForRequests`/`UpdatingPrices`); unknown phase bytes return a refuse-to-classify error.
- **L2** — the classifier takes the existing `*rpc.GetEpochInfoResult` and `*shreds.ExecutionController` structs instead of positional `uint64`s.
- d7090f6f also refuses to classify when the wait-start slot is ahead of the classification-time slot (a provably stale slot view, the cheap subset of L3 that needs no extra RPC), and uses `shreds.NewRPCClient` in the no-pool fallback for parity with `shredsClient`.

Deliberately unchanged: **M2** (a near-epoch-length grace period requires a drastic onchain config change we judge unlikely, and the computed window including the grace period is printed on every classification) and **L3** (a stale-view skip costs at most one hourly cycle before the next run pages; not worth the extra plumbing).

PR #4073, opened 2026-07-22
cli: add access-pass migrate-to-oracle

martinsander00:
## Summary

- Adds `doublezero access-pass migrate-to-oracle`, an operator command that re-owns the shred oracle's validator-seeded access passes back to the oracle (malbeclabs/infra#2031). The creation path was already fixed upstream so fresh subscribers are oracle-owned; this migrates the pre-existing validator-owned population.
- For each pass where `owner == oracle` but `user_payer != oracle` (with a live `Multicast` `User` at that client IP), it assembles the whole re-own as **one atomic transaction per pass**: unsubscribe the user from each group → `DeleteUser` → `CloseAccessPass` (old) → `SetAccessPass` (recreated oracle-seeded) → restore the subscriber allowlist → `CreateSubscribeUser` + re-subscribe the user oracle-owned. The `User` PDA is keyed by `(client_ip, user_type)`, so the account and its `tunnel_id`/`tunnel_net` never move — only ownership changes, and the datapath doesn't flap.
- **Dry run by default**: builds and simulates each transaction, printing its serialized size, CU, and any simulation error, followed by a summary (max size, over-limit count, sim errors). `--execute` sends. `--oracle` selects the target set, `--client-ip` narrows to one; the binary's global `--env` / `--url` / `--keypair` flags apply.
- Must be signed by the oracle key: `SetAccessPass` sets the recreated pass's `owner` to the transaction signer, so the command refuses to run unless the configured signer equals `--oracle` (prevents re-owning to the wrong key).
- A pass still referenced by another, non-migrated user (`connection_count > 1`) is migrated but its old pass is left open rather than closed, and flagged in the output.

## Testing Verification

- Dry-ran the full target set on mainnet-beta (99 passes with a live Multicast user): every transaction fits a legacy transaction (846–1114 bytes, under the 1232-byte limit — no v0/ALT needed) and simulates cleanly end-to-end (≤ ~440k CU).
- The one pass that initially failed simulation surfaced the shared-pass case (`AccessPassInUse` on close, `connection_count == 2`); handled by leaving that pass open, after which the whole set simulates with zero errors.
- Not yet exercised with `--execute` against mainnet-beta.

## Notes / follow-ups

- ~41 additional oracle-owned passes have no live Multicast user (orphaned / probe-first-reuse cases) and are intentionally skipped by target discovery; they need a separate cleanup pass.
- No unit tests yet — the command is validated by the mainnet dry run above; happy to add mock-client coverage if preferred.


smartcontract/cli/src/accesspass/migrate_to_oracle.rs
```diff
@@ -0,0 +1,504 @@
...
+/// re-own sequence: unsubscribe the Multicast user from every group, delete it,
+/// close the old (validator-seeded) pass, recreate the pass oracle-seeded,
+/// restore its subscriber allowlist, and recreate + re-subscribe the user
+/// oracle-owned. Defaults to a dry run that builds and simulates each tx and
+/// prints its serialized size; pass `--execute` to send.
+#[derive(Args, Debug)]
+pub struct MigrateAccessPassToOracleCliCommand {
+    /// The shred oracle pubkey (the passes' current `owner`).
+    #[arg(long)]
+    pub oracle: Pubkey,
```
karl-dz: Not sure how useful this argument is. I can pass in my own keypair's key here and the check below will pass.

Shouldn't we just fetch the oracle permission account and check that the payer's key equals the one that has the shred oracle permission?

smartcontract/cli/src/accesspass/migrate_to_oracle.rs
```diff
@@ -0,0 +1,504 @@
...
+            targets.len(),
+            if self.execute { "" } else { " (dry run)" }
+        )?;
+
+        const LEGACY_TX_LIMIT: usize = 1232;
+        let mut max_bytes = 0usize;
+        let mut over_limit = 0usize;
+        let mut sim_errors = 0usize;
+        let mut retained = 0usize;
+        let mut done = 0usize;
```
karl-dz: Do these types need to be explicitly defined? Why can't they just be `0`?

smartcontract/cli/src/accesspass/migrate_to_oracle.rs
```diff
@@ -0,0 +1,504 @@
...
+
+        for target in &targets {
```
karl-dz: Nit: no need to reference. `targets` can just be moved because it isn't used after this loop, right?

karl-dz (commented):
Will we need this command after we do the migration? Just want to know if we can queue up removing this command afterwards

martinsander00:
> Will we need this command after we do the migration? Just want to know if we can queue up removing this command afterwards

yea i'm not sure if we even want to merge it

martinsander00:
Ran the migration already without merging w Karl :)

PR #4075, opened 2026-07-22
serviceability: gate EdgeSeat access-pass writes on min_compatible_version

ben-dz:
## Summary

- Refuse `SetAccessPass` with an EdgeSeat type and `SetAccessPassFeeds` while `min_compatible_version` is below 0.30.0 (`EDGE_SEAT_MIN_COMPATIBLE_VERSION`), with a new error `EdgeSeatCompatibilityWindowNotMet` (Custom 101). EdgeSeat passes serialize as `EdgeSeat(Vec<FeedSeat>)` (#3954); pre-0.30 client decoders misparse every field after the variant tag and can abort on a bogus allowlist preallocation — this is what crash-looped contributor-rewards on testnet (52 GB allocation abort while decoding pass `5x9DTsWCrAQtYKH4W7trEjoaCRHnzywsZp2oFX3sg5Zg`). The gate turns "raise the compat floor before minting EdgeSeat passes" from a runbook step into a program-enforced invariant.
- `min_compatible_version` is mirrored onto `GlobalState` (new trailing field) so the gate reads an account the gated instructions already receive: no new accounts, no SDK or CLI changes, every instruction's account layout untouched, old SDK builders keep working. `SetMinVersion` remains the single writer and updates both the `ProgramConfig` copy and the mirror. GlobalState accounts written before this field decode it as 0.0.0, so the gate fails closed until `SetMinVersion` runs.
- Enablement is per-environment ops with no further deploys: raise the floor to 0.30.0 with `SetMinVersion` and EdgeSeat provisioning resumes there; leave it and the environment stays protected. Once floors are ≥ 0.30.0 everywhere the check is permanently satisfied and dormant.
- Fixes `SetVersionCommand` (SDK): its account list was missing the ProgramConfig account, so SDK/CLI-driven `SetMinVersion` failed with InvalidAccountData before ever updating the floor — consistent with both networks still declaring the 0.21.0 init default. The floor-raise lever this gate depends on now works.

**Deploy note:** both testnet and mainnet currently declare `min_compatible_version = 0.21.0`, so after this program deploys the gate is closed everywhere. Testnet needs `SetMinVersion(0.30.0)` promptly so edge-connect seat provisioning keeps working; mainnet should stay closed until we deliberately raise its floor (which force-upgrades pre-0.30 CLIs via the existing `checkversion` gate).

Context: the incident this hardens against is the do-tn-rewards1 contributor-rewards crash loop. Sibling fixes: doublezerofoundation/doublezero-offchain#402 (merged — rewards SDK pin to client/v0.31.0), #4074 (bounded `deserialize_vec_with_capacity` prealloc), doublezero-offchain#404 (ingest warn+skip). This PR closes the remaining hole: nothing stops a FeedSeat-era program from minting accounts that admitted-but-stale clients cannot decode.

## Testing Verification

- New integration test `test_edge_seat_blocked_below_compat_floor`: EdgeSeat `SetAccessPass` and `SetAccessPassFeeds` rejected with Custom(101) below the floor; Prepaid passes unaffected; `SetMinVersion` re-raising the floor re-admits feed provisioning (exercises the GlobalState mirror sync).
- Existing EdgeSeat/feed suites (`set_access_pass_feeds_test`, `feed_metro_gate_test`, `delete_user_dynamic_accesspass`, accesspass refill) pass with their setups raising the floor — 47/47 program test suites green.
- `make rust-lint`, the workspace test suite (20/20), and the onchain account-compat check (`accounts -ed`: real testnet/devnet accounts decode with the new GlobalState layout) pass; the SBF (`cargo test-sbf`) pass runs in CI.

A minimal hard-kill-switch alternative is up as draft #4076 for comparison — merge exactly one of the two.

### DIFF
--- a/sdk/serviceability/testdata/fixtures/generate-fixtures/src/main.rs
+++ b/sdk/serviceability/testdata/fixtures/generate-fixtures/src/main.rs
@@ -183,6 +183,7 @@ fn generate_global_state(dir: &Path) {
         qa_allowlist: vec![qa_pk],
         feature_flags: 1,
         feed_authority_pk: feed_authority_pk,
+        min_compatible_version: Default::default(),
     };
 
     let data = borsh::to_vec(&val).unwrap();
--- a/smartcontract/cli/src/globalconfig/airdrop/get.rs
+++ b/smartcontract/cli/src/globalconfig/airdrop/get.rs
@@ -80,6 +80,7 @@ mod tests {
             qa_allowlist: vec![],
             feature_flags: 0,
             feed_authority_pk: Pubkey::default(),
+            min_compatible_version: Default::default(),
         };
 
         client
--- a/smartcontract/cli/src/globalconfig/authority/get.rs
+++ b/smartcontract/cli/src/globalconfig/authority/get.rs
@@ -94,6 +94,7 @@ mod tests {
             qa_allowlist: vec![],
             feature_flags: 0,
             feed_authority_pk: feed_authority,
+            min_compatible_version: Default::default(),
         };
 
         client
--- a/smartcontract/cli/src/globalconfig/featureflags/get.rs
+++ b/smartcontract/cli/src/globalconfig/featureflags/get.rs
@@ -77,6 +77,7 @@ mod tests {
             qa_allowlist: vec![],
             feature_flags: 1,
             feed_authority_pk: Pubkey::default(),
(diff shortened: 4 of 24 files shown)


ben-dz:
Shipping min version quickly so I think this isn't necessary.

PR #4076, opened 2026-07-23
serviceability: hard-disable EdgeSeat access-pass writes (kill-switch alternative to #4075)

ben-dz:
## Summary

Minimal-stopgap alternative to #4075 — merge exactly one of the two.

- Hard-disables both EdgeSeat write paths behind a `EDGE_SEAT_WRITES_DISABLED = true` const: `SetAccessPass` with an EdgeSeat type and `SetAccessPassFeeds` return `FeatureNotEnabled` unconditionally. Same protection goal as #4075: EdgeSeat passes serialize as `EdgeSeat(Vec<FeedSeat>)` (#3954), pre-0.30 client decoders misparse every field after the variant tag and can abort on a bogus allowlist preallocation (the do-tn-rewards1 crash loop) — so stop minting those bytes while such clients are admitted.
- ~30 source lines across 3 files; no state changes, no new error variant, no SDK/CLI changes, no account-layout changes.

Trade-offs versus #4075 (the dynamic `min_compatible_version` gate):

| | this PR (kill switch) | #4075 (dynamic floor gate) |
|---|---|---|
| source diff | ~30 lines / 3 files | ~70 lines / 8 files (+field-add churn, 24 files total) |
| lift / enable | flip the const → another program release | `SetMinVersion(0.30.0)` transaction, per environment |
| environments | blocks testnet and mainnet equally; testnet edge-connect provisioning paused until the lift release | testnet can open immediately, mainnet stays closed until chosen |
| test coverage | 15 EdgeSeat/feed-seat/metro-gate tests `#[ignore]`d until lift | full suite stays live (setups raise the floor) |
| end state | switch + ignores must be removed | gate goes dormant once floors ≥ 0.30.0, nothing to remove |
| SetVersionCommand account-list bug | untouched (fix needed eventually for any floor raise) | fixed |

**Deploy note:** the switch is active everywhere the moment this program version deploys.

## Testing Verification

- New `test_edge_seat_writes_disabled` asserts both instructions are refused with `FeatureNotEnabled` (Custom 84).
- Remaining program suites pass with the 15 EdgeSeat-dependent tests `#[ignore]`d (they exercise exactly the flows the switch forbids; un-ignore when lifting).
- `make rust-lint` passes.

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/mod.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/mod.rs
@@ -8,6 +8,13 @@ use solana_program::{
     rent::Rent, sysvar::Sysvar,
 };
 
+// Temporary hard switch: EdgeSeat passes serialize as `EdgeSeat(Vec<FeedSeat>)` (#3954), and
+// pre-0.30 client decoders misparse every field after the variant tag, aborting on a bogus
+// allowlist preallocation (the do-tn-rewards1 crash loop). Flip to false — together with the
+// checks in set.rs/set_feeds.rs and the #[ignore]s on the EdgeSeat test suites — in the release
+// after ProgramConfig.min_compatible_version has been raised to >= 0.30.0 everywhere.
+pub const EDGE_SEAT_WRITES_DISABLED: bool = true;
+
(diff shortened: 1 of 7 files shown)


ben-dz:
Shipping min version quickly so I think this isn't necessary.

PR #4077, opened 2026-07-23
e2e: demonstrate retransmit-only subscription via shred pay

elitegreg:
Resolves: malbeclabs/infra#2025

## Summary

- Add `TestQA_RetransmitOnlySettlement`, a QA end-to-end test that demonstrates the retransmit-only shred subscription: a client pays for a seat on a device in a retransmit-only metro, is charged the discounted price (10 USDC), and ends up subscribed to the retransmit multicast group only — not the leader group, so leader shreds are excluded.
- Select the retransmit-only device by reading the metro's `MetroHistory` retransmit-only flag (auto-discovery), with a `-retransmit-only-device` pin override; the test skips when no such device exists, so it is a no-op on a network where the feature is not yet deployed.
- Add `MetroHistory.IsRetransmitOnlyEnabled()` (flag bit 0) to the shreds Go SDK and a `GetServiceabilityUser` qa helper, so the test can read the seat's onchain multicast `Subscribers` and assert the leader group is absent and the retransmit group is present.
- Environment-agnostic: the leader/retransmit group codes and the expected price are flags, so the same binary validates the testnet QA network and, after that deploy, mainnet.

The test reuses the existing shred-pay flow (`FeedSeatPrice`, `FeedSeatPay`, tunnel-up, device-assignment, `WithdrawSeatWithRetry`) and its cleanup, and is gated behind `-enable-retransmit-only-settlement-tests` like the existing settlement test.

## Testing

- `go build -tags qa ./e2e/...`, `go vet -tags qa ./e2e/...`, and `golangci-lint --build-tags qa` on the touched packages — 0 issues.
- `go test ./sdk/shreds/go/...` — new `TestMetroHistoryFlags` pins the retransmit-only / price-finalized bit truth table; `TestMetroHistoryDeserialization` now also asserts the retransmit-only bit reads false when only the finalized bit is set.
- The live QA run needs the retransmit-only oracle work (malbeclabs/doublezero-shreds #601, #599) and metro flag deployed to the testnet QA network, plus a metro configured retransmit-only there. It runs via the hourly scheduler with `-enable-retransmit-only-settlement-tests` and the group-code flags; not executed in this PR.


e2e/internal/qa/client_settlement.go
```diff
@@ -121,6 +121,69 @@ func (c *Client) ClosestDevice(ctx context.Context) (*Device, error) {
...
+		}
+		if l.AvgLatencyNs < bestAvg {
+			bestAvg = l.AvgLatencyNs
+			bestDevice = device
+		}
+	}
+	if bestDevice != nil {
+		c.log.Debug("Determined closest retransmit-only device", "host", c.Host, "deviceCode", bestDevice.Code, "avgLatencyNs", bestAvg)
+	}
+	return bestDevice, nil
```
nikw9944: `ClosestRetransmitOnlyDevice` returns `(nil, nil)` both when no metro is flagged retransmit-only and when flagged metros exist but none of their devices is reachable (or present in the devices map), and the test skips in both cases. On the testnet QA network after deploy, every device in the retransmit-only metro going unreachable produces a silent hourly skip instead of a failure, so the alert signal for exactly the feature under test disappears. Suggest skipping only when `RetransmitOnlyExchangeKeys` returns an empty set, and failing (naming the flagged metros) when metros are flagged but no reachable device matched — e.g. have the test call `RetransmitOnlyExchangeKeys` itself, or return the flagged set alongside the device.

e2e/qa_retransmit_only_settlement_test.go
```diff
@@ -0,0 +1,448 @@
...
+		var lastSubscriberCount int
+		require.Eventually(t, func() bool {
+			user, err := client.GetServiceabilityUser(ctx)
+			if err != nil {
+				log.Info("serviceability user poll error", "error", err)
+				return false
+			}
+			lastSubscriberCount = len(user.Subscribers)
+			return subscribed(user, retransmitGroup.PK) && !subscribed(user, leaderGroup.PK)
+		}, retransmitSubscribeTimeout, 5*time.Second,
```
nikw9944: When this 90s `require.Eventually` times out, the run log shows nothing about the seat's actual subscription state: only `lastSubscriberCount` is captured, the success `log.Info` never runs, and `DumpDiagnostics` covers tunnel status/routes/latency but not the onchain `User.Subscribers`. On-call can't tell whether the leader group leaked in or the retransmit group never appeared — the distinction between an oracle bug and a leader-exclusion bug. Suggest logging the subscribed group pubkeys (or the two booleans) on each non-matching poll, or capturing the last-seen state and logging it after the timeout.

e2e/qa_retransmit_only_settlement_test.go
```diff
@@ -0,0 +1,448 @@
...
+// TestQA_RetransmitOnlySettlement demonstrates the retransmit-only shred
+// subscription end to end: a client pays for a seat on a device in a
+// retransmit-only metro, is charged the discounted price, and ends up
+// subscribed to the retransmit multicast group only — not the leader group, so
+// leader shreds are excluded. It mirrors TestQA_MulticastSettlement, adding
+// retransmit-only device selection, the discounted-price assertion, and the
+// subscribed-groups assertion. The test is environment-agnostic: the group
+// codes and the expected price are flags, so the same binary validates the
+// testnet QA network and later mainnet.
+func TestQA_RetransmitOnlySettlement(t *testing.T) {
```
nikw9944, line 42: ~350 of the 448 lines here are copied verbatim from `TestQA_MulticastSettlement`, including the epoch-tail classification, escrow-drain guard, self-heal, and withdraw-retry blocks that all changed within the last two weeks (#4066, #4069). Future fixes to that machinery now have to land in both tests or this copy drifts silently. Consider extracting the shared pay/open-phase/balance/withdraw subtest chain into a helper parameterized on device selection and extra assertions; if linear copied QA flows are preferred for on-call readability, at minimum add a matching "mirrored by TestQA_RetransmitOnlySettlement" pointer in `qa_multicast_settlement_test.go` — the cross-reference currently exists only in one direction.

nikw9944 (commented):
Three Medium suggestions, all about the test's observability and maintainability rather than its pass-path correctness: the auto-discovery skip masks an unreachable retransmit-only metro once the feature is deployed, the `assert_subscribed_groups` timeout leaves no record of the seat's actual subscription state in the run log, and ~350 lines of the settlement skeleton (including the recently-churning epoch-tail/escrow-drain/self-heal machinery) are duplicated from `TestQA_MulticastSettlement` with only a one-way cross-reference. The shreds SDK flag bit (bit 0) and the `MetroHistory.ExchangeKey` ↔ serviceability exchange pubkey mapping both match the onchain program source.

e2e/internal/qa/client_settlement.go
elitegreg: Fixed in a1d709fb. `ClosestRetransmitOnlyDevice` now returns the set of flagged retransmit-only metros alongside the device, so the test distinguishes the two former `(nil, nil)` cases: it skips only when the set is empty (feature not configured on this network) and fails — naming the flagged metros via `flaggedMetroCodes` — when metros are flagged but none of their devices was reachable. The silent hourly skip that would have masked an unreachable feature-under-test is now a loud failure.

e2e/qa_retransmit_only_settlement_test.go
elitegreg: Fixed in a1d709fb. `assert_subscribed_groups` now captures the seat's last-seen state on every poll (the retransmit-present / leader-present booleans and the full list of subscribed group pubkeys) and, on timeout, logs it via `log.Warn` before failing (`assert.Eventually` + `require.True`). The run log now distinguishes an oracle bug (retransmit group never appeared) from a leader-exclusion bug (leader group leaked in) without needing a rerun.

e2e/qa_retransmit_only_settlement_test.go
elitegreg: Fixed in a1d709fb by extracting the shared flow rather than just cross-referencing. The pay/open-phase/balance/withdraw settlement skeleton — including the epoch-tail classification, escrow-drain guard, self-heal, and withdraw-retry machinery — now lives once in `runShredSettlement` (new `e2e/qa_shred_settlement_test.go`), parameterized by `shredSettlementParams` on device selection, the seat-price assertion, and an optional post-tunnel-up assertion. `TestQA_MulticastSettlement` and `TestQA_RetransmitOnlySettlement` are now thin wrappers (net +180/-790 lines), so future fixes to that machinery land in one place instead of drifting between two copies.

e2e/qa_shred_settlement_test.go
```diff
@@ -0,0 +1,453 @@
...
+		require.NoError(t, err, "failed to enable reconciler")
+	}) {
+		return
+	}
+
+	if !t.Run(p.selectSubtestName, func(t *testing.T) {
+		device = p.selectDevice(t, ctx, log, test, client)
+	}) {
+		return
+	}
```
nikw9944, line 190: When `selectDevice` calls `t.Skip` (e.g. `selectRetransmitOnlyDevice` at `qa_retransmit_only_settlement_test.go:87` when no metro is flagged retransmit-only), `t.Run` returns `true` — a skipped subtest counts as success, as the epoch-tail comment below already notes — so the flow continues with `device == nil` and the next subtest panics dereferencing `device.PubKey` at line 193. The designed "feature not deployed" skip becomes a nil-pointer crash, which is exactly the state the QA networks will be in when `-enable-retransmit-only-settlement-tests` is first enabled before a metro is flagged onchain. Mirror the `epochTailWindow` pattern: after this `t.Run` returns true, `if device == nil { t.Skip("device selection skipped") }` at the parent — a nil device after a non-failed subtest can only mean the selector skipped. (This was latent in the pre-refactor code too; the extraction makes it reachable from any selector that skips.)

nikw9944 (changes requested):
All three round-1 findings are verified fixed in a1d709fb: ClosestRetransmitOnlyDevice now returns the flagged-metro set so the test fails (naming metros) instead of silently skipping when flagged metros have no reachable device; assert_subscribed_groups logs the seat's last-seen subscription state on timeout; and the duplicated settlement flow is extracted into runShredSettlement with both tests as thin wrappers, faithfully preserving the original subtest chain. One new High: the skip inside selectRetransmitOnlyDevice does not stop the parent flow — t.Run returns true for a skipped subtest (verified) — so runShredSettlement continues with a nil device and panics dereferencing device.PubKey in query_seat_price. That converts the designed 'feature not deployed' skip into a crash, exactly the state testnet/mainnet are in when the enable flag is first turned on before a metro is flagged.

e2e/qa_shred_settlement_test.go
elitegreg: Fixed in c006d30a. Applied exactly the `epochTailWindow` pattern you suggested: after the selection subtest returns, `runShredSettlement` now guards `if device == nil { t.Skip(...) }` at the parent, so a selector that skips stops the whole run instead of falling through to a nil dereference in `query_seat_price`. A nil device there can only mean the selector skipped — a failure makes `t.Run` return false and returns above, and the success path always assigns a device.

Confirmed the mechanism with a throwaway test: `t.Run` returns `true` after a skipped subtest, and with the guard the parent reports SKIP rather than panicking. Good catch on the latent-since-pre-refactor bit — the extraction is what made it reachable from any selector that skips.

nikw9944 (approved):
No code changes since the last round — the head is unchanged at c006d30a, the commit already verified. The only new activity is the Copilot bot's summary-only review, which generated no findings and contains nothing requiring action or affecting this review. All previously verified fixes stand: ClosestRetransmitOnlyDevice returns the flagged-metro set so an unreachable retransmit-only metro fails loudly instead of silently skipping; assert_subscribed_groups logs the seat's last-seen subscription state on timeout; the settlement flow lives once in runShredSettlement with both tests as thin wrappers; and the parent-level nil-device guard converts a selector skip into a parent skip instead of a nil-pointer panic. CI on this head shows 30 checks passed with only e2e shards pending and no failures.

elitegreg:
confirmed working in testnet

PR #4080, opened 2026-07-23
serviceability: change user feed_pk to vec of feed_pks

nikw9944:
## Summary of Changes
* `User` accounts can now hold EdgeSeat feed seats on multiple feeds: `feed_pks: Vec<Pubkey>` records every feed whose per-feed seat the user consumed at connect, and delete releases all of them (via the new `User::release_feed_seats`).
* The generalized-payments model lets one connection use groups from several feeds (one feed per metro, multiple feeds per pass), so a user must hold and release seats on multiple feeds — not just one.
* The vec **replaces the scalar `feed_pk` in place** (no deprecated placeholder slot). This is safe because the slot was never written with a real feed on any cluster: no EdgeSeat access pass has ever existed (verified against the mainnet-beta data lake — the full `dim_dz_access_passes_history` contains only `prepaid` and `solana_validator`), and the write path only triggers on EdgeSeat multicast connect. Existing accounts carry 32 zero bytes there, which parse as an empty vec with the leftover zero bytes ignored as trailing data; accounts predating the slot default to empty as before. Decoders updated in the Go/TypeScript/Python SDKs; fixtures regenerated.
* `AIRDROP_USER_RENT_LAMPORTS_BYTES` resized for the 32-byte-smaller account (302 bytes with pub+sub+feed, ×3 accounts).
* Post-activation multi-feed *ticking* stays out of scope (deferred to doublezero#1699 / infra#1700); this change is the enabler that gives that work the state it needs.

## Testing Verification
* New `state/user.rs` unit tests: `release_feed_seats` multi-feed release and unicast no-op, plus an old-layout compat test proving a 32-zero-byte scalar slot reads as an empty `feed_pks` with trailing bytes ignored. `test_state_compatibility_user` (historical base64 accounts) still passes unchanged.
* The same old-layout zero-slot compat test is mirrored in the Go, TypeScript, and Python SDK fixture suites; backward-compat truncation tests updated for the new tail size (93 bytes).
* `feed_metro_gate_test`: connect records the ticked feed in `feed_pks`.
* Go `TestFixtureUser` decodes the shared `user.bin` fixture and asserts the populated `FeedPks` (and surrounding field order) against `user.json`, matching the TS/Python fixture suites.
* Go `ReadPubkeySlice` regression test for a `uint32`-overflow length prefix (hardened the bound check to `uint64`, mirroring the Rust `deserialize_vec_with_capacity` cap), since `feed_pks` is a new consumer of that reader; the same guard now also covers `ReadIPv4Slice`/`ReadNetworkV4Slice`. `ReadPubkeySlice` distinguishes a present-but-empty vec (empty non-nil slice) from an absent field (nil), with tests for empty/absent/truncated.


nikw9944 (commented):
Two Medium items, both actionable without restructuring. (1) The `AIRDROP_USER_RENT_LAMPORTS_BYTES` comment's 302-byte "largest single create" claim misses the publisher+subscriber+feed case on an EdgeSeat pass, which writes 334 bytes; the pre-PR constant exactly covered the true worst case, the new one no longer does. (2) The Go SDK has no fixture test decoding a User account, so the new `FeedPks` read is never exercised against `user.bin` (TS and Python are), and the PR description's testing claim about Go fixture assertions is inaccurate. The core change is otherwise sound: append-only borsh layout with correct old-account defaults, double-release guarded by the `contains` check, and delete parity preserved.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/mod.rs
```diff
@@ -9,8 +9,10 @@ use solana_program::{
 };
 
 // Value to rent exempt three `User` accounts + configurable amount for connect/disconnect txns.
-// `User` account size assumes a single publisher and subscriber pubkey registered (298 bytes each).
-pub const AIRDROP_USER_RENT_LAMPORTS_BYTES: usize = 298 * 3; // 298 bytes per User account x 3 accounts = 894 bytes
+// 302 bytes is the largest account a single create writes: base (238, empty vecs) + 32 for one
+// multicast group + 32 for one feed seat in `feed_pks` — which also equals the publisher+subscriber
+// worst case. Accounts that later grow more lists/feeds are rent-topped-up on rewrite (try_acc_write).
+pub const AIRDROP_USER_RENT_LAMPORTS_BYTES: usize = 302 * 3; // 302 bytes per User account x 3 accounts = 906 bytes
```
nikw9944, line 16: The 302-byte "largest account a single create writes" claim misses a reachable case: `process_create_subscribe_user` with `publisher: true, subscriber: true` pushes the group into both `publishers` and `subscribers` (EdgeSeat passes skip the allowlist checks in `update_user_multicastgroup_roles`), and on an EdgeSeat pass the same create also ticks a feed — base 238 + 32 + 32 + 32 = 334 bytes. The pre-PR constant (298) exactly covered the true worst case because the feed slot was in the fixed base; 302 now undershoots by 32 bytes/account, and the size test in `accesspass/set.rs` enshrines the claim with an empty-`feed_pks` `user_with_both`. Suggest 334 × 3, or document (and test) why the both-roles-with-feed case is excluded.

smartcontract/sdk/go/serviceability/deserialize.go
```diff
@@ -310,9 +310,12 @@ func DeserializeUser(reader *ByteReader, user *User) {
...
 	user.BgpRttNs = reader.ReadU64()
-	// ReadPubkey returns the zero pubkey on EOF, so old accounts that predate FeedPk deserialize
-	// with it defaulted — matches the Rust append-only contract.
-	user.FeedPk = reader.ReadPubkey()
+	// ReadPubkey returns the zero pubkey on EOF, so old accounts that predate the legacy feed slot
+	// deserialize with it defaulted — matches the Rust append-only contract.
+	user.DeprecatedFeedPk = reader.ReadPubkey()
+	// ReadPubkeySlice returns nil on EOF, so accounts written before FeedPks deserialize to an
+	// empty slice — matches the Rust append-only contract.
+	user.FeedPks = reader.ReadPubkeySlice()
```
nikw9944, line 317: Nothing in the Go SDK decodes real serialized bytes with a populated `feed_pks`: `fixture_test.go` has no `TestFixtureUser` and `deserialize_test.go` doesn't cover `DeserializeUser`, so a field-order regression here (e.g. reading `FeedPks` before `DeprecatedFeedPk`) would pass CI. TS and Python both assert the shared `user.bin` fixture, and the PR description says Go does too. Suggest adding a `TestFixtureUser` (the `loadFixture` helper already exists) asserting `DeprecatedFeedPk`, `FeedPksLen`, and `FeedPks{0,1}` against `user.json`.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/mod.rs
nikw9944: The 334 math applied to the earlier layout that kept a 32-byte deprecated scalar slot alongside the vec (base 238). The reworked layout on this branch replaces the scalar `feed_pk` slot in place, so the base with empty vecs is 206 and the both-roles-with-feed create — which is included, not excluded — serializes to 206 + 3×32 = 302. `test_airdrop_user_rent_lamports_bytes_covers_user_sizes` asserts exactly that case: `user_with_both` carries 1 publisher + 1 subscriber + 1 entry in `feed_pks` and must be 302 bytes, and the constant must equal 302×3. The doc comment above the constant had stale 334/238 numbers left over from the earlier layout; fixed to 302/206 in e8175be9.

smartcontract/sdk/go/serviceability/deserialize.go
nikw9944: Added `TestFixtureUser` (fixture_test.go) via the existing `loadFixture` helper: it decodes the shared `user.bin` and asserts `FeedPksLen` (2) and `FeedPks{0,1}` against `user.json`, plus the surrounding fields (publishers/subscribers/validator/bgp fields) so a field-order regression in `DeserializeUser` fails the test. There is no `DeprecatedFeedPk` in the shipped layout — the vec occupies the former scalar slot in place — so the fixture's trailing `FeedPks` entries are what pin the position. The PR description's Go testing claim now matches. On the current branch head (5d1ff7a6).

smartcontract/sdk/go/serviceability/bytereader.go
```diff
@@ -111,7 +111,9 @@ func (br *ByteReader) ReadPubkey() [32]byte {
 
 func (br *ByteReader) ReadPubkeySlice() [][32]byte {
 	length := br.ReadU32()
-	if length == 0 || (length*32) > br.Remaining() {
+	// Compare in uint64 so a garbage length >= 2^27 cannot wrap length*32 back under
+	// Remaining() and reach a multi-GiB make() — mirrors the Rust deserialize_vec_with_capacity cap.
+	if length == 0 || uint64(length)*32 > uint64(br.Remaining()) {
 		return nil
```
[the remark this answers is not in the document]
nikw9944: Fixed in e8175be9: `ReadPubkeySlice` now returns an empty non-nil slice for a present length prefix of 0, and nil only when the field is absent (EOF or truncated prefix) or the length is corrupt. `TestReadPubkeySliceEmptyVsAbsent` covers empty/absent/truncated, and the `GetProgramData` expectations were updated for the present-but-empty vectors. Left `ReadIPv4Slice`/`ReadNetworkV4Slice` collapsing empty to nil since no caller needs the distinction there (they did get the same uint64 length-overflow guard in 5d1ff7a6).

nikw9944:
Addressed the review feedback and rebased onto main (f1ce14f2 carries the original change; no conflicts).

* **Rent-bytes worst case (@nikw9944):** the 334 figure belonged to the earlier deprecated-slot layout; with the in-place vec the worst-case single create (pub + sub + feed) is 302 bytes, which the constant and the `user_with_both` size test already assert. Fixed the stale 334/238 prose in the constant's comment — e8175be9. Replied in-thread.
* **Go fixture coverage (@nikw9944):** `TestFixtureUser` now decodes `user.bin` and asserts `FeedPks{0,1}` and field order against `user.json`, mirroring TS/Python. PR description updated to match.
* **Empty vs absent vec (@Copilot):** `ReadPubkeySlice` returns an empty non-nil slice for a present zero-length prefix and nil only for absent/truncated input, with an empty/absent/truncated test — e8175be9.
* **Self-review extra:** `ReadIPv4Slice`/`ReadNetworkV4Slice` had the uint32-wrap allocation hole `ReadPubkeySlice` guards against; gave them the same uint64 comparison plus wrap-regression tests — 5d1ff7a6.

Not changed: the IPv4/NetworkV4 slice readers still collapse empty to nil — no caller needs the empty-vs-absent distinction for those types, so aligning them would be change for its own sake.

Verified locally: `go test ./serviceability/...` (Go SDK), the serviceability program size test, and the unit suites of downstream SDK consumers (client, controlplane, telemetry) — the only failures are pre-existing environment ones (raw-socket and Docker-dependent tests).

elitegreg (approved):

smartcontract/programs/doublezero-serviceability/src/state/user.rs
```diff
@@ -282,17 +282,22 @@ pub struct User {
...
-    /// `Pubkey::default()` for non-EdgeSeat/unicast users. Used to release exactly that seat at
-    /// delete, so the release is bound to what was ticked rather than a caller-supplied account.
+    /// The EdgeSeat `Feed`s whose per-feed seats this user consumed at connect (multicast only).
+    /// Empty for non-EdgeSeat/unicast users. Every entry is released at delete, so the release is
+    /// bound to exactly what was ticked rather than a caller-supplied account. A user may hold
+    /// seats on multiple feeds (one feed per metro; multiple feeds per pass).
+    ///
+    /// Replaces the former 32-byte `feed_pk` slot at the same serialized position. Safe because no
+    /// account ever recorded a feed there (no EdgeSeat pass has existed on any cluster): existing
+    /// accounts carry 32 zero bytes here, which parse as an empty vec plus ignored trailing bytes.
```
karl-dz: Can probably remove this comment. Preserving the history of this struct member in this docstring comment doesn't seem necessary

nikw9944: Removed in e45164a1. The compat rationale still lives where it's load-bearing: the deserialize path comment and the old-layout zero-slot tests (Rust `test_old_layout_zero_feed_slot_reads_as_empty_vec` and the Go/TS/Python fixture mirrors).

nikw9944:
Two follow-ups pushed:

* e45164a1 drops the layout-history paragraph from the `feed_pks` doc comment (per @karl-dz — replied in-thread).
* ffca0c16 adds `impl Default for User` (same pattern as `Device`/`GlobalState`) and ends the CLI/SDK/daemon-cli test literals with `..Default::default()` instead of naming `feed_pks`, so the next `User` field change won't touch those tests. `cargo test` and clippy pass on all four crates.

PR #4081, opened 2026-07-24
controller: dampen transient account fetch error noise, retry truncated RPC bodies

ben-dz:
## Summary

- Escalate the controller's on-chain account fetch failures to `ERROR` only when failures are *sustained*; transient blips that self-heal now log at `WARN`, so a single flaky poll no longer trips the generic ERROR-level Loki alert.
- Classify truncated/partial RPC response bodies (`unexpected end of JSON input`, `unexpected EOF`) as retryable in the JSON-RPC retry layer, so a cut-off 200 response is retried in-call instead of failing immediately.

## Why

Two `error fetching accounts` ERRORs fired overnight and both self-healed on the next 10s poll tick with no operational impact. They paged because the controller logged every failed fetch at `ERROR` and the alert is a generic catch-all on any ERROR-level line from `doublezero-controller.service`. One of the two errors was a truncated response body (`status code: 200 … unexpected end of JSON input`), which the retry classifier treated as non-retryable, so it got zero transport retries.

No infra alert-rule change is needed: once transient failures log at `WARN`, the existing ERROR catch-all behaves correctly.

## How it works

**Fetch-loop severity** (`controlplane/controller/internal/controller/server.go`): a weighted failure score, `+1` per failed fetch and `-0.5` per success, floored at 0, celinged at `cacheFetchErrorThreshold`. At/above `cacheFetchErrorThreshold` (6.0) the fetch logs `ERROR`, otherwise `WARN`. Both the startup fetch and every tick run through one `recordFetch` closure. At the 10s poll interval this means:

- 6 consecutive failures (~60s) → `ERROR`.
- A lone blip decays back to 0 after two good ticks → never pages.
- Something like a sustained every-other-tick failure still escalates (crosses 6.0 at the 11th failure, ~3.7 min) → `ERROR`, so a persistently flaky endpoint is not silently masked.

**Retry classification** (`tools/solana/pkg/jsonrpc/retry.go`): the truncated-body match is placed ahead of the existing `*json.SyntaxError` non-retryable check, so a truncated body is retried while genuinely malformed but complete JSON (e.g. `invalid character …`) stays non-retryable.

## Testing Verification

- `TestController_NextFetchFailScore`: 6 consecutive failures reach the threshold; a blip followed by two successes decays to 0 (floored, never negative); the alternating fail/success pattern crosses the threshold at the 11th failure.
- `TestTools_Solana_JSONRPC_IsRetryableJSONRPC`: added cases for a truncated-body message and a real truncated `json.Unmarshal` decode (both retryable), plus a malformed-but-complete JSON decode that stays non-retryable.


tools/solana/pkg/jsonrpc/retry.go
```diff
@@ -131,7 +131,13 @@ func isRetryableJSONRPC(err error) bool {
 	if strings.Contains(msg, "connection reset by peer") ||
 		strings.Contains(msg, "broken pipe") ||
 		strings.Contains(msg, "use of closed network connection") ||
-		strings.Contains(msg, "rate limited") {
+		strings.Contains(msg, "rate limited") ||
+		// Truncated/partial response body (e.g. a 200 whose body is cut off
+		// mid-stream). Matched before the *json.SyntaxError check below so a
+		// truncated body is retried while genuinely malformed but complete JSON
+		// (e.g. "invalid character ...") stays non-retryable.
+		strings.Contains(msg, "unexpected end of json input") ||
```
nikw9944, line 141: Truncation coverage is partial for the real decoder, and the new tests exercise a decoder the client never uses. gagliardetto solana-go v1.21.0 `rpc/jsonrpc` decodes bodies with goccy/go-json (streaming, `DisallowUnknownFields`+`UseNumber`), not `encoding/json`. Empirically decoding a representative RPC response truncated at every one of its 113 byte positions through that configuration: 102 positions yield `... unexpected end of JSON input` (matched, retried), but 11 yield `invalid character '\x00' looking for beginning of value`, `expected colon after object key`, or `expected comma after object element` — none matched, so those truncated bodies still fail with zero in-call retries. goccy also swallows reader-level `io.ErrUnexpectedEOF` (it resurfaces as those same syntax-style messages), so typed `errors.Is` matching can't close the gap — the substring approach is right, just incomplete. The `truncatedJSONErr`/`malformedJSONErr` fixtures in retry_test.go build errors with stdlib `encoding/json`, which this client never produces; they pass only because stdlib and goccy share the dominant message. Suggested fix: also match goccy's `invalid character '\x00' looking for beginning of value` (a raw NUL can't appear in valid JSON, so it's distinctive of goccy's truncated buffer), and generate the truncated-body fixture through a goccy decoder mirroring the solana-go setup. The controller-side WARN dampening still keeps the unmatched positions from paging, so this is a completeness gap, not a paging regression.

controlplane/controller/internal/controller/server.go
```diff
@@ -40,6 +40,14 @@ const (
...
 	bgpCommunityMinValid = 10000
 	bgpCommunityMaxValid = 10999
+
+	// cacheFetchErrorThreshold is the weighted failure score at which on-chain
+	// fetch failures escalate from WARN to ERROR. Each failed fetch adds 1 and
+	// each success subtracts 0.5 (floored at 0), so at the 10s poll interval a
+	// lone blip that recovers decays back to 0 and stays WARN, while a
+	// persistently flaky endpoint — even one failing every other tick — still
+	// crosses the threshold and pages.
+	cacheFetchErrorThreshold = 6.0
```
nikw9944, line 50: Escalation counts failures, not time, so a slow-failing outage delays the first ERROR by ~6x. The controller builds its RPC client with `NewWithRetries(endpoint, nil)` (default 4 attempts) over an HTTP client with a 5-minute `Timeout` (tools/solana/pkg/rpc/retry.go:16), and `updateStateCache` runs on the process-lifetime ctx with no per-call deadline. An endpoint that blackholes traffic (accepts connections, never responds — a common LB failure mode) makes each failed fetch take ~20 minutes; the ticker drops ticks while the fetch blocks, so the score climbs at 1 per fetch, and the 6th failure / first ERROR arrives ~2 hours into the outage versus ~20 minutes pre-PR. The "6 consecutive failures ≈ 60s" calibration only holds for fast-failing errors (DNS, connection refused). Suggested fix: wrap each tick's fetch in `context.WithTimeout` (30–60s — a 10s poll loop has no business waiting 20 minutes), which also restores the tick≈failure calibration; alternatively escalate on failure-streak duration rather than count.

tools/solana/pkg/jsonrpc/retry.go
```diff
@@ -131,7 +131,13 @@ func isRetryableJSONRPC(err error) bool {
 	if strings.Contains(msg, "connection reset by peer") ||
 		strings.Contains(msg, "broken pipe") ||
 		strings.Contains(msg, "use of closed network connection") ||
-		strings.Contains(msg, "rate limited") {
+		strings.Contains(msg, "rate limited") ||
+		// Truncated/partial response body (e.g. a 200 whose body is cut off
+		// mid-stream). Matched before the *json.SyntaxError check below so a
```
nikw9944: The ordering claim in this comment is misleading: the `*json.SyntaxError` check below returns `false` exactly like the function's final fallthrough, so being "matched before" it is not load-bearing — and the goccy-based client never produces `*encoding/json.SyntaxError`, so that check is doubly dead. Reword to "complete-but-malformed JSON falls through to the default non-retryable" (or delete the dead check along with the ordering claim).

controlplane/controller/internal/controller/server.go
```diff
@@ -40,6 +40,14 @@ const (
 
 	bgpCommunityMinValid = 10000
 	bgpCommunityMaxValid = 10999
+
+	// cacheFetchErrorThreshold is the weighted failure score at which on-chain
```
nikw9944: New text uses "on-chain"; repo CLAUDE.md: 'Use "onchain" (one word, no hyphen), never "on-chain"'. Three occurrences in text added by this PR: server.go:44, server.go:739, CHANGELOG.md:14.

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ All notable changes to this project will be documented in this file.
 
 - Serviceability
   - Bound the preallocation in `deserialize_vec_with_capacity` against the remaining input. A garbage or attacker-controlled u32 length prefix in an account (e.g. a pre-FeedSeat SDK misparsing an EdgeSeat AccessPass) could request tens of GiB via `Vec::with_capacity`, aborting the process through the uncatchable alloc-error handler; the capacity is now capped at the remaining byte count. Decoding of valid accounts is unchanged. (#4072)
+- Device controller
+  - Escalate on-chain account fetch failures to `ERROR` only when sustained; a transient blip that recovers on the next poll now logs at `WARN`, so a single flaky fetch no longer pages via the generic ERROR-level alert. A weighted score (+1 per failure, -0.5 per success, capped at 0) crosses the threshold on a persistently failing endpoint, so real outages still surface. (#4081)
```
nikw9944: "-0.5 per success, capped at 0" misstates the mechanism — the score is floored at 0 (and capped at the threshold). Change "capped at 0" to "floored at 0".

nikw9944 (commented):
Two medium findings worth addressing before or shortly after merge: the truncated-body retry classification only covers ~90% of real truncation positions because the solana-go client decodes with goccy/go-json (whose error messages differ from stdlib at some cut points, and the new test fixtures use stdlib `encoding/json`, a decoder the client never uses); and the WARN→ERROR escalation counts failures rather than time, so with the 5-minute HTTP client timeout and 4 retry attempts a blackholed endpoint delays the first ERROR from ~20 minutes to ~2 hours — bounding each fetch with a context timeout fixes both the delay and the tick≈failure calibration. Plus three small cleanups (misleading ordering comment in retry.go, "on-chain" vs the repo's "onchain" rule in new text, and a "capped at 0" → "floored at 0" changelog wording fix).

tools/solana/pkg/jsonrpc/retry.go
ben-dz: Good catch, and confirmed — fixed in e452613e. Verified solana-go decodes via `goccy/go-json` (`rpc/jsonrpc/jsonrpc.go`: `NewDecoder` + `DisallowUnknownFields` + `UseNumber`), then probed a representative RPC body truncated at all 174 byte positions: 114 yield `unexpected end of JSON input` (matched), 27 yield the NUL invalid-character (previously unmatched). Added a match for `invalid character '\x00'` (a raw NUL never appears in valid JSON, so it's distinctive of goccy's leftover decode buffer), and the `retry_test` fixtures now build errors through a goccy decoder mirroring the client config instead of stdlib `encoding/json`. The ambiguous `expected colon/comma after …` positions stay unmatched by design — they also occur in complete-but-malformed JSON — and the controller-side WARN dampening keeps them from paging.

controlplane/controller/internal/controller/server.go
ben-dz: Fixed in e452613e. Each state-cache update now runs under a 30s `context.WithTimeout`, so a blackholed endpoint fails the tick in 30s instead of ~20 min, and the ticker stops dropping ticks — restoring the ≈1-failure-per-tick calibration. Our own deadline surfaces as `context.DeadlineExceeded`, which the classifier already treats as non-retryable, so the in-flight attempt aborts promptly rather than burning all 4 retries.

tools/solana/pkg/jsonrpc/retry.go
ben-dz: Done in e452613e — deleted the `*json.SyntaxError` check (goccy never produces it, and it returned the same default `false` as the fallthrough) along with the now-unused `encoding/json` import, and rewrote the comment to drop the ordering claim.

controlplane/controller/internal/controller/server.go
ben-dz: Fixed in e452613e — the two occurrences this PR introduced (the const comment and the `nextFetchFailScore` doc) and the changelog now use `onchain`. Left the pre-existing log string and other pre-existing occurrences untouched to keep the diff scoped.

CHANGELOG.md
ben-dz: Fixed in e452613e — now reads "floored at 0, capped at 6".

ben-dz:
Thanks for the thorough review @nikw9944 — all five findings addressed in e452613e.

**Medium findings**
- **Truncation coverage / wrong decoder:** confirmed solana-go decodes with `goccy/go-json`, not stdlib. Probed a representative body truncated at every position — added a match for goccy's NUL invalid-character (the ~10% you flagged), and the `retry_test` fixtures now run through a goccy decoder mirroring the client config (`DisallowUnknownFields` + `UseNumber`). The ambiguous `expected colon/comma` positions are left unmatched on purpose, since they're indistinguishable from complete-but-malformed JSON.
- **Escalation counts failures, not time:** each fetch now runs under a 30s `context.WithTimeout`, so a blackholed endpoint fails the tick in 30s instead of ~20 min and the score climbs ≈1-per-tick again.

**Small cleanups**
- Dropped the dead `*json.SyntaxError` check + `encoding/json` import and the misleading ordering comment.
- `onchain` wording in the text this PR added, plus the changelog.
- Changelog now says "floored at 0, capped at 6".

`go.mod` gained one line — `goccy/go-json` moved from indirect to direct now that the test imports it.

nikw9944 (approved):
lgtm — all five findings from the previous round are verified fixed in e452613e (goccy NUL-truncation match with fixtures built through the real decoder, 30s per-fetch timeout restoring the escalation calibration, dead `*json.SyntaxError` check removed, terminology and changelog wording corrected); no new issues in the follow-up commit.

PR #4083, opened 2026-07-24
serviceability: add dzf_locked access-pass flag and set-flags instruction

elitegreg:
## Summary of Changes

- Add a `dzf_locked` flag to `AccessPass` (bit 0 of the `flags` byte) that marks a pass as foundation-managed, so automated reconcilers such as the Feed Oracle leave it alone instead of tearing it down. This gives the oracle an onchain signal to skip side-deal passes whose payment is settled out of band.
- New `SetAccessPassFlags` instruction (variant 116), gated on `ACCESS_PASS_ADMIN`, that sets/clears access-pass flags independently via `Option<bool>` per flag (`None` = leave, `Some(v)` = set/clear) and never clobbers the other bits. An all-`None` call is rejected.
- `SetAccessPass` now preserves the `dzf_locked` bit across unrelated updates (previously it rebuilt the whole `flags` byte from `allow_multiple_ip` on every call, which would silently clear the lock).
- CLI `doublezero access-pass dzf-lock` / `dzf-unlock` verbs and `access-pass set --dzf-locked` (applied as a second transaction on create), plus the SDK `SetAccessPassFlagsCommand`.

The feed authority (oracle) holds `ACCESS_PASS_ADMIN`, so the existing "feed authority may only touch passes it owns" guard is what prevents the oracle from clearing the lock on a pass it did not create.

Note: re-enabling the Feed Oracle's validator-owned teardown to actually honor this flag is a separate follow-up in the `doublezero-shreds` / feeds repo (and needs coordination on which users have side deals before deploying).

## Diff Breakdown

| Core logic   |     3 | +169 / -2   | +167 |
| Scaffolding  |    11 | +405 / -6   | +399 |
| Tests        |     1 | +227 / -0   | +227 |
| **Total**    |    16 | +802 / -8   | +794 |

Mostly wiring and tests around a small core: one new processor, a flag definition, and a one-line preservation fix.

<details>

- `smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs` — new `SetAccessPassFlags` processor: per-flag `Option<bool>` read-modify-write, `ACCESS_PASS_ADMIN` gate, feed-authority-owns guard, all-`None` rejection.
- `smartcontract/programs/doublezero-serviceability/src/state/accesspass.rs` — `DZF_LOCKED` constant, `dzf_locked()` accessor, `flags_string()`, and a flags unit test.
- `smartcontract/cli/src/accesspass/dzf_lock.rs` — `dzf-lock` / `dzf-unlock` CLI verbs over a shared helper.
- `smartcontract/cli/src/accesspass/set.rs` — `--dzf-locked` on create, fired as a second transaction after the pass exists.
- `smartcontract/sdk/rs/src/commands/accesspass/set_flags.rs` — `SetAccessPassFlagsCommand` instruction builder.
- `smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set.rs` — preserve the `DZF_LOCKED` bit across unrelated `SetAccessPass` updates.
- `smartcontract/programs/doublezero-serviceability/src/instructions.rs` — `SetAccessPassFlags` enum variant, unpack, name, and pack/unpack round-trip test.

</details>

## Testing Verification

- New processor integration test (`accesspass_dzf_locked.rs`): sets then clears `dzf_locked`; verifies the unrelated `ALLOW_MULTIPLE_IP` bit is preserved when only `dzf_locked` changes; verifies an unrelated `SetAccessPass` update preserves `DZF_LOCKED`; rejects an all-`None` call with `InvalidArgument`; rejects a signer without `ACCESS_PASS_ADMIN` with `NotAllowed`.
- Unit tests for `dzf_locked()` / `flags_string()`, the instruction pack/unpack round-trip, the SDK command's account layout, and the CLI `dzf-lock` / `dzf-unlock` verbs and the `--dzf-locked` create path (asserts the second transaction fires).
- `make rust-fmt`, `make rust-lint`, and the full serviceability + SDK + CLI test suites pass.

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/instructions.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/instructions.rs
@@ -2,6 +2,7 @@ use crate::processors::{
     accesspass::{
         check_status::CheckStatusAccessPassArgs, close::CloseAccessPassArgs,
         set::SetAccessPassArgs, set_feeds::SetAccessPassFeedsArgs,
+        set_flags::SetAccessPassFlagsArgs,
     },
     allowlist::{
         foundation::{add::AddFoundationAllowlistArgs, remove::RemoveFoundationAllowlistArgs},
@@ -251,6 +252,7 @@ pub enum DoubleZeroInstruction {
     UpdateFeed(FeedUpdateArgs),                 // variant 113
     DeleteFeed(FeedDeleteArgs),                 // variant 114
     SetAccessPassFeeds(SetAccessPassFeedsArgs), // variant 115
+    SetAccessPassFlags(SetAccessPassFlagsArgs), // variant 116
 }
 
 impl DoubleZeroInstruction {
@@ -397,6 +399,7 @@ impl DoubleZeroInstruction {
             113 => Ok(Self::UpdateFeed(FeedUpdateArgs::try_from(rest).unwrap())),
             114 => Ok(Self::DeleteFeed(FeedDeleteArgs::try_from(rest).unwrap())),
             115 => Ok(Self::SetAccessPassFeeds(SetAccessPassFeedsArgs::try_from(rest).unwrap())),
+            116 => Ok(Self::SetAccessPassFlags(SetAccessPassFlagsArgs::try_from(rest).unwrap())),
 
             _ => Err(ProgramError::InvalidInstructionData),
         }
@@ -546,6 +549,7 @@ impl DoubleZeroInstruction {
             Self::UpdateFeed(_) => "UpdateFeed".to_string(), // variant 113
             Self::DeleteFeed(_) => "DeleteFeed".to_string(), // variant 114
             Self::SetAccessPassFeeds(_) => "SetAccessPassFeeds".to_string(), // variant 115
+            Self::SetAccessPassFlags(_) => "SetAccessPassFlags".to_string(), // variant 116
         }
     }
 
@@ -687,6 +691,7 @@ impl DoubleZeroInstruction {
             Self::UpdateFeed(args) => format!("{args:?}"), // variant 113
             Self::DeleteFeed(args) => format!("{args:?}"), // variant 114
             Self::SetAccessPassFeeds(args) => format!("{args:?}"), // variant 115
+            Self::SetAccessPassFlags(args) => format!("{args:?}"), // variant 116
         }
     }
 }
@@ -1396,6 +1401,13 @@ mod tests {
                 }],
             }),
             "SetAccessPassFeeds",
+        );
+        test_instruction(
+            DoubleZeroInstruction::SetAccessPassFlags(SetAccessPassFlagsArgs {
+                allow_multiple_ip: None,
+                dzf_locked: Some(true),
+            }),
+            "SetAccessPassFlags",
         );
     }
 }
--- a/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set.rs
@@ -282,7 +282,7 @@ pub fn process_set_access_pass(
         };
         accesspass.last_access_epoch = value.last_access_epoch;
         // `flags` here only carries ALLOW_MULTIPLE_IP. The DZF_LOCKED bit is managed solely by
-        // the foundation-gated SetAccessPassFlags instruction, so preserve any existing lock
+        // the `ACCESS_PASS_ADMIN`-gated SetAccessPassFlags instruction, so preserve any existing lock
         // rather than clobbering it on an unrelated update.
         accesspass.flags = (accesspass.flags & DZF_LOCKED) | flags;
         accesspass.max_unicast_users = value.max_unicast_users;
--- a/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set.rs
@@ -6,7 +6,7 @@ use crate::{
     seeds::{SEED_ACCESS_PASS, SEED_PREFIX},
     serializer::{try_acc_create, try_acc_write},
     state::{
-        accesspass::{AccessPass, AccessPassStatus, AccessPassType, ALLOW_MULTIPLE_IP},
+        accesspass::{AccessPass, AccessPassStatus, AccessPassType, ALLOW_MULTIPLE_IP, DZF_LOCKED},
         accounttype::AccountType,
         globalstate::GlobalState,
         permission::permission_flags,
@@ -281,7 +281,10 @@ pub fn process_set_access_pass(
             _ => value.accesspass_type.clone(),
         };
         accesspass.last_access_epoch = value.last_access_epoch;
-        accesspass.flags = flags;
+        // `flags` here only carries ALLOW_MULTIPLE_IP. The DZF_LOCKED bit is managed solely by
+        // the foundation-gated SetAccessPassFlags instruction, so preserve any existing lock
+        // rather than clobbering it on an unrelated update.
+        accesspass.flags = (accesspass.flags & DZF_LOCKED) | flags;
         accesspass.max_unicast_users = value.max_unicast_users;
         accesspass.max_multicast_users = value.max_multicast_users;
 
--- a/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
@@ -41,8 +41,13 @@ pub fn process_set_access_pass_flags(
     let accesspass_account = next_account_info(accounts_iter)?;
     let globalstate_account = next_account_info(accounts_iter)?;
     let payer_account = next_account_info(accounts_iter)?;
-    let _system_program = next_account_info(accounts_iter)?;
+    let system_program = next_account_info(accounts_iter)?;
 
+    assert_eq!(
+        *system_program.unsigned_key(),
+        solana_system_interface::program::ID,
+        "Invalid System Program Account Owner"
+    );
     #[cfg(test)]
     msg!("process_set_access_pass_flags({:?})", value);
 
--- a/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
@@ -66,7 +66,7 @@ pub fn process_set_access_pass_flags(
 
     // Reject a no-op call rather than charging for a write that changes nothing.
     if value.allow_multiple_ip.is_none() && value.dzf_locked.is_none() {
-        msg!("SetAccessPassFlags requires at least one flag to set");
+        msg!("SetAccessPassFlags requires at least one flag to update");
         return Err(DoubleZeroError::InvalidArgument.into());
     }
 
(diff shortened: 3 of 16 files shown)


smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
```diff
@@ -0,0 +1,120 @@
...
+        globalstate_account.owner, program_id,
+        "Invalid GlobalState Account Owner"
+    );
+    assert!(
+        accesspass_account.is_writable,
+        "AccessPass Account is not writable"
+    );
+
+    // Reject a no-op call rather than charging for a write that changes nothing.
+    if value.allow_multiple_ip.is_none() && value.dzf_locked.is_none() {
```
karl-dz, line 73: We may want to do `value == Default::default()` in case we add more flags. But this is fine

karl-dz (approved):

juan-malbeclabs (commented):
## Review

Solid, convention-following change. The core is small (one new processor, a flag constant, a one-line preservation fix) and the scaffolding is mechanical. Splitting `dzf_locked` out of `SetAccessPass` into a separate `ACCESS_PASS_ADMIN`-gated instruction is the right call: `SetAccessPass` is also reachable by a non-privileged tenant admin, so keeping the flag in a gated instruction (plus the preservation mask) is what keeps the foundation lock admin-only.

### Permission-model compatibility (verified)
Compatible with **both** the current model and the not-yet-activated strict model:
- This is a new *instruction* on the *existing* `ACCESS_PASS_ADMIN` flag, not a new flag, so `AUTHORIZE_GATED_FLAGS` / `legacy_keys_for_flags` / `check_legacy_any` need no changes (and correctly get none). `ACCESS_PASS_ADMIN` is already in `AUTHORIZE_GATED_FLAGS`, so `permission audit` already models it.
- Legacy mode: authorizes via foundation ∪ sentinel ∪ feed-authority fallback — the historical accesspass authority.
- Strict mode (`RequirePermissionAccounts`): requires a Permission account granting `ACCESS_PASS_ADMIN`, exactly like the sibling instructions. The account layout lets the SDK conditionally append the Permission PDA and `authorize()` consume it.

**One forward-compat caveat** (see inline on `set_flags.rs`): the ownership guard that enforces the feature's core property is keyed on the legacy `globalstate.feed_authority_pk` field. It holds today and under strict mode only while the oracle's key stays in that field; once the oracle is migrated to authorize purely via a Permission account and the legacy field is cleared/reassigned, the guard silently no-ops. This should be documented and coordinated with the feed-authority → Permission migration.

### Also verified OK
- Bit 0 (`DZF_LOCKED = 1 << 0`) was never used historically, so all legacy passes read `dzf_locked = false` — no migration hazard.
- Variant 116 is the next free variant; pack/unpack/name/debug and the round-trip test are all updated.
- No realloc / vec growth (`flags` is a fixed `u8`); `AccessPass::validate()` only checks `account_type` + `accesspass_type`, which this instruction never touches, so it can't spuriously revert a valid pass.
- Owner/writable/signer/`data_is_empty` checks match the `set_feeds` sibling; error codes (`NotAllowed`=8, `InvalidArgument`=65, `AccessPassNotFound`=53) match the test.

The inline comments mark the places worth a change. None is a hard merge blocker on correctness; the feed-authority-guard item is the one I'd want addressed (at least documented + tested) before the oracle's authority is migrated off the legacy field, since it's the exact property the feature exists to guarantee.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
```diff
@@ -0,0 +1,120 @@
...
+        &globalstate,
+        permission_flags::ACCESS_PASS_ADMIN,
+    )?;
+
+    let mut accesspass = AccessPass::try_from(accesspass_account)?;
+
+    // Mirror the sibling accesspass handlers: the feed authority may only touch passes it owns.
+    // This is what stops the oracle (which holds ACCESS_PASS_ADMIN) from clearing dzf_locked on a
+    // pass it did not create.
+    if globalstate.feed_authority_pk == *payer_account.key && accesspass.owner != *payer_account.key
```
juan-malbeclabs, line 94: This guard is the feature's core security boundary (oracle can't clear `dzf_locked` on passes it doesn't own), but it keys on the legacy `globalstate.feed_authority_pk` field, which is orthogonal to the Permission model.

It works today and under `RequirePermissionAccounts` **only while the oracle's key stays in `feed_authority_pk`**. Once the oracle is migrated to authorize via a Permission account granting `ACCESS_PASS_ADMIN` and this legacy field is cleared/reassigned, the guard silently becomes a no-op and the oracle regains the ability to clear the lock on any pass — defeating the feature.

Suggest at minimum a comment noting this dependency and coordinating with the feed-authority → Permission migration. A more durable formulation would fence *any* non-foundation/sentinel `ACCESS_PASS_ADMIN` caller to passes they own, rather than singling out the legacy feed key.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set.rs
```diff
@@ -281,7 +281,10 @@ pub fn process_set_access_pass(
             _ => value.accesspass_type.clone(),
         };
         accesspass.last_access_epoch = value.last_access_epoch;
-        accesspass.flags = flags;
+        // `flags` here only carries ALLOW_MULTIPLE_IP. The DZF_LOCKED bit is managed solely by
+        // the `ACCESS_PASS_ADMIN`-gated SetAccessPassFlags instruction, so preserve any existing lock
+        // rather than clobbering it on an unrelated update.
+        accesspass.flags = (accesspass.flags & DZF_LOCKED) | flags;
```
juan-malbeclabs, line 287: This whitelists which bit to preserve (`DZF_LOCKED`) and clears every other bit not in `flags`. Correct today (only two bits), but the next flag added will be silently clobbered by `SetAccessPass` unless someone remembers to extend this mask.

Since `SetAccessPass` only owns `ALLOW_MULTIPLE_IP`, blacklisting what it owns is more future-proof — any bit managed elsewhere survives automatically:

```suggestion
        accesspass.flags = (accesspass.flags & !ALLOW_MULTIPLE_IP) | flags;
```

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
```diff
@@ -0,0 +1,120 @@
...
+    pubkey::Pubkey,
+};
+
+/// Surgically set or clear individual bits of an existing access pass's `flags` byte. Each field is
+/// tri-state: `None` leaves that flag untouched, `Some(true)` sets it, `Some(false)` clears it. Bits
+/// not named here are always preserved, so this never clobbers unrelated flags. Adding a future flag
+/// is just another `Option<bool>` field.
+///
+/// Gated on `ACCESS_PASS_ADMIN` (like the sibling SetAccessPass path).
+#[derive(BorshSerialize, BorshDeserializeIncremental, Debug, PartialEq, Clone, Default)]
```
juan-malbeclabs, line 26: `Default` is derived but never constructed anywhere. Deriving `Default` on an args struct is a known fragility: if a future flag field is added and the struct is built with `..Default::default()`, there's no compile error to catch the omission. Here `default()` is also exactly the all-`None` payload the processor rejects. Recommend dropping `Default`.

```suggestion
#[derive(BorshSerialize, BorshDeserializeIncremental, Debug, PartialEq, Clone)]
```

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
```diff
@@ -0,0 +1,120 @@
...
+pub fn process_set_access_pass_flags(
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    value: &SetAccessPassFlagsArgs,
+) -> ProgramResult {
+    let accounts_iter = &mut accounts.iter();
+
+    // Account layout: [accesspass, globalstate, payer, system, (permission)] — matches the sibling
+    // accesspass handlers, with authorize() consuming the trailing Permission account.
+    let accesspass_account = next_account_info(accounts_iter)?;
```
juan-malbeclabs, line 41: Both siblings (`set.rs`, `set_feeds.rs`) assert the accesspass account matches its derived PDA. This handler can't, because the args carry no `client_ip` / `user_payer` seeds. That's acceptable — the account is fully validated (program-owned + `AccessPass` discriminator) and an `ACCESS_PASS_ADMIN` caller can already target any pass — but please add a one-line comment recording that the missing PDA check is intentional, so nobody later "fixes" it by threading the seeds back into the args (which the argument-vs-account guidance discourages).

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
```diff
@@ -0,0 +1,120 @@
...
+    assert_eq!(
+        globalstate_account.owner, program_id,
+        "Invalid GlobalState Account Owner"
+    );
+    assert!(
+        accesspass_account.is_writable,
+        "AccessPass Account is not writable"
+    );
+
+    // Reject a no-op call rather than charging for a write that changes nothing.
```
juan-malbeclabs, line 72: Minor: the no-op rejection runs before `authorize()`, so an unauthenticated caller sending an all-`None` payload gets `InvalidArgument` before the auth check (leaks that the account parsed). Moving this below `authorize()` is marginally cleaner. Non-blocking.

smartcontract/programs/doublezero-serviceability/tests/accesspass_dzf_locked.rs
```diff
@@ -0,0 +1,227 @@
...
+        vec![
+            AccountMeta::new(accesspass_pubkey, false),
+            AccountMeta::new_readonly(globalstate_pubkey, false),
+        ],
+        &payer,
+    )
+    .await;
+    assert_custom_error(result, CODE_INVALID_ARGUMENT, "all-None SetAccessPassFlags");
+
+    // A signer without ACCESS_PASS_ADMIN (not foundation/sentinel/feed, no Permission) is rejected.
```
juan-malbeclabs, line 200: This exercises a random outsider that fails `authorize()` outright — a different branch than the feed-authority-owns guard the PR relies on. That guard (the one flagged inline in `set_flags.rs`) is currently untested. Please add a case: set `globalstate.feed_authority_pk` to a signer, create a pass owned by someone else, then have the feed authority attempt `dzf_locked` and assert `NotAllowed`.

Also worth adding (cheap, since this is a new instruction): a `RequirePermissionAccounts`-on case proving the grant path succeeds with a Permission account and is denied without one.

smartcontract/programs/doublezero-serviceability/src/instructions.rs
```diff
@@ -1397,5 +1402,12 @@ mod tests {
             }),
             "SetAccessPassFeeds",
         );
+        test_instruction(
+            DoubleZeroInstruction::SetAccessPassFlags(SetAccessPassFlagsArgs {
```
juan-malbeclabs, line 1406: The round-trip only covers `{ allow_multiple_ip: None, dzf_locked: Some(true) }`. Given the `BorshDeserializeIncremental` derive, please also cover both-`Some`, all-`None`, and an incremental-truncation case (an older payload carrying only the first field) so the tri-state wire format and forward-compat defaulting are pinned.

PR #4085, opened 2026-07-26
e2e/qa: verify the shred seat price against chain before paying

ben-dz:
## Summary

- `TestQA_MulticastSettlement/pay_for_seat` funded the escrow from the CLI's `shreds price` quote, so whenever that quote diverged from what the program charges the run died at the first payment with an opaque `invalid account data for instruction` and every downstream assertion (tunnel up, device assignment, withdraw, prorated refund) never executed. The probe now computes the charge from chain state and funds from that, so the lifecycle runs regardless of what the CLI quotes.
- The divergence is a real product bug: `shreds price` reported the metro price at `current_subscription_epoch` while `request_instant_seat_allocation` charges the price at `last_settled_epoch` (one ring entry back). Testnet `lax-dz001` was repriced 43 -> 10, so after the epoch rolled the quote read 10 and the program charged 43 prorated.
- Quote verification points at `instant_allocation_price`, the new field from doublezerofoundation/doublezero-offchain#405, **not** at `epoch_price`. 405 deliberately keeps `epoch_price` meaning what a recurring subscriber pays next epoch, so asserting the settled-epoch charge against it would fail at every future reprice forever and look like the fix never landed. The plumbing for the new field (proto, QA agent) is in this PR.
- A second assertion pins `epoch_price` against the chain's current-epoch price. That closes the other half of the invariant: repointing `epoch_price` at `last_settled_epoch` to "fix" the divergence would otherwise go green here while silently breaking recurring subscribers.
- Verification goes through the in-repo Go SDK rather than trusting the CLI. QA hosts install `doublezero-solana` from a version-pinned apt package (`doublezero_solana_version` in `malbeclabs/infra` `ansible/inventory/*/group_vars/all.yml`, `0.5.10-1` on both testnet and mainnet-beta), so a CLI-side field reaches QA only after a release is published and the pin bumped. Reading chain state also gives QA an oracle independent of the CLI under test, which is the only reason a probe can catch a quote/charge divergence at all.
- Standalone SDK correctness fix, worth having on its own: `sdk/shreds/go`'s `ExecutionController` was 144 bytes against a 152-byte onchain body — `total_devices` was decoded as padding and `last_settled_epoch` was missing entirely. `deserializeAccount` tolerates trailing bytes for forward compatibility, so the truncation was silent.

### Rollout behavior

The quote check is gated on field presence, three ways:

| CLI reports | Probe |
| --- | --- |
| key absent (pinned 0.5.10-1 today) | skip, naming the version gap |
| explicit `null` | fail — the CLI could not find the settled-epoch ring entry, which means the program would reject the allocation too |
| a number | assert against chain |

So the subtest is green (skipped) until the CLI rolls out, then starts asserting. That is the right shape: this PR's real job — funding from chain so the lifecycle runs — works the moment it merges, and the quote check activates once there is a quote worth checking.

Carrying all three states end to end needs one bit more than a `*uint64`, which collapses absent and null into `nil`. The agent decodes the key as `json.RawMessage` and sets a companion `reports_instant_allocation_price` bool on the wire.

## Testing Verification

- Ring-buffer `Find` unit tests cover an exact hit, wraparound past index 0, the `TotalCount` bound (an epoch parked in a slot beyond `TotalCount` must not match), `epoch == 0` against a zero-initialized buffer, and a miss. The `TotalCount` bound is the one that matters: scanning all 32 slots would spuriously match `epoch == 0` on unwritten entries.
- Table test for the metro-price + signed-premium combination: positive premium, negative premium (discount), and saturation at both ends including `premium == -32768`.
- `parseDevicePrices` table test pins all three presence states plus `instant_allocation_price: 0`, which must read as a real price rather than as absent or null.
- Decoded the new `ExecutionController` layout against live mainnet-beta account data via the opt-in compat test (`SHREDS_COMPAT_TEST=1`): `last_settled_epoch=1007` against `current_subscription_epoch=1008`, `total_devices=97` against `total_enabled_devices=95` — both cross-checked against raw byte reads at their offsets. The compat test now pins those two offsets and asserts `0 < LastSettledEpoch <= CurrentSubscriptionEpoch`, since a re-truncated struct would silently read zero for the trailing field rather than fail to deserialize.
- Checked the override-floor concern against live mainnet-beta state: of 607 client seats, 3 carry a price override and all 3 are set to 0 (comped seats), none above the device's base price at the settled epoch. Funding `base_at_last_settled` therefore sits at or above 405's `seat_price_override.unwrap_or(base_at_last_settled)` floor for every seat that exists, so the QA payment clears the new preflight. Could not run the same scan against the DZ testnet cluster from CI's network, so the evidence is mainnet-only.
- `epoch_price` vs current-epoch chain price is the one assertion not yet observed against a live CLI response — it assumes `shreds price` derives `epoch_price` from the ring entry stamped with `current_subscription_epoch`. If it instead reads the ring's newest entry, the two agree whenever prices have been updated for the epoch (all of `OpenForRequests`), and the probe skips when the current-epoch entry is missing mid-`UpdatingPrices`.
- The e2e suite was not run (it needs live testnet infrastructure and cEOS containers); both QA tests were verified to build under the `qa` tag.

### DIFF
--- a/e2e/qa_shred_settlement_test.go
+++ b/e2e/qa_shred_settlement_test.go
@@ -68,10 +68,10 @@ type shredSettlementParams struct {
 // runShredSettlement drives the full shred-pay settlement flow: pick a device,
 // query its seat price from the CLI and cross-check it against the price the
 // program will charge (read from chain), wait for the program's
-// open-for-requests phase, pay, verify the debit and that the multicast tunnel
-// comes up on the right device, (optionally) assert extra invariants, then
-// withdraw and verify the refund accounting. Both settlement QA tests are thin
-// wrappers over this.
+// open-for-requests phase, re-read that price and pay it, verify the debit and
+// that the multicast tunnel comes up on the right device, (optionally) assert
+// extra invariants, then withdraw and verify the refund accounting. Both
+// settlement QA tests are thin wrappers over this.
 func runShredSettlement(t *testing.T, p shredSettlementParams) {
 	if !p.enabled {
 		t.Skip(p.skipReason)
@@ -227,19 +227,15 @@ func runShredSettlement(t *testing.T, p shredSettlementParams) {
 	}
 
 	if !t.Run("query_onchain_seat_price", func(t *testing.T) {
-		// Read the prices the program itself computes straight off the chain. The
-		// escrow is funded from the instant-allocation figure rather than from the
-		// CLI quote, so a quote/charge divergence surfaces as a named assertion
-		// below instead of killing the run at pay time with a raw simulation dump
-		// — and everything downstream (tunnel up, device assignment, withdraw,
-		// prorated refund) still gets exercised.
+		// Read the prices the program itself computes straight off the chain, as
+		// an oracle for the CLI quote. This snapshot is taken next to the quote so
+		// the two comparisons below are as close to simultaneous as possible; the
+		// amount actually funded is re-read just before paying, since the wait for
+		// the open phase can outlive this read.
 		var err error
 		onchain, err = client.SeatPrices(ctx, device.PubKey)
 		require.NoError(t, err, "failed to compute the onchain seat prices")
 		require.NotZero(t, onchain.InstantAllocationDollars, "onchain instant-allocation price is zero for device %s", device.Code)
-
-		amount = strconv.FormatUint(onchain.InstantAllocationDollars, 10)
-		fundedAmount = onchain.InstantAllocationDollars * 1_000_000 // dollars to USDC raw units (6 decimals)
 
 		// The price assertion belongs here, not on the CLI quote: what a caller
 		// means by "this seat costs the discounted price" is what the program
@@ -247,7 +243,8 @@ func runShredSettlement(t *testing.T, p shredSettlementParams) {
 		if p.assertPrice != nil {
 			p.assertPrice(t, device, onchain.InstantAllocationDollars)
 		}
-		log.Info("Onchain seat prices", "device", device.Code, "amount", amount,
+		log.Info("Onchain seat prices", "device", device.Code,
+			"instant_allocation_price", onchain.InstantAllocationDollars,
 			"last_settled_epoch", onchain.LastSettledEpoch,
 			"current_epoch_price", onchain.CurrentEpochDollars,
 			"current_subscription_epoch", onchain.CurrentSubscriptionEpoch)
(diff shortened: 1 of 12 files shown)


ben-dz:
Before Merging this:
> Merge [offchain 405](https://github.com/doublezerofoundation/doublezero-offchain/pull/405).
>Cut and publish a doublezero-solana release containing it.
>Merge 4085 with the above. QA stops failing at pay time immediately — funding comes from chain, independent of the CLI. Quote subtest skips with a warning.
>Bump doublezero_solana_version in infra testnet (and mainnet-beta) group_vars, re-run the qa_agent role.

e2e/qa_shred_settlement_test.go
```diff
@@ -205,26 +210,100 @@ func runShredSettlement(t *testing.T, p shredSettlementParams) {
...
+		// CLI quote, so a quote/charge divergence surfaces as a named assertion
+		// below instead of killing the run at pay time with a raw simulation dump
+		// — and everything downstream (tunnel up, device assignment, withdraw,
+		// prorated refund) still gets exercised.
+		var err error
+		onchain, err = client.SeatPrices(ctx, device.PubKey)
+		require.NoError(t, err, "failed to compute the onchain seat prices")
+		require.NotZero(t, onchain.InstantAllocationDollars, "onchain instant-allocation price is zero for device %s", device.Code)
+
+		amount = strconv.FormatUint(onchain.InstantAllocationDollars, 10)
```
nikw9944: The funding amount is computed here, but `wait_for_open_phase` runs after it and can span a settlement boundary (the wait is up to 2 minutes, and the epoch-tail skip only covers waits that stay closed for the whole span). If settlement completes between this read and `pay_for_seat`, `last_settled_epoch` advances and the program charges the newer ring entry — at epoch start proration is ~1, so an upward reprice leaves the escrow underfunded and the run dies with the same opaque `invalid account data` failure this PR set out to eliminate. Minimal fix: after `WaitForOpenForRequests` succeeds, re-read `client.SeatPrices` and, if `LastSettledEpoch` moved, recompute `amount`/`fundedAmount` (and the value later fed to `GetEffectiveSeatPrice`) from the fresh read; keep the two CLI-vs-chain comparison subtests on the original snapshot so quote and chain are compared at consistent times.

nikw9944 (commented):
One improvement worth making: the escrow funding amount is read from `last_settled_epoch` before `wait_for_open_phase`, so a settlement completing during the wait can still underfund the escrow and reproduce the opaque pay-time failure this PR eliminates. The layout fix checks out against the offchain SDK's documented byte offsets (total_devices at 38, last_settled_epoch at 144 past the discriminator), and the failing `e2e (shard 4)` check is the known-flaky `TestE2E_MultiClientIBRL_RouteLiveness` (untouched by this PR; the sibling workflow's shard 4 passed) — a re-run should clear it.

- The failing `e2e (shard 4)` check is `TestE2E_MultiClientIBRL_RouteLiveness` — a file this PR does not touch, with a long flake history (#3950, #3270, #2895); the sibling workflow's shard 4 passed on the same head. A re-run should clear it.

ben-dz:
> the escrow funding amount is read from `last_settled_epoch` before `wait_for_open_phase`, so a settlement completing during the wait can still underfund the escrow

Good catch — fixed in 4908e11. The funded amount now comes from a `refresh_onchain_seat_price` subtest that runs after the open-phase wait, immediately before `record_balance_before_pay`.

The pre-wait read stays put as the oracle for the two quote comparisons, since those want a snapshot taken as close to the CLI quote as possible. When the refreshed price differs from it, the run logs a warning naming both prices and both settled epochs, so a quote-comparison failure can be read in the light of a rollover that happened after it.

A rollover between the refresh and the transaction landing is irreducible — any payer races it — but the window drops from up to two minutes to seconds.

The `e2e (shard 4)` failure: the push re-ran the full suite, so that should clear on its own.

nikw9944 (approved):
No changes since the previous round: the head is still 4908e11e (verified directly against the PR), with no new commits, replies, or comments. The single Medium finding from round 1 remains verified fixed in code at this head — escrow funding comes from a `refresh_onchain_seat_price` subtest that re-reads `client.SeatPrices` after `wait_for_open_phase` immediately before paying, reassigns `onchain` so `GetEffectiveSeatPrice` uses the refreshed value, keeps the pre-wait snapshot as the oracle for the CLI-vs-chain quote comparisons, and logs a warning when the price moved across the wait. All CI checks pass on this head, including e2e shard 4, whose earlier failure was the known-flaky TestE2E_MultiClientIBRL_RouteLiveness (untouched by this PR). Layout verification from round 1 stands: the new ExecutionController fields match the offchain SDK's documented byte offsets (total_devices at 38, last_settled_epoch at 144 past the discriminator).

- Round-1 finding M1 (escrow funded from a pre-wait `last_settled_epoch` read) remains verified addressed by 4908e11e: funding and the effective-price input come from a post-wait re-read, with a logged warning when the price moved during the wait.
- All CI checks pass on the current head; the earlier `e2e (shard 4)` failure was the known-flaky `TestE2E_MultiClientIBRL_RouteLiveness` (untouched by this PR) and cleared on re-run.

martinsander00 (approved):

PR #4086, opened 2026-07-26
client/liveness: bound route reconvergence after transient loss (#3935)

### LINKED ISSUE #3935: e2e: flaky TestE2E_MultiClientIBRL_RouteLiveness — route restoration 'Condition never satisfied'
## Summary

`TestE2E_MultiClientIBRL_RouteLiveness` intermittently fails with a route-convergence timeout. After the test blocks and then **unblocks** a path between IBRL clients, it waits (via `require.Eventually`) for connectivity to be restored, and that condition never becomes true within the timeout. Surfaced on an unrelated dependency-bump PR (#3924, `tabled` — a table-formatting crate with no connection to routing), so it is not caused by that change.

This is a **third, distinct e2e flake class**, separate from the multicast one (#3907) and the Docker network address-pool one (`TestE2E_BackwardCompatibility`, "Pool overlaps").

## Failing assertion

```
multi_client_ibrl_liveness_test.go:449:
    Error:    Condition never satisfied
    Test:     TestE2E_MultiClientIBRL_RouteLiveness
    Messages: pass %d: unblock c1: c1->c3 restored
    Error Trace:
        e2e/multi_client_ibrl_test.go:514
        e2e/multi_client_ibrl_liveness_test.go:402
        e2e/multi_client_ibrl_liveness_test.go:449
        e2e/multi_client_ibrl_liveness_test.go:249
--- FAIL: TestE2E_MultiClientIBRL_RouteLiveness (338.65s)
```

The condition that never satisfied is the **`c1->c3 restored`** check after un-blocking the `c1` path — i.e. the route did not reconverge in time.

## Diagnostic context

At failure, the affected client's diagnostic dump showed a healthy tunnel and BGP session:
```
doublezero status: tunnel doublezero0, "session_status":"BGP Session Up", user_type":"IBRL"
```
So the daemon/BGP session was up; what lagged was the specific route restoration between clients after the unblock.

## Evidence

- Run: https://github.com/malbeclabs/doublezero/actions/runs/27981405704 (dispatched `/run-e2e`, shard 2; job 82813640956)
- PR it surfaced on: #3924 (dependency bump, unrelated)
- Other "Error:" lines in the same shard log ("Contributor cannot be deleted", "not allowed", "Allocation failed, resource exhausted") are **expected** negative-path assertions from sibling tests, not part of this failure.

## To investigate

Open question for the investigator: is this purely a **timing flake** (the `Eventually` window at `multi_client_ibrl_liveness_test.go:402`/`:449` too tight for reconvergence under `-parallel=12` load), or a **real reconvergence gap** (route genuinely not restored after unblock)? The diagnostic dump (BGP up, but path not restored) leans toward a real-but-slow reconvergence worth confirming before just widening the timeout.

- Reproduce by running `TestE2E_MultiClientIBRL_RouteLiveness` repeatedly under shard load (it's in e2e shard 2).
- Inspect the block/unblock + restore poll logic at `e2e/multi_client_ibrl_liveness_test.go:249,402,449` and the helper at `e2e/multi_client_ibrl_test.go:514`.

## Acceptance criteria

- `TestE2E_MultiClientIBRL_RouteLiveness` passes reliably across repeated runs.
- If it's a timing issue: the restore-wait is hardened (appropriate timeout/interval) with a clear rationale.
- If it's a real reconvergence delay/bug: the product issue is fixed (or documented), not just the test.

ben-dz:
## Summary of Changes

* Route liveness now advertises an RX-driven state change immediately and re-arms the normal transmit cadence, instead of waiting out the transmit timer armed while the session was `Down` (exponentially backed off up to `-route-liveness-backoff-max`). Recovery from a transient loss costs one backed-off interval plus two round trips rather than three such intervals.
* Two defects, both in the scheduler: `HandleRx` never enqueued a TX on a state change, and `Scheduler.Run` slept until the deadline of the queue head it last observed with no wakeup path — so an event pushed earlier could not shorten that sleep. The second also delayed the first probe of a newly registered route by up to *another* peer's backed-off interval, withholding its kernel route for that whole time in active mode.
* The daemon logs its effective `minTxFloor`/`maxTxCeil`/`backoffMax` at startup. A wrong `backoffMax` was previously visible only as slow reconvergence, which is what made #3935 take a month of inference to narrow down.
* `TestE2E_MultiClientIBRL_RouteLiveness` now diagnoses its own timeouts: `requireEventuallyRoute` renders the pass number (its `pass %d` messages printed literally, visible in the issue and both recent CI failures) and dumps the failing client's `/routes` view, liveness counters and iptables INPUT counters instead of only `Condition never satisfied`.
* Fixes #3935

RFC: [`rfcs/rfc7-client-route-liveness.md`](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc7-client-route-liveness.md) — line 289 requires "resume normal cadence on first valid RX"; this brings the code up to that, no RFC change.

**Not** in scope, deliberately: the production `backoffMax` default (still 60s), widening the e2e `Eventually` window, or asserting the daemon's effective config from the harness.

### Honest status on the flake itself

This fixes a real, spec-mandated reconvergence gap, but I can't claim it closes #3935 on its own. The plan's evidence read the observed 63s restoration as proof of a 60s effective `backoffMax` (from the cumulative schedule 1+2+4+8+16+32). That arithmetic isn't conclusive: with `txMin`=1s and the harness's 3s cap the schedule is 1,2,3,3,3… which also reaches 63s cumulative. I verified the `-route-liveness-backoff-max=3s` plumbing end to end statically (`ClientSpec` → `DZ_CLIENT_EXTRA_ARGS` → entrypoint → flag → `ManagerConfig.BackoffMax` → `Session.backoffMax`) and it is intact, which points away from a dropped override and toward packets genuinely not flowing on the `c1↔c3` path for ~60s after the unblock — consistent with the other observation that c1's route was still absent 7.6s after c3 reached Up.

The startup log plus the debug logging and richer failure dump are what settle this on the next occurrence: probe gaps of ~60s mean the override isn't in effect; probes sent but never received means the cause sits below liveness. **This needs `/run-e2e` dispatches to confirm**

## Testing Verification

* **Two-manager flap regression test** (`faults_test.go`) over the existing `fakeUDPSwitch`: hard-loss, wait for the Down backoff to saturate, heal, then assert **zero** post-recovery `Up→Down` transitions over `2 × backoffMax`. It pins `backoffMax` = 20× `detectTime` in the test body rather than from `-backoff-max`, because the suite's flag defaults make those two equal and hide the window entirely. Reverting the fix produces **21 flaps per side**; with it, zero.
* **Each fix was verified to fail without itself**, not just to pass with it: removing the immediate TX, the `wake` nudge, or the pending-TX-slot takeover each fails its corresponding test. Both halves of the original fix are load-bearing — without the wakeup, `scheduleTxNow` has no observable effect at all.
* Manager-level tests for cadence resumption after recovery and for prompt probing of a route registered while another peer is backed off; scheduler-level tests for marker takeover, coalescing, and dead-session skip.
* Fault suite green at CI scale (`-clients=100 -stable-duration=10s`) across 5 runs, and the full package green under `-race` across 6 runs — these are the regression guard for a dataplane FSM change, and the soft-loss/flap cases have their own flake history (#2833, #2965, #3212). New tests stable over 20 `-race` iterations.
* Pre-existing unrelated failures in `internal/{latency,manager,services}` (raw sockets need `CAP_NET_RAW`) confirmed byte-identical against `origin/main` in a scratch worktree.


client/doublezerod/internal/liveness/manager.go
```diff
@@ -634,6 +640,20 @@ func (m *manager) HandleRx(ctrl *ControlPacket, peer Peer) {
...
+		// the Down->Init->Up handshake would stall for a full backed-off interval,
+		// taking recovery to ~3x that interval instead of one plus a couple of round
+		// trips. Session.HandleRx has already reset backoffFactor, but that only
+		// takes effect on the *next* scheduleTx, and scheduleTx cannot pull an
+		// already-queued event earlier.
+		//
+		// Note this does not shorten the *first* probe after the path recovers,
+		// which is still bounded by the backed-off interval; see rfcs/
+		// rfc7-client-route-liveness.md "resume normal cadence on first valid RX".
+		m.sched.scheduleTxNow(now, sess)
```
packethog, line 655: **should-fix:** this immediate TX has no rate floor, so a peer that keeps re-triggering a state change drives our transmit rate 1:1 with its receive rate, bypassing `MinTxFloor`/`MaxTxCeil` and the peer's advertised RX interval.

`Session.HandleRx` treats a repeating `State=Down` packet as a state change every time: from `Down`, a packet with `PeerDiscr=0` passes the discriminator guard, falls through to `case StateDown` and takes the `else` branch to `Init` (`changed=true`); the very next identical packet has `prev==Init` and hits `if ctrl.State == StateDown && (prev == StateUp || prev == StateInit)` -> `Down` (`changed=true`). The `upSince`/`detectTime` suppression above it only covers `prev == StateUp`, not `Init`, so the pair alternates indefinitely at one state change per received packet.

Measured on this branch against the base commit, with `MinTxFloor=TxMin=MaxTxCeil=BackoffMax=5s` and 500 identical `{State: Down, PeerDiscr: 0}` packets fed over ~500ms:

- base commit: 250 `Down->Init` + 250 `Init->Down` flips, **0** packets transmitted
- this branch: the same 250/250 flips, **256+** packets transmitted (the count is clipped by the harness's 256-slot buffer)

The oscillation itself is therefore pre-existing; the transmit response to it is new here.

The `!nextTxScheduled.After(now)` check in `scheduleTxNow` does not bound this. It only coalesces while an immediate TX is still queued, and the `wake` nudge added to `push` in this same PR makes `Run` drain that event almost immediately, collapsing the coalescing window to microseconds.

This does not require a hostile peer: a client whose inbound liveness traffic is blocked keeps emitting `Down`, which is the exact scenario this PR targets. Consider gating `scheduleTxNow` on a minimum spacing since the last transmit (derived from `txInterval()`), so a state change still short-circuits the backed-off deadline but cannot transmit faster than the configured floor.

Possibly related, offered as a lead rather than a conclusion: `TestClient_Liveness_Faults_SoftLoss_TimeoutAtHighLoss` failed 3 of 9 full fault-suite `-race` runs on this branch and 0 of 9 on the base commit, always on `non-degraded sessions must remain Up under high soft loss`. It passed 5/5 in isolation on both. That is precisely the workload where degraded pairs oscillate against each other over the shared `fakeUDPSwitch`, so the extra volume is a plausible cause, but the sample is too small to attribute it with confidence.

client/doublezerod/internal/liveness/scheduler.go
```diff
@@ -412,6 +435,43 @@ func (s *Scheduler) doTX(ctx context.Context, sess *Session) {
...
+// then go silent for up to backoffMax, while the peer's detect timer is only
+// detectMult x rxInterval — the peer times us out, withdraws the route, and the
+// session flaps. Taking over the marker means popping this event re-arms the cadence
+// at the new state's interval instead.
+//
+// Claiming the marker also coalesces: while an immediate TX is already pending there
+// is nothing to gain from queueing another, which keeps a burst of state-changing RX
+// packets from driving our TX rate 1:1.
+//
+// The displaced far-future event stays in the queue and costs one extra packet when
```
packethog: **should-fix:** the cost is bounded per takeover, but the takeover rate is not, and a displaced TX event is never rejected as stale.

The `eventTypeDetect` branch in `Run` drops events whose `when` no longer matches the session marker (`if !ev.when.Equal(ev.session.detectDeadline) { ...drop... }`) and counts them under the `stale` metric. The `eventTypeTX` branch has no equivalent check, so a displaced event still calls `doTX` when it pops — the "one extra packet" described here. `maybeDropOnOverflow` additionally exempts TX from `maxEvents` outright ("never drop TX"), so these orphans cannot be shed under queue pressure either.

Each `scheduleTxNow` takeover leaves exactly one orphan behind, so this is self-limiting only while the takeover rate is bounded — and per the comment on `manager.go:655` it is not. Under a sustained state-change stream the queue accumulates orphaned TX events alongside the extra packets.

Rejecting a TX event whose `when` no longer matches `nextTxScheduled`, mirroring the detect branch, would remove both the duplicate packet and the backlog, and would extend the existing `stale` metric to cover TX.

client/doublezerod/internal/liveness/scheduler_test.go
```diff
@@ -67,6 +67,73 @@ func TestClient_Liveness_Scheduler_TryExpireEnqueuesImmediateTX(t *testing.T) {
...
+		state:      StateDown,
+		alive:      true,
+		detectMult: 1,
+		minTxFloor: time.Millisecond,
+		peer:       &Peer{Interface: "eth0", LocalIP: "192.0.2.1"},
+	}
+
+	// A burst of state changes must not queue a TX each: the first pending immediate
+	// TX carries the latest state, so our TX rate stays decoupled from the RX rate.
+	s.scheduleTxNow(now, sess)
```
packethog, line 117: **nit:** this test cannot substantiate the property the comment above claims. It never starts `Scheduler.Run`, so nothing drains the queue between the three calls — the second and third bail only because the first event is still sitting in it.

In production `push` now nudges `Run` through `wake`, so that event is consumed almost immediately and the next state change takes the marker over again. The A/B measurement noted on `manager.go:655` shows the transmit rate does track the receive rate 1:1 once `Run` is live, so "our TX rate stays decoupled from the RX rate" does not hold outside this test's static queue.

The marker-takeover contract this pins is worth keeping. It is the rate-decoupling claim that needs either a rate floor in `scheduleTxNow` or a `Run`-driven test to back it up.

packethog (changes requested):
The reconvergence fix and the `wake` protocol itself look sound — no lost wakeup, timer-reset, deadlock or heap-invariant problem, and the new tests are load-bearing (removing `scheduleTxNow` from `HandleRx` fails both manager tests). Requesting changes on one issue: the immediate TX has no rate floor, so a repeating peer-`Down` stream drives transmits at the receive rate. Details and an A/B measurement inline.

client/doublezerod/internal/liveness/scheduler.go
```diff
@@ -412,6 +435,43 @@ func (s *Scheduler) doTX(ctx context.Context, sess *Session) {
...
+// is what Run clears before re-arming the periodic cadence, so leaving the old
+// (possibly heavily backed-off) deadline in place would make Run's scheduleTx bail
+// and starve the session: after a Down->Init->Up recovery we would advertise Up and
+// then go silent for up to backoffMax, while the peer's detect timer is only
+// detectMult x rxInterval — the peer times us out, withdraws the route, and the
+// session flaps. Taking over the marker means popping this event re-arms the cadence
+// at the new state's interval instead.
+//
+// Claiming the marker also coalesces: while an immediate TX is already pending there
+// is nothing to gain from queueing another, which keeps a burst of state-changing RX
```
nikw9944: This coalescing claim only holds while the immediate TX is still pending — with the new wake channel, Run pops it in sub-milliseconds and re-arms a future cadence marker, so the next state-flipping packet takes the marker over again. A sender alternating valid Down/non-Down packets (source-spoofed to the session peer; `PeerDiscr==0` is accepted at session.go:253) flips Init<->Down per packet, driving one immediate TX plus one orphaned heap event per flip. TX events are exempt from `maybeDropOnOverflow`, so sustained flipping at a few kpps can push the heap past `MaxEvents` (default 10240), at which point `scheduleDetect` drops detect events for *every* session, delaying failure detection client-wide until the flood subsides. Low severity — an on-path attacker could more cheaply just drop packets — but worth (a) softening this comment to state the actual window, and optionally (b) rate-limiting marker take-overs per session (e.g. skip `scheduleTxNow` if the last take-over was < `minTxFloor` ago).

nikw9944 (commented):
One low-severity finding; otherwise the scheduler marker/wake changes held up under adversarial interleaving analysis (marker starvation, missed wakes, duplicate cadence, deadlock, AdminDown races, and the tryExpire behavior change were all traced and refuted), and the new regression tests genuinely pin the failure modes they describe.

The one item worth addressing: the `scheduleTxNow` coalescing comment claims a remote peer cannot drive our TX rate 1:1, but that only holds while the immediate TX is still pending — sub-milliseconds now that the wake channel exists. Sustained state-flipping traffic can still drive one TX plus one orphaned heap event per packet, and TX events bypass the `MaxEvents` overflow cap, so a flood can push the heap past the cap and cause `scheduleDetect` to drop detect events for every session. Low severity (needs an on-path/spoofing sender at sustained kpps), but the comment should state the actual window, and a per-session take-over rate limit would close the edge.

client/doublezerod/internal/liveness/manager.go
ben-dz: Fixed in 3cbfe745. `scheduleTxNow` now clamps to `lastTx + txInterval()` instead of sending at `now`, with `lastTx` recorded in `doTX` (on attempt, not on success, so a failing socket can't be spun on either). Clamping rather than dropping keeps the point of the change: a state change still short-circuits a backed-off deadline, it just can't outrun the configured floor.

Your A/B reproduces exactly. I added `TestClient_Liveness_Manager_HandleRx_RateFloorBoundsStateChangeTX` — a live scheduler, 500 identical `{State: Down, PeerDiscr: 0}` packets over ~500ms with everything pinned to 5s. It asserts the oscillation actually happened (>125 flips) and then that TX ≤ 3. With the floor removed it fails at **sent 256**, the same buffer-clipped number you measured.

You were right about the lead, and it wasn't a small sample artifact. `TestClient_Liveness_Faults_SoftLoss_TimeoutAtHighLoss` is now **9 of 9** green on full fault-suite `-race` runs, against your 3-of-9 failures on the previous commit. The extra transmit volume was the cause.

Also confirmed your reading of the FSM: from `Down` a `PeerDiscr=0` packet promotes to `Init`, and the next identical one hits the `ctrl.State == StateDown && (prev == StateUp || prev == StateInit)` branch back to `Down`, because the `upSince`/`detectTime` suppression only guards `prev == StateUp`. I left that oscillation alone — it's pre-existing and orthogonal — and bounded the transmit response to it, which is what this PR introduced.

client/doublezerod/internal/liveness/scheduler.go
ben-dz: Fixed in 3cbfe745, exactly as you suggested. The `eventTypeTX` branch now mirrors the detect branch: an event whose `when` no longer matches `nextTxScheduled` is dropped and counted under the existing `stale` label, so a takeover emits no duplicate packet and leaves no orphan backlog.

I audited the safety of dropping before doing it, since a wrongly-dropped TX could wedge a session permanently. `nextTxScheduled` is written in exactly three places — `Run`'s clear, `scheduleTx`, and `scheduleTxNow` — and every marker write is paired with a push while every clear is followed by a re-arm, so a dropped event always has a live successor. The two deliberate non-re-arm paths are safe: `scheduleTx` bails only when `!alive`, and `Run` skips the re-arm for `AdminDown`, which wants no periodic TX. I also traced `AdminDownRoute`'s "at least one AdminDown packet promptly" path — its event matches the marker, so it still fires.

`TestClient_Liveness_Scheduler_Run_DropsStaleTX` covers it with a live `Run` over real UDP: a non-matching event must not transmit, and a matching one must, so the first assertion is the marker check rather than a wedged scheduler. It fails on the old code.

With the rate floor also in, the takeover rate is now bounded too, so both halves of your point are closed.

client/doublezerod/internal/liveness/scheduler_test.go
ben-dz: Correct, and the criticism generalizes: a static-queue test can't say anything about rate. Reframed in 3cbfe745.

The test is now `..._ScheduleTxNow_PacedByTransmitFloor` and pins only what a static queue can actually show — the floor arithmetic, i.e. the event lands at `lastTx + txInterval()` rather than `now`, and the marker follows it. The rate-decoupling claim moved to `TestClient_Liveness_Manager_HandleRx_RateFloorBoundsStateChangeTX`, which runs a live scheduler and is backed by a real floor rather than by a coalescing window that `wake` collapses. I left a pointer between the two so the split is obvious.

The marker-takeover contract you wanted kept is still pinned separately in `..._ScheduleTxNow_EnqueuesImmediateTX`.

ben-dz:
All three addressed in 3cbfe745, pushed. Thanks — the rate-floor finding was the right call, and the lead you offered as "too small a sample to attribute" turned out to be the same bug.

**Rate floor (`manager.go:655`).** `scheduleTxNow` clamps to `lastTx + txInterval()` instead of sending at `now`. Clamping rather than dropping keeps the reconvergence win while respecting the configured pacing. Your A/B reproduces exactly: with the floor removed my new test reports `sent 256` for 500 received packets, the same buffer-clipped figure you measured; with it, ≤3.

**Stale TX (`scheduler.go:454`).** `Run`'s TX branch now rejects events whose `when` no longer matches `nextTxScheduled` and counts them under the existing `stale` label, mirroring the detect branch. I audited every `nextTxScheduled` write first, since a wrongly-dropped TX would wedge a session: all three writers pair a marker write with a push, and every clear is followed by a re-arm, so a dropped event always has a live successor.

**Test nit (`scheduler_test.go:113`).** Reframed to pin the floor arithmetic, which is all a static queue can substantiate; rate decoupling moved to a live-scheduler test.

**On your soft-loss lead:** it was real, not sample noise. `TestClient_Liveness_Faults_SoftLoss_TimeoutAtHighLoss` is now **9/9** green on full fault-suite `-race` runs, versus your 3-of-9 failures on the previous commit and 0-of-9 on base. The extra transmit volume was the cause, so the rate floor closed it.

Each of the three fixes across this PR was verified by reverting it individually and confirming the corresponding test fails — including the two from the earlier round (marker takeover, `wake` nudge on all push sites). Also re-ran the fault suite at CI scale and the full package under `-race` three times.

One thing worth flagging that I have *not* changed, since it's adjacent and I can't e2e-verify it here: `AdminDownRoute` calls `scheduleTx` under a comment promising to "send at least one AdminDown packet promptly", but `scheduleTx` bails if a TX is already queued and otherwise schedules a full interval out — same defect class, and `scheduleTxNow` is now the right call there. Happy to fold it in if you'd prefer, otherwise I'll file it separately.

Still outstanding on the issue itself, unchanged from the PR description: this fixes a real reconvergence gap but I can't claim it closes #3935. The 63s observation doesn't uniquely imply a 60s effective `backoffMax` (with `txMin`=1s the harness's 3s cap gives 1,2,3,3,3… which also sums to 63s), and I verified the `3s` override's plumbing is intact end to end — which points toward packets genuinely not flowing after the unblock rather than a dropped override. The new startup log and the richer e2e dump are what settle that on the next occurrence, and it needs `/run-e2e` dispatches to confirm.

client/doublezerod/internal/liveness/scheduler.go
ben-dz: Both points are addressed in 3cbfe745, which landed for the same finding raised on `manager.go:655` — you and that review converged on this independently.

**(b)** is now in place, essentially as you specified. `scheduleTxNow` clamps the TX to `lastTx + txInterval()` rather than sending at `now`, and since `txInterval()` is itself clamped to `[minTxFloor, maxTxCeil]`, it is a per-session floor of at least `minTxFloor` on both transmits and marker takeovers. Your heap-growth path closes with it: takeovers now happen at most once per `txInterval` per session, so orphans can no longer accumulate at packet rate and push the heap toward `MaxEvents` — which was the part that would have degraded `scheduleDetect` for every other session. I also made `Run` drop TX events whose `when` no longer matches the marker (mirroring the detect branch), so the orphans that do occur are shed when they pop instead of transmitting.

Measured: 500 flip-inducing packets over ~500ms produced 256 transmits before the fix (clipped by the test recorder's buffer) and ≤3 after. `TestClient_Liveness_Manager_HandleRx_RateFloorBoundsStateChangeTX` pins it with a live scheduler.

**(a)** the comment is rewritten. It no longer claims coalescing bounds the rate — it now states the actual condition ("if a TX is already pending no later than the time we would pick") and puts the rate bound where it belongs, on the `txInterval()` pacing. It also notes that displaced events are dropped as stale rather than costing a packet.

Your `PeerDiscr==0` reading matches what I traced: from `Down` such a packet reaches `Init`, and the next one hits the `ctrl.State == StateDown && (prev == StateUp || prev == StateInit)` branch back to `Down`, because the stale-`Down` suppression only guards `prev == StateUp`. I left that oscillation as-is — pre-existing and orthogonal — and bounded the transmit response, which is what this PR introduced.

client/doublezerod/internal/liveness/scheduler_test.go
```diff
@@ -67,6 +67,162 @@ func TestClient_Liveness_Scheduler_TryExpireEnqueuesImmediateTX(t *testing.T) {
...
+}
+
+// TestClient_Liveness_Scheduler_Run_DropsStaleTX pins that a TX event whose time no
+// longer matches the session marker is rejected instead of transmitting, mirroring what
+// the detect branch already does.
+//
+// scheduleTxNow displaces the pending backed-off TX by taking the marker over. Firing
+// the displaced event would emit a duplicate packet, and because maybeDropOnOverflow
+// exempts TX from maxEvents these orphans cannot be shed under queue pressure either.
+func TestClient_Liveness_Scheduler_Run_DropsStaleTX(t *testing.T) {
```
packethog, line 147: **test-coverage:** this pins that a stale event does not transmit, but not the invariant the commit message calls out as this change's main risk — that a session survives its stale orphan *actually popping*. Nothing in the suite currently covers that.

The session constructed here has `localTxMin`/`minTxFloor` of an hour and a zero marker, so it has no cadence to lose, and the "stale" event is a fabricated marker-zero orphan rather than one displaced by a real `scheduleTxNow` takeover. The assertion it makes is the right one; it just cannot speak to the wedge case.

`..._HandleRx_ResumesCadenceAfterRecovery` does produce genuine displaced orphans, but its measurement window closes before they fire: it waits for the marker to exceed 1s, takes it over, then feeds for exactly `feedFor = 1s`. Instrumented on this branch, the orphan's deadline landed at **T+1.618s** — after the test had finished asserting. So the "orphan pops and is dropped" path is not exercised by either test.

To be clear about severity: I checked the behaviour and it is correct. Driving a session past its orphan's deadline and then measuring, it keeps transmitting normally (10 packets over 1s at a 100ms cadence, no wedge), and a 40-session concurrent state-churn run left no session without a scheduled TX. This is a missing regression guard, not a defect.

Worth committing something equivalent: hold a session Up until past the displaced deadline, then assert it still transmits. Without it, a future change to the marker bookkeeping could reintroduce a permanent-silence wedge with the suite still green — which is the failure mode the audit in your reply was guarding against.

packethog (changes requested):
Both findings from the previous pass are fixed, and I verified the fix independently rather than taking the numbers on trust: the transmit-rate repro drops from 256+ packets to **1** under an unchanged 5s floor, and the fault suite goes from **3-of-9 failing to 9-of-9 green**. Reverting each half of the fix on its own fails its own test (the rate-floor revert reproduces `sent 256` exactly). No defect found in either the `lastTx` pacing or the stale-TX drop — the no-wedge invariant holds under every interleaving I could construct, including a 40-session concurrent-churn stress run.

One test-coverage gap inline, non-blocking in substance: the wedge invariant your reply identifies as this change's main risk is not actually pinned by the suite.

client/doublezerod/internal/liveness/scheduler_test.go
ben-dz: Agreed on all of it, and your diagnosis of why neither test reaches the case is right. Added `TestClient_Liveness_Manager_HandleRx_SurvivesDisplacedTXOrphan` in 11fe843b.

It does what you suggested. With `backoffMax` at 2s so the deadline lands inside the test's runtime, it lets the Down backoff push the pending TX past 1.5s, **captures that deadline**, then recovers for real via `Down -> Init -> Up` so the orphan is one an actual `scheduleTxNow` takeover displaced (asserted: the new marker is strictly before the captured deadline). It then feeds RX at cadence to hold the session Up until 300ms *past* the orphan's deadline, so the orphan has certainly popped, and only then drains and measures — requiring the session to still be transmitting over the following second.

Capturing the deadline rather than guessing a window is what fixes the problem you found in `ResumesCadenceAfterRecovery`, where the orphan fired at T+1.618s against a 1s window.

Verified it catches the failure mode rather than just passing: injecting the regression it guards against — clearing a live marker on the stale path with no re-arm — fails it with `sent 0`, permanent silence, in 6.5s. Stable over 8 `-race` runs.

Worth noting the fault suite does catch that particular wedge too, but incidentally and in 8–24s of generic timeouts, and one of the tests that trips is `SoftLoss_TimeoutAtHighLoss` — the load-sensitive one from your last pass. So it wasn't a dependable guard for this. I also cross-referenced this test from `Run_DropsStaleTX` so the split in coverage is explicit: that one covers "does not transmit" on a fabricated orphan, this one covers survival of a genuine one.

No production code changed in this commit.

ben-dz:
Addressed in 11fe843b — test-only, no production change.

`TestClient_Liveness_Manager_HandleRx_SurvivesDisplacedTXOrphan` pins the no-wedge invariant directly: it captures the deadline of the event a real `Down -> Init -> Up` recovery displaces, holds the session Up past that deadline so the orphan actually pops, then asserts the session is still transmitting. Capturing the deadline instead of assuming a window is what avoids the problem you found in `ResumesCadenceAfterRecovery`, whose 1s measurement closed before the orphan fired at T+1.618s.

Injecting the regression it guards against — clearing a live marker on the stale path with no re-arm — fails it with `sent 0`, i.e. the permanent silence the audit was worried about. Stable over 8 `-race` runs; fault suite 6/6 green under `-race` plus a CI-scale run; lint and vet clean.

Thanks for verifying the previous round independently rather than taking my numbers, and for pinning down that the orphan fired at T+1.618s — that measurement is what made the gap actionable instead of hypothetical.

One thing still open from earlier, your call: `AdminDownRoute` calls `scheduleTx` under a comment promising to "send at least one AdminDown packet promptly", but `scheduleTx` bails if a TX is already queued and otherwise schedules a full interval out. Same defect class as this PR, and `scheduleTxNow` is now the right call there. I've left it alone since it touches the disconnect path and I can't run e2e here — happy to fold it in, otherwise I'll file it separately.

And unchanged on the issue itself: this fixes a real reconvergence gap but doesn't on its own prove #3935 is closed. The 63s observation doesn't uniquely imply a 60s effective `backoffMax` (with `txMin`=1s the harness's 3s cap gives 1,2,3,3,3… which also sums to 63s), and the override's plumbing verifies intact end to end — so the remaining suspicion is that packets genuinely stopped flowing after the unblock. The new startup log and the richer e2e dump are what settle it on the next occurrence, and that needs `/run-e2e` dispatches.

nikw9944 (approved):
lgtm — no change since the last pass: the head is still 11fe843b, whose test-only delta and the underlying production fixes (transmit-floor pacing of state-change TX, stale-TX drop mirroring the detect branch, corrected comments, and the displaced-orphan survival regression test) were all verified in code in the previous rounds. The new comments are author summaries consistent with the verified state; the only open thread is the deliberately deferred AdminDownRoute follow-up, which is pre-existing and fine to land separately.

packethog (approved):

nikw9944 (approved):
lgtm — the head move is a clean rebase onto newer main: the PR content at 7df96118 is byte-identical to the previously verified state (confirmed via the head-to-head diff containing only main-side files, the merge-base diff matching the verified file set exactly, and the CHANGELOG hunk surviving intact). All earlier findings were verified fixed in code in prior rounds (transmit-floor pacing of state-change TX, stale-TX drop mirroring the detect branch, corrected comments, and the displaced-orphan survival regression test); no findings remain.

PR #4087, opened 2026-07-27
cli: add --feed to `user create-subscribe` for EdgeSeat feed passes

martinsander00:
## Summary

- `doublezero user create-subscribe` gains a `--feed` flag (a Feed pubkey, or an unambiguous feed code). It is passed through as the trailing `Feed` account of `CreateSubscribeUser`.
- Without it, connecting a Multicast user to an EdgeSeat *feed* pass fails: the feed metro gate requires the `Feed` account, but the CLI hardcoded `feed_pk = None`, so it returned `FeedAccountRequired` (custom error `0x5c`). The SDK `CreateSubscribeUserCommand` already appends the account when `feed_pk` is set, so this only wires up the CLI flag.

## Testing

- New unit test `test_cli_user_create_subscribe_with_feed_pubkey` asserts a pubkey `--feed` flows straight through to `CreateSubscribeUserCommand.feed_pk` (no registry lookup); the existing `test_cli_user_create_subscribe` is updated for the added field.
- `cargo test -p doublezero-serviceability-cli create_subscribe` passes.

### DIFF
--- a/smartcontract/cli/src/user/create_subscribe.rs
+++ b/smartcontract/cli/src/user/create_subscribe.rs
@@ -113,7 +113,7 @@ impl CreateSubscribeUserCliCommand {
                             pubkey_or_code: feed.clone(),
                             exchange: None,
                         })
-                        .map_err(|_| eyre::eyre!("Feed not found ({})", feed))?;
+                        .map_err(|e| eyre::eyre!("Feed lookup failed ({feed}): {e}"))?;
                     Some(pubkey)
                 }
             },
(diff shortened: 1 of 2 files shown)


smartcontract/cli/src/user/create_subscribe.rs
```diff
@@ -99,6 +103,22 @@ impl CreateSubscribeUserCliCommand {
...
+            None => None,
+            Some(feed) => match parse_pubkey(&feed) {
+                Some(pk) => Some(pk),
+                None => {
+                    let (pubkey, _) = client
+                        .get_feed(GetFeedCommand {
+                            pubkey_or_code: feed.clone(),
+                            exchange: None,
+                        })
+                        .map_err(|_| eyre::eyre!("Feed not found ({})", feed))?;
```
nikw9944: `.map_err(|_| eyre::eyre!("Feed not found ({})", feed))` collapses every `get_feed` failure into "Feed not found", including the SDK's deliberate ambiguity rejection ("Feed code X is ambiguous: it exists in N metros. Pass --exchange or the feed pubkey."). A user whose feed code exists in two metros is told the feed doesn't exist and never sees the remedy. Unlike the device/multicast-group lookups above (where not-found is essentially the only failure mode), this lookup has two distinct failures needing different user actions. Preserve the underlying error, e.g. `.map_err(|e| eyre::eyre!("Feed lookup failed ({feed}): {e}"))`. One wrinkle: the SDK hint names `--exchange`, which this verb doesn't have — the pubkey remedy still applies, but adding an `--exchange` disambiguator like the `feed get/update/delete` verbs would be the complete fix.

smartcontract/cli/src/user/create_subscribe.rs
```diff
@@ -261,4 +282,122 @@ mod tests {
             output_str,"Signature: 3QnHBSdd4doEF6FgpLCejqEw42UQjfvNhQJwoYDSpoBszpCCqVft4cGoneDCnZ6Ez3ujzavzUu85u6F79WtLhcsv\n"
         );
     }
+
+    // A pubkey `--feed` is passed straight through as the trailing Feed account for an
+    // EdgeSeat feed pass.
+    #[test]
+    fn test_cli_user_create_subscribe_with_feed_pubkey() {
```
nikw9944, line 290: Only the pubkey `--feed` path is tested; the code-resolution branch (bare code → `get_feed` → pubkey) has no CLI test. A short companion test with `expect_get_feed` would cover it and lock in the error mapping once the map_err fix lands.

nikw9944 (commented):
One error-handling fix worth making: the `--feed` lookup's `map_err` reports every failure as "Feed not found", which mislabels the SDK's ambiguous-code rejection and hides its remedy. Plus one optional test suggestion. Otherwise the wiring is correct: the pubkey path passes straight through, and the SDK appends the trailing Feed account as described.

smartcontract/cli/src/user/create_subscribe.rs
martinsander00: Fundamentally, one feed cannot exist in different metros

smartcontract/cli/src/user/create_subscribe.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4087/changes/df5507970f4804d31bd19ef68c6eaed2229fc14e

smartcontract/cli/src/user/create_subscribe.rs
nikw9944: Still unresolved — verified the current diff is unchanged since the last round. `.map_err(|_| eyre::eyre!("Feed not found ({})", feed))` collapses the SDK's ambiguity rejection ("Feed code X is ambiguous: it exists in N metros...") into a false "Feed not found", hiding the remedy. Verified against the program: feed uniqueness is enforced only per `(code, exchange)` PDA (`pda.rs:109`), with no global code-uniqueness check in `processors/feed/create.rs`, so the same code can name distinct feeds in different metros — the exact case `GetFeedCommand` rejects and the SDK tests (`test_commands_feed_get_ambiguous_code_rejected`). Fix remains a one-liner: preserve the underlying error, e.g. `.map_err(|e| eyre::eyre!("Feed lookup failed ({feed}): {e}"))`.

nikw9944 (commented):
Round 4: no changes since round 3 — head is still df550797, the diff is identical (verified by hash and by locating the unchanged map_err), and there are no new comments or reviews; CI is now fully green. M1 remains unresolved: line 116 still collapses the SDK's ambiguous-code rejection into a false "Feed not found" (the author's 'one feed cannot exist in different metros' reply was verified against the program and does not hold — uniqueness is per (code, exchange) PDA with no global code check). L1 remains resolved via the test added in df550797, verified in round 3.

smartcontract/cli/src/user/create_subscribe.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4087/changes/a94bcdd615322b872ee95b6f78d8ac4cfd6ab48c

nikw9944 (approved):
Round 7: the observed delta (head moved to a94bcdd6 + author's reply linking it) is exactly what round 6 reviewed; confirmed nothing newer — all comments/reviews previously seen and the diff is byte-identical (verified by hash). Both findings remain resolved with code-level verification: M1's map_err now preserves the underlying error (`Feed lookup failed ({feed}): {e}`) with a test asserting the ambiguity cause reaches the user, and L1's code-resolution test landed in df550797. All 4 create_subscribe tests pass locally on the current head. No new issues. CI for the a94bcdd6 push was still pending (none failing) at check time — let it finish before merge.

- Re-checked: no changes since the last round — a94bcdd6 and the linked reply were already verified. All findings remain addressed (map_err fix + error-surfacing test, code-resolution test). Only open item is the still-running CI for the latest push (nothing failing); the change is CLI-only and the affected suite passes locally.

PR #4089, opened 2026-07-27
serviceability: address review comments on dzf_locked set-flags

elitegreg:
## Summary of Changes

- `SetAccessPass` now clears only the bit it owns (`ALLOW_MULTIPLE_IP`) rather than whitelisting `DZF_LOCKED` for preservation, so any flag managed by another instruction survives an unrelated pass update automatically instead of needing the mask extended by hand.
- The feed-authority ownership guard in `SetAccessPassFlags` — the check that stops an `ACCESS_PASS_ADMIN` holder from touching a pass it does not own — is now covered by tests, paired with a positive control on a pass it does own so a broken `authorize()` cannot masquerade as the guard firing.
- `SetAccessPassFlags` rejects an all-`None` payload only after `authorize()`, so an unauthorized caller gets `NotAllowed` rather than a diagnostic about the payload.
- `SetAccessPassFlagsArgs` no longer derives `Default` (its `default()` was exactly the payload the processor rejects), and the no-op check destructures the args so adding a flag field is a compile error instead of a silently incomplete check.
- The tri-state wire format is pinned: both-`Some` and all-`None` round-trips, plus incremental decoding of a truncated payload from a sender predating a later field.
- Records that the absent accesspass PDA re-derivation in `set_flags.rs` is intentional, since the args deliberately carry no `client_ip` / `user_payer` seeds.

Follow-up to #4083, which merged before the review landed. Juan's note on the `feed_authority_pk` guard becoming a no-op after the feed-authority → Permission migration is intentionally **not** addressed here: the same guard is copy-pasted in `accesspass/set.rs` and `accesspass/close.rs`, so it belongs with that migration rather than in this PR. The new test pins the guard's current behavior so the change is visible when it happens.

## Diff Breakdown

| Category    | Files | Lines (+/-)  | Net   |
|-------------|-------|--------------|-------|
| Core logic  |     2 | +32 / -14    |   +18 |
| Tests       |     3 | +390 / -1    |  +389 |
| **Total**   |     4 | +422 / -15   |  +407 |

Almost entirely test coverage over a small behavioral core: one mask change and one reordering. (`set_flags.rs` appears in both rows — its processor changes are core logic, its new `mod tests` is tests.)

<details>

- `smartcontract/programs/doublezero-serviceability/tests/accesspass_dzf_locked.rs` — two new integration tests: the feed authority denied on a pass it does not own (with a positive control on one it owns), and the Permission grant path under `RequirePermissionAccounts` proving success with the Permission account appended and `NotAllowed` without it.
- `smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs` — dropped `Default`, destructured no-op check moved below `authorize()`, intentional-no-PDA-check comment, and unit tests for the tri-state encoding and incremental truncation.
- `smartcontract/programs/doublezero-serviceability/src/instructions.rs` — pack/unpack round-trip extended to both-`Some` and all-`None` payloads.
- `smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set.rs` — `flags` mask flipped from whitelist-what-to-keep to clear-only-what-this-instruction-owns.

</details>

## Testing Verification

- Verified the new feed-authority test is a real assertion: temporarily removing the ownership guard makes it fail with `expected Custom(8), got Ok(())`.
- Verified the mask flip is actually future-proof, not just cosmetic: temporarily relocated `DZF_LOCKED` to bit 2 (simulating a third flag) and confirmed it survives an unrelated `SetAccessPass` under the new mask, and is clobbered under the old whitelist mask.
- Confirmed the mask change is behavior-preserving today — the pre-existing "`DZF_LOCKED` must survive an unrelated set" assertion passes unchanged.
- `make rust-lint` and `make rust-test` pass (full serviceability suite, including the 16 existing `accesspass_test` cases).

### DIFF
--- a/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set.rs
@@ -6,7 +6,7 @@ use crate::{
     seeds::{SEED_ACCESS_PASS, SEED_PREFIX},
     serializer::{try_acc_create, try_acc_write},
     state::{
-        accesspass::{AccessPass, AccessPassStatus, AccessPassType, ALLOW_MULTIPLE_IP, DZF_LOCKED},
+        accesspass::{AccessPass, AccessPassStatus, AccessPassType, ALLOW_MULTIPLE_IP},
         accounttype::AccountType,
         globalstate::GlobalState,
         permission::permission_flags,
@@ -281,10 +281,11 @@ pub fn process_set_access_pass(
             _ => value.accesspass_type.clone(),
         };
         accesspass.last_access_epoch = value.last_access_epoch;
-        // `flags` here only carries ALLOW_MULTIPLE_IP. The DZF_LOCKED bit is managed solely by
-        // the `ACCESS_PASS_ADMIN`-gated SetAccessPassFlags instruction, so preserve any existing lock
-        // rather than clobbering it on an unrelated update.
-        accesspass.flags = (accesspass.flags & DZF_LOCKED) | flags;
+        // `flags` here only carries ALLOW_MULTIPLE_IP, the one bit this instruction owns. Clear
+        // just that bit and leave every other flag untouched, so bits managed elsewhere (DZF_LOCKED
+        // via the `ACCESS_PASS_ADMIN`-gated SetAccessPassFlags, and anything added later) survive
+        // an unrelated update automatically.
+        accesspass.flags = (accesspass.flags & !ALLOW_MULTIPLE_IP) | flags;
         accesspass.max_unicast_users = value.max_unicast_users;
         accesspass.max_multicast_users = value.max_multicast_users;
 
--- a/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
@@ -92,12 +92,15 @@ pub fn process_set_access_pass_flags(
 
     // Reject a no-op call rather than charging for a write that changes nothing. Runs after
     // authorize() so an unauthenticated caller gets NotAllowed rather than a payload diagnostic.
-    // Destructured rather than compared against a Default, so adding a flag field is a compile
-    // error here instead of a silently incomplete check.
+    //
+    // Destructured rather than compared against a Default so a future flag field cannot be
+    // silently omitted here: adding one is a hard error at this pattern (E0027), and forgetting to
+    // fold the new binding into the condition below is an unused-variable warning, which CI fails
+    // on under `-Dwarnings`.
     let SetAccessPassFlagsArgs {
         allow_multiple_ip,
         dzf_locked,
-    } = value;
+    } = *value;
     if allow_multiple_ip.is_none() && dzf_locked.is_none() {
         msg!("SetAccessPassFlags requires at least one flag to update");
         return Err(DoubleZeroError::InvalidArgument.into());
@@ -108,20 +111,29 @@ pub fn process_set_access_pass_flags(
     // Mirror the sibling accesspass handlers: the feed authority may only touch passes it owns.
     // This is what stops the oracle (which holds ACCESS_PASS_ADMIN) from clearing dzf_locked on a
     // pass it did not create.
+    //
+    // DEPENDENCY: this holds only while the oracle's key lives in the legacy
+    // `globalstate.feed_authority_pk`. Once the oracle authorizes purely via a Permission account
+    // granting ACCESS_PASS_ADMIN and this legacy field is cleared or reassigned, the condition is
+    // never true and the guard silently no-ops — losing exactly the property dzf_locked exists to
+    // guarantee. The durable fix is to fence *any* non-foundation/sentinel ACCESS_PASS_ADMIN caller
+    // to passes it owns, which belongs with the feed-authority -> Permission migration because the
+    // same guard is duplicated in accesspass/set.rs and accesspass/close.rs. Revisit all three
+    // together at that point.
     if globalstate.feed_authority_pk == *payer_account.key && accesspass.owner != *payer_account.key
     {
         msg!("Feed authority can only modify access passes they own");
         return Err(DoubleZeroError::NotAllowed.into());
     }
 
-    if let Some(allow_multiple_ip) = *allow_multiple_ip {
+    if let Some(allow_multiple_ip) = allow_multiple_ip {
         if allow_multiple_ip {
             accesspass.flags |= ALLOW_MULTIPLE_IP;
         } else {
             accesspass.flags &= !ALLOW_MULTIPLE_IP;
         }
     }
-    if let Some(dzf_locked) = *dzf_locked {
+    if let Some(dzf_locked) = dzf_locked {
         if dzf_locked {
             accesspass.flags |= DZF_LOCKED;
         } else {
--- a/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
@@ -119,7 +119,7 @@ pub fn process_set_access_pass_flags(
     // guarantee. The durable fix is to fence *any* non-foundation/sentinel ACCESS_PASS_ADMIN caller
     // to passes it owns, which belongs with the feed-authority -> Permission migration because the
     // same guard is duplicated in accesspass/set.rs and accesspass/close.rs. Revisit all three
-    // together at that point.
+    // together at that point. Tracked in #4092.
     if globalstate.feed_authority_pk == *payer_account.key && accesspass.owner != *payer_account.key
     {
         msg!("Feed authority can only modify access passes they own");
--- a/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
+++ b/smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
@@ -23,7 +23,11 @@ use solana_program::{
 /// is just another `Option<bool>` field.
 ///
 /// Gated on `ACCESS_PASS_ADMIN` (like the sibling SetAccessPass path).
-#[derive(BorshSerialize, BorshDeserializeIncremental, Debug, PartialEq, Clone, Default)]
+///
+/// Deliberately does not derive `Default`: `default()` would be the all-`None` payload this
+/// instruction rejects, and a `..Default::default()` construction site would silently swallow a
+/// future flag field instead of failing to compile.
+#[derive(BorshSerialize, BorshDeserializeIncremental, Debug, PartialEq, Clone)]
 pub struct SetAccessPassFlagsArgs {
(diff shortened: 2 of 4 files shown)


smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set.rs
```diff
@@ -281,10 +281,11 @@ pub fn process_set_access_pass(
...
         accesspass.last_access_epoch = value.last_access_epoch;
-        // `flags` here only carries ALLOW_MULTIPLE_IP. The DZF_LOCKED bit is managed solely by
-        // the `ACCESS_PASS_ADMIN`-gated SetAccessPassFlags instruction, so preserve any existing lock
-        // rather than clobbering it on an unrelated update.
-        accesspass.flags = (accesspass.flags & DZF_LOCKED) | flags;
+        // `flags` here only carries ALLOW_MULTIPLE_IP, the one bit this instruction owns. Clear
+        // just that bit and leave every other flag untouched, so bits managed elsewhere (DZF_LOCKED
+        // via the `ACCESS_PASS_ADMIN`-gated SetAccessPassFlags, and anything added later) survive
+        // an unrelated update automatically.
+        accesspass.flags = (accesspass.flags & !ALLOW_MULTIPLE_IP) | flags;
```
juan-malbeclabs, line 288: Confirmed this is behavior-preserving today and future-proof as intended: `flags` at this point is built only from `ALLOW_MULTIPLE_IP` (lines 181-184), and both branches that construct a fresh `AccessPass` set `flags` directly, so the mask only ever runs against a stored byte. Any bit added later survives an unrelated `SetAccessPass` automatically.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
```diff
@@ -97,14 +114,14 @@ pub fn process_set_access_pass_flags(
         return Err(DoubleZeroError::NotAllowed.into());
```
juan-malbeclabs, line 126: This is the remaining half of my #4083 comment. Deferring the *durable* fix is right — the same guard is copy-pasted in `accesspass/set.rs:241` and `accesspass/close.rs:96`, so fencing any non-foundation/sentinel `ACCESS_PASS_ADMIN` caller to passes it owns belongs with the feed-authority → Permission migration, not here.

What's still missing is the comment. The block above explains what the guard does but not what it depends on: the property holds only while the oracle's key stays in `globalstate.feed_authority_pk`. Once the oracle authorizes purely via a Permission account granting `ACCESS_PASS_ADMIN` and the legacy field is cleared or reassigned, this condition is never true and the guard silently no-ops — the exact property the feature exists to guarantee. Two lines here, plus a tracked issue for the migration, and this is closed.

juan-malbeclabs (commented):
## Review

All six items from my review of #4083 are addressed, one of them only partially. No new correctness problems; full serviceability suite and clippy are green locally, and the three `accesspass_dzf_locked` tests pass.

### Item by item

| Ask on #4083 | Status |
| --- | --- |
| `set.rs` mask: clear only the bit this instruction owns | Applied verbatim. Verified `flags` is built solely from `ALLOW_MULTIPLE_IP` (`set.rs:181-184`) and both create paths are unaffected, so the flip is behavior-preserving today. |
| Drop `Default` on `SetAccessPassFlagsArgs` | Dropped, with a doc note explaining why. The destructured no-op check is a stronger answer than the `value == Default::default()` suggestion it supersedes. |
| Comment recording the intentionally-absent PDA re-derivation | Done, phrased as a "don't fix this" note. |
| No-op rejection below `authorize()` | Done; an unauthorized caller now gets `NotAllowed`. |
| Round-trip: both-`Some`, all-`None`, incremental truncation | Both-`Some` / all-`None` in `instructions.rs`, truncation as a unit test on the args. Splitting it is fine: `unpack` delegates straight to `SetAccessPassFlagsArgs::try_from` (`instructions.rs:402`), so the args-level test covers the same decode path, and pinning exact bytes is stronger than a round-trip. |
| Feed-authority ownership guard: test it | Done, with the positive control, plus the strict-mode Permission case. |
| Feed-authority ownership guard: document the legacy-field dependency | **Not done** — see inline. |

### The one outstanding item

The rationale for deferring the *code* change is correct: the same guard is copy-pasted at `accesspass/set.rs:241` and `accesspass/close.rs:96`, so a durable reformulation belongs with the feed-authority → Permission migration rather than here. Accepted.

The "at minimum a comment" half is still missing, though, and the hazard is now sharper rather than softer — see the inline comment on the strict-mode test. Nothing in the tree tracks it either: `PERMISSION.md:98` documents only the legacy fallback table.

Not blocking merge. Happy to approve as-is if you'd rather carry the note with the migration work; I'd just prefer two lines at the guard so the dependency is visible where the guard lives.

smartcontract/programs/doublezero-serviceability/tests/accesspass_dzf_locked.rs
```diff
@@ -225,3 +232,301 @@ async fn test_accesspass_dzf_locked() {
...
+        DoubleZeroInstruction::SetFeatureFlags(SetFeatureFlagsArgs {
+            feature_flags: FeatureFlag::RequirePermissionAccounts.to_mask(),
+        }),
+        vec![AccountMeta::new(globalstate_pubkey, false)],
+        &payer,
+    )
+    .await;
+
+    // Grant path: the Permission PDA is appended as the trailing account authorize() reads.
+    execute_transaction_with_extra_accounts(
```
juan-malbeclabs, line 483: Worth noting what this assertion becomes later: `pass_admin` is not the feed authority, so the guard below `authorize()` never fires, and it sets `dzf_locked` on a pass owned by `payer`. That is correct and intended today — an `ACCESS_PASS_ADMIN` holder may target any pass.

But it is also, line for line, the shape the oracle takes after the feed-authority → Permission migration. On that day this test asserts the regression as an expected success, with nothing pointing at it. That's the argument for the comment on `set_flags.rs` rather than relying on the PR description alone.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
```diff
@@ -97,14 +114,14 @@ pub fn process_set_access_pass_flags(
...
     }
 
-    if let Some(allow_multiple_ip) = value.allow_multiple_ip {
+    if let Some(allow_multiple_ip) = *allow_multiple_ip {
```
juan-malbeclabs: Nit: `let SetAccessPassFlagsArgs { .. } = *value;` above would drop the deref here and on line 124 — `Option<bool>` is `Copy`, and the compile-error-on-new-field property is unchanged.

Also a small precision on the comment: the destructure makes adding a field a hard error here (E0027), but forgetting to *use* the new binding in the condition is only an unused-variable warning. It still fails CI under `-Dwarnings`, so the property holds — just via a different mechanism than the comment claims.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
elitegreg: Added, plus the tracked issue: #4092.

The comment now records the dependency explicitly — that the property holds only while the oracle's key lives in `globalstate.feed_authority_pk`, that the condition is never true once the oracle authorizes via a Permission account and the legacy field is cleared, and that the durable fix is fencing any non-foundation/sentinel `ACCESS_PASS_ADMIN` caller to passes it owns across all three copies (`set_flags.rs`, `set.rs`, `close.rs`).

#4092 carries the shared-helper sketch, the note that "elevated" has to include a `FOUNDATION` Permission grant (otherwise emptying `foundation_allowlist` fences legitimate foundation admins), and both tests that need revisiting.

smartcontract/programs/doublezero-serviceability/tests/accesspass_dzf_locked.rs
elitegreg: Right, and that framing is better than mine — the test does not just fail to catch the regression, it asserts it as expected behavior.

Added a comment here saying so: `pass_admin` is not the feed authority, so the guard never fires and this succeeds against a pass owned by `payer`; correct today, but the exact shape the oracle takes post-migration, so this assertion is what would need to flip. Points at #4092, same as the `set_flags.rs` DEPENDENCY comment.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
elitegreg: Both taken. Destructuring from `*value` now, and the two derefs at the `if let`s are gone.

And you are right that the comment overclaimed. Reworded to state the two mechanisms separately: adding a field is a hard error at the pattern (E0027), while forgetting to fold the new binding into the condition is an unused-variable warning that CI fails on under `-Dwarnings`. Different mechanism, same outcome — but worth saying accurately, since the second one depends on the lint config rather than the type system.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
juan-malbeclabs: Closed. `set_flags.rs:115-122` records all three parts of the hazard — that the property holds only while the oracle's key lives in `globalstate.feed_authority_pk`, that the condition is never true once it authorizes via a Permission account and the legacy field is cleared, and that the durable fix spans the copies in `set.rs` and `close.rs` so all three get revisited together. Pointing at #4092 from the comment is what I wanted; the dependency is now visible where the guard lives.

#4092 itself is more complete than what I asked for. The `FOUNDATION`-grant point is the one I would have missed: fencing on `foundation_allowlist` / `sentinel_authority_pk` alone means emptying the allowlist fences legitimate foundation admins, so `can_grant_foundation()` has to be part of the "elevated" test. Peeking the trailing account before `authorize()` consumes it is the right mechanism — slice `Iter` is `Clone`, so it costs nothing.

smartcontract/programs/doublezero-serviceability/tests/accesspass_dzf_locked.rs
juan-malbeclabs: Confirmed at `accesspass_dzf_locked.rs:476-482`. It states the mechanism (not the feed authority, so the guard never fires), that it is correct today, that it is the shape the oracle takes post-migration, and that this assertion is the one that flips — with the pointer to #4092. A reader who lands here after the migration has what they need.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_flags.rs
juan-malbeclabs: Both verified: `= *value;` at `:103`, and the derefs at the two `if let`s are gone (`:129`, `:136`). The reworded comment at `:96-99` is accurate now — hard error at the pattern for a new field, unused-variable warning for forgetting to fold it into the condition, the latter load-bearing only under `-Dwarnings`. Worth stating separately for exactly the reason you give: one is the type system, the other is lint config.

juan-malbeclabs (approved):
LGTM

PR #4093, opened 2026-07-27
ci: add copilot code-review instructions

juan-malbeclabs:
## Summary of Changes

* Adds `.github/copilot-instructions.md` (repo-wide review posture + cross-cutting rules) and eight path-scoped `.github/instructions/*.instructions.md` files, so Copilot's automated PR review applies the standards we already enforce by hand instead of the generic ones.
* Scoped by `applyTo` glob so each review only loads what the diff touches: onchain programs, Rust SDK/builders/CLI, Go, E2E/QA, release workflows, the language SDKs and their golden fixtures, tests, and the changelog.
* No effect on builds or runtime — these files are only read by Copilot.

### Where the rules come from

1. **Our own reviews.** I pulled every inline review comment on PRs merged in the last month — 421 comments across PRs #3578–#4089 — and grouped the recurring findings. The threshold for a rule was that the same pattern appeared in **two or more distinct PRs**, or is already codified in an RFC or `docs/cli-standard.md`. Both sides of each thread were read, so suggestions that were deliberately declined (per-environment payload formats, the grace-period bound, hoisting at three copies) did **not** become rules.
2. **Published RFCs.** RFC-1 contributes semver classification and the requirement to state deploy-ordering prerequisites; RFC-10 contributes the compatibility-window schema-evolution invariant (read both formats, keep writing the old one until `min_compatible_version` clears every reader); RFC-20 gates new CLI verbs.
3. **The code itself.** Every factual claim in the rules — symbol locations, `SDK_ATTACHES_PERMISSION_ACCOUNTS`, the absent `overflow-checks` profile, `split_trailing_permission` semantics, the `goccy/go-json` decode path — was verified against the source rather than taken from the review text.

Rules were then pruned to those where a miss ships a real defect: silent feature disable, broadened authorization, a wire break, an unreclaimable reference count, a release tagged against the wrong commit, an alert that never fires. Style, naming, redundancy-removal and log-clarity nits were dropped, since noisy automated review is worse than none.

Notable gap this closes: nothing previously told an automated reviewer how to treat a **pre-existing defect the diff widens** — a framing that appears in six of the PRs sampled and shapes most human comments in the sample.

## Testing Verification

* Verified every `applyTo` glob matches tracked files — no dead patterns.
* Replayed the file lists of five recent PRs (#4086, #4083, #4041, #3964, #4080) to confirm each activates the intended instruction set, which caught two gaps now fixed: `client/doublezero/**` matched nothing, and the Go serviceability decoder lives at `smartcontract/sdk/go/`, not under `sdk/`.
* Fact-checked the rules against the code; this corrected a wrong file path for `AUTHORIZE_GATED_FLAGS`/`legacy_keys_for_flags`/`check_legacy_any` (they are in `authorize.rs`, not `permission/audit.rs`), a stale claim that the SDK still appends the Permission PDA, and a rule that would have flagged the system-program check present in 56 processor files.

## Follow-up

Path-specific instructions are honored by Copilot code review only when custom instructions are enabled for the repo/org. Worth confirming on this PR's own Copilot review — it touches `CHANGELOG.md`, so the `changelog` rules should visibly apply. If they do not, only the repo-wide file is in effect.


.github/instructions/solana-programs.instructions.md
```diff
@@ -0,0 +1,89 @@
...
+  exception: the discriminator check via `try_from` is sufficient because only one valid account
+  exists.
+- `authorize()` should run immediately after `GlobalState` is parsed, before argument validation,
+  PDA derivation, and `data_is_empty` probes, so an unauthorized caller cannot probe for account
+  existence or trip validation errors before being denied. The accepted exception is a cheap
+  all-`None` no-op rejection placed above it.
+- Optional trailing accounts must be disambiguated by PDA match (`split_trailing_permission` in
+  `authorize.rs`), never by position. The on-wire tail is
+  `[...command accounts, payer, system, permission?]`. The Rust SDK's Permission append is currently
+  deferred — `build_with_permission` delegates to `build`, and `SDK_ATTACHES_PERMISSION_ACCOUNTS`
```
martinsander00: i checked and it looks like `SDK_ATTACHES_PERMISSION_ACCOUNTS` doesn't exist on main

.github/copilot-instructions.md
```diff
@@ -0,0 +1,69 @@
+# Copilot instructions — doublezero
+
+DoubleZero is a protocol for building and operating high-performance, permissionless networks — a
+global dedicated-fiber network for distributed systems like blockchains. Service contributors
+deploy devices and register them and their links onchain; users connect over GRE tunnels and receive
+optimized routes via BGP, in unicast (IBRL) or multicast modes.
+
+Two distinct chains are in play. The **DoubleZero Ledger** is the protocol's own Solana-based
+cluster, run by its own validators; the serviceability, telemetry, geolocation, and
+internet-latency programs in this repo are deployed there, reached via `ledger_rpc_url`.
```
martinsander00: this should be `serviceability, telemetry, geolocation, record` right? like CLAUDE.md says

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ All notable changes to this project will be documented in this file.
...
   - `AccessPass` gains a `dzf_locked` flag (bit 0 of `flags`) marking a pass as foundation-managed so automated reconcilers such as the Feed Oracle leave it alone. A new `SetAccessPassFlags` instruction (gated on `ACCESS_PASS_ADMIN`) sets/clears access-pass flags without disturbing the others, and `SetAccessPass` now preserves the flag across unrelated updates. Set it with the `doublezero access-pass dzf-lock` / `dzf-unlock` verbs or `access-pass set --dzf-locked`. (#4083)
+- CI
+  - Add Copilot code-review instructions: a repo-wide `.github/copilot-instructions.md` (review posture plus cross-cutting rules) and eight path-scoped `.github/instructions/*.instructions.md` files covering the onchain programs, the Rust SDK/builders/CLI, Go, E2E/QA, release workflows, the language SDKs and their golden fixtures, tests, and the changelog. Rules were extracted from the team's own inline PR reviews over the previous month and cross-checked against the code, plus the standards in RFC-1 (semver classification, deploy-ordering prerequisites), RFC-10 (compatibility-window schema evolution), and RFC-20. No effect on builds or runtime.
```
martinsander00: should reference the pr number at the end of this bullet point?

.github/instructions/solana-programs.instructions.md
juan-malbeclabs: Good catch on it not being obvious — it does exist, but not where you'd look from this file's scope.

It lives in the CLI crate rather than under `smartcontract/programs/`: `smartcontract/cli/src/permission/audit.rs:49`

```rust
const SDK_ATTACHES_PERMISSION_ACCOUNTS: bool = false;
```

It's on `main` today (last touched by 8051198f). It's used by `doublezero permission audit` — it's surfaced in the report as `sdk_attaches_permission_accounts`, and while it's `false` the audit renders a paragraph explaining that `authorize()` never sees a Permission account through the CLI/SDK send path, so a key whose only grant is a Permission account falls back to the GlobalState allowlists and is denied. It's the flag to flip once the builder-side append is re-enabled (`build_with_permission` in `crates/doublezero-serviceability-instruction/src/common.rs:95` currently just delegates to `build`).

Since the path isn't discoverable from an instructions file scoped to `smartcontract/programs/**`, I'll add the file path inline so a reviewer can find it. Thanks for checking.

.github/copilot-instructions.md
juan-malbeclabs: You're right, thanks — good catch. The four programs under `smartcontract/programs/` are serviceability, telemetry, geolocation and record; there is no internet-latency program (`InternetLatencySamples` is an account type inside the telemetry program, at `smartcontract/programs/doublezero-telemetry/src/state/internet_latency_samples.rs`).

Corrected to `serviceability, telemetry, geolocation, and record`. The same wording slipped into the overview paragraph in the root guidance doc, so I fixed it there too for consistency.

.github/instructions/go.instructions.md
```diff
@@ -0,0 +1,46 @@
+---
+applyTo: "client/**/*.go,controlplane/**/*.go,telemetry/**/*.go,api/**/*.go,tools/**/*.go"
```
martinsander00: should we add `config/**/*.go` here?

CHANGELOG.md
juan-malbeclabs: Yes — and it's the one entry in `## Unreleased` missing it, while every sibling has one. It also breaks the rule this very PR adds in `.github/instructions/changelog.instructions.md`: "Reference the PR number as `(#NNNN)`, not the branch name or the infra ticket it came from."

Added `(#4093)`. Thanks.

.github/instructions/go.instructions.md
juan-malbeclabs: Yes, good catch — that's a real coverage gap. `config/` has four tracked Go files (`constants.go`, `env.go`, `solana.go`, `env_test.go`) and they matched no instructions file at all, so they were getting only the repo-wide rules.

The scoped rules do apply there: `SolanaNetworkConfigForEnv` in `config/solana.go` is a `switch` over env strings, which is exactly the "treat enum, phase, and status values as an allowlist — an unknown value must return an error" rule, and the pubkey parsing in `config/env.go` is the "do not discard errors" one.

Added `config/**/*.go` and widened the description to mention the shared env config.

While checking this I found the same gap on the Rust side: `config/src/` (`constants.rs`, `env.rs`, `lib.rs`) wasn't in `rust-sdk-cli.instructions.md`'s `applyTo` either. That one matters a bit more, since the repo-wide instructions point at `config/src/constants.rs` as the authority on the DZ-env-to-Solana-cluster mapping — so the file the rule cites had no scoped rules of its own. Added `config/src/**` there too.

One related case I left alone deliberately: Go files under `sdk/**` and `smartcontract/examples/go` match only `sdk-fixtures.instructions.md`, so they don't get the Go rules. That seems right for the read-only decoders, but say the word if you'd rather they did.

martinsander00 (approved):

PR #4095, opened 2026-07-27
serviceability: reject duplicate link_topologies and fix reference_count double-decrement

### LINKED ISSUE #4090: serviceability: duplicate link_topologies entries can zero reference_count and bypass the TopologyDelete guard

ben-dz:
## Summary of Changes

* Reject duplicate entries in a Link's `link_topologies` on the `LinkUpdate` instruction argument, and decrement a topology's `reference_count` once per *unique* entry when a link drops it — on both `LinkDelete` and the `LinkUpdate` removal path.
* `LinkUpdate` diffs old vs new as a set, so a duplicate stored two entries against a single increment while the drop paths decremented per entry. That could zero a `reference_count` another link still contributes to and let `TopologyDelete` close the topology PDA that guard exists to protect, leaving a dangling pubkey in the surviving link and a permanently retired admin-group bit. One comma typo by a `NETWORK_ADMIN` or foundation key (`--link-topology "TOPO-A,TOPO-A"`) was enough.
* The duplicate check is on the instruction argument, deliberately **not** in `Link::validate()`: `validate()` runs inside `try_acc_write`, so a link already holding duplicates would become unwritable and therefore unrepairable.
* Fixes #4090

Related: RFC-18 (`rfcs/rfc-0018-flex-algo.md`).

## Testing Verification

* `test_link_update_rejects_duplicate_topologies` — re-submitting an already-tagged topology as a duplicate is rejected with `InvalidArgument`. The link is first tagged with `topo-a` so old ∪ new is `{topo-a}`, meaning nothing is added or removed and the pre-existing missing-account check cannot fire; the rejection has to come from the duplicate check.
* `test_link_delete_with_duplicate_topology_decrements_once` and `test_link_update_clearing_duplicate_topology_decrements_once` — the full L1/L2 exploit path against both drop paths. Two links tagged into `topo-a` (count 2), L1's stored vector seeded to `[topo-a, topo-a]`, then dropped via delete and via `link update --link-topology default`. Each asserts the count lands at 1, not 0, and that `TopologyDelete` stays blocked with `ReferenceCountNotZero`. The delete test also drives a real `UpdateLink` write against the duplicate-holding link first, pinning that such a link stays writable and therefore repairable.
* Each test was verified to fail without its corresponding source change (count 0 instead of 1; duplicate update succeeding).
* The `[T, T]` state is seeded with `ProgramTestContext::set_account` rather than an instruction, because with the fix in place no instruction can produce it — that is the point of the delete-side and update-removal-side handling.
* **No existing skew to repair on mainnet-beta:** queried the data lake for links whose `link_topologies` array contains repeated entries — 168 current links and all 2,649 historical snapshots, zero duplicates, max array length 1. Testnet could not be reached from the build environment and is worth a spot check before deploy.
* `make rust-lint` clean. `cargo test -p doublezero-serviceability` (51 binaries) and the workspace portion of `make rust-test` pass. The `cargo test-sbf` program step could not run locally — the SBF toolchain is not installed in this environment — so those integration tests were exercised natively via `solana-program-test` instead; CI covers the SBF build.


nikw9944 (approved):
lgtm — duplicate rejection is on the instruction argument (so skewed links stay repairable), both drop paths decrement once per unique topology, `TopologyClear` was already duplicate-safe, and the tests pin exact errors and post-state for the rejection and both skew-repair paths. Only note: e2e shards were still running at review time.

PR #4096, opened 2026-07-28
serviceability: bump min_compatible_version floor to 0.30.0

ben-dz:
Raises the serviceability compatibility floor from `0.21.0` to `0.30.0`, retiring every client older than v0.30.0.

- The floor has not moved since the v0.29.0 bump (3d15b8c56, Jul 2) while releases advanced to v0.31.0. Live `ProgramConfig` today: testnet `version=0.31.0 min=0.21.0`, mainnet-beta `version=0.30.0 min=0.21.0`.
- v0.30.0 is the first release whose `AccessPass` decoder understands `EdgeSeat(Vec<FeedSeat>)` (#3954, #4030). Older decoders misparse every field after the variant tag — the failure mode behind the `do-tn-rewards1` 52 GB abort, mitigated defensively in #4074.
- The e2e backward-compatibility matrix floor (`globalMinVersions`) moves in lockstep, matching how the last two bumps were done.

**This does not take effect at merge.** `MIN_COMPATIBLE_VERSION` is stamped onto `ProgramConfig` by `doublezero init`, which the testnet release runs as its final manual deploy step (`docs/testnet-release.md`). Each cluster picks the new floor up at its next program deploy. Once it lands, a pre-0.30.0 CLI fails at startup with "A new version of the client is available", and the compatibility window becomes `[0.30.0, <deployed version>]`.

Design doc: [rfcs/rfc10-version-compatibility-windows.md](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc10-version-compatibility-windows.md)

## Testing Verification

- New unit test `min_compatible_version_parses_and_does_not_exceed_current` guards the two ways a bad const bites: `process_initialize_global_state` `unwrap()`s the parse, and `SetMinVersion` rejects a floor above `ProgramConfig.version`. A floor above the program's own version would also reject the client shipping alongside it. Verified the test fails when the const is set above the workspace version (`9.0.0` → `MIN_COMPATIBLE_VERSION 9.0.0 exceeds the program version 0.31.0`).
- `initialize_global_state_test` and `accounts_test` pass — these exercise the `ProgramConfig` create and realloc/rewrite paths that stamp the const.

## Follow-ups (not in this PR)

- Every entry in `knownIncompatibilities` in `e2e/compatibility_test.go` has an upper bound at or below 0.24.0, so the whole map goes inert at a 0.30.0 floor. Clearing it would strand `versionRange`/`before()`/`isKnownIncompatible` as dead code, so it is left intact and documenting history.
- The client-side gate wraps its `ProgramConfig` fetch in `if let Ok(...)` (`smartcontract/cli/src/checkversion.rs`), so an RPC failure silently skips the version check entirely. Nothing on the Go side (controller, agent, telemetry) enforces `MinCompatVersion` at all.


e2e/compatibility_test.go
```diff
@@ -34,14 +34,12 @@ import (
...
-// workspace Cargo version was still "0.14.0", so the v0.14.1 CLI binary
-// self-reports its embedded version as 0.14.0 in the startup version check —
-// v0.15.0 is the first release whose embedded version is actually ≥ 0.14.1.
+// The 0.30.0 floor matches the program's MIN_COMPATIBLE_VERSION. Until each cluster is
+// redeployed, its onchain min_compatible_version still admits older clients, so this floor
+// is what keeps the matrix off them.
 var globalMinVersions = map[string]string{
-	"mainnet-beta": "0.21.0",
-	"testnet":      "0.21.0",
+	"mainnet-beta": "0.30.0",
```
nikw9944: This floor is compared against release **tag** versions (`EnumerateCompatibleVersions`, `e2e/internal/devnet/compat.go:149`), so tag `v0.30.0` enters the matrix — but that release was built from a commit whose workspace Cargo version was still `0.29.0`, so its binary self-reports 0.29.0 to the startup gate (which compares embedded versions, `smartcontract/cli/src/checkversion.rs:15`). Against the branch-built program (min=0.30.0 stamped by `doublezero init`), every `doublezero-0.30.0` command fails with `A new version of the client is available: 0.29.0 → 0.30.0` — this is the `e2e (shard 1)` failure on this PR. Set both entries to `"0.31.0"` (the first tag whose embedded version is ≥ 0.30.0; `v0.31.0` embeds `0.30.0`), and re-document the tag-vs-embedded offset here — the comment this PR deleted recorded exactly this failure mode for v0.14.1/v0.15.0.

smartcontract/programs/doublezero-serviceability/src/min_version.rs
```diff
@@ -1 +1,25 @@
-pub const MIN_COMPATIBLE_VERSION: &str = "0.21.0";
+// v0.30.0 is the first release whose `AccessPass` decoder understands
+// `EdgeSeat(Vec<FeedSeat>)` (#3954, #4030); older clients misparse every field after the
+// variant tag.
+pub const MIN_COMPATIBLE_VERSION: &str = "0.30.0";
```
nikw9944, line 10: The stated retirement boundary is off by one release: the `client/v0.30.0` binaries embed workspace version 0.29.0, so once a cluster deploys with min=0.30.0 they are hard-blocked at startup too — the earliest surviving release is v0.31.0 (embeds 0.30.0). No floor value can keep v0.30.0 while dropping v0.29.0, since both tags' binaries self-report 0.29.0; retiring v0.30.0 is forced by the EdgeSeat rationale, but the comment here and the CHANGELOG entry (line 13) should say the effective window in release terms is `[v0.31.0, <deployed>]` and note that v0.30.0 binaries self-report 0.29.0.

nikw9944 (changes requested):
The bump has a one-tag off-by-one against release tags: the `client/v0.30.0` release was built from a commit whose workspace version was still 0.29.0, so its binaries self-report 0.29.0 to the startup gate. That breaks the e2e compatibility matrix on this branch (shard 1 is red: tag v0.30.0 is enumerated as compatible, then every command fails `0.29.0 → 0.30.0`), and it means the floor retires the v0.30.0 release binaries too — the CHANGELOG/const-comment claim of "retiring every client older than v0.30.0" understates the impact. Fix the e2e floor to 0.31.0 and correct the stated retirement boundary. The `rust-test` failure is an unrelated RPC 503 in the accounts-compat step (all workspace tests, including the new unit test, passed) — re-run it.

ben-dz:
Both findings check out. Pushed 8a06916 with the e2e floor at 0.31.0 and corrected wording. Keeping the const at 0.30.0, for a reason worth writing down since it looks wrong at a glance.

## The tag/version drift is wider than one tag

Embedded `[workspace.package] version` at each release tag:

| tag | commit | date | self-reports |
|---|---|---|---|
| client/v0.26.0 | 54713b8fa | 06-05 | 0.26.0 |
| client/v0.27.0 | 2d0a9fd4b | 06-10 | 0.27.0 |
| client/v0.27.1 | — | 06-10 | 0.27.1 |
| client/v0.28.0 | ea7225304 | 06-26 | 0.28.0 |
| client/v0.29.0 | 3d15b8c56 | 07-02 | 0.29.0 |
| client/v0.30.0 | 354190e83 | 07-10 | **0.29.0** |
| client/v0.31.0 | 25ae6ee72 | 07-16 | **0.30.0** |

Both affected tags were cut by the release orchestrator (#3879, merged 07-08) before its version-bump PR merged:

- v0.30.0 tagged at 354190e83 (07-10 14:28); bump 23048e954 merged 15:49, 81 minutes later. `git merge-base --is-ancestor 23048e954 client/v0.30.0` is false.
- v0.31.0 tagged at 25ae6ee72 (07-16 02:46); bump f2248b0b9 merged 07-17 16:27.

Fixed by #4068 (merged 07-17 19:27), which tags the version-bump merge commit (305ae0c2f) and makes gate 1 wait for the version PRs (e96f49480). Every release before the orchestrator has embedded == tag, and v0.32.0 onward should too.

## Why the floor stays at 0.30.0

The gate compares self-reported version strings only, and two releases report the same one:

| release | self-reports | has EdgeSeat decoder (#3954 + #4030) |
|---|---|---|
| v0.29.0 | 0.29.0 | **no** |
| v0.30.0 | 0.29.0 | yes |
| v0.31.0 | 0.30.0 | yes |

Verified: neither f58432155 nor 8b795a3c8 is an ancestor of `client/v0.29.0`; both are ancestors of `client/v0.30.0`.

So v0.29.0 (crashes) and v0.30.0 (fine) are indistinguishable to the gate. Any floor at or below 0.29.0 admits the crasher, which makes 0.30.0 simultaneously the lowest safe floor and the widest one. Lowering it isn't available; the v0.30.0 release is excluded by a version string that lies, not by a real incompatibility.

## This is a live crash on testnet, not a hypothetical

`getProgramAccounts` on the testnet serviceability program, filtered to `AccountType::AccessPass` and decoded on the `accesspass_type` tag:

- **testnet**: 1314 passes — 1253 Prepaid, 60 SolanaValidator, **1 EdgeSeat** (`5x9DTsWCrAQtYKH4W7trEjoaCRHnzywsZp2oFX3sg5Zg`, `feed_seats` len 1)
- mainnet-beta: 2715 passes — 2581 Prepaid, 134 SolanaValidator, 0 EdgeSeat

That is the pass behind the `do-tn-rewards1` 52 GB abort. Any pre-decoder client enumerating access passes on testnet hits it today, which is the argument for taking the floor forward now rather than waiting.

## What changed in 8a06916

- `globalMinVersions` → 0.31.0 for both clusters, with the reason inline: the 0.30.0 tag can never pass the matrix because the gate rejects its binaries. Note the e2e floor is redundant with `MIN_COMPATIBLE_VERSION` in the local-devnet path (`doublezero init` stamps the const, so `max()` already yields 0.30.0) — it is load-bearing only for skipping the v0.30.0 tag, and should track the const again once v0.30.0 falls out of range.
- CHANGELOG and const comment now say the oldest admitted release is v0.31.0 and explain why v0.30.0 is collateral.

Also worth a look outside this PR: #4080's CHANGELOG entry justifies the in-place `feed_pk` → `feed_pks` change with "no EdgeSeat access pass has existed on any cluster — verified against the mainnet-beta data lake". True for mainnet-beta, not for testnet. Its own deploy note ("confirm no nonzero `feed_pk` account exists on each cluster at upgrade time") is therefore load-bearing on testnet rather than a formality, since a multicast connect against that pass would have recorded a feed the new layout reads differently.

`rust-test` was a 503 from the testnet ledger RPC in `make rust-program-accounts-compat`, unrelated to the diff — the new CI run covers it.

nikw9944 (approved):
lgtm — both round-one findings are fixed in 8a06916 (e2e floor now 0.31.0 with the tag-drift documented; CHANGELOG and const comment now state the oldest admitted release is v0.31.0 and why). Note `rust-test` is still red, but on the same external 503 from doublezerolocalnet.rpcpool.com in the accounts-compat step as before the push — all workspace tests pass; it needs that endpoint to recover, not a code change.

PR #4099, opened 2026-07-28
serviceability: add multi feed user subscription support

### LINKED ISSUE #4097: serviceability: add multi feed user subscription support
- program: the subscribe.rs re-gate + feed_pks append + seat accounting
- a CLI/SDK surface to actually drive "existing user subscribes to another feed" (there's no such command today — create-subscribe only creates)
- onchain tests, IDL bump, and Go SDK regen
- an e2e once the surface exists

martinsander00:
Closes #4097

## Summary

- `user subscribe --feed <PUBKEY|CODE>` subscribes a user to every multicast group a feed carries in their metro. It is repeatable, so one command can join several feeds, and each group is charged against the seat of the feed it came from. A group two feeds both carry is joined once, against the first of them.
- `--feed` is mutually exclusive with `--group`, which names its own groups, and rejects `--publisher`: a feed is a receive-only SKU, so publishing stays per-group. Both are rejected at parse time.
- `UpdateMulticastGroupRoles` now runs the full EdgeSeat feed gate instead of only re-validating coverage, so joining a group drawn from a second feed on the pass takes that feed's seat. Before this, it consumed that feed's capacity for free.
- Seats are held per feed, not per group: a second group inside a feed the user already holds does not tick again, and a removal releases the seat once the user's last group in that feed goes away. Whether a role change counts as an add or a removal is decided against current membership, so dropping one role while keeping the other neither re-gates nor frees the seat.
- An EdgeSeat pass admits multicast users only (`EdgeSeatIsMulticastOnly`, 103), both at connect and on a role add, and grants subscriber roles only (`EdgeSeatIsSubscribeOnly`, 106): a seat is per feed and no other user type takes one, and a feed is receive-only. Other pass types are unaffected and still allow a multicast role on any user type, gated by the multicast-group allowlist. Removals stay allowed everywhere so a user already in one of those states can be cleaned up.
- The SDK still resolves the gate itself when no feed is named (device from the user, feed from the pass's seat serving that metro and carrying the group), which is what the `--group` path and all the daemon role changes use. It preferentially picks a feed the user already holds, so a later `--group` on a group two feeds cover does not tick a second seat.
- New errors: `UserDeviceMismatch` (101), `UserFeedLimitExceeded` (104), `DeviceAccountRequired` (105).
- Out of scope: the e2e, which belongs in the feed-subscription repo driving `feed-subscription-admin` for the pass plus this CLI for the connect and role change.

## Example

```console
# join every group in two feeds, in the metro the user's device sits in
$ doublezero user subscribe --user <USER_PUBKEY> --feed shreds --feed blocks --subscriber
Updated roles for <GROUP_1>: <SIG>
Updated roles for <GROUP_2>: <SIG>
Updated roles for <GROUP_3>: <SIG>

# leave one of them again, releasing its seat
$ doublezero user subscribe --user <USER_PUBKEY> --feed blocks --subscriber false

# both rejected before any transaction
$ doublezero user subscribe --user <USER_PUBKEY> --feed shreds --group <GROUP_1> --subscriber
$ doublezero user subscribe --user <USER_PUBKEY> --feed shreds --publisher
```

## Testing Verification

- CLI: a feed expands to every group it carries with the seat attributed per group; two feeds sharing a group join it once against the first; `--group` with `--feed` and `--publisher` with `--feed` are rejected both at parse time and on a programmatically built command; a feed pubkey serving another metro is rejected before any transaction.
- Onchain (`feed_metro_gate_test.rs`): a group from a second feed ticks that feed's own seat; a second group inside a held feed does not tick again (that seat is `max_users: 1`, so a double tick would fail as `FeedSeatFull`); missing metro gate, feed not on the pass, group not in the feed, and a device that is not the user's are each rejected on their exact error code; the seat is released on the last group's removal, kept while another group in the feed remains, and kept when a removal omits the gate.
- SDK: metro-gate resolution for the no-feed path, including skipping a feed that carries the group but serves another metro.


smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
```diff
@@ -279,25 +280,36 @@ pub fn process_update_multicastgroup_roles(
...
             program_id,
             &accesspass,
             &device.exchange_pk,
             Some(mgroup_account.key),
-            feed_account,
+            Some(feed_account),
         )?;
+        // A seat is held per feed, not per group: joining a second group inside a feed the user
+        // already holds must not tick again.
+        if !user.feed_pks.contains(feed_account.key) {
```
nikw9944: This tick site never checks `user.user_type`, but `User::release_feed_seats` (state/user.rs:519) early-returns for non-multicast users. A unicast (IBRL) user on an EdgeSeat pass can be granted a multicast role (nothing in the SDK/CLI blocks it), which ticks the seat and records the feed in `feed_pks`; if the role is later removed without a resolvable gate (group dropped from the feed, feed deprovisioned, or the gate-less removal `DeleteUserCommand`'s cleanup sends), `DeleteUser` skips the release and the seat's `current_users` stays elevated permanently — recoverable only by rewriting seats via `SetAccessPassFeeds`. The connect-time tick in `create_user_core` is gated on `UserType::Multicast`, so the guard in `release_feed_seats` was safe before this second tick site. Fix: drop the `user_type != Multicast` early-return in `release_feed_seats` (feed_pks records exactly what was ticked and is empty for legitimate unicast users), and optionally also gate this tick on `UserType::Multicast` to mirror create_core.

smartcontract/sdk/rs/src/commands/multicastgroup/subscribe.rs
```diff
@@ -55,24 +55,68 @@ impl UpdateMulticastGroupRolesCommand {
...
+            .execute(client)?;
+
+            // Pick the pass's feed that serves the device's metro and carries the target group. A
+            // role add needs one; a removal does not, since the group may already have been dropped
+            // from its feed's group set.
+            let feed_pk = match self.feed_pk {
+                Some(feed_pk) => Some(feed_pk),
+                None => {
+                    let mut resolved = None;
+                    for seat in accesspass.feed_seats() {
```
nikw9944, line 105: Resolution takes the first seat whose feed serves the metro and carries the group, without consulting `user.feed_pks`. If two feeds on the pass serve the same metro and both carry the group: on an add this can pick a feed the user doesn't hold, ticking a second seat for a group the held feed already admits (and failing `FeedSeatFull` when that feed is at capacity even though the held feed would admit the change for free — the processor skips the tick for feeds already in `feed_pks`); on a removal it can resolve a feed the user never ticked, so nothing is released and the actually-held seat stays consumed until delete. Prefer seats whose `feed_key` is in `user.feed_pks` (the user is already fetched), falling back to the current first-match.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
```diff
@@ -279,25 +280,36 @@ pub fn process_update_multicastgroup_roles(
...
+    // EdgeSeat passes derive joinable groups from their feeds' metro→group map, so a role add is
+    // gated on a feed that is provisioned on the pass, serves the user's metro, and carries the
+    // target group.
+    if is_edge_seat && adding_role {
         let device_account = device_account.ok_or(DoubleZeroError::MetroMismatch)?;
         validate_program_account!(device_account, program_id, writable = false, "Device");
-        if user.device_pk != *device_account.key {
-            return Err(ProgramError::InvalidAccountData);
+        if device_account.key != &user.device_pk {
+            msg!("Device {} is not the user's device", device_account.key);
```
nikw9944: The log names the wrong key but not the expected one (same at line 334). House style is to state the expected value: `msg!("Device {} is not the user's device {}", device_account.key, user.device_pk)`.

nikw9944 (changes requested):
One High: the new post-activation seat tick is not gated on `user.user_type`, but `release_feed_seats` no-ops for non-multicast users, so a unicast EdgeSeat user granted a multicast role can permanently strand a feed seat at delete — a one-line fix in `release_feed_seats` restores the tick/release invariant. Two Mediums: SDK metro-gate resolution should prefer feeds the user already holds (`user.feed_pks`) to avoid ticking a redundant second seat on adds and silently skipping the release on removals, and the `FeedAccountRequired`/`FeedSeatFull` rejection paths on the new add path need integration tests. One Low on an error log omitting the expected device key.

- Medium: two rejection paths the new add path introduces are untested in feed_metro_gate_test.rs — `FeedAccountRequired` (device passed without the feed account, subscribe.rs:298) and `FeedSeatFull` on the post-activation tick (a second feed's seat already at max_users; the max_users:1 test proves no double tick, not that a genuinely full seat rejects). Per smartcontract/programs/CLAUDE.md Testing §4, both deserve integration tests; each is a few lines with the existing try_update_roles helper.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4099/changes/ed3c8c3c38a31c0a7c8bf81994d37f0529011275

smartcontract/sdk/rs/src/commands/multicastgroup/subscribe.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4099/changes/ed3c8c3c38a31c0a7c8bf81994d37f0529011275

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4099/changes/ed3c8c3c38a31c0a7c8bf81994d37f0529011275

nikw9944: **Double tick / missed release when a group is covered by two feeds on the pass.** This tick fires on any covering feed not yet in `feed_pks`, even when the call changes no membership (idempotent re-add, or a publisher-add on a group already held as subscriber — what daemon-cli `connect.rs:754-769` sends). Combined with the SDK resolving the gate by seat order without consulting `user.feed_pks` (`smartcontract/sdk/rs/src/commands/multicastgroup/subscribe.rs:84`), a user holding group g via F1 whose next call resolves to F2 consumes a second seat for the same group — or reverts with spurious `FeedSeatFull` — and a removal resolved to the wrong covering feed skips the release, stranding the held seat until delete. Fix both sides: tick only when the call adds the user's first membership of the target group (check `user.publishers`/`user.subscribers` before `update_user_multicastgroup_roles`), and have the SDK prefer feeds already in `user.feed_pks`. Add an SVM test for the two-covering-feeds shape.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
```diff
@@ -279,25 +280,36 @@ pub fn process_update_multicastgroup_roles(
...
-        && (value.publisher || value.subscriber)
-    {
+    let is_edge_seat = matches!(accesspass.accesspass_type, AccessPassType::EdgeSeat(_));
+    let adding_role = value.publisher || value.subscriber;
+    let mut accesspass_changed = false;
+
+    // EdgeSeat passes derive joinable groups from their feeds' metro→group map, so a role add is
+    // gated on a feed that is provisioned on the pass, serves the user's metro, and carries the
+    // target group.
+    if is_edge_seat && adding_role {
```
nikw9944: **Unicast EdgeSeat user ticks a seat that `DeleteUser` never releases.** There is no `user_type` check in this processor, so an Activated IBRL/unicast user under an EdgeSeat pass can take multicast roles; this block then ticks a seat and pushes onto `user.feed_pks`. But `User::release_feed_seats` early-returns for non-`Multicast` users (`state/user.rs:519-521`), so `DeleteUser` never gives the seat back — a permanent capacity leak, violating the documented multicast-only invariant on `feed_pks` and breaking parity with the connect path's `user_type == UserType::Multicast` gate (`create_core.rs:688`). Reject EdgeSeat role adds for non-`Multicast` users here (mirroring the create-path condition).

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
```diff
@@ -308,6 +320,42 @@ pub fn process_update_multicastgroup_roles(
...
+            &device.exchange_pk,
+            None,
+            Some(feed_account),
+        )?;
+        let feed = Feed::try_from(feed_account)?;
+        let holds_group_in_feed = user
+            .get_multicast_groups()
+            .iter()
+            .any(|group| feed.groups.contains(group));
+        if !holds_group_in_feed && user.feed_pks.contains(feed_account.key) {
```
nikw9944, line 392: **A membership no-op removal frees a billed seat after a routine feed-group rotation.** The release fires even when the removal changed nothing. Since `feed.groups` is freely replaceable via `FeedUpdate`, once group g rotates out of feed F, a user holding g through F can send a removal for any group they never held with the `(device, F)` gate: membership is untouched, the intersection is empty, F's seat frees — service retained on an uncounted seat, and the slot is re-tickable. Available to the pass's own `user_payer`. Fix: gate the release on the call actually removing one of the user's roles in the target group (return a membership-removed flag from `update_user_multicastgroup_roles`); or record the group→feed association at tick time instead of re-deriving from mutable `feed.groups`.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
```diff
@@ -279,25 +280,36 @@ pub fn process_update_multicastgroup_roles(
...
     }
 
-    // EdgeSeat passes derive joinable groups from their feeds' metro→group map. The seat was
-    // ticked at connect (CreateSubscribeUser), so post-activation role adds only re-validate
-    // coverage; they do not re-tick.
-    if matches!(accesspass.accesspass_type, AccessPassType::EdgeSeat(_))
-        && (value.publisher || value.subscriber)
-    {
+    let is_edge_seat = matches!(accesspass.accesspass_type, AccessPassType::EdgeSeat(_));
+    let adding_role = value.publisher || value.subscriber;
```
nikw9944: **Partial role removals are classified as adds and wedge once a group leaves its feed.** A call dropping one role while carrying the other — what daemon-cli sends on unsubscribe/unpublish (`multicast.rs:180-188, 345-353`) — runs the full add gate including group-in-feed. After the same `FeedUpdate` drift the release path deliberately tolerates, an EdgeSeat user who is both publisher and subscriber can no longer drop just one role (`GroupNotInFeed` onchain; the SDK bails before building). Compute `adding_role` relative to current membership: `(value.publisher && !user.publishers.contains(..)) || (value.subscriber && !user.subscribers.contains(..))`.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
```diff
@@ -308,6 +320,42 @@ pub fn process_update_multicastgroup_roles(
...
+    if let Some((device_account, feed_account)) = releasing_seat {
+        validate_program_account!(device_account, program_id, writable = false, "Device");
+        if device_account.key != &user.device_pk {
+            msg!("Device {} is not the user's device", device_account.key);
+            return Err(DoubleZeroError::UserDeviceMismatch.into());
+        }
+        let device = Device::try_from(device_account)?;
+        // Coverage is checked without a target group: the group being removed may already have been
+        // dropped from the feed's mutable group set, and that must not wedge the removal.
+        check_feed_metro_coverage(
```
nikw9944: **Stranded `feed_pks` entries can never be drained and later corrupt fresh counts.** Requiring `check_feed_metro_coverage` here means a feed de-provisioned from the pass wedges every gated removal (`FeedNotOnAccessPass`), stranding the `feed_pks` entry until `DeleteUser`. `SetAccessPassFeeds` resets `current_users` for a dropped-then-re-added feed (`set_feeds.rs:151-155`), so the stale entry then suppresses ticks and its eventual release/delete decrements a count belonging to other users — `remove_feed_user`'s `saturating_sub` hides the underflow. Coverage (and the device account, whose only use is the exchange for it) buys nothing on releases. Fix: drop the device account and coverage call on the removal path — owner-check the feed, then gate purely on `user.feed_pks.contains(feed)` + `!holds_group_in_feed`, letting stale entries drain. Consider `SetAccessPassFeeds` refusing to drop seats with `current_users > 0`, and `checked_sub` semantics in `remove_feed_user`.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
```diff
@@ -279,25 +280,36 @@ pub fn process_update_multicastgroup_roles(
...
             &device.exchange_pk,
             Some(mgroup_account.key),
-            feed_account,
+            Some(feed_account),
         )?;
+        // A seat is held per feed, not per group: joining a second group inside a feed the user
+        // already holds must not tick again.
+        if !user.feed_pks.contains(feed_account.key) {
+            accesspass.try_add_feed_user(feed_account.key)?;
+            user.feed_pks.push(*feed_account.key);
```
nikw9944, line 362: `user.feed_pks` grows here with no cap enforced at the mutating instruction. Concurrent growth is bounded by seats on the pass, but stranded entries across `SetAccessPassFeeds` rotations make it unbounded over the account's lifetime — the account-stored-Vec heap-brick class. Reject the push when `user.feed_pks.len() >= MAX_ACCESS_PASS_FEEDS` (`set_feeds.rs:27`), with a boundary test.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
```diff
@@ -279,25 +280,36 @@ pub fn process_update_multicastgroup_roles(
...
-    {
+    let is_edge_seat = matches!(accesspass.accesspass_type, AccessPassType::EdgeSeat(_));
+    let adding_role = value.publisher || value.subscriber;
+    let mut accesspass_changed = false;
+
+    // EdgeSeat passes derive joinable groups from their feeds' metro→group map, so a role add is
+    // gated on a feed that is provisioned on the pass, serves the user's metro, and carries the
+    // target group.
+    if is_edge_seat && adding_role {
         let device_account = device_account.ok_or(DoubleZeroError::MetroMismatch)?;
```
nikw9944: A missing device account surfaces as `MetroMismatch`, which will misdirect every pre-gate client hitting this path; the missing feed already gets a precise `FeedAccountRequired`. Use a dedicated variant (e.g. `DeviceAccountRequired`) while the path is new and no client matches on the code yet.

smartcontract/sdk/rs/src/commands/multicastgroup/subscribe.rs
```diff
@@ -21,9 +22,8 @@ pub struct UpdateMulticastGroupRolesCommand {
     pub user_pk: Pubkey,
     pub publisher: bool,
     pub subscriber: bool,
-    /// Reserved for the EdgeSeat feed metro gate (the user's device + covering Feed). Not appended
-    /// by this builder — the authorized-transaction layout has no slot after the trailing
-    /// `[payer, system, permission]`; post-activation re-gating is deferred to #1699.
+    /// EdgeSeat feed metro gate. Left as `None`, this command resolves the device from
+    /// `user.device_pk` and the feed from the pass's seats; set either to override.
     pub device_pk: Option<Pubkey>,
```
nikw9944: The `device_pk` override can only reproduce the default (`user.device_pk`) or guarantee a `UserDeviceMismatch` revert; no in-repo caller sets it. Remove the field and always use `user.device_pk`.

smartcontract/cli/src/user/subscribe.rs
```diff
@@ -75,6 +81,22 @@ impl SubscribeUserCliCommand {
             group_pks.push(group_pk);
         }
 
+        // An omitted --feed leaves the metro gate to the SDK, which picks the pass's feed serving
+        // the user's metro.
+        let feed_pk = match &self.feed {
```
nikw9944: One `--feed` is applied to every group in the multi-group loop, so groups outside that feed fail `GroupNotInFeed` mid-loop after earlier groups already committed; and on the pubkey branch the `GetDeviceCommand` fetch is wasted (`GetFeedCommand` ignores `exchange` for pubkeys). Reject `--feed` with more than one group, and fetch the device only on the code branch.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
```diff
@@ -194,7 +195,7 @@ pub fn process_update_multicastgroup_roles(
     validate_program_account!(
         accesspass_account,
         program_id,
-        writable = false,
+        writable = true,
```
nikw9944: The accesspass account is now asserted writable for every caller, including non-EdgeSeat calls that never write it. Nothing in-repo passes it read-only, but this tightens the external instruction interface and the CHANGELOG entry doesn't mention it. Either scope the assert to EdgeSeat or add one changelog sentence.

smartcontract/sdk/rs/src/commands/multicastgroup/subscribe.rs
nikw9944, line 105: Gate resolution issues one sequential `GetFeedCommand` RPC per seat (up to 64), round-tripping `seat.feed_key` through `to_string()`/parse; `DeleteUserCommand`/`RequestBanUserCommand` invoke it per group, making an EdgeSeat delete cascade O(groups × seats) RPCs. Call `client.get(seat.feed_key)` directly, or batch via `get_multiple_accounts`.

nikw9944 (changes requested):
Two High findings need fixing before merge, both in the new seat accounting. (1) When two feeds on one pass cover the same group, a membership-no-op role add ticks a second seat (the processor only checks `feed_pks`, and the SDK resolves the gate by seat order without consulting `user.feed_pks`), and a removal resolved to the wrong covering feed skips the release — seat counts silently drift, no adversary needed. (2) The processor has no `user_type` check, so a unicast user under an EdgeSeat pass can tick a seat and gain `feed_pks` entries that `DeleteUser` never releases (`release_feed_seats` early-returns for non-Multicast users) — a permanent seat leak that also breaks parity with the connect path's `user_type == Multicast` gate. Mediums cover the seat lifecycle under feed mutation: a membership-no-op removal can free a billed seat after a `FeedUpdate` group rotation while keeping the membership; partial role removals (carry one role) are classified as adds and wedge once a group leaves its feed; stranded `feed_pks` entries are un-drainable and later corrupt re-provisioned counts; and `feed_pks` has no cap at the mutating instruction.

- low: tests/feed_metro_gate_test.rs — no test asserts `FeedSeatFull` (Custom(95)) surfacing through UpdateMulticastGroupRoles, the one new error `try_add_feed_user` can raise on a post-activation add; add a `seat(feed2, 0)`-style case.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
martinsander00: 1ddd714ba46d2edfc507f76b5280c8c11c547ada

smartcontract/sdk/rs/src/commands/multicastgroup/subscribe.rs
martinsander00: 1ddd714ba46d2edfc507f76b5280c8c11c547ada

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
martinsander00: 1ddd714ba46d2edfc507f76b5280c8c11c547ada

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
martinsander00: e7964b61ef6a8c1f9c591c40f7a5f916a17bf66c

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
martinsander00: a1016dd333b0da3dc63a1260043948a6bbec2443

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
martinsander00: a1016dd333b0da3dc63a1260043948a6bbec2443

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
martinsander00: a1016dd333b0da3dc63a1260043948a6bbec2443

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
martinsander00: a1016dd333b0da3dc63a1260043948a6bbec2443

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
martinsander00: a1016dd333b0da3dc63a1260043948a6bbec2443

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
martinsander00: a1016dd333b0da3dc63a1260043948a6bbec2443

smartcontract/sdk/rs/src/commands/multicastgroup/subscribe.rs
martinsander00: a1016dd333b0da3dc63a1260043948a6bbec2443

smartcontract/cli/src/user/subscribe.rs
martinsander00: a1016dd333b0da3dc63a1260043948a6bbec2443

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
martinsander00: a1016dd333b0da3dc63a1260043948a6bbec2443

smartcontract/sdk/rs/src/commands/multicastgroup/subscribe.rs
martinsander00: a1016dd333b0da3dc63a1260043948a6bbec2443

nikw9944:
Note for reviewers: my 19:18 review (One High) is superseded by the 19:40 review (Two High findings) — two Roadbike instances reviewed this PR concurrently, apologies for the noise. Please work against the 19:40 review.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
```diff
@@ -308,6 +372,30 @@ pub fn process_update_multicastgroup_roles(
...
+    if let Some(feed_account) = releasing_feed {
+        if feed_account.owner != program_id {
+            return Err(DoubleZeroError::InvalidAccountOwner.into());
+        }
+        let feed = Feed::try_from(feed_account)?;
+        let holds_group_in_feed = user
+            .get_multicast_groups()
+            .iter()
+            .any(|group| feed.groups.contains(group));
+        if !holds_group_in_feed && user.feed_pks.contains(feed_account.key) {
```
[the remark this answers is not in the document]
martinsander00: issue should be solved in the UpdateFeed admin cmd, opening an issue in the infra repo: https://github.com/malbeclabs/infra/issues/2113

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
```diff
@@ -308,6 +372,30 @@ pub fn process_update_multicastgroup_roles(
...
+        leading.last().copied()
+    } else {
+        None
+    };
+    if let Some(feed_account) = releasing_feed {
+        if feed_account.owner != program_id {
+            return Err(DoubleZeroError::InvalidAccountOwner.into());
+        }
+        let feed = Feed::try_from(feed_account)?;
+        let holds_group_in_feed = user
```
nikw9944, line 388: **Known residual, tracked in malbeclabs/infra#2113 — flagging so it stays visible in this PR's record.** `holds_group_in_feed` derives from the feed's *current* `groups` set, so when a user holds two groups charged to one feed and the catalog rotates the set, dropping one group releases the seat while the user keeps the other (verified still reproducible at this head). The in-PR fixes (`dropping_group`, no-op protection) close the cheaper levers; the remaining case needs the admin-side ordering the open infra issue describes (unsubscribe users before rotating a feed's groups) or per-group attribution on `User`. No action required in this PR given the tracker — consider linking infra#2113 in the code comment above this block so the residual is discoverable.

nikw9944 (approved):
Re-check confirmed no drift: the PR head is still `b3f8464b` — the exact commit fully verified in the previous round against a synced local checkout with the SVM gate/delete suites (16 gate tests) and the 12 reworked CLI subscribe tests passing locally. No new commits, no new review comments beyond the already-processed author deferral (tracking issue malbeclabs/infra#2113, verified open), and no steering attempts. All previously blocking findings remain verified fixed in this code: the seat tick is membership-gated with the SDK preferring held feeds (double-tick closed); EdgeSeat passes admit multicast users only (103) and grant subscriber roles only (106), eliminating both the seat-leak and the carry-role wedge class; releases fire only on a full group drop (`dropping_group`), closing the no-op-removal and partial-removal levers; the removal path is coverage-free so stale `feed_pks` entries drain; the `feed_pks` cap (104) and `DeviceAccountRequired` (105) are in place; the accesspass-writable assert is scoped to the EdgeSeat write path; and the CLI `--feed` expansion rework has solid coverage with e2e on the new flag syntax. Two Lows carry over unchanged, re-confirmed at this head: the 95/104 error paths lack tests through this instruction, and the multi-group feed-rotation undercount is deliberately deferred to the admin-side fix tracked in malbeclabs/infra#2113.

- low: smartcontract/programs/doublezero-serviceability/tests/feed_metro_gate_test.rs — two new error paths still lack a test through UpdateMulticastGroupRoles: `FeedSeatFull` (Custom(95)) on a post-activation add (e.g. second feed provisioned with `seat(feed2, 0)` — currently only referenced in a comment), and `UserFeedLimitExceeded` (Custom(104)) at the `MAX_ACCESS_PASS_FEEDS` boundary. The other new paths (103, 106, rotation) have tests; these two would complete the set per the repo's all-error-paths convention.
- note: re-check outcome — no changes since the previous round: PR head unchanged at `b3f8464b` (open, not merged), local checkout at the same SHA where all suites passed, and the only comment since the prior cutoff is the author's already-verified deferral referencing the open malbeclabs/infra#2113. All fixed-finding verifications from the previous round therefore remain valid byte-for-byte; the two Lows above are the complete outstanding set.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
martinsander00: created this to track it: https://github.com/malbeclabs/infra/issues/2113

martinsander00:
will write a new `doublezero user feed` command instead of messing around with `doublezero user subscribe`

PR #4100, opened 2026-07-28
tools/solana: fix JSON-RPC retry classification and consolidate RPC construction

### LINKED ISSUE #4098: Go JSON-RPC retry classification is broken; Go and Rust disagree on what's retryable
# Go JSON-RPC retry classification is broken; Go and Rust disagree on what's retryable

**This is a problem brief, not an implementation plan.** Design the fix yourself, including
where the shared code should live. The "Open design questions" section is the part we
deliberately have not decided.

## What happened

On 2026-07-28 the DZ ledger RPC endpoint (RPCPool host `doublezerolocalnet.rpcpool.com`,
shared by DZ testnet and devnet) began returning intermittent HTTP 503 "Service unavailable"
on `getProgramAccounts`. It lasted 4+ hours and eventually hit ~20 components across both
envs. RPCPool confirmed it as a routing/networking fault in their load-balancer layer — the
chain itself was healthy the whole time.

The Rust components rode it out. `shred-oracle` logged `WARN Retrying RPC read … 503 Service
Unavailable` and its work completed normally.

Every Go component that reads the ledger logged one ERROR per poll and gave up. The
controller sat at **100% fetch failure for over three hours**, serving devices purely from a
cache populated before the incident started. Nothing was user-visible only because every
affected component happens to keep a last-good cache — but any restart during that window
would have converted graceful degradation into a hard outage (controller `GetConfig` returns
`NotFound` on a cache miss; `state-ingest` starts with `ready=false`; `doublezerod` with no
cache returns a hard error).

Correct retry behavior would have absorbed nearly all of this.

## Root cause

`tools/solana/pkg/jsonrpc/retry.go`, `isRetryableJSONRPC`. Two of its three classification
branches are unreachable in production:

- **Line 148** — `type hasStatusCode interface{ StatusCode() int }`. In solana-go v1.21.0,
  both `*jsonrpc.RPCError` and `*jsonrpc.HTTPError` expose `Code` as a struct **field**, not
  a method. No production error type satisfies this interface, so the entire
  429/500/502/503/504 branch is dead.
- **Line 162** — `type hasCode interface{ Code() int }`. Same field-vs-method problem, and
  even if satisfied the switch only lists `-32005/-32004/-32003/-32429`.

That leaves only the substring match at lines 129–145, and `"service unavailable"` is not in
that list. So the 503s were classified non-retryable and `doRetry` returned after one attempt.

**Why no test caught it:** `retry_test.go:230` defines a local `statusCodeErr` type *with* a
real `StatusCode() int` method and asserts `{"http 503", statusCodeErr(503), true}`. The
fixture satisfies an interface that no production error resembles, so the test passes while
production does not retry.

**Empirically confirmed:** every `state-ingest` failure was logged at a fixed `:53.16–.19s`
offset on its 1-minute ticker. Four attempts with 0.5s/1.0s/1.5s backoff would have pushed
the log ≥3s later. Single attempt, no retry.

## Three inconsistent implementations already in the monorepo

This is why a shared piece is probably the right shape.

**1. JSON-RPC layer — broken for HTTP status codes.**
`tools/solana/pkg/jsonrpc/retry.go` + `tools/solana/pkg/rpc/retry.go` (`NewWithRetries`).
Callers: `controlplane/controller/cmd/controller/main.go:216,225`,
`telemetry/global-monitor/cmd/global-monitor/main.go:189,212`,
`telemetry/flow-enricher/cmd/flow-enricher/main.go:139`,
`telemetry/state-ingest/cmd/server/main.go:83`.

**2. HTTP transport layer — works, but duplicated.**
`sdk/shreds/go/rpc.go` and `sdk/revdist/go/rpc.go` each define a near-identical
`retryHTTPClient` wrapping `http.Client.Do`, retrying on `resp.StatusCode == 429 ||
resp.StatusCode >= 500`. This approach is correct — it inspects the real HTTP response before
solana-go converts it into an `RPCError`. Weaknesses: copy-pasted twice, uses `time.Sleep`
rather than honoring `ctx`, fixed `2s × attempt` backoff, no jitter.

**3. No retry at all.** ~12 production call sites construct bare `rpc.New` / `solanarpc.New`:

| Call site | Notes |
|---|---|
| `client/doublezerod/internal/runtime/run.go:80` | client daemon, ~60 ho

ben-dz:
## Summary of Changes

* Retry HTTP 429/5xx from a Solana JSON-RPC endpoint. `isRetryableJSONRPC` asserted on `StatusCode() int` / `Code() int` interfaces that no solana-go error type implements — both expose their code as a struct field — so the whole status-code branch was dead and every Go ledger reader gave up after one attempt.
* On 2026-07-28 that meant 4+ hours of 100% fetch failure across ~20 components against a healthy chain, absorbed only because each happens to keep a last-good cache. A restart in that window would have been a hard outage.
* Classification now reads the concrete `*jsonrpc.HTTPError` and `*jsonrpc.RPCError` values, covering both the undecodable-body case and the decodable JSON-RPC error envelope the incident actually arrived in.
* `sendTransaction` and `requestAirdrop` are never retried regardless of the options a caller passes, so nothing can resubmit a transaction the endpoint already accepted.
* Backoff is jittered, so the ~60 `doublezerod` hosts and dozen services reading one endpoint don't retry in lockstep against an endpoint already shedding load.
* One shared constructor (`tools/solana/pkg/rpc.New`) replaces the three duplicated transport-level retry loops (`sdk/shreds`, `sdk/revdist`, internet-latency collector — each ignored `ctx` and slept a fixed unjittered backoff) and adds retries to the ~13 ledger readers that had none, including `doublezerod` and the on-device telemetry agent's netns path.
* New `doublezero_solana_rpc_retries_total` / `doublezero_solana_rpc_retries_exhausted_total` counters, labelled by method, make the next episode visible without reading logs.
* Fixes #4098

## Testing Verification

* `TestNewWithRetries_RetriesServiceUnavailable` reproduces the incident end-to-end through real solana-go against an `httptest` server: HTTP 503 with a JSON-RPC error body, HTTP 503 with an HTML body, and 429. Verified to **fail against the pre-fix classifier** (1 attempt) and pass now (full 4-attempt budget). A 400 `-32602` case asserts a permanent error is still not retried.
* Every classifier branch has a positive and negative case built from real `*jsonrpc.HTTPError` / `*jsonrpc.RPCError` values rather than hand-rolled fixtures — the old test's `statusCodeErr` fixture had a real `StatusCode()` method, which is why it passed while production didn't retry.
* Write safety is pinned at both layers: `sendTransaction` / `requestAirdrop` single-attempt, a batch containing one non-idempotent member not retried, and `SendEncodedTransaction` against a 503 server hitting it exactly once.
* Retry exhaustion asserted to surface as an error, never a silent empty result — the `getProgramAccounts` empty-result guard and the controller's `0 devices found on-chain` check both depend on that.
* Jitter test asserts waits stay within `[d/2, d]` and that not every wait hits the ceiling, so an unjittered regression fails.
* ctx-cancel-during-backoff asserts the returned error matches *both* the ctx error and the 503 behind it, so an operator sees the cause and not just "context deadline exceeded".
* `RequestTimeout` enforcement tested behaviorally (slow server, request aborts) at the shared layer, replacing the equivalent test deleted with `sdk/shreds`' private HTTP client; `sdk/shreds` separately pins that its request bound stays under the ~56s blockhash window.
* Pre-existing environmental failures (root for netlink/ICMP, Docker for containers) confirmed identical on a clean tree.


tools/solana/pkg/jsonrpc/retry.go
```diff
@@ -76,31 +80,115 @@ func (c *retryingJSONRPCClient) CallBatch(ctx context.Context, requests jsonrpc.
...
+		return true
+	}
+	return false
+}
+
+// isRetryableRPCCode covers the provider-specific "busy / retry later" JSON-RPC
+// codes (node is behind, rate limited, transaction history not available yet).
+func isRetryableRPCCode(code int) bool {
+	switch code {
+	case -32005, -32004, -32003, -32429:
```
nikw9944: `-32003` is Agave's `JSON_RPC_SERVER_ERROR_TRANSACTION_SIGNATURE_VERIFICATION_FAILURE` — a deterministic rejection, not a busy/retry-later code. This branch was dead before this PR, so the list was never vetted by production traffic; now it's live, a simulateTransaction-style call that fails signature verification burns the full retry budget (3 extra requests + ~3s) against the endpoint every time. The doc comment also mislabels the set: "transaction history not available yet" is `-32011`, not any listed code. Suggest dropping `-32003` (or substituting `-32011` if that was the intent) and correcting the comment.

tools/solana/pkg/jsonrpc/retry.go
```diff
@@ -111,6 +199,31 @@ func isRetryableJSONRPC(err error) bool {
...
+	// *HTTPError carries the real HTTP status and is returned when a >=400 response
+	// body did not decode as a JSON-RPC envelope. *RPCError carries the code from a
+	// body that did decode; a provider fronting a load balancer commonly puts the
+	// HTTP status there, which is how those 503s actually arrived.
+	var httpErr *jsonrpc.HTTPError
+	if errors.As(err, &httpErr) && isRetryableHTTPStatus(httpErr.Code) {
+		return true
+	}
+
+	//
```
nikw9944: Stray empty `//` line orphaned above the `*RPCError` comment block — editing leftover, delete it.

nikw9944 (commented):
One classification fix worth making before merge: `-32003` in the now-live retryable-code list is Agave's transaction-signature-verification failure — a deterministic rejection — so retrying it burns the full budget against an already-degraded endpoint (bounded to simulateTransaction-style calls since sendTransaction is excluded). Drop it or substitute `-32011` if history-not-available was the intent, and fix the `isRetryableRPCCode` comment, which describes `-32011` rather than any listed code. Plus one nit.

nikw9944 (approved):
Approving — this closes exactly the classifier gap that one-shot the QA payments probe on 07-27 (malbeclabs/infra#2100), and the incident-reproducing test (verified failing pre-fix) is the right way to pin it. Deleting the interface branches rather than keeping them, and hardcoding `sendTransaction`/`requestAirdrop` as never-retried, are both better calls than what I'd sketched independently.

Three things before/around merge:

1. #4098 frames the problem as Go *and* Rust disagreeing on retryability, and this fixes Go only — the Rust SDK (`is_retryable_rpc_error` in `smartcontract/sdk/rs/src/client.rs`) still won't retry a decoded JSON-RPC envelope like `-32603 "Service unavailable"`; Rust survived 07-28 only because those 503s arrived at transport level. Can we file the Rust-parity follow-up now while the context is fresh?
2. lake uses `NewWithRetries` in 4 places and will inherit never-retried sends on its next SDK bump — worth one line in the PR body/changelog as a heads-up to that team.
3. `shard-e2e (shard 3)` / `TestE2E_UserFailurePaths` looks unrelated (Rust CLI/shred-oracle path; the overlapping `e2e (shard 3)` job passed), but let's re-run it to green or note the triage before merging.

tools/solana/pkg/jsonrpc/retry.go
ben-dz: Dropped in 708041bb.

Also rewrote the doc comment to describe the set that is actually listed, with a one-line justification per code: `-32005` NODE_UNHEALTHY (node catches up, or the LB routes the retry to a backend that already has), `-32004` BLOCK_NOT_AVAILABLE (slot arrives, or another backend holds it), and `-32429` — flagged explicitly as **not** an Agave code, minted by providers fronting Agave to mirror HTTP 429 inside a JSON-RPC envelope, which is why it reaches us as `*RPCError` rather than `*HTTPError`.

I did **not** substitute `-32011`. It describes a node that carries no long-term history at all rather than one that is momentarily busy, so retrying the same node cannot change the answer — adding it would repeat the mistake being fixed. Both exclusions are recorded in the comment under "Deliberately absent" so the next person does not re-add them.

Tests at both layers: a classifier unit case, and an end-to-end case through real solana-go asserting exactly one HTTP attempt. Mutation-checked — re-adding `-32003` to the switch fails both.

tools/solana/pkg/jsonrpc/retry.go
ben-dz: Deleted in 708041bb.

tools/solana/pkg/rpc/retry_test.go
```diff
@@ -121,10 +124,130 @@ func TestNewWithHeadersAndRetries_SendsHeaders(t *testing.T) {
...
+				var req struct {
+					ID any `json:"id"`
+				}
+				require.NoError(t, json.Unmarshal(body, &req))
+
+				out := tc.body
+				if strings.Contains(out, "%q") {
+					out = fmt.Sprintf(out, fmt.Sprint(req.ID))
+					w.Header().Set("Content-Type", "application/json")
+				}
```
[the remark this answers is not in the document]
ben-dz: Fixed in 708041bb. The helpers now decode the id as `json.RawMessage` and echo it back byte-for-byte with `%s`, so a numeric request id comes back numeric. Verified the actual wire change: `{"id":"1"}` → `{"id":1}`. Applied to all three server helpers in the file, not just this one — leaving two with the same pattern invites the same confusion later.

One correction to the diagnosis, since it affects whether the existing cases can be trusted: this was **not** reducing the fidelity of the test. solana-go types `RPCResponse.ID` as `any` and does not verify it for single (non-batch) calls, so the `*jsonrpc.RPCError` path was genuinely exercised and every table case was valid as written. The tests were not rewritten on the assumption they were broken.

Worth fixing regardless, for the reason the report gets right in substance: echoing a mismatched-type id is non-conformant to JSON-RPC 2.0, and it is a latent trap because `CallBatch` correlates responses to requests by id — anyone extending these cases to cover batch retry would have hit silent misbehavior.

ben-dz:
## Review feedback addressed in 708041bb

Inline replies on the three code comments. Three further items below.

### Default `RequestTimeout`: 5 min → 10s

The classifier fix made retries live without bounding a single attempt, so a hung endpoint could consume `MaxAttempts` × 5 min on one logical call. Fixed. I picked the value against measurements rather than copying 15s from #3973.

Heaviest calls we actually make, measured against live mainnet:

| call | payload | wall |
| --- | --- | --- |
| unfiltered `getProgramAccounts`, serviceability (DZ ledger) | 2.90 MB raw / ~1.0 MB gzipped | 0.39–0.44s |
| same, testnet | 0.96 MB | 0.33s |
| `getLeaderSchedule` (Solana mainnet) | 1.85 MB | 0.27s |
| `getSignaturesForAddress` limit 1000 | 100 KB | 0.27s |
| `getAccountInfo`, largest telemetry PDA (111 KB) | 111 KB | 0.20s |

10s leaves >20x headroom on a full program scan, so it will not manufacture false failures. Worst case for one logical call is ~43s (4 × 10s + ~3s jittered backoff).

On the controller: `cacheFetchTimeout` **does** already protect it — `fetch()` wraps `updateStateCache` in a 30s context, and doublezerod has the same 30s via `onchain.DefaultRPCTimeout`. But at a 5-minute per-attempt bound those callers got exactly one attempt before their own budget expired, so retry never actually fired for them. The caller that needs this default is **state-ingest**: `Refresh` hands its root context straight to `GetProgramData` with no call-site bound at all, on a 60s tick. ~43s fits inside that; 20 minutes does not.

One coupling bug found while doing this: `IdleConnTimeout` reused the same `defaultTimeout` constant, so lowering it would have churned a pooled connection every poll tick. Split onto its own `defaultIdleConnTimeout = 5 * time.Minute`.

`sdk/shreds` and the internet-latency collector keep their explicit 15s. Note `sdk/revdist` was passing `Options{}`, not a short value, so it had been inheriting the 5-minute default — it now gets 10s. `TestNew_DefaultRequestTimeoutIsBounded` pins the value and both bounds.

### lake — checked, no action needed

Built and tested lake against this branch via a local `replace`. Findings:

- **lake never sends transactions.** No `SendTransaction` / `SendEncodedTransaction` / `RequestAirdrop` anywhere, so the never-retry-sends rule is a no-op there.
- **Six call sites, all `NewWithRetries(url, nil)`**, so all inherit the new 10s default: `indexer/cmd/indexer/main.go` (×4), `admin/internal/admin/backfill_{device_link,internet_metro}_latency.go`. Every real call path is in the table above; the heaviest is 0.44s.
- The telemetry SDK only issues `GetAccountInfo` on specific PDAs, never a program scan. Worth noting because an unfiltered `getProgramAccounts` on the telemetry program returns **2.29 GB in 55.7s** (200,696 accounts, 3.5 GB total) — that would break at 10s, but nothing in either repo makes that call. Flagging it as a landmine for anyone tempted to add one.
- lake's unfiltered shreds-program scan goes through `shreds.NewRPCClient` (explicit 15s), unaffected by this change; measured 0.3 MB / 0.28s.
- Test results are **identical** with and without the change (the only diff is `(cached)` markers). Remaining failures in my sandbox are all `rootless Docker not found` from testcontainers, present on baseline too.

lake pins `doublezero v0.8.1-0.20260724204114-f7981ab6f7a3` (f7981ab6, 07-24), so nothing reaches it until someone bumps the dep. Still worth a nod from whoever owns lake, since I verified behavior rather than intent.

### Not changed: `sendTransaction` / `requestAirdrop` retry policy

Left exactly as-is — unconditional `maxAttempts = 1`, no escape hatch, no parameterization, test untouched. For awareness only: this conflicts with a planned change in `malbeclabs/doublezero-shreds` that wants retries on `SendTransactionWithOpts` in the QA payments e2e test (malbeclabs/infra#2100). Whether send-retry should become opt-in is a human decision that has not been made, and this PR does not presuppose one.

### CHANGELOG

Extended with the `-32003` classification correction, the timeout default change, and a note that ~30s is the established retry-budget convention for QA-test callers — those callers should pass `Retry` with a larger `MaxAttempts` explicitly rather than silently inheriting the shorter default.

---

Unrelated: `shard-e2e (shard 3)` is red for a pre-existing cross-repo skew — doublezero-shreds' e2e calls `doublezero user create-subscribe --feed`, a flag the CLI does not have yet. Same failure on `main` at 3f9f420e and 9d28b9f0. #4087 is the fix.

PR #4105, opened 2026-07-29
ci: retry go module downloads in the daily release build

nikw9944:
## Summary of Changes

- Pre-populate the Go module cache with up to 3 attempts before GoReleaser runs, so a dropped module zip no longer fails the daily release build. GoReleaser has no download-retry knob of its own, and the `go` command does not retry a mid-transfer HTTP/2 `INTERNAL_ERROR`, so the retry has to wrap a separate, idempotent `go mod download`.
- `release.daily.yml` is the shared build job for all devnet component releases, so this covers all 10 components in the daily matrix.

Observed in [run 30472842137 attempt 2](https://github.com/malbeclabs/doublezero/actions/runs/30472842137), where `Build client` failed with:

```
github.com/google/gopacket@v1.1.19: read "https://proxy.golang.org/github.com/google/gopacket/@v/v1.1.19.zip":
stream error: stream ID 5; INTERNAL_ERROR; received from peer
```

The same run also logged `Failed to restore: "/usr/bin/tar" failed with exit code 2` from `actions/setup-go`, i.e. the module cache restore failed — which is why the build had to reach the network for every module in the first place. Retrying the download covers that combination too.

Retrying the GoReleaser step itself was the other option, but it publishes a GitHub release, uploads to Cloudsmith, and announces to Slack; re-running it on any failure is not side-effect free.

## Testing Verification

- Confirmed against the run's failure annotations that the failing download was a module zip fetch during GoReleaser's Go build, not the GoReleaser binary download or the toolchain install.
- `go mod download` at the repo root is the correct invocation for the modules these builds need (verified it resolves the full build list; the sandbox blocks `storage.googleapis.com`, where `proxy.golang.org` redirects zip fetches, so the download itself could not complete locally).
- The retry loop follows the existing `pull_with_retry` pattern in `e2e.yml`.


.github/workflows/release.daily.yml
```diff
@@ -29,6 +29,21 @@ jobs:
...
       - uses: Swatinem/rust-cache@v2
+      # proxy.golang.org occasionally drops a module zip mid-transfer (HTTP/2
+      # INTERNAL_ERROR). The go command does not retry, so the failure surfaces
+      # as a GoReleaser build error partway through the release. Populate the
+      # module cache up front, where a retry is cheap and side-effect free.
+      - name: Download Go modules
+        run: |
+          for attempt in 1 2 3; do
+            echo "go mod download (attempt $attempt/3)"
+            if go mod download; then
```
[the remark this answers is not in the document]
nikw9944: `release.daily.yml` never builds `s3-uploader`, so there is nothing to pre-populate for it here.

The workflow builds whatever `inputs.component` names (`args: release -f release/.goreleaser.devnet.${{ inputs.component }}.yaml`), and no caller passes `s3-uploader`. The full set of callers is the 10-entry matrix in `release.devnet.all.daily.yml` (agent, controller, funder, monitor, client, device-telemetry-agent, internet-latency-collector, qa-agent, geoprobe-agent, geoprobe-target) plus the per-component `release.devnet.*.yml` wrappers, which pass those same names.

Every one of those components resolves to a directory inside the root module — `monorepo.dir` is `controlplane/agent`, `controlplane/telemetry`, `controlplane/internet-latency-collector`, `e2e/`, etc. `controlplane/s3-uploader/go.mod` is the repo's only second module, and it is built by `release.s3-uploader.yml` against `release/.goreleaser.testnet.s3-uploader.yaml`, a separate workflow this PR does not touch. `release/.goreleaser.devnet.s3-uploader.yaml` exists but is only exercised by `release.pipeline.validation.yml` for config validation, not by this build path.

So a root `go mod download` covers every module this workflow actually builds, and keying `setup-go`'s cache on a second `go.sum` would have no effect on it. If the s3-uploader release wants the same retry, that belongs in `release.s3-uploader.yml` as its own change.

.github/workflows/release.daily.yml
```diff
@@ -29,6 +29,21 @@ jobs:
           go-version-file: go.mod
           cache: true
       - uses: Swatinem/rust-cache@v2
+      # proxy.golang.org occasionally drops a module zip mid-transfer (HTTP/2
+      # INTERNAL_ERROR). The go command does not retry, so the failure surfaces
+      # as a GoReleaser build error partway through the release. Populate the
+      # module cache up front, where a retry is cheap and side-effect free.
+      - name: Download Go modules
```
ben-dz, line 44: This is a best-effort cache warm-up, but as written it is a hard, unbounded gate — it can fail the daily release in two ways it previously could not.

1. **No bound on the fetch.** The retry covers a *dropped* transfer, but a *stalled* one (connection open, no bytes, no error) makes `go mod download` hang. The `release` job sets no `timeout-minutes`, so the step runs to GitHub's six-hour job default. Previously this fetch happened inside the GoReleaser step, which GoReleaser bounds with its own `--timeout` for the whole release. And `release.devnet.all.daily.yml:75-89` gates the Slack failure alert on `needs: [release, ...]`, so a hung warm-up suppresses the daily-failure notification for those six hours. `.github/instructions/github-actions.instructions.md:37-38` asks for this bound.

2. **Wider failure surface than the release needs.** `go mod download` with no arguments fetches every module explicitly required by the main module — 187 requires in `go.mod`, including test-only ones no released binary links (`testcontainers-go` and its clickhouse/redpanda modules, `docker/docker`, both influx clients, `clickhouse-go`). A durable proxy failure on any of those now fails all 10 daily component releases, where before GoReleaser fetched only what each binary imports and the release succeeded.

Both are covered by two lines on this step: `timeout-minutes: 10` and `continue-on-error: true`. GoReleaser needs a strict subset of what this downloads and will fetch anything still missing, so a failed warm-up should degrade to the pre-PR behavior rather than block the release. The `exit 1` after three attempts stays useful as a visible warning and diagnostic.

ben-dz (commented):
One medium finding: the new `Download Go modules` step is a best-effort cache warm-up but is wired as a hard, unbounded gate, so it can fail the daily release in two ways it previously could not — an unbounded stall (no `timeout-minutes` on the step or job, where the fetch was previously bounded by GoReleaser's own `--timeout`, and the Slack failure alert is `needs: release` so it stays silent meanwhile), and a hard failure on the full 187-module explicit require set including test-only modules no released binary links. Adding `timeout-minutes: 10` and `continue-on-error: true` to the step fixes both and degrades a failed warm-up to the pre-PR behavior.

Separately, the earlier automated comment about `controlplane/s3-uploader` needing its own retry does not apply: s3-uploader is not in the daily matrix (`release.devnet.all.daily.yml:24-44`) and does not call `release.daily.yml` — it releases from `release.s3-uploader.yml`. All 10 daily components build in the root module, so the root `go mod download` covers them.

.github/workflows/release.daily.yml
nikw9944: Both taken, as `continue-on-error: true` + `timeout-minutes: 10` on the step. Pushed in 4bf02ec6 (branch also rebased onto main).

I verified the reasoning rather than just the suggestion: `go mod download` with no arguments covers the modules explicitly required by the main module, which is 187 requires here including the test-only ones you named (`testcontainers-go` and its clickhouse/redpanda modules, `docker/docker`, both influx clients, `clickhouse-go`) — none of which any released binary links. And `notify` is `needs: [release, ...]` with `if: failure()`, so a hung warm-up does suppress the daily-failure alert for as long as it hangs, since the `release` job carries no `timeout-minutes`.

The step comment now records why the two lines are there, so the next person does not "tidy" a bounded best-effort step into a gate. The `exit 1` after three attempts stays, so an exhausted retry is still a visible red step and a diagnostic without failing the release. Changelog entry updated to state the best-effort/bounded semantics rather than implying a hard gate.

ben-dz (approved):
lgtm — the medium finding from the previous round is fixed: the `Download Go modules` step now carries `continue-on-error: true` and `timeout-minutes: 10`, so a stalled or exhausted warm-up degrades to the pre-PR behavior instead of hanging to the six-hour job default or blocking the release on a test-only module.

Issue #4110, opened 2026-07-29
serviceability: make CreateUser naked and idempotent


martinsander00:
Part 2 of #4108: after #4113 (`SubscribeFeed`), before #4111 (`connect --subscribe-feed`).

The #4111 connect flow is: create a user, poll to `Activated`, then one `SubscribeFeed` covering every feed. `CreateUser` blocks both halves of that. Under an EdgeSeat pass a multicast create runs the feed metro gate and fails with `FeedAccountRequired` unless it names a feed and group, so a user that holds nothing cannot exist. And a re-run fails with `AccountAlreadyInitialized`, so `connect` cannot call create unconditionally and retry; it has to read the chain first to decide, and a crash between steps leaves it guessing.

Fix:

- Naked: move `enforce_feed_metro_gate` out of `create_user_core` into `CreateSubscribeUser`, its only real caller. `CreateUser` never feed-gates; a new user starts with no groups, no feeds, and no seat charged.
- Idempotent: an existing user matching the requested owner, device, and user type makes `CreateUser` a no-op that succeeds. A mismatch still fails with `AccountAlreadyInitialized`; a banned user fails with `InvalidStatus`.

`CreateSubscribeUser` keeps its semantics: a duplicate is an error and the feed gate applies as before.

nikw9944:
<details>
<summary>System Prompt</summary>

# Code Review

You are a Claude Code instance managed by Roadbike, performing a thorough code review.
You must NOT make any code changes — review only. Do NOT post the review to GitHub —
the operator will read your findings in the dashboard and decide what to do.

## Assignment

- **PR**: [malbeclabs/doublezero#4116](https://github.com/malbeclabs/doublezero/pull/4116)
- **Title**: serviceability: make CreateUser naked and idempotent
- **Repository**: malbeclabs/doublezero
- **Branch**: `ms/4110` → `main`
- **Linked Issue**: #4110 — serviceability: make CreateUser naked and idempotent

## PR Description

Resolves #4110. Part 2 of #4108.

## Summary

- `CreateUser` no longer runs the EdgeSeat feed metro gate, so it can create a bare multicast user holding no group and no feed. The gate moved to `CreateSubscribeUser`, its only real caller; feed seats are otherwise charged by `SubscribeFeed`.
[...]

PR #4113, opened 2026-07-29
serviceability: add SubscribeFeed and UnsubscribeFeed for EdgeSeat feeds

### LINKED ISSUE #4109: serviceability: add UpdateFeedSubscription instruction for EdgeSeat feed joins
Part 1 of #4108.

Add a dedicated instruction for EdgeSeat feed subscriptions, so a user joins every group across the feeds they paid for in a single atomic transaction, with one seat charged per feed.

## Instruction

`UpdateFeedSubscription`, variant **117** (116 `SetAccessPassFlags` is the last used discriminant).

Args: `{ client_ip, subscriber: bool, feed_count: u8 }`. There is no publisher field: a feed is a receive-only SKU, so publishing stays group-level and allowlist-gated.

Accounts:

```
[accesspass(w), user(w), globalstate, multicast_publisher_block(w), device]
[feed_0 .. feed_{F-1}]      readonly     F = feed_count
[group_0 .. group_{G-1}]    writable
[payer(s,w), system, permission?]        existing trailer
```

`split_trailing_permission` already returns the variable `leading` slice; split it at `feed_count` to separate feeds from groups.

## Validation

- Authorization matches `UpdateMulticastGroupRoles`: the pass's `user_payer`, a foundation member, or `ACCESS_PASS_ADMIN`
- Access pass must be `EdgeSeat`; user must be `UserType::Multicast` and `Activated` when granting
- The device account must equal `user.device_pk`
- Per feed: provisioned on the pass, and `feed.exchange == device.exchange_pk` (reuse `check_feed_metro_coverage`)
- Per group: `Activated`, and present in at least one passed feed's group set, otherwise `GroupNotInFeed`
- Seats: tick each passed feed not already in `user.feed_pks` **and** contributing at least one of the passed groups, so a stray feed cannot consume a seat for nothing. Cap at `MAX_ACCESS_PASS_FEEDS`
- `subscriber: false` drops the passed groups and releases the seat for any feed in which the user no longer holds a group

Atomicity comes for free: any failure reverts the whole transaction.

Reuses `check_feed_metro_coverage`, `update_user_multicastgroup_roles`, `try_add_feed_user`, `remove_feed_user`, and `split_trailing_permission`.

## Also in this change

Delete the `!is_edge_seat &&` guards in `update_user_multicastgroup_roles`:

```rust
if publisher && !is_edge_seat && !accesspass.mgroup_pub_allowlist.contains(...) { NotAllowed }
if subscriber && !is_edge_seat && !accesspass.mgroup_sub_allowlist.contains(...) { NotAllowed }
```

With feed joins handled by the new instruction, `UpdateMulticastGroupRoles` no longer needs to know EdgeSeat exists and the `is_edge_seat` local disappears. Allowlist enforcement becomes uniform across pass types, which closes the existing publish gap: an EdgeSeat holder can still publish, but only to a group a foundation member explicitly granted via `AddMulticastGroupPublisher`.

## Known edge

A group that is in both a feed and the allowlist, joined through the new instruction and then removed through `UpdateMulticastGroupRoles`, leaves its seat held. `UpdateMulticastGroupRoles` has no feed account to check against. The seat is released when the new instruction runs with `subscriber: false`, or on disconnect, since `DeleteUser` releases everything in `user.feed_pks`. Document rather than pull feed logic back into the older instruction.

## Testing

- A group drawn from a second feed on the pass takes that feed's seat
- A second group inside an already-held feed does not tick again (use a `max_users: 1` seat, so a double tick surfaces as `FeedSeatFull`)
- Group outside every passed feed, feed not on the pass, feed serving another metro, and a device that is not the user's each rejected on their exact error
- Removal releases the seat only when the last group in that feed is dropped, and keeps it while another remains
- Exceeding the account ceiling fails with a clear error rather than truncating
- Allowlist subscribe and publish still work on an EdgeSeat pass after the guards are deleted

martinsander00:
Resolves #4109

## Summary

- New `SubscribeFeed` (117) and `UnsubscribeFeed` (118) instructions join or leave whole feeds on an EdgeSeat access pass, in one atomic transaction. Seats are charged per feed, not per group.
- A feed is all-or-nothing: the caller names feeds, the processor derives which groups change, and a group list that does not match is rejected so a stale client cannot half-apply. Because a seat follows the feed the caller named rather than being inferred from group membership, two feeds carrying the same group stay unambiguous — leaving one keeps the shared group and releases only that feed's seat.
- An unsubscribe must pass every feed the user holds. Without the retained feeds' group sets the processor cannot tell that a departing feed is dropping a group another held feed still covers, so it rejects rather than stranding that feed's seat.
- `UpdateMulticastGroupRoles` now enforces the multicast-group allowlists for every access-pass type, EdgeSeat included, which previously bypassed both. Purchased groups go through the feed instructions and charge a seat; individually comped groups go through the allowlist and charge nothing.
- `CreateSubscribeUser` skips the allowlist only for the case `create_user_core`'s feed gate actually covers, and always checks the publisher allowlist. Two paths could previously join a group with no check at all: an EdgeSeat pass creating a non-multicast user, and an EdgeSeat multicast user taking a publisher role on a feed group it only bought receive rights to.

Not breaking: no EdgeSeat pass exists on mainnet-beta, and the EdgeSeat path through `UpdateMulticastGroupRoles` was already unreachable — no builder emitted the device account its coverage check required, so every such call failed with `MetroMismatch`.

Must land before #4110 (naked `CreateUser`), or naked EdgeSeat users could reach any group through an ungated `UpdateMulticastGroupRoles`.

## Testing Verification

- Seat accounting: a feed's whole group set joins for one seat (cap set to 1, so a per-group tick would surface as `FeedSeatFull`); a second feed takes its own seat; two feeds join in one transaction; leaving releases the seat.
- `test_leaving_a_feed_keeps_a_group_a_retained_feed_covers`: two feeds both carry g0, the user leaves one, keeps g0 via the retained feed, and only the departing feed's seat is released.
- `test_leave_omitting_a_held_feed_rejected`: a leave that omits a held feed is refused instead of stranding its seat.
- `test_seat_cap_rejects_a_second_machine` fills a one-user feed with a real second user on the same dynamic pass, rather than a synthetic full seat.
- Regression suites green for the paths whose authorization moved: `create_subscribe_user_test`, `multicastgroup_subscribe_test`, `feed_metro_gate_test`.


smartcontract/programs/doublezero-serviceability/src/processors/user/create_subscribe.rs
```diff
@@ -135,10 +140,25 @@ pub fn process_create_subscribe_user(
         feed_account,
     )?;
 
+    // Authorize the group before joining it. An EdgeSeat pass derives joinable groups from the feeds
+    // provisioned on it, and `create_user_core` above already enforced that gate (and charged the
+    // feed's seat), so the allowlist does not apply. Every other pass type grants groups
+    // individually and is checked against the allowlist here.
+    if !matches!(
```
nikw9944: This skip keys only on **pass type**, but the gate it defers to (`enforce_feed_metro_gate` via `create_user_core`) runs only for `AccessPassType::EdgeSeat(_) && user_type == UserType::Multicast` (`create_core.rs:325`) and never distinguishes the publisher role from the subscriber role. Two paths fall through:

**(a) EdgeSeat pass + non-Multicast `user_type`.** `CreateSubscribeUser` accepts any `UserType` and never restricts it. With `user_type: IBRL`, `subscriber: true`, and any Activated `MulticastGroup`: `create_core.rs:325` skips the feed gate (type is not `Multicast`) and this line skips the allowlist (pass is EdgeSeat), so `update_user_multicastgroup_roles` joins the group with no authorization at all. The comment above ("`create_user_core` above already enforced that gate") and the doc on `check_mgroup_allowlists` ("a caller runs one or the other, never neither") are both false here. The controller gates multicast rendering on `UserType == UserTypeMulticast`, so this stops short of the data plane — but it inflates `mgroup.subscriber_count` and defeats the allowlist that `create_core.rs:235` names as the governing check.

**(b) EdgeSeat + Multicast + `publisher: true`.** The feed gate passes (group is in the feed) and the pub allowlist is skipped, so the caller becomes a publisher on a feed group and gets a `dz_ip` from `MulticastPublisherBlock`. The controller does program `MulticastPublishers` for a Multicast user, so this is live publish rights on content a feed sells receive-only — the same hole the PR description reports as closed, closed in `UpdateMulticastGroupRoles` only. `subscribe_feed.rs` is explicit that a feed grants no publisher role, and no test in `feed_metro_gate_test.rs` exercises an EdgeSeat publisher.

Suggested fix — skip only the case the feed gate actually covers, and always check the pub allowlist:

```rust
let feed_gated = matches!(result.accesspass.accesspass_type, AccessPassType::EdgeSeat(_))
    && value.user_type == UserType::Multicast;
// A feed sells receive only, so publishing still needs the pub allowlist.
let subscriber_gated = value.subscriber && !feed_gated;
check_mgroup_allowlists(&result.accesspass, mgroup_account.key, value.publisher, subscriber_gated)?;
```

Worth resolving here rather than deferring: #4110 is sequenced behind this PR precisely so no ungated path into a group survives, and (a) is that same shape.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe_feed.rs
```diff
@@ -0,0 +1,261 @@
...
+            carry_publisher,
+            value.subscriber,
+        )?;
+        try_acc_write(&result.mgroup, group_account, payer_account, accounts)?;
+    }
+
+    // Reconcile seats against the user's *final* membership rather than ticking as we go. A feed is
+    // held exactly while the user is in at least one of its groups, so a second group inside a held
+    // feed is free, dropping one of several groups keeps the seat, and dropping the last releases it.
+    let held_groups = user.get_multicast_groups();
```
nikw9944: `get_multicast_groups()` is publishers ∪ subscribers, so a publisher-only membership pins the feed seat permanently. An operator comps `mgroup_pub_allowlist` for a group a feed also carries (a combination this PR explicitly supports), the user publishes via `UpdateMulticastGroupRoles`, then leaves every feed group as a subscriber — `holds_group` stays true, the seat stays charged on the pass, and only user delete releases it. Reconcile against `user.subscribers` instead, matching this instruction's receive-only semantics.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe_feed.rs
```diff
@@ -0,0 +1,261 @@
...
+    for (feed_key, feed) in &feeds {
```
nikw9944: This loop iterates only the *passed* feeds, so a feed the user holds whose membership changed as a side effect of this call is never reconciled. With two feeds on the pass carrying an overlapping group, `subscriber: false` over feed B's groups can drop the user's last group in feed A; A's seat and its `user.feed_pks` entry stay behind. Recoverable by calling again with A, but silent. Reject the call when a passed group belongs to a feed in `user.feed_pks` that is not in `feeds`.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe_feed.rs
```diff
@@ -0,0 +1,261 @@
...
+        let holds_group = feed.groups.iter().any(|group| held_groups.contains(group));
+        let seat_recorded = user.feed_pks.contains(feed_key);
+
+        if holds_group && !seat_recorded {
+            if user.feed_pks.len() >= MAX_ACCESS_PASS_FEEDS {
+                return Err(DoubleZeroError::UserFeedLimitExceeded.into());
```
nikw9944: This reverts with no `msg!`, unlike every other error in the function — an operator who hits it gets `Custom(103)` and nothing else. Log the current count and the cap.

nikw9944 (changes requested):
One authorization gap to close before merge, plus two seat-accounting cases and a test-coverage gap.

The EdgeSeat allowlist skip in `create_subscribe.rs` keys only on pass type, while the feed gate it defers to keys on pass type **and** `user_type == Multicast` and never separates the publisher role from the subscriber role. That leaves two ungated paths: an EdgeSeat pass creating a non-Multicast user can join any Activated multicast group with no check at all, and an EdgeSeat Multicast user can take a **publisher** role on a feed group without being on `mgroup_pub_allowlist` — live publish rights on content a feed sells receive-only. The second is the same hole the PR description says it closed; it was closed in `UpdateMulticastGroupRoles` only, not at connect. Since #4110 is sequenced behind this PR precisely so no ungated path into a group survives, this is worth fixing here.

Separately, the seat reconcile in the new processor holds a feed seat on publisher-only membership and only reconciles the feeds passed in this call, so both a comped-publisher user and overlapping feed group sets can leave a seat silently charged on the pass until user delete. Three of the four new rejection paths have no test.

- `tests/feed_subscription_test.rs`: `EdgeSeatRequired` (101) and `EdgeSeatIsMulticastOnly` (104) are introduced with no test, as are the two `InvalidArgument` arg-shape rejections (`feed_count == 0` / no group accounts, `subscribe_feed.rs:108`; duplicate feed, `subscribe_feed.rs:201`). 101 is the primary pass-type gate on the new instruction. CLAUDE.md Testing §4 asks for all error paths; the rest of this file already does that well.

smartcontract/programs/doublezero-serviceability/src/processors/user/create_subscribe.rs
martinsander00: Fixed in 52c47dffeea939b4292d9da07c0824b7b6d49ba7.

Both paths are closed by keying the skip on the same condition `create_user_core` uses for the gate, and by never skipping the publisher allowlist:

```rust
let feed_gated = matches!(result.accesspass.accesspass_type, AccessPassType::EdgeSeat(_))
    && value.user_type == UserType::Multicast;
check_mgroup_allowlists(
    &result.accesspass,
    mgroup_account.key,
    value.publisher,
    value.subscriber && !feed_gated,
)?;
```

So (a) an EdgeSeat pass creating a non-Multicast user now falls to the allowlist check, and (b) a publisher role always needs `mgroup_pub_allowlist` regardless of pass type. Thanks for catching that the publish hole was only half closed.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe_feed.rs
martinsander00: Fixed in 52c47dffeea939b4292d9da07c0824b7b6d49ba7, and the underlying inference is gone rather than corrected.

Seats no longer derive from group membership at all. The caller names feeds and the seat follows the named feed, so a comped publisher role on a group a feed happens to carry cannot pin that feed's seat. `get_multicast_groups()` is no longer consulted for seat state.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe_feed.rs
martinsander00: Fixed in 52c47dffeea939b4292d9da07c0824b7b6d49ba7, though not by rejecting.

Leaving is now `UnsubscribeFeed`, which takes the departing feeds plus every feed the user retains. A group a retained feed still carries is kept rather than dropped, so the case that stranded feed A's seat no longer arises — "leave feed B" means stop paying for B, not lose a group A entitles you to. A leave that omits a held feed is rejected with `FeedAccountRequired` rather than silently reconciling a partial view, covered by `test_leave_omitting_a_held_feed_rejected`; the kept-group case is `test_leaving_a_feed_keeps_a_group_a_retained_feed_covers`.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe_feed.rs
martinsander00: Fixed in 52c47dffeea939b4292d9da07c0824b7b6d49ba7 — logs the current count and the cap before reverting.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe_feed.rs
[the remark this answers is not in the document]
martinsander00: Fixed in 52c47dffeea939b4292d9da07c0824b7b6d49ba7. Seat state no longer derives from group membership at all: the caller names feeds and the seat follows the named feed, so a publisher-only membership cannot hold a feed seat.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs
```diff
@@ -52,20 +50,45 @@ impl fmt::Debug for UpdateMulticastGroupRolesArgs {
...
+        return Err(DoubleZeroError::NotAllowed.into());
+    }
+    Ok(())
+}
+
 /// Toggle a user's multicast group roles.
 ///
+/// Mechanics only: this does NOT authorize the change. Callers must first run either
+/// [`check_mgroup_allowlists`] or the EdgeSeat feed metro gate, so that the authorization a
+/// processor performed is visible in that processor rather than claimed through an argument here.
```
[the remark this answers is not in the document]
martinsander00: Fixed in 86f05d2b375cac9f942b14b49c363344f8b56fc0. Reworded to say `publisher` and `subscriber` are desired states rather than toggle signals, and dropped "toggle" from the description of post-activation changes.

smartcontract/programs/doublezero-serviceability/tests/feed_subscription_test.rs
```diff
@@ -0,0 +1,708 @@
...
+}
+
+/// Assert the transaction failed on a specific `ProgramError::Custom` code, so a test cannot pass
+/// because the instruction failed for an unrelated reason.
+fn assert_custom_error(err: &BanksClientError, code: u32) {
+    assert!(
+        format!("{err:?}").contains(&format!("Custom({code})")),
+        "expected Custom({code}), got: {err:?}"
+    );
+}
```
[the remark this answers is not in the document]
martinsander00: Fixed in 86f05d2b375cac9f942b14b49c363344f8b56fc0. Now matches `TransactionError::InstructionError(0, InstructionError::Custom(code))` structurally, following `accesspass_dzf_locked.rs`.

One note on the framing: both styles exist in this suite — 14 test files match on the debug text, including `feed_metro_gate_test.rs`, which is where I copied the pattern from. The structural form is still the better one, so the change stands.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/unsubscribe_feed.rs
```diff
@@ -0,0 +1,129 @@
...
+    let (target_accounts, rest) = ctx.variable.split_at(feed_count);
+    let (retained_accounts, group_accounts) = rest.split_at(retained_count);
+
+    let targets = load_feeds(
+        program_id,
+        &ctx.accesspass,
+        &ctx.device.exchange_pk,
+        target_accounts,
+    )?;
+    let retained = load_feeds(
```
nikw9944, line 82: `targets` and `retained` come from two independent `load_feeds` calls. `load_feeds` dedupes only within a single call, and nothing checks that the two lists are disjoint or that a retained feed is one the user actually holds. With `targets = [T]`, `retained = [T]`, `groups = []`:

- the held-feed coverage loop (85-92) passes, since `T` is in `targets`;
- `expected` (94-102) ends **empty**, because `still_covered` consults `retained`, which *is* `T`, so every group of `T` looks covered;
- `check_group_accounts(&[], &[])` passes and `apply_groups` is a no-op — no group is unsubscribed;
- the seat loop (114-119) still fires: `T` is in `feed_pks`, so `remove_feed_user(T)` runs and `T` is dropped.

Net: `user.subscribers` keeps every group the feed carries, `user.feed_pks` is empty, and the pass's seat is released. The controller renders `MulticastSubscribers` straight from `user.Subscribers` for a Multicast user (`controlplane/controller/internal/controller/server.go:662`) and never consults `feed_pks`, so this is live multicast receive with the seat metering fully bypassed. Reachable by the pass's own `user_payer` (`authorize_payer` returns `Ok` immediately for them) and expressible with the existing helper as `leave(&mut f, &[feed], &[feed], &[])`.

Duplicate account metas do survive into the program: this repo's `UpdateDevice` fixture passes `location_pubkey` twice (`feed_subscription_test.rs:129-130`) and `process_update_device` consumes both slots (`device/update.rs:186-187`).

A second variant needs no duplicate at all: any feed provisioned on the pass that serves the user's metro and overlaps the target's groups works as a bogus `retained`, even if the user never held it.

Fix — both checks are needed (disjointness alone leaves the not-held variant; the held check alone leaves target-as-retained):

```rust
for (key, _) in &retained {
    if targets.iter().any(|(t, _)| t == key) {
        msg!("feed {} passed as both a target and retained", key);
    }
    if !ctx.user.feed_pks.contains(key) {
        msg!("retained feed {} is not held by this user", key);
    }
}
```

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/unsubscribe_feed.rs
```diff
@@ -0,0 +1,129 @@
...
+        program_id,
+        &ctx.accesspass,
+        &ctx.device.exchange_pk,
+        retained_accounts,
+    )?;
+
+    for held in &ctx.user.feed_pks {
```
nikw9944, line 117: This "every held feed must be passed" requirement combines with `load_feeds`'s provisioning check to make leaving permanently impossible once any held feed is dropped from the pass. `SetAccessPassFeeds` replaces the whole seat list and never refuses to drop a feed with `current_users > 0` (`set_feeds.rs:136-141` only carries counts forward for feeds still listed). Afterwards the held-but-unprovisioned feed cannot be passed at all — `load_feeds` returns `FeedNotOnAccessPass` (93) — so the user can never release the seats on their *other* feeds either, and only `DeleteUser` clears them.

Skip the coverage requirement for a held feed absent from `accesspass.feed_seats()` (there is no seat left to strand) and prune it from `feed_pks`.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe_feed.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    let mut expected: Vec<Pubkey> = Vec::new();
+    for (_, feed) in &feeds {
+        for group in &feed.groups {
+            if !ctx.user.subscribers.contains(group) && !expected.contains(group) {
+                expected.push(*group);
+            }
+        }
+    }
+    check_group_accounts(group_accounts, &expected)?;
```
nikw9944, line 83: `SubscribeFeed` is exactly all-or-nothing — `check_group_accounts` rejects any list that is not the full un-held set — and with `UpdateMulticastGroupRoles` now allowlist-only there is no incremental path. A feed created at `MAX_FEED_GROUPS = 64` (`feed/create.rs:26`) can therefore never be joined: 4 fixed accounts + 1 feed + 64 groups + payer + system + permission + program id exceeds Solana's 64-account transaction lock limit, and a legacy non-ALT transaction caps near 26 groups on the 1232-byte size limit.

Either bound `MAX_FEED_GROUPS` to what one transaction can carry, or accept a partial group list. The largest fixture feed carries 3 groups, so nothing tests near the boundary.

nikw9944 (changes requested):
All five findings from the previous round are verified fixed. The redesign that fixed them — splitting into `SubscribeFeed` (117) and `UnsubscribeFeed` (118), with seats following the named feed rather than inferred group membership — introduces a seat-metering bypass that must be closed before merge.

**`UnsubscribeFeed` lets a pass holder keep every group in a feed while releasing its seat.** `targets` and `retained` are loaded by two independent `load_feeds` calls with no disjointness check, and a retained feed is never required to be one the user holds. Passing the same feed as both target and retained makes every one of its groups "still covered", so `expected` is empty and no group is unsubscribed — but the seat loop still releases the seat and clears `feed_pks`. The controller renders multicast subscribers straight from `user.Subscribers` and never consults `feed_pks`, so the result is live multicast receive with the paid-capacity metering fully bypassed, reachable by the pass's own `user_payer`.

Two further limits in the new design: the mandatory retained-feed coverage makes leaving permanently impossible once an admin drops a held feed from the pass, and `SubscribeFeed`'s exactly-all-or-nothing group list means a feed created at `MAX_FEED_GROUPS` can never be joined at all.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/unsubscribe_feed.rs
martinsander00: https://github.com/malbeclabs/doublezero/commit/7f763fc452ee4dfa6f7f5612642520c3e6574daf

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe_feed.rs
martinsander00: https://github.com/malbeclabs/doublezero/commit/7f763fc452ee4dfa6f7f5612642520c3e6574daf

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/unsubscribe_feed.rs
martinsander00: Tracking separately: cascading the cleanup needs every affected User and MulticastGroup account, which does not fit one instruction, so the fix is orchestration in the access-pass CLI (unsubscribe the holders, then set the feeds). Filing an issue.

martinsander00: Filed as #4114.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/unsubscribe_feed.rs
```diff
@@ -0,0 +1,141 @@
...
+            msg!("feed {} passed as both a target and retained", key);
+            return Err(DoubleZeroError::InvalidArgument.into());
+        }
+        if !ctx.user.feed_pks.contains(key) {
+            msg!("retained feed {} is not held by this user", key);
+            return Err(DoubleZeroError::InvalidArgument.into());
+        }
+    }
+
+    for held in &ctx.user.feed_pks {
```
nikw9944, line 117: Unchanged, and deferred to #4114 — but the deferral answers a different fix than the one needed.

`SetAccessPassFeeds` replaces the whole seat list and never refuses to drop a feed with `current_users > 0` (`set_feeds.rs:136-141` only carries counts forward for feeds still listed). Afterwards this loop requires every entry in `user.feed_pks` to be passed as a target or retained, but `load_feeds` rejects the now-unprovisioned feed with `FeedNotOnAccessPass` (93) at lines 72-83 — before the new retained-validation block. The user can then never release the seats on their *other* feeds either; only `DeleteUser` clears them via `release_feed_seats`.

The reply describes cascading cleanup across every affected User and MulticastGroup account. The fix asked for here is local to this processor and touches no other account: a held feed absent from `accesspass.feed_seats()` has no seat left to strand, so exempt it from the coverage requirement and prune it from `feed_pks` in the same pass. If the admin-side orchestration in #4114 is the preferred long-term shape, this processor still needs the exemption so an operator mistake is not a permanent lockout.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe_feed.rs
```diff
@@ -0,0 +1,116 @@
...
+        &mut ctx.user,
+        ctx.payer_account,
+        true,
+    )?;
+
+    for (feed_key, _) in &feeds {
+        if ctx.user.feed_pks.contains(feed_key) {
+            continue;
+        }
+        if ctx.user.feed_pks.len() >= MAX_ACCESS_PASS_FEEDS {
```
nikw9944: Lowering `MAX_FEED_GROUPS` to 20 does fix the join path — a 1-feed/20-group `SubscribeFeed` is 29 unique accounts, ≈1063 bytes, inside both the 64-account lock limit and the 1232-byte legacy limit. Two gaps remain:

**The sibling cap is unbounded.** `UnsubscribeFeed` must name *every* held feed, so its account count is `8 + feeds_held + groups_departing`. `MAX_ACCESS_PASS_FEEDS = 64` — borrowed here as the `feed_pks` cap — permits 64 held feeds. Past roughly 6 held feeds at max group size a legacy transaction no longer fits (`feeds_held + groups ≤ 26`), and past ~36 no transaction fits at all (`≤ 56` under the lock limit). That is the same un-leavable end state as the dropped-feed case, reached through the cap this commit did not revisit. Bound this cap by what one leave transaction can name rather than reusing `MAX_ACCESS_PASS_FEEDS`.

**No boundary test.** The new `MAX_FEED_GROUPS = 20` is justified by transaction capacity but nothing exercises it; the largest fixture feed carries 3 groups. A test that joins a 20-group feed in one transaction pins the constant to its stated reason and stops it drifting back up.

nikw9944 (approved):
The seat-metering bypass from the last round is fixed in code — `unsubscribe_feed.rs:85-94` now rejects a feed in both lists and any retained feed not in `user.feed_pks`, with tests covering both variants and asserting the seat and subscriptions survive the rejected call.

Two Mediums remain. The dropped-feed lockout is unchanged and deferred to #4114, but the deferral answers a different fix: exempting a held feed that is absent from `accesspass.feed_seats()` is local to this processor and needs no cascade, and without it a routine `SetAccessPassFeeds` leaves a user's other seats permanently unreleasable. And the transaction-capacity bound added to `MAX_FEED_GROUPS` was not applied to the sibling cap — `UnsubscribeFeed` must name every held feed, so the 64-feed `feed_pks` cap permits a held-feed count no leave transaction can carry, reaching the same un-leavable end state. The new constant also has no boundary test.

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/unsubscribe_feed.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4113/changes/e7a7a060b8f4d7d4f58e62c7f68b46df566065d6

smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe_feed.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4113/changes/e7a7a060b8f4d7d4f58e62c7f68b46df566065d6

Issue #4111, opened 2026-07-29
sdk/client: add connect multicast --subscribe-feed


martinsander00:
Part 3 of #4108. Depends on parts 1 and 2.

Expose the new instruction through the Rust SDK and give `doublezero connect` a way to subscribe by feed, so a buyer runs one command per machine and never has to look up which groups a feed carries.

## CLI

```console
$ doublezero connect multicast --subscribe-feed shreds-nyc shreds-ams
```

`--subscribe-feed` is variadic and takes feed pubkeys or codes, matching how `--publish` and `--subscribe` already work. It is mutually exclusive with both of them.

The name deliberately carries the role. Under `connect multicast` the flags are what select publisher versus subscriber, so a bare `--feed` would be the only role-neutral flag in a role-bearing set. `--subscribe-feed` also makes publish-by-feed unrepresentable rather than something a conflict rule has to reject, and leaves room for a `--publish-feed` if a future pass type ever sells it.

Flow: `CreateUser` (naked, idempotent) then poll to `Activated` then one `UpdateFeedSubscription` covering every feed and group.

## SDK

- `UpdateFeedSubscriptionCommand`: takes feed pubkeys, reads each `Feed`, filters to the ones serving the user's device metro, unions and dedupes their groups, and builds the account list. The caller names feeds explicitly, so there is no resolution guesswork.
- `LedgerClient` gains `get_feed` and `list_feed`. Neither exists today, which is why the client cannot resolve feed codes at all right now.

## Testing

- A feed expands to every group it carries in the user's metro
- Two feeds sharing a group join it once
- A feed serving another metro is rejected before any transaction
- `--subscribe-feed` with `--subscribe` or `--publish` is rejected at parse time and on a programmatically built command
- Connect against an existing activated user skips creation and goes straight to the subscription

Issue #4108, opened 2026-07-29
serviceability: feed-scoped multicast subscription for EdgeSeat access passes


martinsander00:
## Goal

An EdgeSeat access-pass holder connects to everything they paid for with one command per machine, with seats billed per feed:

```console
$ doublezero connect multicast --subscribe-feed shreds-nyc shreds-ams
```

## Problem

The EdgeSeat feed gate only runs at connect time in `create_user_core`, and only for a single group. `UpdateMulticastGroupRoles` runs no feed gate and skips both multicast-group allowlists for EdgeSeat passes, so a second feed's groups are joinable for free and publishing is ungated. The client `connect` path fails with `FeedAccountRequired`, and `CreateUser` cannot create a bare multicast user under an EdgeSeat pass at all.

## Design

Purchased and comped subscriptions become two separate paths:

- Purchased feeds: new `SubscribeFeed` / `UnsubscribeFeed` instructions. Gate: feed on the pass, serves the device's metro. One seat per feed.
- Comped groups: `UpdateMulticastGroupRoles`, allowlist-gated uniformly across pass types (the EdgeSeat bypass is deleted, which closes the publish gap by removal).

`CreateUser` becomes naked (no group, no feed) and idempotent, so `connect` can call it unconditionally: create, poll to `Activated`, then one atomic `SubscribeFeed` covering every feed and group.

## Breakdown

1 must land before 2, or naked EdgeSeat users can join any group through an ungated `UpdateMulticastGroupRoles`.

1. onchain: `SubscribeFeed` / `UnsubscribeFeed`, delete the `is_edge_seat` guards (#4113)
2. onchain: naked and idempotent `CreateUser` (#4110)
3. sdk/client: `connect multicast --subscribe-feed`

Out of scope: `CreateSubscribeUser` and its `--feed` flag (#4087) stay as they are.

Supersedes #4097. PR #4099 (optional accounts retrofitted onto `UpdateMulticastGroupRoles`) was closed unmerged in favour of this design.

PR #4116, opened 2026-07-30
serviceability: make CreateUser naked and idempotent

### LINKED ISSUE #4110: serviceability: make CreateUser naked and idempotent

martinsander00:
Resolves #4110. Part 2 of #4108.

## Summary

The #4111 connect flow needs `CreateUser` to mean "make sure my user exists": create it bare, poll to `Activated`, then join feeds with one `SubscribeFeed`. Today a multicast create under an EdgeSeat pass fails without a feed and group, and a re-run fails with `AccountAlreadyInitialized`, so `connect` can neither create a user that holds nothing nor retry blindly.

- The feed gate moved out of `create_user_core` into `CreateSubscribeUser`, its only real caller. A new user starts bare: no groups, no feeds, no seat charged; `SubscribeFeed` charges seats.
- Re-running `CreateUser` for an existing user matching the requested owner, device, user type, and tenant is a no-op success. A mismatch still fails with `AccountAlreadyInitialized`; a banned user fails with `InvalidStatus`. Owner matters because the user PDA carries no owner in its seeds.
- The exists check sits after the pass identity checks and before every counter bump and gate, so a retry survives an epoch rollover or a device that filled up.
- The pass-level `max_multicast_users` cap is enforced again; it had been vestigial under the feed-seat model, and bare users would otherwise be unbounded. The oracle already provisions it as the total purchased seats.
- Controller: a multicast user without a publisher role now gets the subscriber ingress ACL and a deny-all announce prefix-list; a role-less user previously fell through to the publisher versions of both.
- `CreateSubscribeUser` is unchanged: a duplicate still errors and the gate still applies.

## Testing Verification

- Re-run asserts full-struct equality on User, Device, and AccessPass; a re-run against a device shrunk to `max_users = 1` still succeeds.
- Rejections: different device, tenant, or owner, and a banned user all fail and leave the user untouched; a duplicate `CreateSubscribeUser` errors without ticking a seat; a second bare user past `max_multicast_users` is rejected (90).
- `test_naked_create_then_subscribe_feed` runs the #4111 flow: bare create takes no seat, one `SubscribeFeed` then charges it.
- Controller render and server tests gain a zero-role multicast tunnel asserting the subscriber ACL and deny-all prefix-list; `feed_metro_gate_test` and `create_subscribe_user_test` cover the moved gate.


smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs
```diff
@@ -61,11 +58,14 @@ pub struct CreateUserCoreResult {
 /// PDA derivation, device validation, max users checks, epoch check) and sets up the
 /// initial User struct with Pending status.
```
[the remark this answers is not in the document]
martinsander00: https://github.com/malbeclabs/doublezero/commit/1724eb510021f0561cd34b7f41468ab782c29274

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
...
+  - `CreateUser` can now create a bare multicast user (no group, no feed) under an EdgeSeat pass: the feed gate moved to `CreateSubscribeUser`, its only caller, and feed seats are otherwise charged by `SubscribeFeed`. `CreateUser` is also idempotent: re-running it for an existing user matching the requested owner, device, and user type succeeds without changing state, so a client can retry safely; a mismatch still fails with `AccountAlreadyInitialized`. (#4110)
```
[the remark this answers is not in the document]
martinsander00: https://github.com/malbeclabs/doublezero/commit/1724eb510021f0561cd34b7f41468ab782c29274

smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs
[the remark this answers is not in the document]
martinsander00: https://github.com/malbeclabs/doublezero/commit/a56ce0dc40d039865ec50a01e3a0e5e5f201a456

smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs
```diff
@@ -204,6 +198,39 @@ pub fn create_user_core(
         return Err(DoubleZeroError::Unauthorized.into());
     }
 
+    // Idempotent create: an existing user matching the request is a no-op, so a caller can retry
+    // safely. Checked after the pass validation so a no-op still needs the caller's own pass, and
+    // before the tenant, epoch, device-status, and capacity gates: those apply to adding a user,
+    // not to retrying one.
```
nikw9944: The exists-match never checks `status`, so a **banned** user's re-run returns success.

`RequestBanUser` leaves the account onchain: `requestban.rs:124-142` deallocates the user's resources, zeroes `tunnel_net`/`tunnel_id`/`dz_ip`, sets `status = UserStatus::Banned`, and writes with `try_acc_write` (not `try_acc_close`). So for a banned user this branch returns `Ok(None)` and `create.rs:120-123` turns it into `Ok(())` — where it previously returned `AccountAlreadyInitialized`.

Repro: create and activate a user, `RequestBanUser` it, re-run `CreateUser` with the same owner/device/type — the transaction succeeds while the account stays `Banned` with a zeroed tunnel.

Nothing is corrupted (this path writes nothing), and `find_or_create_user` (`connect.rs:519-523`) still pre-checks `Banned` today. But the guarantee the PR advertises — `connect` calling this unconditionally and retrying safely — is the one that breaks: on the #4111 flow the program is what has to refuse, and it doesn't. Ban enforcement shouldn't rest on a client-side check.

Suggested fix, after the owner check at `:207`:

```rust
if existing.status != UserStatus::Activated {
    msg!("user exists with status {}; not re-creatable", existing.status);
}
```

`UserStatus` is already in scope via `use crate::state::user::*`. Worth an SVM test that bans an activated user, re-runs `CreateUser`, and asserts the specific custom code (CLAUDE.md Testing §1).

smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs
```diff
@@ -204,6 +198,39 @@ pub fn create_user_core(
...
+    if already_exists {
+        if core.user_account.owner != program_id {
+            return Err(ProgramError::IncorrectProgramId);
+        }
+        let existing = User::try_from(core.user_account)?;
```
nikw9944, line 209: `tenant_pk` is missing from the equality check, so a re-run naming a **different tenant** succeeds and leaves the user in the old VRF.

`CreateUser` takes an optional tenant account (`create.rs:84-90`) and the shipped SDK appends it whenever `tenant_pk` is `Some` (`crates/doublezero-serviceability-instruction/src/user.rs:180-185`, from `sdk/rs/src/commands/user/create.rs:57-71`); `connect.rs:589-596` resolves it from `--tenant`, then the config file, then the pass allowlist. This branch returns at `:224` — before the `tenant_allowlist` enforcement at `:229-252` — and never compares `existing.tenant_pk`.

Tenant drives VRF ID and metro routing in the generated device config (`controlplane/controller/internal/controller/server.go:646-660`), so the caller believes it moved the user into a new VRF and it didn't. That is verbatim the reasoning this PR gives for rejecting a different `device_pk`, and tenant is a routing-isolation boundary. Changing tenant has a legitimate path (`UpdateUser`, `update.rs:318`); `CreateUser` should reject rather than absorb it.

```rust
let requested_tenant = core.tenant_account.map(|a| *a.key).unwrap_or_default();
if existing.owner != effective_owner
    || existing.device_pk != *core.device_account.key
    || existing.user_type != user_type
    || existing.tenant_pk != requested_tenant
{
```

plus a test mirroring `test_user_create_existing_different_device_rejected`.

smartcontract/programs/doublezero-serviceability/src/processors/user/create_subscribe.rs
```diff
@@ -136,18 +138,30 @@ pub fn process_create_subscribe_user(
         value.tunnel_endpoint,
         value.publisher,
         owner_override,
-        Some(mgroup_account.key),
-        feed_account,
-    )?;
+    )?
+    else {
```
nikw9944, line 142: This arm has no test coverage, though the PR description claims "a duplicate user is still an error" on this path.

It is also load-bearing rather than merely conservative, which is worth saying in the comment: if it fell through, `enforce_feed_metro_gate` would tick a **second** feed seat and push a duplicate `feed_pks` entry, which `User::release_feed_seats` (`state/user.rs:519-526`) would then double-release on delete.

Add a duplicate-`CreateSubscribeUser` test asserting `AccountAlreadyInitialized`, and reword the comment to state the reason rather than just "only CreateUser is idempotent".

smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs
```diff
@@ -204,6 +198,39 @@ pub fn create_user_core(
         return Err(DoubleZeroError::Unauthorized.into());
     }
 
+    // Idempotent create: an existing user matching the request is a no-op, so a caller can retry
```
nikw9944, line 201: "Checked after the pass validation so the no-op still requires the caller's own pass" overstates what has run at this point — the epoch check (`:256-266`) and the tenant allowlist (`:229-252`) both come after this branch. Narrowing it to what actually precedes it (the `user_payer` and `client_ip` checks) keeps a future reader from assuming expiry is covered.

nikw9944 (changes requested):
Four things to resolve before this lands.

**The idempotent path is under-specified.** It matches on `owner`/`device_pk`/`user_type` but ignores `existing.status`, so a re-run for a user that `RequestBanUser` left onchain as `Banned` (with `tunnel_net`/`tunnel_id`/`dz_ip` zeroed) returns a successful transaction. It also ignores `existing.tenant_pk`, so a re-run naming a different tenant succeeds and silently leaves the user in the old VRF — the same failure mode the PR correctly rejects for `device_pk`. Both matter specifically because the stated goal is to let `connect` call this unconditionally (#4111); at that point the program is the only thing that can say no.

**Removing the feed gate removes the only bound on EdgeSeat multicast user count.** `try_add_user` is an unconditional `Ok(())` for EdgeSeat + Multicast and `max_multicast_users` is enforced nowhere, so one dynamic EdgeSeat pass can now create a bare multicast user per client IP until a device's `max_multicast_subscribers` is exhausted, charging no seat. Either enforce the pass-level cap or state in the changelog that it is intentionally unbounded.

**The bare multicast state is over-permissioned on the device.** The controller template selects the subscriber ingress ACL only when `len(MulticastSubscribers) > 0`, so a zero-role multicast user falls through to `SEC-USER-PUB-MCAST-IN`, which permits `ip any <MulticastGroupBlock>`. That condition predates this PR, but this PR makes zero-role the normal intermediate state of every EdgeSeat multicast connect.

The mechanical feed-gate relocation itself is correct: the seat tick persists (both callers write `result.accesspass` after the gate), `feed_pks.push` mutates the struct that is actually created, and the reordering relative to `try_add_user` is observationally neutral.

- **High — removing the feed gate leaves nothing bounding how many multicast users an EdgeSeat pass can create.** `AccessPass::try_add_user` (`state/accesspass.rs:326`) returns `Ok(())` for EdgeSeat + Multicast without incrementing `multicast_user_count` or consulting `max_multicast_users`, and `remove_user` skips it too (`:341-348`). The doc comment at `:318-321` justifies that on the grounds that the per-feed `FeedSeat` supersedes it — which held only because every EdgeSeat multicast create had to pass `enforce_feed_metro_gate` and tick a seat. This PR removes that gate, so the premise is gone and no bound replaces it; `max_multicast_users`' only remaining consumer anywhere is the airdrop multiplier (`processors/accesspass/set.rs:377`).

Concretely: one EdgeSeat pass at the UNSPECIFIED PDA with zero `FeedSeat`s can call `CreateUser { user_type: Multicast, client_ip: <any> }` repeatedly. The IP match is waived for an UNSPECIFIED / `allow_multiple_ip` pass (`create_core.rs:189-199`), multicast is neither epoch- (`state/user.rs:48-50`) nor tenant-gated (`create_core.rs:229`), and each distinct `client_ip` yields a distinct User PDA. Each user bumps `device.users_count` and `device.multicast_subscribers_count`, takes a global `UserTunnelBlock` allocation plus a device tunnel ID, and materializes a Tunnel interface, BGP neighbor and ACLs on the box — until `max_multicast_subscribers` is exhausted. Cost is refundable rent plus ledger fees; no seat is charged. The updated test at `tests/delete_user_dynamic_accesspass.rs:743-768` is the before/after of exactly this, now asserting `connection_count == 2` with `multicast_user_count == 0`.

Two qualifications: it needs a hand-built transaction (`doublezero user create` only emits IBRL / IBRLWithAllocatedIP, `smartcontract/cli/src/user/create.rs:100-106`), and the same gap already exists for every non-EdgeSeat pass type since `try_add_user` early-returns for those (`accesspass.rs:323-325`). But EdgeSeat is the metered SKU where the seat cap *is* the product, and this is the change that removes its only bound.

Suggested fix — make the pass-level cap real, orthogonal to the per-feed cap:

```rust
UserType::Multicast => {
    // FeedSeat caps how many users a *feed* admits; this caps how many multicast
    // users the pass may create at all, bare ones included (reachable since #4110).
    if self.multicast_user_count >= self.max_multicast_users {
        return Err(DoubleZeroError::AccessPassMaxMulticastUsersExceeded);
    }
    self.multicast_user_count += 1;
    Ok(())
}
```

with the matching decrement in `remove_user`, the stale supersede paragraph at `:318-321` rewritten, and `tests/delete_user_dynamic_accesspass.rs:766` updated. If leaving it unbounded is a deliberate product call, please say so in the changelog — "EdgeSeat multicast user count is unbounded" is materially different from what the entry currently implies.
- **High — a bare multicast user gets the *publisher* ingress ACL on the device.** The bare multicast user this PR makes creatable reaches the controller with `IsMulticast = true` and empty subscriber/publisher/boundary lists (`controlplane/controller/internal/controller/server.go:662-693`). `controlplane/controller/internal/controller/templates/tunnel.tmpl:174` selects the subscriber ACL only when `len(MulticastSubscribers) > 0`:

```
{{- if and .IsMulticast (gt (len .MulticastSubscribers) 0) (eq (len .MulticastPublishers) 0) }}
   ip access-group SEC-USER-SUB-MCAST-IN in
{{- else }}
   ip access-group SEC-USER-PUB-MCAST-IN in
```

Zero roles falls into the `else`, so the tunnel gets `SEC-USER-PUB-MCAST-IN` — which permits `ip any <MulticastGroupBlock>` (`tunnel.tmpl:447-454`) — instead of `SEC-USER-SUB-MCAST-IN`, which does not (`:456-462`). The egress boundary ACL is empty and denies `224.0.0.0/4`, so *receive* is correctly dead; it is the ingress filter that no longer distinguishes this user from a paying publisher. Whether packets actually reach a group tree depends on PIM/RPF state (`pim ipv4 border-router` is not applied without publishers), so this is not a proven injection — but the packet filter is the wrong one by design intent, not by accident of state.

The template condition predates this PR (an unsubscribe-to-zero reaches the same state). What changes is that zero-role multicast becomes the *normal* intermediate state of every EdgeSeat multicast connect — create bare, then `SubscribeFeed` — and a user can sit in it indefinitely, since there is no CLI verb or SDK command for `SubscribeFeed` yet (only the raw builder at `crates/doublezero-serviceability-instruction/src/multicastgroup.rs:210`). Keying the condition on the publisher count alone fixes it:

```
{{- if and .IsMulticast (eq (len .MulticastPublishers) 0) }}
```

smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4116/changes/fa2071a9bc69de04abb242e831e5adb4674e187e

smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4116/changes/fa2071a9bc69de04abb242e831e5adb4674e187e

smartcontract/programs/doublezero-serviceability/src/processors/user/create_subscribe.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4116/changes/fa2071a9bc69de04abb242e831e5adb4674e187e

smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4116/changes/fa2071a9bc69de04abb242e831e5adb4674e187e

nikw9944 (changes requested):
All four inline findings from the last round are fixed and verified at `fa2071a9`, with tests:

- The idempotent path now rejects a banned user with `InvalidStatus` (`create_core.rs:230-235`), covered by `test_user_create_rerun_on_banned_user_rejected` asserting `Custom(7)`. Narrowing to `== Banned` rather than `!= Activated` is complete — `Banned` and `Activated` are the only statuses any live writer sets, and `DeleteUser` closes the account atomically.
- `existing.tenant_pk` is now part of the exists-match (`create_core.rs:209-213`), covered by `test_user_create_existing_different_tenant_rejected`.
- The `CreateSubscribeUser` duplicate arm now documents why it must error and is covered by `test_duplicate_create_subscribe_user_rejected`, which asserts the seat is not double-ticked rather than just the error code.
- The placement comment now names the gates on each side accurately.

Two findings from the review body remain open and unanswered, both unchanged in the current tree:

**The bare multicast state is over-permissioned on the device.** `tunnel.tmpl:174` selects the subscriber ingress ACL only when `len(MulticastSubscribers) > 0`, so a zero-role multicast user falls through to `SEC-USER-PUB-MCAST-IN`, which permits `ip any <MulticastGroupBlock>`. That condition predates this PR, but this PR makes zero-role the normal intermediate state of every EdgeSeat multicast connect — and with no CLI verb for `SubscribeFeed` yet, a user can sit in it indefinitely.

**Nothing bounds EdgeSeat multicast user count.** `try_add_user` is still an unconditional `Ok(())` for EdgeSeat + Multicast and `max_multicast_users` is enforced nowhere, so one dynamic EdgeSeat pass can create a bare multicast user per client IP until a device's `max_multicast_subscribers` is exhausted, charging no seat. The doc comment at `accesspass.rs:316-321` still asserts the per-feed-seat supersede premise that this PR invalidates for the bare-create path. The changelog was revised in these commits without addressing this, so neither enforcing the cap nor documenting it as intentional has happened.

- **High — a bare multicast user gets the *publisher* ingress ACL on the device.** Unchanged at `fa2071a9`. The bare multicast user this PR makes creatable reaches the controller with `IsMulticast = true` and empty subscriber/publisher/boundary lists (`controlplane/controller/internal/controller/server.go:662-693`). `controlplane/controller/internal/controller/templates/tunnel.tmpl:174` selects the subscriber ACL only when `len(MulticastSubscribers) > 0`:

```
{{- if and .IsMulticast (gt (len .MulticastSubscribers) 0) (eq (len .MulticastPublishers) 0) }}
   ip access-group SEC-USER-SUB-MCAST-IN in
{{- else }}
   ip access-group SEC-USER-PUB-MCAST-IN in
```

Zero roles falls into the `else`, so the tunnel gets `SEC-USER-PUB-MCAST-IN` — which permits `ip any <MulticastGroupBlock>` (`tunnel.tmpl:447-454`) — instead of `SEC-USER-SUB-MCAST-IN`, which does not (`:456-462`). The egress boundary ACL is empty and denies `224.0.0.0/4`, so *receive* is correctly dead; it is the ingress filter that stops distinguishing this user from a paying publisher. Whether packets actually reach a group tree depends on PIM/RPF state (`pim ipv4 border-router` is not applied without publishers), so this is not a proven injection — but the packet filter is the wrong one by design intent, not by accident of state.

The condition predates this PR. What this PR changes is that zero-role multicast becomes the *normal* intermediate state of every EdgeSeat multicast connect — create bare, then `SubscribeFeed` — and a user can sit in it indefinitely, since there is still no CLI verb or SDK command for `SubscribeFeed` (only the raw builder at `crates/doublezero-serviceability-instruction/src/multicastgroup.rs:210`). Keying the condition on the publisher count alone fixes it:

```
{{- if and .IsMulticast (eq (len .MulticastPublishers) 0) }}
```
- **High — nothing bounds how many multicast users an EdgeSeat pass can create.** Unchanged at `fa2071a9`. `AccessPass::try_add_user` (`smartcontract/programs/doublezero-serviceability/src/state/accesspass.rs:322`) still returns an unconditional `Ok(())` for EdgeSeat + Multicast without incrementing `multicast_user_count` or consulting `max_multicast_users`, and `remove_user` still skips it. The doc comment at `:316-321` still justifies that with the supersede argument — "for EdgeSeat **multicast** the authoritative cap is the per-feed `FeedSeat`" — which held only while every EdgeSeat multicast create had to clear `enforce_feed_metro_gate` and tick a seat. This PR removes that gate from `CreateUser`, so the comment is now stale and the bound is gone. `max_multicast_users`' only remaining consumer anywhere is the airdrop multiplier (`processors/accesspass/set.rs:377`).

One EdgeSeat pass at the UNSPECIFIED PDA with zero `FeedSeat`s can call `CreateUser { user_type: Multicast, client_ip: <any> }` repeatedly: the IP match is waived for an UNSPECIFIED / `allow_multiple_ip` pass (`create_core.rs:189-199`), multicast is neither epoch- (`state/user.rs:48-50`) nor tenant-gated (`create_core.rs:240`), and each distinct `client_ip` yields a distinct User PDA. Each user bumps `device.users_count` and `device.multicast_subscribers_count`, takes a global `UserTunnelBlock` allocation plus a device tunnel ID, and materializes a Tunnel interface, BGP neighbor and ACLs on the box — until `max_multicast_subscribers` is exhausted. Cost is refundable rent plus ledger fees; no seat is charged.

Qualifications unchanged: it needs a hand-built transaction (`doublezero user create` only emits IBRL / IBRLWithAllocatedIP, `smartcontract/cli/src/user/create.rs:100-106`), and the same gap already exists for every non-EdgeSeat pass type since `try_add_user` early-returns for those. But EdgeSeat is the metered SKU where the seat cap *is* the product, and this is the change that removes its only bound.

The original ask offered two ways out — enforce the cap, or state in the changelog that it is intentionally unbounded. `CHANGELOG.md:13` *was* revised in these commits (adding the tenant, `InvalidStatus`, and V1-PDA details) without addressing this, so neither branch has been taken. Enforcing it:

```rust
UserType::Multicast => {
    // FeedSeat caps how many users a *feed* admits; this caps how many multicast
    // users the pass may create at all, bare ones included (reachable since #4110).
    if self.multicast_user_count >= self.max_multicast_users {
        return Err(DoubleZeroError::AccessPassMaxMulticastUsersExceeded);
    }
    self.multicast_user_count += 1;
    Ok(())
}
```

plus the matching decrement in `remove_user`, the stale paragraph at `:316-321` rewritten, and `tests/delete_user_dynamic_accesspass.rs:766` updated.

smartcontract/programs/doublezero-serviceability/src/state/accesspass.rs
```diff
@@ -311,21 +311,25 @@ impl AccessPass {
...
     ///
-    /// Per the feed-scoped supersede model (#1700): for EdgeSeat **multicast** the authoritative
-    /// cap is the per-feed [`FeedSeat`] (see [`Self::try_add_feed_user`]), so `max_multicast_users`
-    /// is no longer enforced here and is retained only for layout/back-compat. The per-category
-    /// **unicast** cap is still enforced.
+    /// The multicast cap counts users on the pass, bare ones included (reachable since #4110);
+    /// how many users each purchased feed admits is capped separately by its [`FeedSeat`]
+    /// (see [`Self::try_add_feed_user`]).
     pub fn try_add_user(&mut self, user_type: UserType) -> Result<(), DoubleZeroError> {
         if !matches!(self.accesspass_type, AccessPassType::EdgeSeat(_)) {
```
nikw9944, line 322: Enforcing this cap is the right fix, but it is a live-state change worth a pre-deploy check.

The field wasn't merely unenforced before — the code it replaces documented it as inert ("no longer enforced here and is retained only for layout/back-compat"), so nothing had a reason to keep it accurate. Three things make that matter now:

- `SetAccessPassFeeds` — the instruction the oracle uses to provision EdgeSeat feed seats (CHANGELOG:84) — has zero references to `max_multicast_users`. Feed seats and this cap are set by different instructions (`SetAccessPass`, `set.rs:290`), so nothing onchain keeps them consistent.
- `max_multicast_users` deserializes with `unwrap_or(1)` (`:259`) when the field is absent from the stored bytes.
- CHANGELOG:25 notes testnet already holds an EdgeSeat pass.

So an existing EdgeSeat pass whose value is stale or absent (→ 1) starts rejecting multicast connects the feed-seat model admitted, from the moment the program is deployed: a customer with three purchased feed seats gets one multicast user.

The changelog's mitigation — "the oracle already provisions it as the total purchased seats" — is a claim about a service outside this repo, so I couldn't confirm it from the workspace; it needs the oracle's provisioning code or the actual pass state on testnet/mainnet.

Suggested: before deploying, read `max_multicast_users` on every EdgeSeat pass per cluster and confirm it is at least the sum of that pass's feed-seat caps, backfilling via `SetAccessPass` where it isn't — then record it as a deploy note on the changelog entry. There's precedent at CHANGELOG:26: "Deploy note: confirm no nonzero `feed_pk` account exists on each cluster at upgrade time."

nikw9944 (approved):
Both High findings are fixed and verified at `77fa9c24`.

- **Bare multicast user got the publisher ingress ACL** — `tunnel.tmpl:174` now keys on `len(MulticastPublishers) == 0`, so a role-less user gets `SEC-USER-SUB-MCAST-IN`. The same bug shape in the announce prefix-list at `:406` was also fixed, which the finding did not name: a role-less user previously got `permit <DzIp>/32`, and since `dz_ip = client_ip` for a non-publisher it was announcing its client IP into the DZ multicast community. Both changes are narrow — only the zero-role case moves. Pinned by a golden fixture asserting the subscriber ACL and the deny-all prefix-list, driven from new role-less tunnel cases in `render_test.go` and `server_test.go`.
- **Nothing bounded EdgeSeat multicast user count** — `try_add_user` now enforces the pass-level cap with a matching decrement in `remove_user`, the stale supersede doc comment is rewritten, and there is no double-count with the per-feed `FeedSeat` layer. Covered by a unit test asserting `AccessPassMaxMulticastUsersExceeded` and by `test_edge_seat_user_caps_enforced` asserting `Custom(90)` on the second bare multicast user.

Full serviceability suite (50 binaries), clippy, and the controller Go tests are all green.

One new item, Medium: enabling `max_multicast_users` makes a previously-inert field load-bearing against passes that already exist. `SetAccessPassFeeds` never sets it, so nothing onchain keeps it consistent with the feed seats, and it defaults to 1 when absent from the stored bytes. Worth a pre-deploy value check plus a changelog deploy note — details inline.

smartcontract/programs/doublezero-serviceability/src/state/accesspass.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4116/changes/976c347bdcc27990bd2af40d220316d747d3f8b2

PR #4118, opened 2026-07-30
build: bump rust toolchain to 1.97.1

ben-dz:
Pinned at 1.90.0 (Sept 2025) — seven releases behind stable and below rust-analyzer's supported floor of 1.94.0, so the shipped language server warns on every workspace load. That floor tracks stable at roughly minus-three, so the gap only widens.

Four places have to move together: `rust-toolchain.toml`, the twelve `dtolnay/rust-toolchain` CI pins across seven workflows, the devcontainer rust feature, and the rustup pin in `release/Dockerfile.release`. **A devcontainer rebuild is needed after this merges**; the release image rebuilds itself, since `scripts/build-snapshot.sh` keys that on the Dockerfile hash.

The rest is clippy fallout: `useless_borrows_in_formatting`, `unnecessary_sort_by`, `iter_kv_map`, `to_string_in_format_args`, `collapsible_match`, `result_large_err`.

Two spots worth a reviewer's attention, both in `disconnect.rs`:

- Clippy wanted a match guard for `collapsible_match`, but a guard forces a `_ => {}` catch-all and loses exhaustiveness over `DzMode` — a new mode would silently skip filtering instead of failing the build. Hoisted into a `mode_mismatch` bool to keep the match exhaustive.
- `result_large_err` flagged a closure returning `Err(user)` (272 bytes) purely as a keep-retrying signal; the payload is never read. Changed to `Err(())` rather than boxing.

The 18 serviceability changes are redundant `&` removals inside `Display`/`Debug` impls — `format_args!` auto-references, so formatted output is byte-identical. They were required, not optional: `lint-programs` runs from `smartcontract/` and so picks up the root toolchain. Program codegen is unaffected; `cargo build-sbf --tools-version v1.54` supplies its own rustc.

## Testing Verification

Green locally under 1.97.1: `make rust-fmt`, `make rust-lint` (including `lint-programs`), 20 workspace test binaries, 400 program lib tests, and `rust-program-accounts-compat` — zero failures throughout.

`cargo test-sbf` could not be run locally, as Anza publishes no agave installer for `aarch64-unknown-linux-gnu`. As a substitute I ran the serviceability integration tests natively (660 tests across 50 binaries, zero failures), but CI's x86 runners are the real check on the SBF path.


nikw9944 (commented):
One miss: `release/Dockerfile.release:8` is a fourth place the toolchain pin has to move, and it's still on 1.90.0 under a comment claiming it matches `rust-toolchain.toml`.

`scripts/build-snapshot.sh:113` rebuilds the release image only when the Dockerfile's sha256 changes, so leaving the file untouched means contributors keep the cached 1.90.0 image. `goreleaser` then runs from `/workspace`, where `rust-toolchain.toml` says 1.97.1, and rustup downloads a second full toolchain. `RUSTUP_HOME=/usr/local/rustup` is not among the named volumes (`build-snapshot.sh:150-154` mounts only the cargo registry, cargo git, target dir, and Go cache) and the container is `--rm`, so that download repeats on every snapshot build and is never cached — and the build newly depends on reaching static.rust-lang.org.

Fix: bump line 8 to `1.97.1`. That changes the Dockerfile hash too, so the image rebuilds itself on the next `build-snapshot.sh` / `deploy-snapshot.sh` run with no extra step for contributors.

Everything else checks out. The `mode_mismatch` hoist in `disconnect.rs` produces the same decision for every `(dz_mode, user_type)` pair and keeps the match exhaustive; `Err(())` is safe because the only consumer discards the payload and `backon` retries on any error type; the three `sort_by` → `sort_by_key` rewrites preserve ordering and stability; `.values()` never needed the map key; and the `&` removals in the `Display`/`Debug` impls are byte-identical output, as described.

- `release/Dockerfile.release:8` still pins `--default-toolchain 1.90.0`, under a comment that says it matches `rust-toolchain.toml`. This is a fourth place that needs to move with the bump. `scripts/build-snapshot.sh:113` rebuilds the image only when the Dockerfile's sha256 changes, so an untouched file means contributors keep the cached 1.90.0 image; `goreleaser` then runs from `/workspace` where `rust-toolchain.toml` says 1.97.1, and rustup downloads a second full toolchain. Since `RUSTUP_HOME=/usr/local/rustup` is not one of the named volumes (`build-snapshot.sh:150-154` mounts only the cargo registry, cargo git, target dir, and Go cache) and the container runs `--rm`, that download repeats on every snapshot build and is never cached — and the build newly depends on reaching static.rust-lang.org. Fix: bump line 8 to `1.97.1`, which also changes the Dockerfile hash so the image rebuilds itself on the next `build-snapshot.sh` / `deploy-snapshot.sh` run.

martinsander00 (approved):

nikw9944 (approved):
lgtm. The stale `release/Dockerfile.release` rustup pin is fixed — all five toolchain surfaces (`rust-toolchain.toml`, the 12 workflow pins, the devcontainer feature, and the release image) now read 1.97.1, with no 1.90.0 left outside the CHANGELOG's own before/after prose. Bumping the Dockerfile changes its hash, so `scripts/build-snapshot.sh` rebuilds the image itself and the repeated in-container toolchain download is gone.

PR #4119, opened 2026-07-30
cli: unsubscribe orphaned users before rotating or deleting a feed

nikw9944:
## Summary of Changes

* `doublezero feed update` (when `--group` drops a group) and `doublezero feed delete` now unsubscribe the EdgeSeat users the change would orphan — or fail closed listing them, unless `--force-unsubscribe` is passed.
* An EdgeSeat user's multicast groups must stay a subset of the groups carried by the feeds on their access pass; serviceability relies on that when deciding whether an unsubscribe releases a feed's seat. `UpdateFeed`/`DeleteFeed` can break it and the program cannot notice — there is no reverse index from a `Feed` to its subscribers — and the resulting state lets a later unsubscribe release a seat the user still needs.
* Unsubscribe-then-rotate is the ordering that only ever passes through legal states: a user holding fewer groups than their feeds offer is fine. Rotate-then-unsubscribe would publish the illegal state for as long as the cleanup takes.
* A rename and an update without `--group` are unchanged and scan nothing. Any `--group` set runs the guard — it re-derives the dropped set from its own fresh scan, so an additive change costs one scan and finds nothing to do.
* Coverage is per role and mirrors seat release: a feed covers a group only when it is seated on an accepted pass *and* among the seats the user consumed (`user.feed_pks`), and a role the pass's own `mgroup_pub_allowlist`/`mgroup_sub_allowlist` still authorizes is not orphaned (the program runs that check for every pass type). A mixed case — one role removable while the allowlist still covers the other — fails closed even with the flag, since preserving a role through `UpdateMulticastGroupRoles` would be a grant with different authorization; the automated path stays removal-only.

The removals go through `UpdateMulticastGroupRoles`, which needs `USER_ADMIN` (or foundation membership) on the payer in addition to the `FEED_AUTHORITY` these commands already require. The first removal failing leaves the feed untouched, so that surfaces before anything is submitted.

### Two notes for reviewers

* **`UpdateFeedSubscription` follow-up.** Once the #4108 series lands a CLI/SDK surface for instruction 117, the unsubscribe leg should switch to it: `subscriber: false` drops the groups *and* releases the seat, and its `GroupNotInFeed` validation passes because removals run pre-rotation. That switch also moves the required permission from `USER_ADMIN` to `ACCESS_PASS_ADMIN`. Nothing here conflicts with #4113 — it rewrites the *add* path and the internal authorization plumbing; the guard only issues removals through the public SDK command.
* **Deleted-feed seat residue is deliberately out of scope.** The guard covers subscriptions only. `FeedSeat` entries naming a deleted feed stay on existing access passes; that is a known bug with a fix in progress elsewhere, and this PR leaves it there so the two fixes do not collide.

## Testing Verification
* New tests
* Fixed three `user::create_subscribe` tests that were latently order-dependent: they passed a `Pubkey::new_unique()` string as `--subscriber` and relied on its base58 length landing outside `parse_pubkey`'s 43–44 char window, which varies with the process-global counter. Pinned to a fixed short pubkey.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 ### Breaking
 
 - CLI
-  - `doublezero feed update` (with `--group`) and `doublezero feed delete` now fail closed if the change would leave an EdgeSeat user holding a multicast group role that no feed on their access pass carries and the pass's own allowlists don't authorize, printing the affected users, groups, and roles and submitting nothing. Pass `--force-unsubscribe` to remove those roles first (allowlist-covered roles are re-asserted, not removed) and then apply the change. A rename or an update without `--group` is unaffected; an additive `--group` set scans but finds nothing to do. The removals need `USER_ADMIN` (or foundation membership) on the payer in addition to `FEED_AUTHORITY`. (malbeclabs/infra#2113, #4119)
+  - `doublezero feed update` (with `--group`) and `doublezero feed delete` now fail closed if the change would leave an EdgeSeat user holding a multicast group role that no consumed feed seat on their access pass covers and the pass's own allowlists don't authorize, printing the affected users, groups, and roles and submitting nothing. Pass `--force-unsubscribe` to remove those roles first and then apply the change; a group where one role must go while the allowlist still authorizes the other fails closed even with the flag and needs manual cleanup. A rename or an update without `--group` is unaffected; an additive `--group` set scans but finds nothing to do. The removals need `USER_ADMIN` (or foundation membership) on the payer in addition to `FEED_AUTHORITY`. (malbeclabs/infra#2113, #4119)
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 ### Breaking
 
 - CLI
-  - `doublezero feed update` (with `--group`) and `doublezero feed delete` now fail closed if the change would leave an EdgeSeat user holding a multicast group role that no feed on their access pass carries and the pass's own allowlists don't authorize, printing the affected users, groups, and roles and submitting nothing. Pass `--force-unsubscribe` to remove those roles first (allowlist-covered roles are re-asserted, not removed) and then apply the change. A rename or an update without `--group` is unaffected; an additive `--group` set scans but finds nothing to do. The removals need `USER_ADMIN` (or foundation membership) on the payer in addition to `FEED_AUTHORITY`. (malbeclabs/infra#2113)
+  - `doublezero feed update` (with `--group`) and `doublezero feed delete` now fail closed if the change would leave an EdgeSeat user holding a multicast group role that no feed on their access pass carries and the pass's own allowlists don't authorize, printing the affected users, groups, and roles and submitting nothing. Pass `--force-unsubscribe` to remove those roles first (allowlist-covered roles are re-asserted, not removed) and then apply the change. A rename or an update without `--group` is unaffected; an additive `--group` set scans but finds nothing to do. The removals need `USER_ADMIN` (or foundation membership) on the payer in addition to `FEED_AUTHORITY`. (malbeclabs/infra#2113, #4119)
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 ### Breaking
 
 - CLI
-  - `doublezero feed update` (when `--group` drops a group) and `doublezero feed delete` now fail closed if the change would leave an EdgeSeat user subscribed to a multicast group no feed on their access pass carries, printing the affected users and groups and submitting nothing. Pass `--force-unsubscribe` to unsubscribe them first and then apply the change. A rename, a purely additive `--group` set, and an update without `--group` are unaffected. The removals need `USER_ADMIN` (or foundation membership) on the payer in addition to `FEED_AUTHORITY`. (malbeclabs/infra#2113)
+  - `doublezero feed update` (with `--group`) and `doublezero feed delete` now fail closed if the change would leave an EdgeSeat user holding a multicast group role that no feed on their access pass carries and the pass's own allowlists don't authorize, printing the affected users, groups, and roles and submitting nothing. Pass `--force-unsubscribe` to remove those roles first (allowlist-covered roles are re-asserted, not removed) and then apply the change. A rename or an update without `--group` is unaffected; an additive `--group` set scans but finds nothing to do. The removals need `USER_ADMIN` (or foundation membership) on the payer in addition to `FEED_AUTHORITY`. (malbeclabs/infra#2113)
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file.
 
 ### Breaking
 
+- CLI
+  - `doublezero feed update` (when `--group` drops a group) and `doublezero feed delete` now fail closed if the change would leave an EdgeSeat user subscribed to a multicast group no feed on their access pass carries, printing the affected users and groups and submitting nothing. Pass `--force-unsubscribe` to unsubscribe them first and then apply the change. A rename, a purely additive `--group` set, and an update without `--group` are unaffected. The removals need `USER_ADMIN` (or foundation membership) on the payer in addition to `FEED_AUTHORITY`. (malbeclabs/infra#2113)
+
 ### Changes
 
 - Serviceability
--- a/smartcontract/cli/src/feed/guard.rs
+++ b/smartcontract/cli/src/feed/guard.rs
@@ -62,7 +62,7 @@ impl Orphan {
     }
 }
 
-/// The onchain state [`plan`] reads, fetched once per round by [`scan`].
+/// The onchain state [`find_orphans`] reads, fetched once per round by [`scan`].
 pub struct Snapshot {
     pub users: HashMap<Pubkey, User>,
     pub accesspasses: HashMap<Pubkey, AccessPass>,
@@ -105,7 +105,7 @@ pub fn unsubscribe_orphans<C: CliCommand, W: Write>(
             .copied()
             .collect();
 
-        let orphans = plan(feed_pk, &dropped, &snap)?;
+        let orphans = find_orphans(feed_pk, &dropped, &snap)?;
         if orphans.is_empty() {
             return Ok(());
         }
@@ -180,19 +180,15 @@ fn scan<C: CliCommand>(client: &C) -> eyre::Result<Snapshot> {
 
 /// The memberships that dropping `dropped` from `feed_pk` would orphan.
 ///
-/// A membership survives when some *other* feed still carries the group in the user's metro,
-/// provided that feed is both seated on an accepted pass and among the seats the user actually
-/// consumed (`user.feed_pks`) — the same set `UnsubscribeFeed`'s retained-feeds check releases
-/// seats against. A seated feed the user never consumed is not coverage: keeping the group on its
-/// strength would leave usage no seat accounts for. The rotated feed itself never counts: every
-/// group in `dropped` is by construction absent from its post-change set, and a delete removes it
-/// entirely.
-///
-/// A role also survives when an accepted pass's own allowlist authorizes it
-/// (`mgroup_pub_allowlist` for publisher, `mgroup_sub_allowlist` for subscriber): the program runs
-/// that check for every pass type, so such a membership is legal with no feed involved, and seat
-/// release never keys on a group the feeds don't carry.
-pub fn plan(feed_pk: &Pubkey, dropped: &[Pubkey], snap: &Snapshot) -> eyre::Result<Vec<Orphan>> {
+/// A role survives when another feed the user holds a consumed seat on (`user.feed_pks`, the set
+/// seat release retains against) still carries the group in their metro, or when the pass's own
+/// per-role allowlist authorizes it. The rotated feed itself never counts: every group in
+/// `dropped` is by construction absent from its post-change set.
+pub fn find_orphans(
+    feed_pk: &Pubkey,
+    dropped: &[Pubkey],
+    snap: &Snapshot,
+) -> eyre::Result<Vec<Orphan>> {
     let mut orphans = Vec::new();
 
     // Index once so the per-user pass lookup doesn't rescan every pass (O(users × passes) when a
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file.
 
 ### Breaking
 
+- CLI
+  - `doublezero feed update` (with `--group`) and `doublezero feed delete` now fail closed if the change would leave an EdgeSeat user holding a multicast group role that no feed on their access pass carries and the pass's own allowlists don't authorize, printing the affected users, groups, and roles and submitting nothing. Pass `--force-unsubscribe` to remove those roles first (allowlist-covered roles are re-asserted, not removed) and then apply the change. A rename or an update without `--group` is unaffected; an additive `--group` set scans but finds nothing to do. The removals need `USER_ADMIN` (or foundation membership) on the payer in addition to `FEED_AUTHORITY`. (malbeclabs/infra#2113)
```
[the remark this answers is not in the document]
nikw9944: Added `(#4119)` alongside the infra ticket in 8112c226.

smartcontract/cli/src/feed/guard.rs
```diff
@@ -0,0 +1,820 @@
...
+
+        // Every pass the program would accept for this user. Not the single pass the SDK read
+        // path prefers: `create_user` and `update_multicastgroup_roles` both accept either the
+        // exact-IP PDA or the shared UNSPECIFIED one, so a dynamic Prepaid pass must not be
+        // allowed to hide a live EdgeSeat pass at the other PDA.
+        let passes: Vec<&AccessPass> = accepted_passes(&snap.accesspasses, user).collect();
+        let seated: Vec<Pubkey> = passes
+            .iter()
+            .flat_map(|pass| pass.feed_seats().iter().map(|seat| seat.feed_key))
+            .collect();
```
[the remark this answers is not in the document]
nikw9944: Done in 8112c226 — `plan()` now builds a `HashMap<Pubkey, Vec<&AccessPass>>` keyed by `user_payer` once per call and filters by `client_ip` per user. One note on the premise: the pass lookup only ran for users holding a dropped group (the `held.is_empty()` continue precedes it), so the quadratic term was bounded by affected users rather than all users — but a mass rotation makes those the same thing, so the index is worth having.

nikw9944:
Addressed all review feedback.

**Allowlist coverage (guard.rs)** — fixed in 34919671. Coverage is now per role: a role the pass's own `mgroup_pub_allowlist`/`mgroup_sub_allowlist` still authorizes (on any accepted pass, either PDA) is not orphaned, and a forced removal re-asserts the covered role so only the uncovered one is removed. The re-assertion is safe onchain: a kept role is only ever allowlist-covered — feed coverage is role-agnostic, so it can never produce a mixed case — and `check_mgroup_allowlists` passes exactly for allowlist-covered roles. `Orphan` now carries explicit `remove_*`/`keep_*` fields and the printed plan shows both. New planner tests cover sub-allowlist coverage, the pub/sub role mismatch, the mixed removal that keeps the covered role, and an allowlist on the other PDA.

**Stale drop-vs-additive decision (update.rs/delete.rs)** — fixed in 34919671. The guard now runs whenever `--group` is present and on every delete; its per-round scan re-derives the dropped set, so a group added between the initial `get_feed` read and the submit is caught. An additive update costs one scan and finds nothing to do.

**Changelog PR reference** and **pass indexing in `plan()`** — fixed in 8112c226, replied in-thread.

`cargo test -p doublezero-serviceability-cli` (394 passing), `make rust-fmt`, and clippy `-Dwarnings` are clean. The runbook in malbeclabs/infra#2129 was updated to match the per-role coverage and always-scan behavior.

smartcontract/cli/src/feed/guard.rs
```diff
@@ -0,0 +1,820 @@
...
+        for orphan in &orphans {
+            client
+                .update_multicastgroup_roles(UpdateMulticastGroupRolesCommand {
+                    user_pk: orphan.user_pk,
+                    group_pk: orphan.group_pk,
+                    client_ip: orphan.client_ip,
+                    // Desired absolute state, not a toggle: an allowlist-covered role is
+                    // re-asserted so only the orphaned role is removed.
+                    publisher: orphan.keep_publisher,
+                    subscriber: orphan.keep_subscriber,
```
martinsander00: when a user holds both the publisher and the subscriber role on a group being dropped, and only one of them still has to go, this keeps the other one by sending it as `true`. ix 116 reads `true` as "grant this role", not "leave it alone", so the call is no longer a removal: `removal_only` goes false and the payer now needs `ACCESS_PASS_ADMIN` instead of the `USER_ADMIN` the hint below names (subscribe.rs:254-259), the user now has to be Activated (subscribe.rs:217-218), and the allowlist gets re-checked against whichever single pass the sdk resolved, which prefers the UNSPECIFIED pda while `plan` looked at both.

so if the pass allowlists the subscriber role but no feed covers the publisher role, this submits `publisher: false, subscriber: true` and comes back with "User not allowed to subscribe multicast group". it fails closed, but there is no way to finish the run and the hint sends the operator after the wrong permission.

should we just leave those users to the operator instead of unsubscribing them automatically?

smartcontract/cli/src/feed/guard.rs
```diff
@@ -0,0 +1,820 @@
...
+        if held.is_empty() {
+            continue;
+        }
+
+        // Every pass the program would accept for this user. Not the single pass the SDK read
+        // path prefers: `create_user` and `update_multicastgroup_roles` both accept either the
+        // exact-IP PDA or the shared UNSPECIFIED one, so a dynamic Prepaid pass must not be
+        // allowed to hide a live EdgeSeat pass at the other PDA.
+        let passes: Vec<&AccessPass> = accepted_passes(&snap.accesspasses, user).collect();
+        let seated: Vec<Pubkey> = passes
```
martinsander00, line 229: `seated` here is every feed on the pass, but when `UnsubscribeFeed` later decides whether a group is still covered, it only counts feeds that are on the pass *and* listed in `user.feed_pks` (unsubscribe_feed.rs:96-98). so this is more permissive than the check whose invariant the guard is protecting.

say a pass has seats on two feeds that both carry group G, but the user only ever connected through the first one, so `feed_pks` holds just that one. rotating G out of the first feed looks fine to the guard, because the second feed still carries it. when the user later unsubscribes the first feed, the program finds nothing to strip, releases that seat, and the user keeps G with the second seat never counted.

should `seated` be intersected with `user.feed_pks`?

smartcontract/cli/src/feed/guard.rs
```diff
@@ -0,0 +1,820 @@
...
+) -> eyre::Result<()> {
+    let mut rounds_done = 0;
+    loop {
+        let snap = scan(client)?;
+
+        // Re-derive the dropped set from the freshly scanned feed rather than from the group set
+        // the command read before the loop: a group added to the feed mid-run is dropped by this
+        // change too, and pinning the earlier set would let its subscribers through unseen.
+        let feed = snap.feeds.get(feed_pk).ok_or_else(|| {
+            eyre::eyre!("feed '{feed_code}' ({feed_pk}) disappeared mid-run; nothing was changed")
```
martinsander00: from round 2 onward the removals above have already been submitted, so this isn't quite accurate. the sibling at :112 says "nothing was changed on the feed", which reads right for both cases.

smartcontract/cli/src/feed/update.rs
```diff
@@ -71,18 +92,214 @@ impl UpdateFeedCliCommand {
...
+        // with the membership gone.
+        f.expect_scan(&mut client, vec![]);
+        client
+            .expect_update_multicastgroup_roles()
+            .with(predicate::eq(UpdateMulticastGroupRolesCommand {
+                user_pk: f.user_pk,
+                group_pk: g2,
+                client_ip: f.client_ip,
+                publisher: false,
+                subscriber: false,
```
martinsander00, line 174: both force tests pin `publisher: false, subscriber: false`, so the `keep_*` path never reaches the sdk in a test. that's the branch where the required flag and the pass resolution both change, so a case that sends `subscriber: true` would be worth having.

smartcontract/cli/src/feed/update.rs
```diff
@@ -71,18 +92,214 @@ impl UpdateFeedCliCommand {
...
+            UpdateFeedCliCommand {
+                pubkey: f.feed_pk.to_string(),
+                exchange: None,
+                name: None,
+                groups: vec![g1.to_string()],
+                force_unsubscribe: true,
+            }
+            .execute(&ctx, &client, &mut output),
+        );
+        assert!(res.is_err(), "{res:?}");
```
martinsander00: could this pin the error message? right now it passes on any failure, including a removal that errored or the feed going missing between rounds, so it would keep passing if the round limit stopped working.

also the comment at :258 says the test avoids hard-coding the round count, but `.times(3)` at :285 hard-codes it. that is actually what makes the run end on the round limit, so it is doing useful work, just not what the comment says.

martinsander00 (commented):
read through the guard and the ix 116 side. two things about the keep_* path and the coverage set, rest is minor.

smartcontract/cli/src/feed/guard.rs
nikw9944: You're right on all three flips — verified against `subscribe.rs` (`removal_only` → `ACCESS_PASS_ADMIN`, the `has_role` Activated gate, and the single-pass allowlist recheck). Fixed in 4876bec0 by taking your first option: mixed-role orphans now fail the run closed even with `--force-unsubscribe`, with the plan flagging the still-allowlisted role, and the automated path stays removal-only (`publisher: false, subscriber: false`), so the USER_ADMIN hint stays accurate. Computing coverage against the SDK-resolved pass would have fixed the allowlist recheck but not the permission/Activated flips, so I didn't take that path.

smartcontract/cli/src/feed/guard.rs
nikw9944: Intersecting with `feed_pks` is right — the invariant the guard protects is the one seat release enforces, and `UnsubscribeFeed` retains only against feeds the user holds (`retained feed … is not held by this user` at unsubscribe_feed.rs:95-98). In your F1+F2 scenario the connect-time test would call G covered, but the user never ticked F2's seat, so G would outlive its accounting exactly as you describe. Fixed in 4876bec0: feed coverage now requires the covering feed to be seated on an accepted pass **and** in `user.feed_pks`, with a planner test for the seated-but-unconsumed case.

smartcontract/cli/src/feed/guard.rs
nikw9944: Fixed in 4876bec0 — it now says "the feed change was not submitted", which is accurate on every round.

smartcontract/cli/src/feed/update.rs
nikw9944: With the mixed-role fix (4876bec0) the `keep_*` path no longer reaches the SDK by design — the run fails closed before submitting anything. The new `test_cli_feed_update_force_fails_closed_on_a_mixed_role_orphan` pins that: a publisher+subscriber orphan with the sub allowlisted bails with `update_multicastgroup_roles` and `update_feed` both at `.times(0)`, and the plan output flags the kept role.

smartcontract/cli/src/feed/update.rs
nikw9944: Both fixed in 4876bec0: the test now asserts the error contains "kept appearing" (the round-limit bail specifically), and the comment is reworded — the uncapped scans are what let the guard run to its own limit, and the `.times(3)` on the removals then pins one removal pass per round.

nikw9944:
Addressed the full review in 4876bec0 (replies in each thread):

- **Mixed-role orphans fail closed.** The `keep_*` re-assertion was broken exactly as described — a `true` is a grant (`removal_only` flips to `ACCESS_PASS_ADMIN`, the user must be Activated, and the allowlist is re-checked against the single SDK-resolved pass). The automated path is now strictly removal-only; an orphan whose other role is still allowlisted fails the run closed with the user flagged for manual cleanup, so the USER_ADMIN hint stays accurate.
- **Feed coverage requires the consumed seat.** Coverage now intersects with `user.feed_pks`, mirroring `UnsubscribeFeed`'s retained-feeds check — a seated feed the user never consumed no longer keeps a group alive with no seat accounting for it.
- **Disappeared-feed message** no longer claims "nothing was changed"; it says the feed change was not submitted.
- **Tests:** new mixed-role force test proves nothing reaches the SDK on that path; the round-limit abort pins its error message and the misleading comment is reworded; planner tests cover consumed vs. unconsumed seats (including the rotated feed's own consumed seat still not counting).

396 tests passing, fmt/clippy clean. The runbook in malbeclabs/infra#2129 (b7c1686) and the changelog entry were updated to match.

smartcontract/cli/src/feed/guard.rs
```diff
@@ -0,0 +1,885 @@
...
+/// seats against. A seated feed the user never consumed is not coverage: keeping the group on its
+/// strength would leave usage no seat accounts for. The rotated feed itself never counts: every
+/// group in `dropped` is by construction absent from its post-change set, and a delete removes it
+/// entirely.
+///
+/// A role also survives when an accepted pass's own allowlist authorizes it
+/// (`mgroup_pub_allowlist` for publisher, `mgroup_sub_allowlist` for subscriber): the program runs
+/// that check for every pass type, so such a membership is legal with no feed involved, and seat
+/// release never keys on a group the feeds don't carry.
+pub fn plan(feed_pk: &Pubkey, dropped: &[Pubkey], snap: &Snapshot) -> eyre::Result<Vec<Orphan>> {
```
martinsander00: i wonder if we can rename this function to something clearer and remove a lot of the explanation above

smartcontract/cli/src/user/create_subscribe.rs
```diff
@@ -588,7 +594,10 @@ mod tests {
             .times(1)
             .returning(move |_| Ok((device_pubkey, device.clone())));
 
-        let mgroup_pubkey = Pubkey::new_unique();
+        // Fixed short base58 (< 43 chars) so `--subscriber` deterministically takes the
```
martinsander00: same comment in 3 different parts, should we add it somewhere else?

martinsander00 (approved):

smartcontract/cli/src/feed/guard.rs
nikw9944: Renamed to `find_orphans` in a56974d9 and trimmed the doc comment to three lines (the coverage rule and the rotated-feed exclusion); the seat-accounting rationale lives inline at the `feed_covered` check where it applies.

smartcontract/cli/src/user/create_subscribe.rs
nikw9944: Hoisted into a single `short_mgroup_pubkey()` helper at the top of the test module in a56974d9; the three sites now just call it.

nikw9944:
Both follow-up nits addressed in a56974d9: `plan()` is now `find_orphans()` with its doc comment cut to the coverage rule, and the repeated pinned-pubkey comment in the `create_subscribe` tests is hoisted into a single `short_mgroup_pubkey()` helper. 396 tests passing, fmt/clippy clean.

PR #4121, opened 2026-07-30
ci: report failed jobs from dispatched runs to the release Slack thread

ben-dz:
## Summary

When a `release.testnet` stage dispatches a workflow into `malbeclabs/infra` and that run fails, the orchestrator exits 1 with the failing component buried in a `gh run watch` dump and posts nothing to the release thread. In v0.32.0 the failing leg (`deploy / onchain-collector-to-s3`) was known at 18:02Z but appeared nowhere a human would look.

- `dispatch-and-wait` is split into steps so a reporter can run under `if: failure()`. It lists the run's failed job names (`gh run view --json jobs`, which yields the useful `deploy / onchain-collector-to-s3` form) and posts them as a threaded reply, capped at 10 names plus `… and N more`.
- The reporter is best-effort: it only warns and always exits 0, so it cannot replace or mask the original failure. A `gh` error, a cancelled run, or a run with no failed job all post the run URL with a note instead of an empty list.
- The inlined `curl` to `chat.postMessage` is gone; both posts now go through the existing `slack-thread-post` action, keeping its `.ok` check (Slack returns HTTP 200 with `ok=false`).
- `qa` was the only dispatching stage with no Slack wiring, so a QA stall or failure was invisible in the thread. It now passes the same three Slack inputs as `deploy-clients`.

**Reviewer trap:** `qa` had `needs: deploy-clients`, and `needs` only exposes *direct* dependencies — reading `needs.preflight.outputs.thread_ts` without adding `preflight` to `needs` yields an empty `thread_ts` and silently degrades to top-level posts. Hence `needs: [deploy-clients, preflight]`, matching `gate-tags`, `stage-programs`, and `deploy-core`.

## Testing Verification

`actionlint` is clean on `release.testnet.yml`; both files parse as YAML.

The reporter's shell body was extracted and run against a stubbed `gh` for five shapes: two failed legs of nine (the v0.32.0 `deploy-clients` case), zero failed jobs, 13 failed (cap plus `… and 3 more`), `gh run view` erroring, and a missing `RUN_ID` from a failed dispatch. All exit 0 with the intended message, checked through `slack-thread-post`'s `jq` filter.

**Not verified:** the Slack failure post end to end — that needs a `release.testnet` dry run or a real release. The v0.32.0 infra run is not readable with this token, so the `--json jobs` shape is taken as reported, not re-confirmed against that run.


.github/actions/dispatch-and-wait/action.yml
```diff
@@ -71,21 +71,86 @@ runs:
...
+      run: |
+        set -uo pipefail
+        # Empty when the dispatch step itself failed: there is no run to report on.
+        if [ -z "${RUN_ID:-}" ]; then
+          echo "::warning::no dispatched run to report on"
+          exit 0
+        fi
+        # Reusable-workflow children come back as e.g. "deploy / onchain-collector-to-s3".
+        NAMES=$(gh run view "$RUN_ID" -R "$REPO" --json jobs \
+          --jq '.jobs[] | select(.conclusion=="failure") | .name') || {
```
nikw9944: This filter misses `timed_out`. GitHub job conclusions are `success | failure | cancelled | timed_out | action_required | neutral | skipped`. A job that exceeds its `timeout-minutes` concludes `timed_out`, not `failure` — it still fails the run, so `gh run watch --exit-status` exits 1 and this reporter runs, matches nothing, and posts `_No failing jobs were reported — see the run for details._` with no job name. Verified: a jobs payload whose only non-success entries are `timed_out` and `cancelled` yields zero names through this jq. A hung deploy or QA leg is at least as common as one that errors out, and it's the case where the job name matters most.

Invert the filter to "concluded and not fine":

```
--jq '.jobs[] | select(.conclusion != null and .conclusion != "success" and .conclusion != "skipped") | .name'
```

That covers `failure`, `timed_out`, `cancelled`, and `action_required` in one expression and needs no maintenance as GitHub adds conclusions. `"\(.name) (\(.conclusion))"` costs nothing and makes a timeout distinguishable from an error. The `TOTAL -eq 0` branch then genuinely means "the run failed with no failed job", which is what its comment claims.

.github/actions/dispatch-and-wait/action.yml
```diff
@@ -71,21 +71,86 @@ runs:
...
-            ERR=$(echo "$RESP" | jq -r '.error // "unknown"' 2>/dev/null | tr -d '\r\n')
-            echo "::warning::failed to post Slack approval ping for $RUN_URL: ${ERR:-unparseable response}"
-          fi
-        fi
+        # env for the shell steps below; output for the Slack steps, whose
+        # `with:` expressions cannot read values set through GITHUB_ENV.
+        {
+          echo "RUN_ID=$RUN_ID"
+          echo "RUN_URL=$RUN_URL"
+        } >> "$GITHUB_ENV"
```
nikw9944: The comment implies GITHUB_ENV is needed for the shell steps, but `env:` fed from `steps.dispatch.outputs.*` works for `run:` steps too — the output/GITHUB_ENV split isn't required. As written the composite mutates the caller's job environment for every step after it, which is a leak a reusable action shouldn't have. Emit `run_id` alongside `run_url` as outputs, drop this block, and set `RUN_ID`/`RUN_URL` from those outputs in the `env:` of the watch and report steps. One mechanism instead of two, and the comment goes away.

.github/actions/dispatch-and-wait/action.yml
```diff
@@ -71,21 +71,86 @@ runs:
...
         gh run watch "$RUN_ID" -R "$REPO" --exit-status --interval 30
+
+    # `if: failure()` in a composite scopes to this composite's own earlier
+    # steps, so this fires when the watch (or the dispatch) above failed. It is
+    # best-effort by design: it only warns and always exits 0 — hence no
+    # `set -e` — so it can never replace or mask the failure the job is already
+    # reporting.
+    - name: Collect failing jobs
+      id: report
+      if: failure() && inputs.slack-bot-token != '' && inputs.slack-channel != ''
```
nikw9944: `failure()` is false when the job is cancelled, and a job that exceeds `timeout-minutes` is cancelled rather than failed. For `qa` (`timeout-minutes: 90`, commented as expecting to queue behind an hourly cron) that means the stall path posts nothing: the last message in the thread is the approval ping from 90 minutes earlier, and `notify-failure` (`if: failure()`) doesn't fire either. Gating on `(failure() || cancelled()) && …` would at least post the run URL with the note so the thread says the stage died.

nikw9944 (commented):
One logic gap worth fixing before merge: the failing-job filter only matches `conclusion == "failure"`, so a job that hits its `timeout-minutes` (conclusion `timed_out`) or one stuck on `action_required` fails the run but produces the "_No failing jobs were reported_" note — the exact case where the job name is most valuable, since the run URL alone doesn't say which leg stalled. Two medium notes on the GITHUB_ENV plumbing and the cancellation path.

.github/actions/dispatch-and-wait/action.yml
ben-dz: Fixed in c2fc502, with one deviation worth flagging: I used `(.conclusion // "") | . != "" and . != "success" and . != "skipped"` rather than the `!= null` form.

`gh run view --json` unmarshals into a Go struct with a `string` conclusion, so a not-yet-concluded job comes back as `""`, not `null` — only the raw REST API gives `null`. Checked against a live in-progress run: `{"status":"in_progress","conclusion":"","ctype":"string"}`. The `!= null` filter therefore passes in-progress jobs through and emits `deploy / qa ()`. That is not hypothetical here, because it collides with the third comment: on the `cancelled()` path the dispatched run is usually still running, so every in-flight leg would get listed with an empty conclusion.

Also took the `(\(.conclusion))` suffix. Note it means `neutral` gets listed too — accepted, since excluding it would reintroduce the maintenance you were trying to remove.

.github/actions/dispatch-and-wait/action.yml
ben-dz: Fixed in c2fc502 — gated on `(failure() || cancelled())` for both the collect and post steps.

One knock-on: on that path the dispatched run is often still in flight, so the old `*$WORKFLOW failed in $REPO*` header would assert a failure that had not happened. The header now carries the run's own outcome (`failure`, `timed_out`, `in_progress`, `unknown` if the query fails), so a cancelled caller posts `qa.testnet.yml in malbeclabs/infra: in_progress` with the run URL.

I left `notify-failure` on bare `if: failure()`. Firing the "release failed" alert on a manual cancel seemed like the wrong trade, and the composite change already covers the silent-stall case you described. Say the word if you want that one moved too.

.github/actions/dispatch-and-wait/action.yml
ben-dz: Fixed in c2fc502. `run_id` is emitted alongside `run_url`, the `GITHUB_ENV` block and its comment are gone, and the watch and report steps take `RUN_ID`/`RUN_URL` from `steps.dispatch.outputs.*` via `env:`. The report step's `[ -z "${RUN_ID:-}" ]` guard still holds for a failed dispatch, since the output resolves to an empty string.

.github/actions/dispatch-and-wait/action.yml
```diff
@@ -71,21 +71,102 @@ runs:
...
+
+    # Empty text means the step above was skipped (no Slack config) or bailed out.
+    - name: Post failure report to Slack
+      if: (failure() || cancelled()) && steps.report.outputs.text != ''
+      uses: ./.github/actions/slack-thread-post
+      with:
+        token: ${{ inputs.slack-bot-token }}
+        channel: ${{ inputs.slack-channel }}
+        thread-ts: ${{ inputs.slack-thread-ts }}
+        color: danger
```
nikw9944, line 171: Now that the report also fires on `cancelled()`, a hand-cancelled release puts a red-bordered `*qa.testnet.yml in malbeclabs/infra: in_progress*` in the thread. The rest of this pipeline deliberately avoids red-on-cancel — `pipeline-complete` uses `!cancelled()` so "deliberately cancelled runs don't end with a red guard or a spurious failure alert", and `notify-failure` stayed on bare `failure()` for the same reason — so this is the one place a clean cancel reads as a failure at a glance. Derive the color from `$OUTCOME` (`danger` for `failure`/`timed_out`, `warning` otherwise) and emit it as a second step output.

nikw9944 (approved):
`c2fc502` resolves all three findings from the previous round, verified against the current diff rather than the replies: the job filter now catches `timed_out`, `cancelled`, and `action_required` while excluding in-flight jobs (the `(.conclusion // "")` form is more robust than the `!= null` version originally suggested); `run_id`/`run_url` are step outputs with the `GITHUB_ENV` block gone; and both report steps are gated on `(failure() || cancelled())` with a header that carries the run's real outcome instead of asserting a failure.

One small thing introduced by that commit: `color: danger` is now fixed on a report that also fires when the caller is cancelled, so a deliberate cancel puts a red attachment in the release thread — the one place in this pipeline where a clean cancel reads as a failure.

.github/actions/dispatch-and-wait/action.yml
ben-dz: I'm okay leaving this. A canceled deploy in red is okay.

PR #4122, opened 2026-07-30
sdk/client: add connect multicast --subscribe-feed and --unsubscribe-feed

### LINKED ISSUE #4111: sdk/client: add connect multicast --subscribe-feed

martinsander00:
Resolves #4111. Part 3 of #4108, the last one.

## Summary

An EdgeSeat buyer joins and leaves whole feeds with one command per machine: `doublezero connect multicast --subscribe-feed <code|pubkey>...` and `--unsubscribe-feed ...`. Joining creates the bare multicast user if none exists (#4116), polls it to Activated, and subscribes. With both flags the leave runs first, and if one half's transaction fails the output says which flag to rerun.

- New SDK commands `SubscribeFeedCommand` and `UnsubscribeFeedCommand` derive the exact group lists the program demands, check the pass, metro, and feed cap before sending, refuse the lookup's pass when a sibling pass is the one seating the user's held feeds (an all-stale cleanup still goes through), and split a join or multi-target leave that does not fit one transaction, naming the parts already applied if a later one fails. The one unsplittable shape, a single-target leave, always fits now that `MAX_USER_FEEDS` is 5; a size test including the compute-budget prelude pins that.
- Every deterministic check runs before any transaction, and on a fresh IP the feeds choose the metro: device selection filters to the requested feeds' exchange, so `--subscribe-feed shreds` picks a device the feed serves instead of failing on the nearest one. An existing user's device is fixed, so a conflicting `--device` fails explicitly.
- The feed flags conflict with `--publish`/`--subscribe` and the legacy positionals, so purchased feeds and comped groups stay separate paths. Feed codes match exactly, like group codes, and the same feed in both flags is rejected on resolved pubkeys.

About 800 lines of non-test code, over the repo guideline; kept as one PR since the two commands mirror each other.

## Testing Verification

- SDK: exact-instruction assertions for group expansion and dedup, retained coverage, stale feeds excluded from retained, an already-held feed retrying as a no-op, splitting in both commands, and metro/cap/not-held/duplicate/wrong-pass rejections that send nothing.
- Connect: a fresh IP creates a bare user then joins on a device in the feed's metro (pinned against a lower-latency device elsewhere); an existing user skips creation; a metro with no eligible device fails before anything is created; overlap and flag mixing are rejected; a failed half prints which flag to rerun; a failed leave with a successful join still provisions the daemon.


crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -370,9 +627,29 @@ impl Connect {
...
+                // programmatically built command.
+                if has_feeds && (has_legacy || has_new) {
+                    eyre::bail!("Cannot mix --subscribe-feed/--unsubscribe-feed with --publish/--subscribe or positional group arguments");
+                }
+                if has_feeds {
+                    if let Some(feed) = sub_feeds.iter().find(|f| unsub_feeds.contains(f)) {
+                        eyre::bail!(
+                            "feed {feed} is in both --subscribe-feed and --unsubscribe-feed"
+                        );
+                    }
```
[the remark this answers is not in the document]
martinsander00: https://github.com/malbeclabs/doublezero/pull/4122/changes/d2aff275da1732ce37a1323671c4548cb59ef3bf

smartcontract/sdk/rs/src/commands/multicastgroup/subscribe_feed.rs
```diff
@@ -0,0 +1,432 @@
...
+    pub fn execute(&self, client: &dyn DoubleZeroClient) -> eyre::Result<Signature> {
+        if self.feed_pks.is_empty() {
+            eyre::bail!("no feeds given");
+        }
+
+        let (_, user) = GetUserCommand {
+            pubkey: self.user_pk,
+        }
+        .execute(client)
+        .map_err(|_err| eyre::eyre!("User not found"))?;
```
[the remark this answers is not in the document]
martinsander00: https://github.com/malbeclabs/doublezero/pull/4122/changes/d2aff275da1732ce37a1323671c4548cb59ef3bf

smartcontract/sdk/rs/src/commands/multicastgroup/unsubscribe_feed.rs
```diff
@@ -0,0 +1,386 @@
...
+        if self.feed_pks.is_empty() {
+            eyre::bail!("no feeds given");
+        }
+
+        let (_, user) = GetUserCommand {
+            pubkey: self.user_pk,
+        }
+        .execute(client)
+        .map_err(|_err| eyre::eyre!("User not found"))?;
+        if user.user_type != UserType::Multicast {
```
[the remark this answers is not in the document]
martinsander00: https://github.com/malbeclabs/doublezero/pull/4122/changes/d2aff275da1732ce37a1323671c4548cb59ef3bf

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -350,6 +406,245 @@ impl Connect {
...
+            }
+            // A pure leave changes routes only; the daemon reconciler picks that up on its own.
+            _ => Ok(false),
+        }
+    }
+
+    /// Resolve everything the join needs without sending anything: the device (the existing
+    /// user's, or the best one for a new user), the feed names against that device's metro, and
+    /// whether a multicast user already exists or one must be created.
+    #[allow(clippy::too_many_arguments)]
```
nikw9944, line 540: `CreateUser` is sent before anything that decides whether the join can succeed: `resolve_feeds_for_metro` runs at line 550 and the EdgeSeat-pass / feed-seated / metro / `MAX_USER_FEEDS` checks are all inside `SubscribeFeedCommand` at line 554. The top-level `check_accesspass` only proves *a* pass exists (`enforce_epoch = false` here), so a Prepaid-pass operator also reaches this line.

Repro: fresh IP with an EdgeSeat pass, `doublezero connect multicast --subscribe-feed shreds`. `find_or_create_device` selects purely on latency with no feed-metro input, so it picks the nearest device (say `dz-ams`) while `shreds` serves another exchange. The create transaction lands, then the command fails with "feed shreds does not serve the metro of device dz-ams…". Same for a typo'd code, a feed not seated on the pass, or a non-EdgeSeat pass.

The leftover user costs real capacity: `create_core.rs:349-360` calls `accesspass.try_add_user(user_type)` (EdgeSeat per-category cap), bumps `connection_count`, flips the pass to `Connected`, and increments `device.users_count` / `multicast_subscribers_count`. With `max_multicast_users == 1` the operator's only multicast seat now belongs to a user joined to nothing, and no output line mentions `doublezero disconnect multicast`.

Fix: resolve the feed names to pubkeys first, derive the target exchange from them, and `devices.retain(|_, d| d.exchange_pk == feed.exchange)` before `find_or_create_device` — metro should be an input to device selection, not a post-hoc assertion — and hoist the EdgeSeat-pass check ahead of the create. Add a fresh-IP + out-of-metro-feed test with `expect_create_user().times(0)`; `test_connect_command_subscribe_feed_wrong_metro_rejected` pre-creates the user, so it cannot catch this ordering.

nikw9944 (commented):
Two things should change before merge, both in the `--subscribe-feed` create path.

1. `CreateUser` is sent before feed resolution, the EdgeSeat-pass check, and the metro check, so a failed join leaves a bare Multicast user onchain holding an EdgeSeat multicast seat and a device slot — with nothing in the output pointing at `doublezero disconnect multicast` as the cleanup. Device auto-selection is pure latency and takes no feed-metro input, so an out-of-metro feed is a routine way to hit this.
2. Once a Multicast user exists for the IP, `--device` is silently ignored (the device comes from `user.device_pk`), yet the metro error tells the operator to pass `--device`. The advice is a no-op, so the same error repeats on every rerun.

Both have the same root fix: resolve the feeds first, derive the target exchange from them, and filter candidate devices to it before creating anything; then make the metro message reflect which branch chose the device.

Medium items: the AccessPass is resolved from a caller-supplied `client_ip` with no check that the resolved pass actually seats the user's feeds (on unsubscribe that can strand seats and wipe `user.feed_pks`); no account-count preflight, so a legitimate large leave exceeds the legacy transaction size with an opaque RPC error; the (leave failed, join succeeded) path skips tunnel provisioning; and the documented retry-idempotency behavior is untested.

- `crates/doublezero-daemon-cli/src/connect.rs:1327` — low: Feed codes are matched case-insensitively here (and at line 581), while multicast group codes in the same file match exactly (lines 346, 361). Feed PDA seeds are `(code, exchange)` (`pda.rs:109`) and codes are stored case-preserving, so `Shreds` and `shreds` can coexist in one exchange and `find` over a `HashMap` then picks one arbitrarily. Match exactly, consistent with group codes, or bail when more than one code folds to the input.

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -350,6 +406,245 @@ impl Connect {
...
+            (Some(Err(unsub_err)), Some(Ok(()))) => {
+                // The join landed, so still hand the daemon its provisioning work.
+                self.user_activated(daemon, UserType::Multicast, spinner, out)
+                    .await?;
+                writeln!(out, "❌  --unsubscribe-feed failed: {unsub_err:#}")?;
+                writeln!(
+                    out,
+                    "    --subscribe-feed succeeded. Rerun with only --unsubscribe-feed {} to finish.",
+                    unsub_feeds.join(" ")
+                )?;
```
nikw9944, line 509: This branch derives the device solely from `user.device_pk`; `self.device` is only consulted in the `None` branch. But the metro error at line 1335 tells the operator to "pass `--device` to pick a device in the feed's metro", so the printed remedy is a silent no-op whenever a Multicast user already exists for the IP.

Repro: an IP with an existing Multicast user on a device in metro A and a purchased feed in metro B. `doublezero connect multicast --device dz-tyo --subscribe-feed shreds` drops the flag, resolves the device back to `dz-ams`, and prints the same error — on every rerun. The `CreateUser`-before-validation ordering above puts operators into exactly this state on their first failed run.

Fix: if `self.device` is set and resolves to something other than `user.device_pk`, fail explicitly ("existing Multicast user {pk} is on device {code}; run `doublezero disconnect multicast` first, or omit --device"), and make the metro message depend on which branch chose the device — with an existing user the remedy is disconnect-and-reconnect in the feed's metro.

smartcontract/sdk/rs/src/commands/multicastgroup/unsubscribe_feed.rs
```diff
@@ -0,0 +1,517 @@
...
+            eyre::bail!(
+                "user {} is a {} user; only a Multicast user holds feeds",
+                self.user_pk,
+                user.user_type
+            );
+        }
+        for feed_pk in &self.feed_pks {
+            if !user.feed_pks.contains(feed_pk) {
+                eyre::bail!("user does not hold feed {}", feed_pk);
+            }
```
nikw9944, line 59: The pass is resolved from the caller-supplied `self.client_ip` rather than the `user.client_ip` loaded one statement earlier, and nothing checks that the resolved pass is the one seating the user's feeds (same pattern at `subscribe_feed.rs:55`).

`GetAccessPassCommand` prefers the dynamic `0.0.0.0` PDA whenever it exists, and `load_context` (`processors/multicastgroup/feed.rs:111-118`) accepts either PDA. So for a payer holding an exact-IP EdgeSeat pass with the seats plus a second seat-less dynamic EdgeSeat pass, `--unsubscribe-feed f1` resolves the dynamic pass, `retained_pks` comes out empty, and the processor prunes every held feed from `user.feed_pks` (`unsubscribe_feed.rs:109-115`) then skips `remove_feed_user` because the list is already empty — the real pass keeps every seat charged with no user holding them, and the user's feed list is gone.

Fix: derive the lookup from `user.client_ip` (or bail when `self.client_ip != user.client_ip`), and bail when the resolved pass seats none of `user.feed_pks` rather than sending a coincidentally-empty retained list.

smartcontract/sdk/rs/src/commands/multicastgroup/unsubscribe_feed.rs
```diff
@@ -0,0 +1,517 @@
...
+        // Exactly the groups this call drops, mirroring the processor: subscribed, carried by a
+        // departing feed, and not covered by any retained feed.
+        let mut groups: Vec<Pubkey> = Vec::new();
+        for feed in &targets {
+            for group in &feed.groups {
+                let still_covered = retained.iter().any(|r| r.groups.contains(group));
+                if user.subscribers.contains(group) && !still_covered && !groups.contains(group) {
+                    groups.push(*group);
+                }
+            }
```
nikw9944, line 145: No account-count preflight before sending. `send_transaction` builds a legacy `Transaction` with the compute-budget prelude and no address lookup table (`sdk/rs/src/client.rs:188-191`), which caps the variable section at ~25 accounts (33 bytes per meta against the 1232-byte packet limit).

With `MAX_FEED_GROUPS = 20` and `MAX_USER_FEEDS = 6`, leaving one 20-group feed while holding five others needs 1 target + 5 retained + 20 groups = 26 variable accounts and can never be sent; the operator gets an opaque RPC serialization error instead of a diagnosis. Same exposure on the join side (2 feeds × 12 distinct groups already overflows). Note the `MAX_USER_FEEDS` doc comment ("bounded by what a single UnsubscribeFeed transaction can name") counts only feed accounts, not the group fan-out.

Fix: count `4 + targets + retained + groups + 2` and bail with an actionable message (leave/join in smaller batches).

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -350,6 +406,245 @@ impl Connect {
...
+                    client_ip,
+                    spinner,
+                    out,
+                )
+                .await?,
+            )
+        };
+
+        // Catches a code in one flag naming the same feed as a pubkey (or another casing) in the
+        // other; the raw-string check in parse_dz_mode only catches identical spellings.
```
nikw9944, line 466: This early `return Err(unsub_err)` bypasses the `user_activated` call at line 486, so on the (leave failed, join succeeded) path a join that just created the user never enables the reconciler and the tunnel is not provisioned. The accompanying "Rerun with only --unsubscribe-feed {names}" is also unsatisfiable in the common version of this case — the leave failed precisely because the user never held that feed (line 573 or 588), so the suggested rerun can never succeed.

Fix: resolve and validate both flags' names up front, before executing either half, so a bad name fails the whole command before any transaction; and run `user_activated` whenever the join half succeeded, regardless of the leave's outcome.

smartcontract/sdk/rs/src/commands/multicastgroup/subscribe_feed.rs
```diff
@@ -0,0 +1,518 @@
...
+                );
+            }
+            if feed.exchange != device.exchange_pk {
+                eyre::bail!(
+                    "feed {} serves exchange {}, but the user's device {} is in exchange {}",
+                    feed.code,
+                    feed.exchange,
+                    device.code,
+                    device.exchange_pk
+                );
```
nikw9944, line 105: The `!user.feed_pks.contains(pk)` filter is what makes the doc comment's claim true ("a retry against a user that already holds some of the feeds sends the right list and succeeds"), but no test requests an already-held feed: `test_..._expands_to_new_groups` holds none, and `test_..._past_user_cap_rejected` holds six *unrelated* feeds. Deleting this filter leaves every test green.

Fix: add a case with `held = [f1..f6]` re-requesting `f1` — expect a transaction with `feed_pks = [f1]`, no cap error, and only the groups not already in `subscribers`.

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -350,6 +406,245 @@ impl Connect {
...
+                user_pk
+            }
+        };
+
+        self.poll_for_user_activated(ledger, &user_pk, spinner)?;
+        spinner.set_message("Joining feed(s)...");
+        ledger.subscribe_feed(SubscribeFeedCommand {
+            user_pk,
+            client_ip,
+            feed_pks: join.feed_pks,
```
nikw9944, line 750: The overlap check compares raw user-supplied strings, but both resolvers accept pubkeys and match codes with `eq_ignore_ascii_case`. `--subscribe-feed SHREDS --unsubscribe-feed shreds` (or a pubkey on one side) passes this guard, and the command then leaves and immediately rejoins the same feed — dropping its groups in between, and losing the feed entirely if another user takes the freed seat in that window. Compare resolved pubkeys instead.

smartcontract/sdk/rs/src/commands/multicastgroup/unsubscribe_feed.rs
```diff
@@ -0,0 +1,517 @@
...
+        let (_, user) = GetUserCommand {
+            pubkey: self.user_pk,
+        }
+        .execute(client)
+        .map_err(|err| eyre::eyre!("failed to fetch user {}: {err}", self.user_pk))?;
+        if user.user_type != UserType::Multicast {
+            eyre::bail!(
+                "user {} is a {} user; only a Multicast user holds feeds",
+                self.user_pk,
+                user.user_type
```
nikw9944, line 48: Targets are checked for being held but never deduped, unlike `subscribe_feed.rs:76`. An SDK caller passing `[F1, F1]` gets an onchain `InvalidArgument` from `load_feeds` instead of a local "feed given more than once". Mirror the subscribe check.

crates/doublezero-daemon-cli/src/connect.rs
martinsander00: https://github.com/malbeclabs/doublezero/commit/d2aff275d9edc06ac31c78d868b83593e69f7c66

crates/doublezero-daemon-cli/src/connect.rs
martinsander00: https://github.com/malbeclabs/doublezero/commit/d2aff275d9edc06ac31c78d868b83593e69f7c66

smartcontract/sdk/rs/src/commands/multicastgroup/unsubscribe_feed.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4122/changes/14275f11ed72a1ed1e449a5b012ff9b5456e42f5

smartcontract/sdk/rs/src/commands/multicastgroup/subscribe_feed.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4122/changes/14275f11ed72a1ed1e449a5b012ff9b5456e42f5

smartcontract/sdk/rs/src/commands/multicastgroup/unsubscribe_feed.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4122/changes/14275f11ed72a1ed1e449a5b012ff9b5456e42f5

smartcontract/sdk/rs/src/commands/multicastgroup/unsubscribe_feed.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4122/changes/14275f11ed72a1ed1e449a5b012ff9b5456e42f5

crates/doublezero-daemon-cli/src/connect.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4122/changes/14275f11ed72a1ed1e449a5b012ff9b5456e42f5

crates/doublezero-daemon-cli/src/connect.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4122/changes/14275f11ed72a1ed1e449a5b012ff9b5456e42f5

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -350,6 +409,336 @@ impl Connect {
...
+            if !accesspass
+                .feed_seats()
+                .iter()
+                .any(|seat| seat.feed_key == *feed_pk)
+            {
+                eyre::bail!("feed {feed_pk} is not provisioned on the access pass");
+            }
+        }
+        let new_feeds = join.feed_pks.iter().filter(|pk| !held.contains(pk)).count();
+        if held.len() + new_feeds > MAX_USER_FEEDS {
```
nikw9944: This cap check runs against the **pre-leave** feed list, so a swap at the cap is now rejected before the leave frees its slot. `held` comes from `user.feed_pks` at line 576 and the resolved `unsub_pks` (already available at line 436, and executed first) is never subtracted.

Repro: a user holding `MAX_USER_FEEDS` (5) feeds runs `doublezero connect multicast --unsubscribe-feed old --subscribe-feed new`. `held.len() = 5`, `new_feeds = 1`, so this bails "user holds 5 feeds and this join adds 1; a user may hold at most 5" and the `?` propagates out of `execute_multicast_feeds` — neither half runs.

That is the case the doc comment at line 415 and the PR description both describe as working ("the leave runs first, freeing its slot for a swap at the per-user feed cap"). `SubscribeFeedCommand`'s own cap check re-reads the user post-leave and would have passed; the CLI never reaches it.

Fix: pass the resolved `unsub_pks` into `resolve_feed_join` and compute the cap against `user.feed_pks` minus `unsub_pks` — the state the join will actually see. No test covers a swap at the cap (`test_connect_command_feed_swap_reports_failed_half` holds one feed); add one asserting both transactions are sent.

smartcontract/sdk/rs/src/commands/multicastgroup/unsubscribe_feed.rs
```diff
@@ -0,0 +1,586 @@
...
+            eyre::bail!(
+                "the access pass is {}; only an EdgeSeat pass carries feeds",
+                accesspass.accesspass_type
+            );
+        }
+        // EdgeSeat passes carry their feeds on the dynamic (0.0.0.0) pass, but nothing forbids a
+        // second pass for the same payer. If the resolved pass seats none of the held feeds it is
+        // the wrong one, and sending would make the program prune every held feed as stale while
+        // the seat-bearing pass keeps its seats charged.
+        if !user.feed_pks.iter().any(|held| {
```
nikw9944: This guard also fires in a legitimate case: a user whose *only* held feeds have been de-provisioned from the pass. `--unsubscribe-feed <stale feed>` now bails "access pass … seats none of the feeds user … holds" instead of letting the processor prune it, so the operator cannot drop that feed's group subscriptions without `doublezero disconnect multicast`. The wrong-pass case and the all-stale case look identical from the resolved pass alone — discriminate by reading the sibling PDA (the exact-IP one when the dynamic pass was chosen, or vice versa) and bail only when *that* pass seats a held feed.

smartcontract/sdk/rs/src/commands/multicastgroup/subscribe_feed.rs
```diff
@@ -0,0 +1,554 @@
...
+                    std::mem::take(&mut chunk_feeds),
+                    std::mem::take(&mut chunk_groups),
+                ));
+            }
+            chunk_feeds.push(*feed_pk);
+            chunk_groups.extend(new_groups);
+        }
+        chunks.push((chunk_feeds, chunk_groups));
+
+        let mut signature = Signature::default();
```
nikw9944, line 150: An oversized join is now several transactions (same for the per-feed leave loop at `unsubscribe_feed.rs:139`), so a failure mid-sequence leaves the change partly applied while the CLI reports only a blanket "--subscribe-feed failed: {err}" — the operator cannot tell which feeds landed, even though a rerun is idempotent. One concrete way to hit it on the leave side: a still-held but de-provisioned target is pruned by the first transaction, so its own transaction later fails `FeedNotOnAccessPass`. Report per-chunk progress, or name the chunks already applied in the error.

nikw9944 (commented):
All eight findings from the previous round are addressed in code with tests. One new regression came in with the validation rework:

The cap check now runs against the pre-leave feed list, so a swap at `MAX_USER_FEEDS` is rejected before the leave frees its slot — the exact case the code and PR description document as the reason the leave runs first. `resolve_feed_join` has the resolved `unsub_pks` available and should subtract them.

Two Low notes on the new code: the wrong-pass guard also blocks the legitimate all-feeds-de-provisioned cleanup, and an oversized join/leave is now several transactions whose partial application isn't reported.

Confirmed fixed: `CreateUser` now runs only after feed/metro/pass/cap validation with device selection driven by the feeds' metro; `--device` conflicts fail explicitly; both commands resolve the pass from `user.client_ip` and refuse a pass that seats none of the held feeds; `MAX_USER_FEEDS` 6→5 plus splitting at 25 accounts with a worst-case size test; `user_activated` runs on the leave-failed arm; overlap is checked on resolved pubkeys; feed codes match exactly; duplicate targets are rejected locally; and the retry-idempotency path has a test.

smartcontract/sdk/rs/src/commands/multicastgroup/subscribe_feed.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4122/changes/1d6f6dc069970c701132603802ce090df583e790

smartcontract/sdk/rs/src/commands/multicastgroup/unsubscribe_feed.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4122/changes/1d6f6dc069970c701132603802ce090df583e790

crates/doublezero-daemon-cli/src/connect.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4122/changes/1d6f6dc069970c701132603802ce090df583e790

nikw9944 (approved):
lgtm — the swap-aware feed cap, the sibling-pass check, and the split-progress errors each address the previous round's findings, and each has a test pinning it.

PR #4142, opened 2026-08-01
inet-collector: stop recreating the RIPE Atlas fleet on metadata loss

### LINKED ISSUE #4131: internet-latency-collector: missing measurement metadata triggers full RIPE Atlas fleet teardown and sustained recreation churn
## Summary

On 2026-07-31 a deploy restart of `doublezero-internet-latency-collector` on `aws-mn-internet-latency-collector1` caused the collector to tear down and recreate its entire RIPE Atlas measurement fleet (29/29), which then settled into a self-sustaining recreation loop. Mainnet RIPE Atlas metro-pair coverage has been degraded ever since, and does not recover on its own.

WheresItUp, running on the same host through the same code path, held a perfect 435/435 the entire time, which isolates this to the RIPE Atlas measurement lifecycle rather than ingestion, ClickHouse, or the ledger submitter.

## Impact

Mainnet internet-latency coverage, ripeatlas vs wheresitup as control:

| Hour (UTC) | ripeatlas pairs | ripeatlas rows | wheresitup pairs |
| --- | --- | --- | --- |
| 06:00-14:00 | 435 | ~2800 | 435 |
| 14:00 | 435 | 2335 | 435 |
| 15:00 | 412 | 1485 | 435 |
| 16:00 | 415 | 1927 | 435 |
| 17:00 | 287 | 1562 | 435 |
| 18:00 | 380 | 2058 | 435 |
| 19:00 | 380 | 2564 | 435 |

Row volume down 30-45%. Six hours in, coverage was still 380-415 of 435 with no trend toward recovery.

Downstream, this fired 33 alert instances across `Internet Telemetry / No Samples` and `Internet Telemetry: Account Not Found`, currently silenced (see Mitigation).

### 1. Trigger: metadata loss on restart

The collector restarted twice, 11 seconds apart (14:52:08Z and 14:52:19Z), the only restarts in 24h. The state file loaded without error — there is no `"Failed to load measurement state"` warning anywhere in 7 days — but the metadata map contained no entry for any of the 29 live measurements, producing 29 `"No metadata found for measurement"` warnings. That hour is the only hour in 48 with any such warning.

The collector's response to a measurement it cannot identify is to destroy it, not skip it:

```go
// controlplane/internet-latency-collector/internal/ripeatlas/collector.go:972-981
for _, measurement := range doubleZeroMeasurements {
    _, hasMetadata := measurementState.GetMetadata(measurement.ID)
    if !hasMetadata {
        c.log.Info("Marking measurement for removal due to missing metadata", ...)
        toRemove = append(toRemove, measurement)
        continue
    }
```

Result at 14:52:34Z: `wanted:29 existing:29 to_create:29 to_remove:29`. Entire fleet rebuilt in one cycle.

### 2. Why it does not self-recover

Recreation resets a measurement's result stream, which feeds back into probe selection:

1. Fresh measurement has no exports yet
2. `"Marking probe as unresponsive - no exports after 1 hour"` fires (e.g. probe 54140, target `slc`, created 17:53:47Z, `last_export_at` never, still nothing 2h later)
3. Unresponsive probes are excluded from the wanted source set
4. Changed selection fails the `sourcesMatch` check, marking the measurement for recreation
5. Back to step 1

Hourly management cycles show the loop:

| Time (UTC) | wanted | existing | to_create | to_remove |
| --- | --- | --- | --- | --- |
| 14:52:34 (restart) | 29 | 29 | 29 | 29 |
| 15:53:15 | 29 | 29 | 0 | 0 |
| 16:53:15 | 27 | 29 | 27 | 29 |
| 17:53:16 | 29 | 27 | 22 | 20 |
| 18:53:16 | 29 | 29 | 0 | 0 |
| 19:53:17 | 29 | 29 | 22 | 22 |

Unresponsive probes were still accruing hours later (4 new at 16:53, 1 new at 19:53). `slc` and `pit` have exhausted their entire known probe pools, logging `"All known probes unresponsive for location, fetching non-anchor fallback probes"`. `slc` is the endpoint on 8 of the alerting circuits.

### 3. Amplifier: recreation granularity

At 19:53 all 22 recreated measurements logged `existing_sources == wanted_sources` (29/29, 28/28, 27/27, 26/26, ...). The comparison at `collector.go:944-963` is itself correct, so equal counts with a mismatch means exactly one probe was substituted per target. A single probe swap out of 29 sources destroys the whole target's measurement and discards continuity for all of its pairs. This is what turns routine probe flapping into fleet-wide churn.

### 4. Where the data loss surfaces

Destroyed measurements lose incremental timestamp continuity, so `exportSingleMeasurementResults` re-fetches from `start=lastTimestamp`, gets only the boundary result back, and re-exports 1-3 duplicates. This is the `"Exported records but timestamp not updated (old results?)"` warning, which went from **zero occurrences in the preceding 24 hours** to 4, then 10, then 20, then 21 per hour. That log line was added in `00483a99` (2025-12-04), so its appearance is a genuine behavior change and not new instrumentation.

## Unresolved

Why the metadata vanished at 14:52 despite a clean `Load()`. The prior process ran 48+ hours with zero metadata warnings, so the file was healthy beforehand. Two candidates the logs do not separate:

- The **double restart** 11 seconds apart. Both processes run `RunRipeAtlasMeasurementCreation` immediately on startup (`collector.go:1386`) and both call `Save()`, so one can overwrite the other's metadata.
- The prior process's final `Save()` never landed before SIGTERM.

Either way `Save()` is unsafe: `os.Create` truncates in place then encodes (`state.go:107-124`), with no temp-file-and-rename. Any interruption mid-write destroys state. Combined with `Load()` failure being warn-only and non-fatal (`collector.go:1377-1379`), a torn state file silently becomes a full fleet teardown.

## Proposed fixes, highest value first

1. **Rehydrate metadata instead of destroying measurements** (`collector.go:975`). The measurement description already carries everything needed: `"DoubleZero [mainnet-beta] to osl probe 7439"`. Parsing that turns a 29-measurement teardown into a no-op. This alone would have prevented the incident.
2. **Make `Save()` atomic** — write to temp, fsync, rename (`state.go:107`). Separately, stop the deploy from restarting the service twice.
3. **Reduce recreation granularity** so one probe substitution does not discard a 29-source measurement, or add hysteresis before recreating.
4. **Fix the unresponsive-probe race.** Measurements are recreated roughly hourly and the no-export threshold is 1 hour, so a probe can be condemned before it has had a chance to report.

## Related alerting defects (separate follow-up)

- `Internet Telemetry: Account Not Found` is misnamed. The name and runbook point at the `ErrAccountNotFound` path in `exporter/submitter.go:139`, which is a self-healing branch that initializes the account and retries. The rule's actual query is a 2-hour ClickHouse data-freshness check over a full metro mesh, unrelated to onchain accounts.
- That rule cross-joins all 435 metro pairs against both providers and fires per-pair, so one collector fault produces dozens of instances and it scales O(metros^2). Alerting on the fraction of pairs missing per provider would collapse it to one actionable signal.

## Mitigation in place

Grafana silence `af07a542-805a-45e6-b614-335611a32130`, matching `alertname=~"Internet Telemetry.*"`, `env="mainnet-beta"`, `data_provider="ripeatlas"`. Suppresses 33 instances across both the legacy and managed rule copies. **Expires 2026-08-01T21:06:30Z**, after which the alerts return unless the loop is fixed. WheresItUp, devnet, and testnet remain unsilenced.

## Notes

- A service restart is **not** a safe mitigation. If metadata is missing again it re-triggers the same teardown.
- `atlas clear-measurements` is destructive (`ClearAllMeasurements` stops everything on the RIPE side) and would make coverage worse.
- Worth checking the RIPE Atlas account for orphaned `Ongoing` measurements left behind by the churn.

elitegreg:
Fixes #4131.

## Summary

A degraded RIPE Atlas measurement listing could destroy the collector's local state and, through that, the entire measurement fleet. On 2026-07-31 this stopped and recreated all 29 mainnet-beta measurements at once and left the collector in a recreation loop for hours, with metro-pair coverage oscillating between 380 and 427 of 435 while WheresItUp held a steady 435.

The trigger is `configureMeasurements`: when `GetAllMeasurements` failed it logged a warning, substituted an **empty** list, and continued reconciling. With no measurements in view, every metadata entry looks orphaned and gets pruned. The next run then finds 29 measurements it cannot recognize, and unrecognized measurements are stopped and recreated.

Changes, in dependency order:

- **Abort reconciliation on a failed listing.** Reconciling against a view we failed to fetch is never safe. Returns an error and retries next cycle.
- **Rebuild metadata instead of recreating measurements.** Everything needed is still available: the description carries the target location and probe (`DoubleZero [mainnet-beta] to osl probe 7439`), and the new `GetMeasurementProbes` API call plus the current location-to-probe mapping recovers the sources. A measurement is only recreated when it genuinely cannot be identified.
- **Refuse to prune all metadata at once.** If every known entry looks orphaned, a degraded listing is far more likely than a vanished fleet.
- **Write the state file atomically.** `Save` used `os.Create`, truncating in place, so an interrupted write left a torn file. Now writes a temp file and renames.
- **Fix the tracked-measurement gauge.** It was `Set` at startup then `Inc`/`Dec` per operation, so the two drifted; the gauge read `-2` during the incident. Now set from state at the end of each run.

Note the atomic-write change alone would not have prevented this incident. The state file was never torn, it was validly overwritten with a pruned map. The listing guard is the actual fix; atomicity closes a related latent hole.

## Not included

Issue #4131 lists two further items, both changes to probe-selection behavior that sustain the loop once started rather than causing it. They warrant separate review and are left for follow-up:

- recreation granularity, where one changed probe out of 29 discards a whole measurement
- the unresponsive-probe race, where the 1-hour no-export deadline races the ~1-hour recreation interval

## Testing Verification

- `parseMeasurementDescription` covers both description forms, large probe IDs, and rejects legacy source-qualified descriptions (`DoubleZero NYC probe 100 to LON probe 200`) that cannot be rebuilt from text alone
- Metadata rebuilding reconstructs target, probe, and sources from a probe list, and the result survives a reload from disk
- Unrecoverable cases invent nothing: an unparseable description never reaches the probes API, and a probe that cannot be placed in a known metro leaves the measurement unrecognized
- A failed listing returns an error, stops no measurements, and leaves metadata intact both in memory and on disk
- A listing that would orphan every known entry leaves metadata intact
- Atomic save verified under 10 concurrent writers and 25 concurrent readers: every read sees a complete file with all entries, and no temporary files remain. A failed save into a read-only directory leaves no partial file
- Full `internet-latency-collector` suite passes; `golangci-lint` reports 0 issues

One existing assertion was updated: `State_FilePermissionError` matched on the old error string. The behavior it checks is unchanged, and the test now also asserts no temp file is left behind.


nikw9944 (changes requested):
This is not possible to implement because the RIPE Atlas API allows measurement changes via the PATCH verb, but it doesn't allow modification of assigned probes and targets. 

https://atlas.ripe.net/docs/apis/rest-api-reference/measurements/measurements_partial_update

Issue #4130, opened 2026-07-31
[TRACKER] Telemetry outage 2026-07-29 — ledger RPC wedge and agent resilience


elitegreg:
Track the fixes coming out of the device telemetry outage of 2026-07-29 → 2026-07-30 on
mainnet-beta.

**Summary:** a one-hour upstream network outage became a 22-hour telemetry outage, because a shared
Go HTTP transport has HTTP/2 enabled without health checks, and because the telemetry agent's probe
loop has a hard dependency on the RPC endpoint that failed. 23 devices lost between 11 minutes and
19 hours of latency samples. The last three required manually bouncing the agent, roughly 7 hours
after the upstream endpoint had fully recovered.

## Timeline

| Time (UTC) | Event |
| --- | --- |
| Jul 29 ~19:15 | TeraSwitch (TSW) outage begins. Path to DZ Ledger LB `64.130.56.79` (Pittsburgh) blackholes. |
| Jul 29 19:09–19:14 | `nyc002-dz002`, `dz-ny7-sw01`, `dfw001-dz002` write their last epoch-193 samples. |
| Jul 29 19:15:55 | `nyc002-dz002` logs `Submission failed after all retries` (5 attempts, 4 samples, epoch 193). |
| Jul 29 19:20 | DZ Ledger LB begins refusing a subset of connections. Geo-DNS keeps returning it; their health checks kept passing because the node was partially reachable. |
| Jul 29 19:26–19:45 | Lake indexer records 503s from the ledger RPC across four activities. |
| Jul 29 21:03 | `chi001-dz001` stops writing. |
| Jul 30 ~09:00–14:00 | Lake indexer ingestion lag grows to 304 minutes. `Lake Indexer: Errors` alerts fire. |
| Jul 30 11:37 | DZ Ledger LB fully recovers. **Wedged agents do not recover.** |
| Jul 30 14:10:37 | Epoch 194 begins. 20 devices appear to recover simultaneously (artifact, see below). |
| Jul 30 17:29 / 17:42 / 18:52 | `nyc002-dz002`, `dfw001-dz002`, `dz-ny7-sw01` resume after the telemetry agent is bounced. |

## Root cause

**1. Trigger.** TSW blackholes the path to one of five DZ Ledger load balancers. DZ Ledger's geo-DNS keeps
pointing at it because the node stayed partially reachable and their health checks passed. The other
four LBs were healthy throughout and were never used.

[...]

PR #4143, opened 2026-08-03
device/telemetry: keep probing when the epoch fetch fails

### LINKED ISSUE #4125: device/telemetry: pinger aborts all probing when the epoch fetch fails
Tracked by #4130.

## Problem

`Pinger.Tick` fetches the current epoch before doing anything else and returns early if that
fails, so a single unreachable ledger RPC endpoint stops **all** TWAMP probing:

```go
// controlplane/telemetry/internal/telemetry/pinger.go:56
func (p *Pinger) Tick(ctx context.Context) {
	epoch, err := p.getCurrentEpoch(ctx)
	if err != nil {
		p.log.Error("failed to get current epoch", "error", err)
		return          // line 59 — no probes at all this tick
	}
	peers := p.cfg.Peers.GetPeers()
```

Probing is pure UDP and needs no ledger access. The epoch is used only to build the
`PartitionKey` for the sample buffer. Losing it should not cost us the measurement.

Everything downstream then goes quiet with no additional signal: no probes means an empty buffer,
and `Submitter.Tick` returns at `submitter.go:177` without logging when there are no partitions.

Note that peer discovery is *not* affected, because `CachingFetcher` serves stale data on RPC
error (`internal/serviceability/cache.go:83-89`). The epoch is the only call in the probe path
with no cache in front of it.

## Impact

During the 2026-07-29 ledger RPC outage this cost us up to 19 hours of latency samples per device
across 23 devices on mainnet-beta. Measured against the 15,844 samples a full epoch holds:

| Device | Missing samples | Equivalent |
| --- | ---: | ---: |
| nyc002-dz002 | 6,844 | 19.0h |
| dz-ny7-sw01 | 6,825 | 19.0h |
| dfw001-dz002 | 6,820 | 18.9h |
| chi001-dz001 | 6,161 | 17.1h |
| (19 more) | 68–1,803 | 11m–5h |

The probes themselves would have succeeded the entire time.

Timing detail worth knowing: `getCurrentEpoch` (`pinger.go:149-165`) wraps the call in 3 tries with
exponential backoff, and each underlying call can burn ~43s (4 jsonrpc retries at a 10s client
timeout plus backoff). A failing Tick therefore takes ~130s, and the 10s ticker coalesces. So the
error at line 58 appears roughly every two minutes rather than every tick, which is sparse enough
that it read as "no obvious errors" during triage.

## Proposed change

Cache the last known epoch on the `Pinger` and fall back to it. Refuse to probe only when there is
no cached epoch at all (agent just started and has never reached the ledger).

```go
type Pinger struct {
	log *slog.Logger
	cfg *PingerConfig

	mu         sync.Mutex
	lastEpoch  uint64
	haveEpoch  bool
	epochStale time.Time   // when we started serving a cached value
}
```

In `Tick`:

```go
epoch, err := p.getCurrentEpoch(ctx)
if err != nil {
	cached, ok := p.cachedEpoch()
	if !ok {
		p.log.Error("no epoch available and none cached, skipping probe tick", "error", err)
		return
	}
	epoch = cached   // probe anyway; samples buffer and flush when RPC returns
}
```

On success, store the value and clear the stale marker.

Two follow-on considerations for whoever picks this up:

1. **Epoch boundary risk.** Probing with a stale epoch across a rollover writes samples into the
   previous epoch's partition. `Submitter.Tick` already removes partitions whose epoch is behind
   the current one (`submitter.go:205-208`), so those samples would be discarded rather than
   corrupt anything. That is still better than not probing, but it caps how long a stale epoch
   should be trusted. A bound of roughly one epoch seems right; beyond that, stop probing and say
   so.
2. **Alternative worth considering.** The epoch is derivable locally from slot height and epoch
   schedule, both of which change predictably. Deriving it would remove the RPC dependency from
   the probe path entirely rather than making it best-effort. Larger change; noting it as an
   option.

## Acceptance

- Probing continues through a total ledger RPC outage as long as the agent has previously
  fetched an epoch.
- Samples buffer during the outage and flush once RPC recovers.
- A test that fails `GetCurrentEpoch` for N ticks asserts that samples are still added to the
  buffer with the last known epoch.
- Refusal to probe only when no epoch has ever been fetched, and that case logs once, not per tick.

Context: incident analysis in the 2026-07-29 telemetry outage. Related: #4098, #4100.

elitegreg:
## Summary of Changes

* `Pinger.Tick` fetched the current epoch first and returned early on failure, so one unreachable ledger RPC endpoint stopped all TWAMP probing. Probing is pure UDP and needs no ledger access — the epoch only builds the sample buffer's `PartitionKey`. During the 2026-07-29 outage ([#4130](https://github.com/malbeclabs/doublezero/issues/4130)) this cost up to 19 hours of latency samples per device across 23 mainnet-beta devices, for probes that would have succeeded the whole time.
* The pinger now caches the last known epoch and falls back to it, so an RPC outage costs epoch precision rather than measurements. Samples buffer through the outage and flush once submission recovers.
* The epoch is refreshed on its own loop instead of inline per tick, and that loop is its sole writer — the probe path only reads the cache. This matters as much as the fallback: a failing fetch burns ~130s across its retries and `time.Ticker` buffers only one tick, so an inline fetch swallowed roughly a dozen probe opportunities per failure. The refresh cadence follows `-probe-interval` and can be set independently with `-epoch-refresh-interval`.
* Probing is refused when no epoch has ever been fetched, when the cached one exceeds `-max-epoch-staleness`, or when the cached epoch's projected end has passed. Both bounds are load-bearing and neither is arbitrary:
  * **The staleness bound is the sample buffer's.** A partition holds 4096 samples, the submitter cannot drain during the same outage, and once it finds the partition over capacity it discards the entire backlog in one tick. Since the onchain account carries one start timestamp and a fixed sampling interval rather than a timestamp per sample, every sample written after that discard is backdated by the whole gap. The default is 10h and the collector clamps whatever is configured to what the buffer holds at the probe interval, so a faster `-probe-interval` narrows it automatically.
  * **The rollover bound is the epoch's own.** `GetEpochInfo` already returns `SlotIndex`/`SlotsInEpoch`/`AbsoluteSlot`, so the epoch's remaining time is free. Probing past it would file samples under the previous epoch with timestamps in the next epoch's window, where `getCircuitLatenciesForTimeRange` never looks — it resolves the range to epoch numbers and iterates only those accounts. The slot rate is measured from `AbsoluteSlot` deltas over a long baseline rather than assumed: the 400ms target overshoots a ~44h ledger epoch by ~9%, up to ~4h right after a boundary.
* Each refusal logs once rather than per tick and is counted separately (`pinger_epoch_never_fetched`, `pinger_epoch_too_stale`, `pinger_epoch_ended`) — a bad ledger URL and a multi-hour outage want different alerts. Repeated fetch failures collapse into fresh→stale and stale→fresh transitions, and the transition itself waits for three consecutive failures so a flapping endpoint stays quiet. `pinger_epoch_fetch_failed` counts every failed attempt, so a partial outage whose retries succeed still leaves a signal. The per-attempt retry warning in the fetch drops to Debug — at a 10s cadence it emitted thousands of lines across the outage, which is what #4130's "single-digit log lines per component" exit criterion is about.
* `doublezero_device_telemetry_agent_epoch_cache_stale_age_seconds` reports the cached epoch's age while the fetch is failing, 0 when fresh, and `+Inf` when no epoch has ever been fetched — the state where nothing is probed at all, which would otherwise read as healthy.
* Epoch age is measured from a monotonic reading, so a device booting with a bad RTC and stepping forward over NTP does not trip the staleness bound instantly.
* Fixes #4125

Follow-ups filed rather than folded in, both pre-existing submitter bugs this change makes reachable at scale: #4148 (retries restart at batch 0, duplicating committed samples) and #4149 (the over-capacity path discards the whole backlog instead of the oldest slice).

## Testing Verification

* Probing continues under the cached epoch while the fetch fails, and the fallback and recovery each log exactly once.
* Probing stops past `MaxEpochStaleness` with an injected clock, and stops at the projected rollover with `MaxEpochStaleness` set far larger — so only the rollover bound can be what stopped it.
* The measured slot rate is actually used: 1200 slots over 600s puts the epoch end past where the 340ms fallback would have called it over, and probing continues to the measured end and no further.
* A `-race` test drives `Tick` concurrently with `Run` while the fetch flaps and asserts no ERROR is emitted and no refusal is counted while an epoch is cached — the regression for the inline fetch making the probe path a second writer of the cache.
* Two flaps that the retries ride out produce nothing above Debug; a sustained outage produces exactly one Warn on the third consecutive failure and one Info on recovery.
* The staleness gauge tracks the cached epoch's age, clears on recovery, and reads `+Inf` when nothing was ever fetched. Each refusal type and each failed attempt lands on its own counter.
* A submitter-level test probes for the whole window the staleness bound allows against a `WriteDeviceLatencySamples` that refuses every write, then asserts every sample survives the flush once writes recover — the regression for the bound outliving the buffer. It fails on its own premise if the bound ever drifts past what the buffer holds.
* Verified the new tests discriminate by mutating each fix back: disabling the epoch-end projection, ignoring the measured slot rate, removing the Warn debounce, removing the gauge sentinel, and no-oping the retention clamp with the default back at 12h. Each mutation failed the tests meant to catch it.
* Ran the pinger suite repeatedly to check for timing flakiness in the `Run`-driven tests.
* `internal/netns/TestRunInNamespace_EmptyNameErrors` fails identically on a clean `main` (needs privileges for `setns`); unrelated to this change.


controlplane/telemetry/internal/telemetry/pinger.go
```diff
@@ -9,10 +9,22 @@ import (
...
 	"github.com/malbeclabs/doublezero/controlplane/telemetry/pkg/buffer"
 	twamplight "github.com/malbeclabs/doublezero/tools/twamp/pkg/light"
 )
 
+const (
+	// DefaultMaxEpochStaleness bounds how long the pinger keeps probing with a cached epoch after
+	// the ledger RPC stops answering. Samples recorded with a stale epoch are written to that
+	// epoch's account, so once we are far enough behind that a rollover is likely they would be
+	// misattributed to the previous epoch.
+	DefaultMaxEpochStaleness = 12 * time.Hour
```
ben-dz: **The 12h bound outlives the buffer, and the overflow silently backdates every later sample in the account.**

`partitionBufferCapacity = 4096` per partition (`collector.go:17`); at the default `-probe-interval=10s` that is **11.4h** of samples for one link — inside this 12h window. During a ledger outage the submitter cannot drain either, so each 60s tick does `CopyAndReset`, fails all 5 attempts, and prepends the batch back only while `Len + len(tmp) < Capacity`. The first time that check trips at ~11.4h (`submitter.go:248`), the **entire** accumulated backlog is discarded in one tick rather than the oldest slice, and accumulation restarts from zero. A 12h outage therefore yields roughly 36 minutes of samples on recovery, not 12h.

What makes this High rather than a tuning nit is what the discard does to the data that *does* survive. The onchain account stores one `start_timestamp_microseconds`, one `sampling_interval_microseconds`, and a flat `samples` array with no per-sample timestamps (`smartcontract/programs/doublezero-telemetry/src/state/device_latency_samples.rs:117-123`), and the reader synthesizes each timestamp as `start + i*interval` (`controlplane/telemetry/internal/data/device/latencies.go:205`). A dropped run leaves no marker, so every sample appended after the discard is backdated by the full 11.4h gap. Wrong data presented as good data is worse than the probe gap this PR removes.

This state was not reachable before the change: the pre-fix pinger stopped adding during an outage, so the partition never grew and the over-capacity path never tripped.

Fix (one line for the immediate hazard): set `DefaultMaxEpochStaleness` below the buffer's real retention — derive it as `partitionBufferCapacity × ProbeInterval`, or hardcode a value with margin (8h) rather than one past the cliff. The comment above should also name the real binding constraint; rollover likelihood is not it, given a ~44h ledger epoch and no onchain epoch check. Worth a follow-up: have the over-capacity path keep the newest `Capacity - Len` samples instead of dropping all of `tmp`.

Test: a submitter-level test with a small `partitionBufferCapacity` and a failing `WriteDeviceLatencySamples`, asserting the observed sample rate stays within tolerance of `ProbeInterval` across the whole allowed staleness window.

controlplane/telemetry/internal/telemetry/pinger.go
```diff
@@ -143,14 +195,155 @@ func (p *Pinger) Tick(ctx context.Context) {
...
+// than MaxEpochStaleness. Both cases log once rather than per tick.
+func (p *Pinger) epochForTick(ctx context.Context) (uint64, bool) {
+	p.mu.Lock()
+	epoch, have, at := p.epoch, p.haveEpoch, p.epochAt
+	p.mu.Unlock()
+
+	if !have {
+		// Nothing cached: either the agent just started and the refresh loop has not produced a
+		// value yet, or Tick is being driven directly. Fetch inline so the tick is not wasted.
+		epoch, err := p.getCurrentEpoch(ctx)
```
ben-dz: **The inline fetch makes `epochForTick` a second concurrent writer to the epoch cache, and each act-on-stale-read is observable.**

`haveEpoch` is read under the lock at lines 205-207, then this fetch runs for tens of seconds when failing. Three consequences follow from acting on that stale read, and from emitting side effects outside the critical section that flipped the corresponding flag:

1. If the concurrent `refreshEpochLoop` lands an epoch during the retry window, the failure branch below still fires — it increments `errors_total{pinger_epoch_unavailable}` and logs `Error("No epoch available and none cached, skipping probes until the ledger answers")` while a fresh epoch sits in the cache. An operator-facing ERROR and an error-counter increment on a healthy agent is a false page source, in exactly the flapping/partially-degraded-RPC scenario this PR targets.
2. `markEpochStale` sets `servingStale` under the lock at 316-317 but logs at 330. A `storeEpoch` from this inline path can run in the gap and log `"Epoch fetch recovered"` (302) *before* the failure it recovered from, with `staleFor ≈ 0`.
3. `markEpochStale` reads `at` under the lock at 315 but calls `EpochCacheStaleAge.Set(age)` outside it at 327, racing `storeEpoch`'s `Set(0)` at 298. The gauge can be left large right after a successful fetch. It self-heals in one refresh interval, but this gauge is the operator's only staleness signal.

Fix: make `refreshEpochLoop` the sole writer and have `epochForTick` read the cache only. The cost is one skipped tick at startup, and `Tick`-without-`Run` no longer self-seeds, which affects the subtest at `pinger_test.go:440`. If the inline fetch stays, re-read the cache under `mu` after it fails and fall through to the staleness check at line 232 when an epoch landed meanwhile, and move each log and `Set` inside the critical section that flips its flag.

Test: a `-race` test driving `Tick` concurrently with `Run` while the fetch flaps, asserting no ERROR is emitted while an epoch is cached, and that the gauge is 0 whenever the last fetch succeeded.

controlplane/telemetry/internal/telemetry/pinger.go
```diff
@@ -143,14 +195,155 @@ func (p *Pinger) Tick(ctx context.Context) {
...
+func (p *Pinger) markEpochStale(err error) {
+	p.mu.Lock()
+	epoch, have, at := p.epoch, p.haveEpoch, p.epochAt
+	first := !p.servingStale
+	p.servingStale = true
+	p.mu.Unlock()
+
+	if !have {
+		// epochForTick reports this case; there is no cached epoch to fall back to.
+		p.log.Debug("Failed to get current epoch, none cached", "error", err)
```
ben-dz, line 361: `markEpochStale` returns here before setting `EpochCacheStaleAge`, so the gauge reads 0 in the worst state — restarted mid-outage, no epoch ever fetched, zero probing — exactly as it does when healthy. Emit a sentinel on this branch, and add the one subtest that would have caught it (`testutil.ToFloat64` asserting the gauge tracks age while stale and returns to 0 on recovery). Neither new metric has any coverage today.

controlplane/telemetry/internal/telemetry/pinger.go
```diff
@@ -143,14 +195,155 @@ func (p *Pinger) Tick(ctx context.Context) {
...
+			}
+			return 0, false
+		}
+		p.storeEpoch(epoch)
+		return epoch, true
+	}
+
+	age := p.cfg.NowFunc().Sub(at)
+	if age > p.cfg.MaxEpochStaleness {
+		metrics.Errors.WithLabelValues(metrics.ErrorTypePingerEpochUnavailable).Inc()
```
ben-dz: "No epoch ever fetched" (boot ordering, bad ledger URL, line 214) and "gave up after `MaxEpochStaleness`" (outage, here) share one `errors_total` value, which is what the fleet alert fires on. Split into two error types while the type is being introduced.

controlplane/telemetry/internal/telemetry/pinger.go
```diff
@@ -143,14 +195,155 @@ func (p *Pinger) Tick(ctx context.Context) {
...
+// and probing continues against the cached epoch in the meantime.
 func (p *Pinger) getCurrentEpoch(ctx context.Context) (uint64, error) {
 	attempt := 0
 	epoch, err := backoff.Retry(ctx, func() (uint64, error) {
 		if attempt > 1 {
-			p.log.Warn("Failed to get current epoch, retrying", "attempt", attempt)
+			// Debug, not Warn: markEpochStale carries the operator-facing signal, collapsed into
+			// the fresh->stale transition. A per-attempt warning here means thousands of lines
+			// across a multi-hour outage.
+			p.log.Debug("Failed to get current epoch, retrying", "attempt", attempt)
```
ben-dz: The demotion itself is right, but the fetch-failure signal now has no coverage in one direction and no debounce in the other. When retries eventually succeed (the partial-outage mode) `markEpochStale` never runs, so there is no signal above Debug and no metric at all. When the RPC flaps at the 10s refresh, each flap emits a Warn (330) plus an Info (302) — against this PR's own single-digit-log-lines goal. Add a per-failed-attempt counter here, and require N consecutive failures before the Warn.

controlplane/telemetry/internal/telemetry/pinger.go
```diff
@@ -30,15 +54,44 @@ type PingerConfig struct {
...
+		cfg.EpochRefreshInterval = cfg.Interval
+	}
+	if cfg.EpochRefreshInterval <= 0 {
+		cfg.EpochRefreshInterval = defaultEpochRefreshInterval
+	}
+	if cfg.MaxEpochStaleness <= 0 {
+		cfg.MaxEpochStaleness = DefaultMaxEpochStaleness
+	}
+	if cfg.NowFunc == nil {
+		cfg.NowFunc = func() time.Time { return time.Now().UTC() }
```
ben-dz: `time.Now().UTC()` strips the monotonic reading, so `NowFunc().Sub(at)` is pure wall-clock arithmetic. A device booting with a bad RTC and then NTP-stepping forward trips the staleness bound instantly and stops probing until the next successful fetch. Keep a monotonic timestamp alongside `epochAt` and measure age from it.

controlplane/telemetry/internal/telemetry/pinger.go
```diff
@@ -22,6 +34,18 @@ type PingerConfig struct {
 	GetSender         func(ctx context.Context, peer *Peer) twamplight.Sender
 	GetCurrentEpoch   func(ctx context.Context) (uint64, error)
 	RecordProbeResult func(peer *Peer, success bool)
+
+	// EpochRefreshInterval is how often the cached epoch is refreshed in the background.
+	// Defaults to Interval, which keeps the epoch RPC rate the same as when the fetch was inline.
+	EpochRefreshInterval time.Duration
```
ben-dz, line 52: `EpochRefreshInterval` is a new config field with no CLI flag and no non-test caller, and its `defaultEpochRefreshInterval` fallback (lines 24-25, 75-77) is unreachable: `Config.Validate` rejects `ProbeInterval <= 0` (`config.go:96`) and `Run` would panic in `time.NewTicker(0)` first. Either wire a flag next to `-max-epoch-staleness` — the epoch resolves once per ~44h, so refreshing every 10s is ~8.6k RPC calls/day/device that an operator now cannot turn down — or drop the field and the constant and read `Interval` directly.

controlplane/telemetry/internal/telemetry/config.go
```diff
@@ -115,6 +119,9 @@ func (c *Config) Validate() error {
 	if c.MaxConsecutiveSenderLosses <= 0 {
 		c.MaxConsecutiveSenderLosses = 30
 	}
+	if c.MaxEpochStaleness <= 0 {
```
ben-dz, line 128: `-max-epoch-staleness=0` silently becomes 12h, the opposite of the natural reading of the flag's "before giving up". `main.go:117` already fails fast on `probe-interval >= submission-interval`; reject `<= 0` there too rather than defaulting silently past an operator's intent.

ben-dz (changes requested):
The core move is right and the central claim holds — `write_device_latency_samples.rs` never reads `Clock` and takes no epoch argument, and `initialize_device_latency_samples.rs:193` uses `args.epoch` only as PDA seed material, so past-epoch writes are accepted.

Two major things to fix:
- The 12h default sits past the buffer underneath it. `partitionBufferCapacity = 4096` at the default `-probe-interval=10s` is 11.4h of samples for one link, inside the 12h window. Once `submitter.go:248`'s over-capacity check trips, the entire accumulated backlog is discarded in one tick rather than the oldest slice, so a 12h outage yields roughly 36 minutes of samples on recovery. One-line fix: set the default below the buffer's real retention (say, 11h)
- Use the 11h as a ceiling for longest it will store, but also store how long the Epoch actually lasts and don't store samples past it. The data is already on the wire and thrown away — main.go:339-345 keeps only epochInfo.Epoch and discards SlotIndex/SlotsInEpoch, so no extra RPC. "Round down a little" needs to be ~15%, not minutes: the 400ms constant overshoots by 9%, i.e. up to ~4h right after a boundary — or derive the slot rate from consecutive absoluteSlot deltas, free at the 10s refresh cadence. CalculateEpochTimes already exists at epoch.go:14 but sits under monitor/internal/, so telemetry can't import it.

- **Stale-epoch samples are unreachable through the time-range read path** — `controlplane/telemetry/internal/data/device/latencies.go:156-193`. `getCircuitLatenciesForTimeRange` resolves `from`/`to` to epoch numbers via `EpochFinder.ApproximateAtTime` and iterates `startEpoch..endEpoch` only (`:168`), then filters by synthesized timestamp (`:193`). When an outage crosses an epoch boundary, post-boundary samples land in epoch N's account with timestamps inside N+1's window; any query scoped entirely to N+1 — a "last 6h" dashboard — resolves `startEpoch == endEpoch == N+1`, never opens N's account, and returns a silent gap. Epoch-scoped queries do return them.

This narrows the tradeoff the description states: "epoch precision, not measurements" holds for the raw account and for epoch-scoped reads, but for the consumer path dashboards actually use, cross-boundary samples are missing rather than mislabeled. Worth stating plainly in the changelog even if the behavior stands. Fix: widen the reader's epoch scan by one on each side, or bound staleness by the time remaining in the last known epoch rather than a flat duration.
- **Submitter retries restart at batch 0, duplicating committed samples once a backlog exists** — `controlplane/telemetry/internal/telemetry/submitter.go:95-170`. `SubmitSamples` batches with `for i := 0; i < len(samples); i += MaxDeviceLatencySamplesPerBatch` and returns on the first failing batch; `Tick` then re-invokes it from scratch, so batches already committed are written again. Pre-existing code, but this PR makes it reachable at scale: `tmp` is normally 6 samples (one batch, no partial-failure window), whereas an hours-long backlog is ~17 sequential transactions per attempt, retried up to 5 times, precisely while the RPC is still flaky. Duplicates inflate `next_sample_index`, which shifts every synthesized timestamp for that account — the same corruption mechanism as the staleness-bound finding — and can push the account to `SamplesAccountFull`, where the submitter drops the whole partition.

Fix: track the batch index across attempts, or retry per batch rather than per call. Reasonable as a follow-up PR given it is pre-existing, but it should not stay unowned once this change makes multi-batch flushes routine.

controlplane/telemetry/internal/telemetry/pinger.go
elitegreg: Fixed, in both halves.

The default is now 10h rather than 11h, and the collector clamps whatever is configured to `partitionBufferCapacity × ProbeInterval × 0.95` (`collector.go`, `maxEpochStaleness`). 10h so the default does not itself trip the clamp warning at the default probe interval — 11h is inside the 11.4h cliff but past the 10.8h the headroom leaves, and a warning on every healthy startup is noise. The clamp is what keeps a non-default `-probe-interval` safe: at 5s the buffer only holds 5.7h, and a flat 10h would be past the cliff again.

The comment above the constant now names the buffer as the binding constraint, and says what the overflow does to the data rather than talking about rollover likelihood — the rollover is bounded separately now, see below.

Tests: `TestMaxEpochStaleness` covers the clamp arithmetic and asserts the default does not need clamping. `TestSubmitter_RetainsEverySampleAcrossTheStalenessBound` is the submitter-level one you asked for — small `capacity`, `WriteDeviceLatencySamples` refusing every write, probing for the full window the bound allows, then asserting every sample survives the flush once writes recover. It also `require.Less(probes, capacity)` up front, so if the bound ever drifts back past the buffer the test fails on the premise rather than the assertion. Verified both fail with the clamp no-oped and the default back at 12h.

The over-capacity path keeping the newest `Capacity - Len` instead of dropping all of `tmp` is filed as #4149.

controlplane/telemetry/internal/telemetry/pinger.go
elitegreg: Fixed the way you suggested: `RefreshEpoch` is the sole writer and `epochForTick` only reads the cache and decides. The inline fetch is gone, so all three consequences go with it — no ERROR or error-counter increment while a fresh epoch is cached, no recovery logged before the failure it recovered from, no gauge left large after a success. Every log line and every `Set` now happens inside the critical section that flips its flag.

On the cost you flagged: `refreshEpochLoop` refreshes immediately on start and the first probe tick is `Interval` later, so a healthy agent skips nothing. `Tick`-without-`Run` no longer self-seeds, so the three affected subtests now call `RefreshEpoch` explicitly — which reads better anyway, since the seeding is now visible instead of a side effect of the first tick. `refreshEpoch` is exported as `RefreshEpoch` for that, mirroring how `Tick` exposes one step of the probe loop.

Test: `TestAgentTelemetry_PingerEpochMetrics/a_flapping_fetch_never_reports_an_error_while_an_epoch_is_cached` drives `Tick` in a loop against a running `Run` with the fetch flapping, under `-race`, and asserts no ERROR is emitted and `pinger_epoch_never_fetched` does not move. It then settles the flapping on a success and asserts the gauge is 0.

controlplane/telemetry/internal/telemetry/pinger.go
elitegreg: Fixed. That branch now sets the gauge to `+Inf`, so an alert on `stale_age > threshold` fires in the state where nothing is being probed at all, rather than reading identically to healthy. `+Inf` rather than `-1` for exactly that reason — a negative sentinel would pass the same alert silently. Noted in the metric's Help string.

Both new metrics now have coverage: `TestAgentTelemetry_PingerEpochMetrics` has one subtest walking the gauge through fresh → stale → staler → recovered with `testutil.ToFloat64`, one asserting the `+Inf` sentinel, and one asserting each refusal and each failed attempt lands on its own counter. Verified the sentinel subtest fails with the `Set` removed.

controlplane/telemetry/internal/telemetry/pinger.go
elitegreg: Split, into three rather than two. `pinger_epoch_never_fetched` for boot ordering or a bad ledger URL, `pinger_epoch_too_stale` for giving up after the bound, and `pinger_epoch_ended` for the new projected-rollover refusal. Covered by `each_refusal_and_each_failed_attempt_is_counted_separately`, which asserts each one moves on its own trigger and not on the others.

controlplane/telemetry/internal/telemetry/pinger.go
elitegreg: Both fixed.

Coverage in the eventually-succeeds direction: `pinger_epoch_fetch_failed` now increments per failed attempt inside `getEpochInfo`, so a partially degraded endpoint whose retries ride it out still leaves a signal even though `markEpochStale` never runs. Kept at Debug for the log line itself, for the reason in the original comment.

Debounce in the other: the fresh→stale Warn now waits for `epochStaleWarnAfter` (3) consecutive failed fetches, and the recovery Info is gated on whether that Warn actually fired — so a flap produces neither, rather than one of each. `storeEpoch` resets the counter.

Test: `does_not_report_a_fallback_the_retries_recovered_from` runs two flaps and asserts nothing above Debug, then a sustained outage and asserts exactly one Warn on the third consecutive failure and one Info on recovery. Verified it fails with the debounce removed.

controlplane/telemetry/internal/telemetry/pinger.go
elitegreg: Fixed. `NowFunc` defaults to `time.Now` rather than `time.Now().UTC()`, so the reading carries monotonic and `NowFunc().Sub(at)` is monotonic arithmetic; the `.UTC()` moved to the log call sites where it belongs. A bad RTC stepping forward over NTP no longer looks like hours of staleness.

The collector deliberately no longer passes `Config.NowFunc` into `PingerConfig` — that one is the UTC wall clock the sender cache uses for its bookkeeping, and it strips exactly the reading this needs. Comment at the call site says so, so it does not get helpfully wired back.

Injected test clocks are wall-only, which `Sub` handles by falling back to wall-clock arithmetic; that is what the staleness and rollover subtests rely on.

controlplane/telemetry/internal/telemetry/pinger.go
elitegreg: Wired a flag: `-epoch-refresh-interval`, next to `-max-epoch-staleness`. It defaults to 0, which follows `-probe-interval`, so the epoch RPC rate is unchanged from before this PR and stays coupled if an operator changes the probe interval — and the ~8.6k calls/day/device can now be turned down without touching the probe cadence.

The unreachable `defaultEpochRefreshInterval` fallback stays, but it is no longer dead: `PingerConfig` is constructed directly by tests with no `Interval` set, and `NewPinger` would otherwise hand `time.NewTicker` a zero. `Config.Validate` still rejects `ProbeInterval <= 0` for the collector path.

controlplane/telemetry/internal/telemetry/config.go
elitegreg: Fixed in `main.go`, next to the `probe-interval >= submission-interval` check:

```go
if *maxEpochStaleness <= 0 {
    fmt.Println("max-epoch-staleness must be greater than 0")
    os.Exit(1)
}
```

Plus the same for a negative `-epoch-refresh-interval`, since 0 there means "follow `-probe-interval`" and is deliberate.

The zero-value default in `Config.Validate` stays. It is a library default for embedded callers who never see a flag, and the flag path can no longer reach it.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
+- Telemetry
+  - A ledger RPC outage no longer stops TWAMP probing on the device telemetry agent: the pinger caches the last known epoch and refreshes it off the probe path, instead of fetching it inline and skipping the tick on failure. Probing stops only when no epoch has ever been fetched or the cached one exceeds the new `-max-epoch-staleness` (default 12h). ([#4125](https://github.com/malbeclabs/doublezero/issues/4125))
```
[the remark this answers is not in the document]
elitegreg: Fixed in 0518c7dd — the entry now references `(#4143)`. The issue link stays in the PR description.

elitegreg:
Thanks — this was a good review. Every inline point is answered in its thread; the three from the review body that had no inline anchor are below.

**Store how long the epoch actually lasts, and don't store samples past it.** Done, and it turned out to be the more important of the two bounds. `Config.GetCurrentEpochFunc` becomes `GetEpochInfoFunc` returning a `telemetry.EpochInfo{Epoch, SlotIndex, SlotsInEpoch, AbsoluteSlot}`, so `main.go` keeps what `GetEpochInfo` was already putting on the wire instead of throwing it away — no extra RPC. On each successful refresh the pinger projects `epochEndsAt` and `epochForTick` refuses to probe past it.

I went with the measured slot rate rather than the ~15% round-down, since you offered both and the measurement is free at the refresh cadence. The baseline is the first observation rather than the previous one — at a 10s cadence a single delta is a few dozen slots, where RPC latency and commitment jitter are a large fraction of the reading, so consecutive deltas are useless. Once the baseline spans 5 minutes the measured duration is used, trimmed 5% so the projection lands before the real boundary if the ledger speeds up during an outage (while the fetch is failing we can't notice). Until then it falls back to 340ms — the 400ms target less 15%, per your number. Clamped to [200ms, 1s], and the reading rebases if the slot counter or the clock moves backwards. `SlotsInEpoch == 0` yields a zero projection, so a caller that doesn't supply a slot position is bounded by `MaxEpochStaleness` alone rather than refusing to probe.

Test: `projects_the_epoch_end_from_the_measured_slot_rate` feeds 1200 slots over 600s (500ms/slot), then advances past where the 340ms fallback would have called the epoch over and asserts probing continues, then past the measured end and asserts it stops. Fails if the measured value is ignored. `stops_probing_once_the_cached_epoch_has_likely_rolled_over` covers the fallback path with `MaxEpochStaleness` set far larger, so only the rollover bound can be what stops it.

Didn't reach for `CalculateEpochTimes` — as you say it's under `monitor/internal/`, and it hardcodes the 400ms constant that's the problem here.

**Stale-epoch samples are unreachable through the time-range read path.** Confirmed, and stated plainly in the changelog: samples taken against a cached epoch land in that epoch's account, so a query scoped to a later epoch won't return them.

No reader change, though. Of your two suggested fixes I took the second — bounding by the epoch's remaining time — and that removes the cause rather than compensating for it downstream: with the rollover bound in place the probe loop stops at the projected boundary, so cross-boundary writes need the projection to be wrong, not merely an outage to be long. Widening the reader's scan by one on each side would still be defensible as belt-and-braces (it would also pick up the ~10s of samples per boundary that normal operation produces, pre-existing), but it costs two extra account fetches on every time-range query and belongs in a PR about the read path rather than this one. Happy to file it if you'd rather have it.

**Submitter retries restart at batch 0.** Agreed on all of it, including that this PR is what makes it routine. Filed as #4148 rather than folded in — it's a pre-existing submitter bug and the fix (batch index across attempts, or per-batch retry) wants its own tests. Not unowned.

While there: your inline note about the over-capacity path discarding all of `tmp` rather than the oldest slice is #4149. The staleness clamp means the probe loop no longer walks into it, but the path itself is still sharp.

---

Two things worth flagging that weren't asked for:

- `DefaultMaxEpochStaleness` is 10h, not the 11h you suggested. 11h is inside the 11.4h cliff but past the 10.8h the clamp's headroom leaves, so every agent would log a clamp warning on a healthy startup. `TestMaxEpochStaleness` asserts the default doesn't need clamping, so the two can't drift apart quietly.
- `refreshEpoch` is now exported as `RefreshEpoch`, because dropping the inline fetch means `Tick` no longer self-seeds and the tests have to drive the refresh explicitly. It mirrors how `Tick` exposes one step of the probe loop.

`internal/netns/TestRunInNamespace_EmptyNameErrors` still fails identically on a clean `main` (needs privileges for `setns`); everything else in `./controlplane/telemetry/...` passes under `-race`.

ben-dz (approved):
All round-1 findings re-verified against `281639a8` in the checkout. Every one is addressed in code, or dispositioned to a filed issue that exists. No new findings.

**Staleness bound past the buffer cliff.** `DefaultMaxEpochStaleness` is 10h (`pinger.go:29`) and `collector.go:34-42` clamps whatever is configured to `capacity × probeInterval × 0.95`, wired at `:116`. The arithmetic checks out — 4096 × 10s × 0.95 = 10.81h, so the default passes unclamped while a 5s probe interval clamps to 5.4h with a Warn. You were right that 11h would have tripped the clamp on every healthy startup. `retention_test.go:57-107` is the submitter-level test asked for, and `require.Less(probes, capacity)` at `:71` fails on the premise if the bound ever drifts back past the buffer.

**Epoch-duration bound.** New `epoch.go`, refused at `pinger.go:256-265`, with `main.go:354-366` reusing the single `GetEpochInfo` call rather than adding an RPC. The measured-slot-rate approach holds up under the cases I pushed on: a lagging or rewound `AbsoluteSlot` rebases or trips the `[200ms, 1s]` bounds and falls back to 340ms, which projects early — the safe direction. Over-estimating is the only way to project late, caps at 950ms, and merely makes the bound no-op back to `MaxEpochStaleness`, which is still enforced independently at `:241`. `slotRateEstimator` is touched only at `:321`, inside `storeEpoch`'s critical section.

**Inline fetch as a second cache writer.** `epochForTick()` no longer takes a context, holds `p.mu` across its whole body, and only reads; `storeEpoch` and `markEpochStale` hold the lock across their log line and gauge write. All three consequences are gone.

**All six Lows fixed** — `+Inf` gauge sentinel (`pinger.go:360`), four error types (`metrics.go:34-37`), per-attempt counter plus a three-failure debounce with the recovery Info gated on it (`:392`, `:369-373`, `:338`), monotonic `NowFunc` with the UTC clock deliberately unwired (`:97-102`, `collector.go:117-118`), flag validation (`main.go:125,130`), and a real `-epoch-refresh-interval` (`main.go:101`).

**Dispositioned, not fixed here.** The time-range read path is unchanged, but this was addressed via the second fix offered — bounding by the epoch's remaining time — which removes the cause rather than compensating downstream, and the tradeoff is stated in `CHANGELOG.md:12`. The submitter batch-0 retry is #4148 and the over-capacity discard is #4149, both open with accurate titles; the only round-1 condition was that the former not stay unowned, which is met.

**Test isolation** is handled correctly: the metrics test is not `t.Parallel()` and asserts on deltas rather than absolute values. Verified by running it plus both retention tests under `-race -count=2`.

One benign behavior change worth naming: on a healthy ledger the projected-end bound can skip a single probe tick in the last ~11s before an epoch boundary. Those samples would otherwise land in the old epoch's account with post-boundary timestamps, so it removes a small pre-existing artifact rather than creating one.

Issue #4127, opened 2026-07-31
device/telemetry: account-full path reports success and logs the wrong count


elitegreg:
Tracked by #4130.

## Problem

Two defects in the account-full handling in `Submitter.SubmitSamples`.

**1. Data loss is reported as success.**

```go
// controlplane/telemetry/internal/telemetry/submitter.go:159
} else if errors.Is(err, telemetry.ErrSamplesAccountFull) {
	log.Warn("Partition account is full, dropping samples from buffer and moving on", "droppedSamples", len(samples))
	s.cfg.Buffer.Remove(partitionKey)
	return nil          // caller counts this as a successful submission
}
```

`return nil` means `Tick` records `success = true`, no error metric fires, and the whole partition
is removed from the buffer. The only trace is a Warn. There is no counter to alert on and no way
to see this in Grafana. The same block is duplicated at lines 151-155 for the post-initialize path.

**2. The logged count is wrong.**

`SubmitSamples` batches at `telemetry.MaxDeviceLatencySamplesPerBatch` (line 95) and the loop
variable is `batch`, but the log reports `len(samples)` — the full partition, not the batch that
actually hit the limit. The number in the log overstates the loss whenever a partition spans more
than one batch.

## Proposed change

Use the shared drop counter from #4126, fix the count, and factor the duplicated
block:

```go
func (s *Submitter) handleAccountFull(log *slog.Logger, partitionKey PartitionKey, batchLen int) {
	metrics.SamplesDropped.WithLabelValues(metrics.DropReasonAccountFull).Add(float64(batchLen))
	metrics.Errors.WithLabelValues(metrics.ErrorTypeSubmitterAccountFull).Inc()
	log.Warn("partition account is full, dropping partition",
		"droppedBatch", batchLen,
		"epoch", partitionKey.Epoch)
	s.cfg.Buffer.Remove(partitionKey)
}
```

Called from both branches, replacing `len(samples)` with `len(batch)`.

Whether `return nil` should become a distinct sentinel is worth deciding during the fix. Returning
an error would trigger 5 pointless retries against an account that cannot accept writes, so `nil`
is probably right; the gap is the missing metric, not the control flow.

## Worth checking separately

An account filling up mid-epoch is itself a signal. If this fires outside of a backlog-flush
scenario it may mean the account sizing does not cover a full epoch at the current 10s sampling
interval, which would be a separate bug.

## Acceptance

- Account-full increments a counter that can be alerted on.
[...]

PR #4144, opened 2026-08-03
telemetry: log and count samples dropped at buffer capacity

### LINKED ISSUE #4126: device/telemetry: samples silently dropped when submission fails at buffer capacity
Tracked by #4130.

## Problem

`Submitter.Tick` discards samples with no log and no metric when submission failed and the
partition is at capacity:

```go
// controlplane/telemetry/internal/telemetry/submitter.go:246
// If submission failed and the buffer is not at capacity, prepend the samples back to the
// buffer. If the buffer is at capacity and we have failed all attempts, don't prepend the
// samples back to the buffer.
overCapacity := s.cfg.Buffer.Len(partitionKey)+len(tmp) >= s.cfg.Buffer.Capacity(partitionKey)
if !success && !overCapacity {
	s.cfg.Buffer.PriorityPrepend(partitionKey, tmp)
}

// Always recycle the slice for reuse
s.cfg.Buffer.Recycle(partitionKey, tmp)
```

When `!success && overCapacity`, `tmp` is recycled and the samples are gone. The
`SubmitterRetriesExhausted` counter fires for the failed submission, so you learn a submission
failed, but nothing anywhere records that the samples were **discarded** rather than requeued.

This is the only path in the submitter that permanently destroys measurement data with zero
signal.

Partition capacity is 4096 (`collector.go:17`, `partitionBufferCapacity`), which at 6 samples per
minute is roughly 11 hours of backlog before this triggers.

## Proposed change

Log and count the drop.

```go
if !success {
	if overCapacity {
		metrics.Errors.WithLabelValues(metrics.ErrorTypeSubmitterBufferFull).Inc()
		metrics.SamplesDropped.WithLabelValues(metrics.DropReasonBufferFull).Add(float64(len(tmp)))
		log.Warn("partition buffer at capacity after failed submission, dropping samples",
			"droppedSamples", len(tmp),
			"bufferLen", s.cfg.Buffer.Len(partitionKey),
			"capacity", s.cfg.Buffer.Capacity(partitionKey))
	} else {
		s.cfg.Buffer.PriorityPrepend(partitionKey, tmp)
	}
}
```

New identifiers needed in `internal/metrics/metrics.go`:

- `ErrorTypeSubmitterBufferFull = "submitter_buffer_full"`
- a `doublezero_device_telemetry_agent_samples_dropped_total` counter with a `reason` label

The `reason` label lets the account-full path (#4127) use the same counter.

## Noise

One line per partition per submission interval, and only while a partition has been backed up for
~11 hours. In steady state this never fires.

## Acceptance

- A test that fails submission and fills the partition to capacity asserts the warning is emitted
  and the counter increments by `len(tmp)`.
- Existing requeue behavior below capacity is unchanged.

elitegreg:
Resolves: #4126

## Summary of Changes

- The telemetry submitter now logs a warning and increments counters when it discards samples because a submission failed and the partition buffer is already at capacity. That path previously recycled the batch silently, so lost measurement data left no trace beyond `submitter_retries_exhausted`, which fires whether the batch was requeued or destroyed.
- New counter `doublezero_device_telemetry_agent_samples_dropped_total` with a `reason` label (`buffer_full`), so the account-full drop path (#4127) can report on the same series with a different reason.
- New `submitter_buffer_full` error type on the existing `doublezero_device_telemetry_agent_errors_total` counter, one increment per dropped batch.
- Requeue behavior below capacity is unchanged. Neither signal fires in steady state: a partition holds 4096 samples, roughly 11 hours of backlog at 6 samples per minute.

## Diff Breakdown

| Category    | Files | Lines (+/-) |  Net |
| Tests       |     1 | +89 / -0    |  +89 |
| Core logic  |     1 | +14 / -5    |   +9 |
| Scaffolding |     1 | +14 / -0    |  +14 |
| Docs        |     1 | +3 / -0     |   +3 |
| **Total**   |     4 | +120 / -5   | +115 |

Mostly tests around a nine-line observability fix; no behavior change to submission or requeueing.

<details>

- [`controlplane/telemetry/internal/telemetry/submitter.go`](https://github.com/malbeclabs/doublezero/pull/4144/files#diff-71c7c65c268c9f40ba1a5a78b3a0740268a543208924b94895f0f1de91a064d4) — the over-capacity branch in `Tick` now warns with the dropped count, buffer length, and capacity, and bumps both counters; the requeue branch is unchanged apart from being nested under the existing `!success` check
- [`controlplane/telemetry/internal/metrics/metrics.go`](https://github.com/malbeclabs/doublezero/pull/4144/files#diff-ef253c293b1d931c3caffa6ec21a1e17db4ef687891d7a47f4b34fdfe68f962e) — registers the `samples_dropped_total` counter, the `reason` label, and the `submitter_buffer_full` error type

</details>

## Testing Verification

- New test fills a partition to exactly its capacity, fails every submission attempt, and asserts the warning is emitted with `droppedSamples=2` and `capacity=2`, that the drop counter rises by the batch length, that the error counter rises once, and that nothing is requeued. Confirmed it fails against the pre-fix submitter.
- Companion test at capacity 1024 asserts the below-capacity path still requeues the batch, emits no drop warning, and leaves the drop counter untouched.
- Both counter assertions measure deltas and the subtests run serially, since the metrics are package-level and shared with the existing over-capacity subtests.
- `go test ./controlplane/telemetry/...` passes except `internal/netns`, which fails identically on `main` in this environment (needs privileges).


controlplane/telemetry/internal/metrics/metrics.go
```diff
@@ -46,6 +52,14 @@ var (
 		[]string{LabelErrorType},
 	)
 
+	SamplesDropped = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: MetricNameSamplesDropped,
+			Help: "Number of samples permanently discarded, by reason",
```
ben-dz: The Help string "Number of samples permanently discarded, by reason" overclaims in two directions, and an operator reading `samples_dropped_total` on a dashboard will treat it as the amount of telemetry lost.

**It under-reports.** `SubmitSamples` handles `telemetry.ErrSamplesAccountFull` by calling `Buffer.Remove(partitionKey)` and returning `nil` (`submitter.go:151-155`, `159-162`), which sets `success = true` and skips the new block entirely, so the account-full discard is uncounted. Deferring that to #4127 is the right scope and the `reason` label exists for it — the point is only that the Help text promises a totality the single `buffer_full` reason does not yet deliver.

**It over-reports.** `SubmitSamples` chunks `tmp` at `MaxDeviceLatencySamplesPerBatch` (239) and returns on the first failing chunk (`submitter.go:95-97`), while the retry loop re-submits all of `tmp` from index 0 on every attempt (`submitter.go:213-214`). If a prefix of chunks is written onchain and a later chunk fails persistently, `submitter.go:253` attributes the whole `len(tmp)` to loss. The two conditions co-occur by construction: `len(tmp) > 239` and `Len + len(tmp) >= 4096` (`collector.go:17`) both require the same sustained-failure regime, so the over-count is likeliest exactly when the metric is being read. An outage starting mid-flush over-reports by a few hundred to a few thousand samples.

Threading per-chunk progress out of `SubmitSamples` is more than the minimum change and would drag in the pre-existing duplicate re-write of already-committed prefixes. The minimum fix is wording that matches what is counted:

```suggestion
			Help: "Number of samples in batches discarded from the buffer without a successful submission, by reason",
```

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
+- Device telemetry
```
ben-dz: `- Device telemetry` is a new component heading. Every prior telemetry entry uses `- Telemetry`, with `- Telemetry (geoprobe)` for the variant (`CHANGELOG.md:264, 266, 328, 452, 521, 666, 987`). Use `- Telemetry` (or `- Telemetry (device)`) so the released section doesn't split into two telemetry groups.

ben-dz (approved):
Two Low findings; nothing blocking.

The core accounting checks out: hoisting `overCapacity` inside `if !success` is semantically identical to the old `!success && !overCapacity`, the `>=` boundary is unchanged, `len(tmp) > 0` is guaranteed by the early return at `submitter.go:195`, and `Recycle` after `PriorityPrepend` stays safe because the prepend copies into a fresh slice. No path emits the new signal without actually discarding, and no sample identifier reaches a metric label.

The one thing worth changing before merge is the counter's Help string, which promises more than the metric delivers in both directions — see the inline comment.

PR #4145, opened 2026-08-03
telemetry: count and correctly report samples dropped when account is full

### LINKED ISSUE #4127: device/telemetry: account-full path reports success and logs the wrong count

elitegreg:
Resolves: #4127

Stacked on #4144 (`gm/telemetry-drop-samples-metric`), which adds the shared `samples_dropped_total` counter this reuses. Base is that branch, so the diff shown is just this change; merge #4144 first.

## Summary of Changes

- Account-full drops are now visible in metrics. `SubmitSamples` returns `nil` on `ErrSamplesAccountFull` so `Tick` counts it as a successful submission, which means data loss previously left nothing but a warning. It now increments `doublezero_device_telemetry_agent_samples_dropped_total{reason="account_full"}` and `submitter_account_full` on the errors counter.
- The logged count is fixed. The old warning reported `len(samples)`, the whole flushed partition, rather than what was actually lost, overstating the loss whenever a partition spans more than one batch (a 300-sample partition that filled the account on its second batch logged `droppedSamples=300` when 61 samples were lost).
- The two duplicated account-full blocks (direct write and post-initialize) are factored into `handleAccountFull`.
- Control flow is unchanged: the account cannot accept further writes, so returning an error would only buy five pointless retries. The partition is still removed and the caller still treats it as done.

### Note on the dropped count

The issue proposed counting `len(batch)`. That undercounts for the same reason `len(samples)` overcounts: when the account fills mid-partition, the batches *behind* the failing one are never attempted and are discarded along with it, and `Buffer.Remove` also throws away whatever the collector buffered since the flush. `handleAccountFull` therefore counts `len(samples)-i` (the failing batch plus everything after it) and adds the partition's buffered length, read before the removal. The log reports the two separately as `unsubmittedSamples` and `bufferedSamples`.

## Diff Breakdown

| Category    | Files | Lines (+/-) |  Net |
| Tests       |     1 | +140 / -0   | +140 |
| Core logic  |     1 | +23 / -4    |  +19 |
| Scaffolding |     1 | +3 / -1     |   +2 |
| Docs        |     1 | +1 / -0     |   +1 |
| **Total**   |     4 | +167 / -5   | +162 |

Mostly tests; the fix itself is one extracted helper and two call sites.

<details>

- [`controlplane/telemetry/internal/telemetry/submitter.go`](https://github.com/malbeclabs/doublezero/pull/4145/files#diff-71c7c65c268c9f40ba1a5a78b3a0740268a543208924b94895f0f1de91a064d4) — new `handleAccountFull` records the error and drop counters, logs the unsubmitted and buffered counts, and removes the partition; both `ErrSamplesAccountFull` branches in `SubmitSamples` now call it with `len(samples)-i`
- [`controlplane/telemetry/internal/metrics/metrics.go`](https://github.com/malbeclabs/doublezero/pull/4145/files#diff-ef253c293b1d931c3caffa6ec21a1e17db4ef687891d7a47f4b34fdfe68f962e) — adds the `account_full` drop reason and the `submitter_account_full` error type

</details>

## Testing Verification

- New test adds a sample from inside the write callback (after the flush, so it is only reachable via `Buffer.Remove`) and asserts the drop counter rises by 3 for a 2-sample flush, the error counter by 1, and the log carries `unsubmittedSamples=2 bufferedSamples=1`.
- New multi-batch test flushes 300 samples with the first batch succeeding and the second hitting a full account, and asserts exactly 61 drops and no `unsubmittedSamples=300`. Against the pre-fix submitter this test sees the old `droppedSamples=300`, confirming the miscount.
- New test covers the post-initialize account-full branch, which the extracted helper now shares.
- All three fail against the pre-fix submitter. Existing account-full tests (`drops_samples_if_account_full`, `initializes_then_drops_samples_if_account_full`) still pass unchanged.
- The new subtests run serially and compare counter deltas, since the metrics are package-level and shared with those existing parallel subtests.


ben-dz (commented):
One thing worth addressing before this lands: the over-count survives on the retry path.

controlplane/telemetry/internal/telemetry/submitter.go
```diff
@@ -149,16 +149,14 @@ func (s *Submitter) SubmitSamples(ctx context.Context, partitionKey PartitionKey
...
+						s.handleAccountFull(log, partitionKey, len(samples)-i)
 						return nil
 					}
 					metrics.Errors.WithLabelValues(metrics.ErrorTypeSubmitterFailedToWriteSamples).Inc()
 					return fmt.Errorf("failed to write device latency samples after init: %w", err)
 				}
 			} else if errors.Is(err, telemetry.ErrSamplesAccountFull) {
-				log.Warn("Partition account is full, dropping samples from buffer and moving on", "droppedSamples", len(samples))
-				s.cfg.Buffer.Remove(partitionKey)
+				s.handleAccountFull(log, partitionKey, len(samples)-i)
```
ben-dz, line 163: `i` is the batch offset within the *current attempt*, but `Tick` re-invokes `SubmitSamples` with the same slice and this loop restarts at `i := 0`. So batches already persisted by an earlier attempt are counted as unsubmitted when a later attempt hits account-full at a lower index — the same over-count this PR fixes, on an adjacent path.

Reproduced on this branch (300 samples, two batches, `MaxAttempts: 2`; attempt 1 writes batch 0 then takes a transient RPC error on batch 1; attempt 2 re-sends batch 0 and gets `ErrSamplesAccountFull`):

```
msg="Partition account is full, dropping partition" unsubmittedSamples=300 bufferedSamples=0
samples_dropped_total{reason="account_full"} += 300
```

61 samples were actually lost. Reaching it needs a flush over `MaxDeviceLatencySamplesPerBatch` (239) — i.e. a `PriorityPrepend` backlog, which is the same sustained-RPC-trouble condition that produces the transient error.

Two ways to land it:

- **Minimum:** amend the `handleAccountFull` doc comment, which currently reads as exact — "Everything that has not been written is gone" is wrong when a prior attempt wrote some of it. Say the count is per-attempt and over-reports after a mid-partition retry.
- **Exact:** hoist the batch cursor into `Tick` so a retry resumes at the last successfully written index.

The second option also fixes the pre-existing bug behind this: any non-account-full error mid-partition makes the next attempt re-send batches already onchain, appending duplicate samples and pushing `next_sample_index` toward the 35,000 cap — manufacturing the very account-full condition this PR instruments. Out of scope here, but worth its own issue.

Either way, pin it with a subtest alongside `counts_only_unsubmitted_samples_when_account_fills_mid_partition` using `MaxAttempts: 2` — the new tests use `MaxAttempts: 1`, so no retry path is covered.

controlplane/telemetry/internal/telemetry/submitter.go
```diff
@@ -172,6 +170,27 @@ func (s *Submitter) SubmitSamples(ctx context.Context, partitionKey PartitionKey
...
+// removal.
+func (s *Submitter) handleAccountFull(log *slog.Logger, partitionKey PartitionKey, unsubmitted int) {
+	buffered := s.cfg.Buffer.Len(partitionKey)
+
+	metrics.Errors.WithLabelValues(metrics.ErrorTypeSubmitterAccountFull).Inc()
+	metrics.SamplesDropped.WithLabelValues(metrics.DropReasonAccountFull).Add(float64(unsubmitted + buffered))
+	log.Warn("Partition account is full, dropping partition",
+		"unsubmittedSamples", unsubmitted,
+		"bufferedSamples", buffered,
+		"epoch", partitionKey.Epoch)
```
ben-dz: Redundant: both callers bind `log` with `.With("partition", partitionKey)`, and `PartitionKey.String()` already ends in `-<epoch>`, so the line reads `partition=…-100 … epoch=100`. The sibling buffer-full warning below doesn't add it either. Drop the key/value pair.

controlplane/telemetry/internal/telemetry/submitter.go
elitegreg: Went with the exact fix. `SubmitSamples` now returns how many samples it wrote, and `Tick` carries that cursor across attempts, so a retry sends `tmp[written:]` instead of restarting at the beginning of the flushed partition. The account-full count is then exact on every path: the sub-slice never contains batches an earlier attempt already put onchain.

I also extended it to the requeue. `PriorityPrepend` was pushing the whole `tmp` back after a failed submission, so even with a per-attempt cursor the *next* tick would re-send batch 0. It now requeues only `tmp[written:]`, and the buffer-full drop counter measures that same remainder rather than the full flush.

That means the pre-existing duplicate-append bug is fixed here rather than deferred — you were right that it's the mechanism manufacturing the account-full condition this PR instruments, and it seemed worse to ship the cursor half-applied. Happy to split it out if you'd rather keep this PR to the metric.

Two subtests, both using your repro (300 samples, two batches, `MaxAttempts: 2`; attempt 1 writes batch 0 then takes a transient error on batch 1):

- `counts_only_unsubmitted_samples_when_account_fills_after_a_retry` — asserts the drop counter rises by 61 and the write sizes are `[239, 61, 61]`. Against the pre-fix code the sizes are `[239, 61, 239]` and the log reads `unsubmittedSamples=300`, exactly what you saw.
- `requeues_only_unwritten_samples_after_a_failed_submission` — all attempts fail; asserts 61 samples are requeued, not 300.

controlplane/telemetry/internal/telemetry/submitter.go
elitegreg: Dropped.

ben-dz (approved):

PR #4147, opened 2026-08-03
telemetry: log rpc endpoint identity and collapse repeated failures

### LINKED ISSUE #4129: device/telemetry: log RPC endpoint identity, collapse repeated failures into transitions
Tracked by #4130.

## Problem

During the 2026-07-29 ledger RPC outage the telemetry agent emitted roughly 600 near-identical
error lines over 22 hours and still did not explain what was wrong. Meanwhile the three facts that
would have identified the cause were either unlogged or invisible.

The pattern to fix is logging per iteration instead of per state change.

### 1. Log which RPC endpoint is actually in use

Nothing anywhere records the resolved remote address. The 2026-07-29 outage was one bad load
balancer IP out of five behind one hostname, and no log line on any host said which one it was
talking to. This is the single fact that would have cut triage from a day to minutes.

Add a `DialContext` wrapper that logs the resolved address on each new connection:

```go
DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
	c, err := dialer.DialContext(ctx, network, addr)
	if err != nil {
		return nil, err
	}
	log.Info("ledger rpc connection established",
		"host", addr, "remote", c.RemoteAddr().String())
	return c, nil
},
```

Noise: with connection pooling this is a handful of lines per day. It also has the useful property
of going quiet exactly when a client has stopped re-dialing, which is itself the symptom.

### 2. Collapse the pinger's epoch failure into a state transition

`pinger.go:58` logs an Error on every failing tick:

```go
epoch, err := p.getCurrentEpoch(ctx)
if err != nil {
	p.log.Error("failed to get current epoch", "error", err)
	return
}
```

Replace with first-failure and recovery only:

```go
// on entering the failed state
p.log.Error("epoch fetch failing, probing degraded", "error", err)
// on recovery
p.log.Info("epoch fetch recovered", "downtime", d, "ticksSkipped", n)
```

Roughly 600 lines become 2, and the recovery line carries the duration and skipped-tick count,
which is the part anyone actually needs. Add an
`errors_total{error_type="pinger_epoch_fetch"}` counter for the middle; the pinger currently has
no metric at all.

(If #4125 lands first, this becomes a warning about serving a
cached epoch rather than about skipping the tick. Either way the transition-only shape applies.)

### 3. Log peer count changes

`peers.go:210` reports the outcome of every refresh at Debug, so the peer list is invisible in
production:

```go
p.log.Debug("Refreshed peers", "devices", len(devices), "links", len(links), "peers", len(peers), "tunnelsNotFound", tunnelsNotFound)
```

Keep the Debug line, and add an Info line when the count changes plus a Warn when it reaches zero.
A refresh that succeeds and returns zero peers is currently indistinguishable from a healthy one at
Info level. Add a `peers_total` gauge alongside the existing
`peer_discovery_not_found_tunnels`.

Noise: one line per change. In steady state, none.

### 4. Route the stale-cache warning through the agent logger

`internal/serviceability/cache.go:87` calls the package-global `slog`:

```go
slog.Warn("telemetry: program data fetch failed, returning stale cached data", "age", cachedAge, "error", err)
```

`cmd/telemetry/main.go:129` builds its own handler and never calls `slog.SetDefault`:

```go
log := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{
	Level:     logLevel,
	AddSource: true,
}))
```

So the most diagnostic signal in the agent — "the ledger is unreachable and I am serving stale
data" — is emitted through a different logger, in a different format, with no `source=` field, and
cannot be leveled with the rest of the agent. That is very likely why it was missed during triage.

Fix: give `CachingFetcher` an injected `*slog.Logger` (`NewCachingFetcher` already takes a config-ish
argument list; add it there), and rate-limit to first occurrence plus recovery. It currently fires
on every refresh, which is every 10s for the whole outage.

The existing `metricStaleCacheAge` gauge already covers the sustained case and is the right thing
to alert on.

### 5. Make silent drops loud

Covered in detail by #4126 (buffer-full) and #4127 (account-full). Listing here
so the logging work is not considered done without them.

## Not proposed

Per-tick logging of successful probes, submissions, or refreshes. The failure mode being fixed is
too much repetition, not too little detail.

## Acceptance

- A 22-hour total RPC outage produces single-digit log lines per component, not hundreds.
- The resolved RPC endpoint appears in logs.
- Peer count reaching zero is visible at Info level.
- Every path that discards samples has both a log line and a counter.

elitegreg:
Resolves: #4129

Third in a chain: #4143 → #4146 → this. Base is `gm/telemetry-peer-cache-refresh`, so the diff shown is just this change; merge #4143 and #4146 first.

Item 5 of the issue (make silent drops loud) is #4126 / #4127, already in #4144 and #4145. Items 1-4 are here.

## Summary of Changes

- **Logs name the RPC endpoint actually in use.** `Options.OnDial` on the shared retrying RPC constructor, and the same hook on the netns client, report the resolved remote address of each new connection. The agent logs it once per distinct address, so a hostname that resolves to several load balancer addresses no longer hides which one this host is talking to — the fact that would have cut 2026-07-29 triage from a day to minutes.
- **The pinger gets a per-failure epoch-fetch counter.** `errors_total{error_type="pinger_epoch_fetch"}`, incremented once per exhausted fetch. #4143 already turned this path's logging into transitions and demoted the per-retry `Warn` to `Debug`, so that work is not repeated here; the counter is what it does not cover. `EpochCacheStaleAge` only says how far behind the cache is and `pinger_epoch_unavailable` only fires once probing is refused, so an endpoint that fails intermittently without ever going stale enough to stop probing is invisible without this. Shutdown is not counted.
- **Peer count is visible above Debug.** An `Info` line when the count changes, a `Warn` when a refresh succeeds with no peers (previously indistinguishable from a healthy one), and a new `doublezero_device_telemetry_agent_peers` gauge. Steady state stays silent.
- **The stale-cache warning goes through the agent's logger.** `NewCachingFetcher` takes a `*slog.Logger` instead of calling package-global `slog`, which the agent never points anywhere: the single most diagnostic line in the agent was being written in a different format, without `source=`, and could not be leveled with the rest. It is also now rate-limited to the first stale read plus a recovery line, rather than firing every 10s for the length of an outage.

## Diff Breakdown

| Category    | Files | Lines (+/-)  |  Net |
|-------------|-------|--------------|------|
| Tests       |     6 | +444 / -4    | +440 |
| Core logic  |     6 | +191 / -20   | +171 |
| Scaffolding |     2 | +20 / -3     |  +17 |
| Docs        |     1 | +1 / -0      |   +1 |
| **Total**   |    15 | +656 / -27   | +629 |

Four independent logging changes, each small; the bulk is tests. Non-test code is ~210 lines, inside the 500-line guideline.

<details>

- `controlplane/telemetry/internal/rpclog/endpoint.go` — new `EndpointLogger`: logs the first connection to a dial target and any later connection resolving somewhere new, silent on repeats
- `controlplane/telemetry/internal/serviceability/cache.go` — injected logger, `enterStale` records the transition and warns once, recovery logs downtime and stale-read count
- `tools/solana/pkg/rpc/retry.go` — `Options.OnDial`, invoked after a successful dial with the target and resolved remote; unset means the transport is built exactly as before
- `controlplane/telemetry/internal/netns/http.go` — same hook for the namespaced client, plus per-field option defaults so setting only `OnDial` does not zero the timeouts
- `controlplane/telemetry/internal/telemetry/peers.go` — peer count gauge and change/zero logging
- `controlplane/telemetry/internal/telemetry/pinger.go` — counts each exhausted epoch fetch, skipping context cancellation

</details>

### Notes for review

- Rebased onto #4143, which moved the epoch fetch off the probe tick. My original transition logging here duplicated what `epochForTick`/`markEpochStale`/`storeEpoch` now do, so it is gone; only the per-failure counter survives, and it moved into `getCurrentEpoch` where both the refresh loop and the first-fetch path pass through it.
- The netns transport does not pool connections, so its `OnDial` fires once per request. That is why the dedupe lives in `EndpointLogger` rather than at the call site — logging every dial there would be thousands of lines a day, not the handful the issue estimated for a pooled client. The tradeoff is that a client which has stopped re-dialing no longer goes conspicuously quiet.
- `OnDial` is not called when a dial fails; the error already reaches the caller.
- `NewCachingFetcher`'s signature changes, and the telemetry agent is its only caller. `client/doublezerod` has a separate copy of this type that still uses package-global `slog`; out of scope here.

## Testing Verification

- `EndpointLogger`: covers first connection, silence across 100 repeat dials, the change line with `previousRemote`, independent tracking per target, and 50 concurrent dials collapsing to one line under `-race`.
- `OnDial` wiring: a real `httptest` server asserts the reported target and resolved remote match the listener, and a connection-refused endpoint asserts nothing is reported for a failed dial.
- Pinger: five failing ticks produce five counter increments, and a tick under a cancelled context produces none. #4143's own subtests still cover the transition logging on this path.
- Peers: first refresh logs the count at Info, ~10 further refreshes with an unchanged list add nothing, dropping every link logs the WARN with `peers=0 previousPeers=2`, and the gauge tracks both.
- Caching fetcher: 10 stale reads produce one warning and keep serving cached data; recovery reports `staleReads=10`; a second outage after recovery warns again; a failure with no cache still returns the error and logs no stale line.
- Package passes under `-race`, including the peer discovery loop reading its log buffer concurrently.
- `internal/netns` has one pre-existing failure in this environment (`TestRunInNamespace_EmptyNameErrors` needs privileges); it fails identically on `main`.


controlplane/telemetry/internal/rpclog/endpoint.go
```diff
@@ -0,0 +1,50 @@
...
+// OnDial reports a newly established connection to addr, resolved to remote. It is safe for
+// concurrent use and does no I/O beyond the log line it may emit.
+func (l *EndpointLogger) OnDial(addr, remote string) {
+	l.mu.Lock()
+	previous, known := l.seen[addr]
+	if known && previous == remote {
+		l.mu.Unlock()
+		return
+	}
+	l.seen[addr] = remote
```
ben-dz: `seen` maps dial target → *last* remote, so any dial resolving differently from the previous one logs `Ledger RPC endpoint changed` at Info. The premise this rests on — same target ⇒ same remote ⇒ silent — doesn't hold for the endpoint it was written for.

The production ledger hostname serves a **single rotating A record**. Eight consecutive `getent ahostsv4 doublezero-mainnet-beta.rpcpool.com` returned `208.91.111.241` ×6 and `204.16.246.49` ×2, interleaved; a 40-lookup run over ~28s produced 16 transitions between `208.91.111.241` and `204.16.246.48`. Go's resolver caches nothing, and `sortByRFC6724` deliberately skips Rule 9 for IPv4, so the dialer takes whatever DNS put first.

Worst on the path production uses: with `-management-namespace` set, `SingleThreadTransport.RoundTrip` dials once per request (`netns/http.go:137`). At the shipped defaults the agent issues ~12 ledger requests/min (epoch refresh every 10s via the `-probe-interval` fallback, plus peer discovery every 10s through a 5s-TTL cache) — order 17k dials/day per device, a large fraction of which emit an Info line under the observed rotation. That inverts the goal, at a higher level than the Debug status quo, on Arista `/var/log/agents` where rotation then evicts the probe-failure and epoch-staleness lines that matter.

It also destroys the diagnostic: with "changed" as normal churn, a genuine pin-to-one-bad-load-balancer — the 2026-07-29 case in the package doc comment — is indistinguishable from noise. The pooled path (`tools/solana/pkg/rpc/retry.go:136`) is lower volume but loses the same meaning on each re-dial, and with `MaxConnsPerHost: 9` can hold connections to several backends at once while `seen` records one.

Fix: dedupe on the *(target, remote)* pair — `seen map[string]map[string]struct{}` (or key `addr+"|"+remote`) — and log `Ledger RPC connection established` the first time each distinct remote is seen for a target, silent thereafter. Output is then bounded by the backend pool size, and a genuinely new backend still logs once. Drop the `endpoint changed`/`previousRemote` branch, which can't be made meaningful against rotating DNS.

Test: alternate `OnDial(host, r1)` / `OnDial(host, r2)` 20 times and assert exactly 2 `Ledger RPC` lines. Today it emits 21.

controlplane/telemetry/internal/serviceability/cache.go
```diff
@@ -27,18 +27,34 @@ type ProgramDataProvider interface {
...
-func NewCachingFetcher(provider ProgramDataProvider, cacheTTL, rpcTimeout time.Duration) *CachingFetcher {
+// NewCachingFetcher creates a CachingFetcher with the given logger, provider, TTL, and RPC
+// timeout.
+//
+// The logger is injected rather than taken from the package-global slog: the agent builds its
+// own handler and never calls slog.SetDefault, so "the ledger is unreachable and I am serving
+// stale data" would otherwise be emitted in a different format, without the agent's source
+// field, and could not be leveled with the rest of the agent's output.
+func NewCachingFetcher(log *slog.Logger, provider ProgramDataProvider, cacheTTL, rpcTimeout time.Duration) *CachingFetcher {
+	if log == nil {
```
ben-dz: This fallback silently restores the package-global `slog` behavior the doc comment just above (`:48-51`) argues is the bug being fixed, and the sibling constructor in this subsystem rejects a nil logger outright (`telemetry/peers.go:71`). Delete the branch — the sole caller always passes `log` (`cmd/telemetry/main.go:296`).

controlplane/telemetry/internal/serviceability/cache.go
```diff
@@ -97,12 +113,37 @@ func (f *CachingFetcher) GetProgramData(ctx context.Context) (*serviceability.Pr
...
+func (f *CachingFetcher) enterStale(cachedAge time.Duration, err error) {
+	f.mu.Lock()
+	first := f.staleSince.IsZero()
+	if first {
+		f.staleSince = time.Now()
+	}
+	f.staleReads++
+	f.mu.Unlock()
+
+	if first {
```
ben-dz, line 143: `enterStale` logs nothing on stale reads after the first, not even Debug, so a failure cause that changes mid-outage (503 → TLS → DNS) is invisible even under `-log-verbose`. The sibling `markEpochStale` handles this with an `else` Debug branch (`telemetry/pinger.go:332`) — mirror it.

controlplane/telemetry/internal/telemetry/main_test.go
```diff
@@ -233,3 +234,28 @@ func loopbackInterface(t *testing.T) string {
 	}
 	return ""
 }
+
+// syncBuffer collects log output for a test that reads it while the code under test is still
+// logging from another goroutine. bytes.Buffer alone races there.
+type syncBuffer struct {
```
ben-dz: `syncBuffer` reimplements `recordingHandler` (`pinger_test.go:655`), already in this test package: same mutex-guarded capture, plus `count(level, message)` for exact level+message assertions rather than the `level=WARN` / `peers=2` substring scraping in the new subtest. Add a `reset()` to `recordingHandler` and delete `syncBuffer`.

controlplane/telemetry/internal/telemetry/peers_test.go
```diff
@@ -696,6 +700,102 @@ func TestAgentTelemetry_PeerDiscovery_Ledger(t *testing.T) {
...
+		require.Eventually(t, func() bool {
+			return strings.Contains(logs.String(), "Peer count changed")
+		}, 2*time.Second, 20*time.Millisecond)
+		assert.Contains(t, logs.String(), "peers=2")
+
+		// Many refreshes later, an unchanged count has still only logged once.
+		require.Eventually(t, func() bool {
+			return testutil.ToFloat64(metrics.Peers.WithLabelValues(localDevicePK.String())) == 2
+		}, 2*time.Second, 20*time.Millisecond)
+		time.Sleep(200 * time.Millisecond)
```
ben-dz: `time.Sleep(200ms)` then `assert.Equal(1, count)` proves "didn't log again", not "refreshed repeatedly and didn't log again" — on a loaded runner the 20ms ticker can deliver one or two ticks and this passes vacuously. The mock already holds a `mu`: count `GetProgramDataFunc` invocations, `require.Eventually` until ≥5, then assert the log count once. Removes the sleep too.

ben-dz (changes requested):
One blocking issue: `EndpointLogger` dedupes on the dial target only, so the rotating single A record the production ledger hostname actually serves makes `Ledger RPC endpoint changed` the steady state rather than a transition. On the netns path — the DZD case, which dials once per request — that is on the order of 17k dials/day per device with a large fraction logging at Info, which both floods `/var/log/agents` and hides the pin-to-one-bad-backend signal the package was written to expose. Keying the dedupe on the (target, remote) pair fixes it and is smaller than the code that's there.

The rest is minor: a nil-logger fallback that reinstates the behavior the PR removes, no Debug line for stale reads after the first (the pinger sibling has one), a test helper that duplicates `recordingHandler` already in the same package, and a sleep-based steady-state assertion that can pass vacuously.

The `NewWithRetries` → `New(…, Options{})` swap is behaviorally identical, the caching fetcher's locking is sound under singleflight, and `loggedPeerCount` is genuinely single-goroutine.

- The PR description's pinger bullet describes the failure/recovery state transitions and the retry `Warn`→`Debug` demotion as part of this change, but both live in the base branch (`pinger.go:284-334`, `:346`) — this diff only adds the `pinger_epoch_fetch` counter. `pinger_epoch_unavailable` also predates it, so it isn't "the pinger's first metric". Worth updating the body so reviewers of a stacked PR aren't hunting for absent changes.

controlplane/telemetry/internal/rpclog/endpoint.go
elitegreg: Fixed — `seen` is now `map[string]map[string]struct{}`, keyed on `(target, remote)`. Dropped the `endpoint changed`/`previousRemote` branch entirely since it can't be made meaningful against rotating DNS; every distinct remote now logs `Ledger RPC connection established` once. Added the alternating-remote test you described (asserts 2 lines over 20 alternating dials, not 21).

controlplane/telemetry/internal/serviceability/cache.go
elitegreg: Deleted the fallback — the sole caller always passes `log`, and the sibling constructor in `peers.go` rejects nil outright, so this now does the same.

controlplane/telemetry/internal/serviceability/cache.go
elitegreg: Added an else branch mirroring `markEpochStale`: reads after the first now log `Program data fetch still failing, serving stale cached data` at Debug, with a test that varies the error across three stale reads and asserts both the single Warn and the two Debug lines carry the changing cause.

controlplane/telemetry/internal/telemetry/main_test.go
elitegreg: Deleted `syncBuffer` and added `reset()` to `recordingHandler`, plus an `attrs(level, message)` helper so callers can assert on specific attribute values instead of scraping formatted text. `peers_test.go` now uses `recordingHandler` directly.

controlplane/telemetry/internal/telemetry/peers_test.go
elitegreg: Replaced the sleep with an atomic counter on `GetProgramDataFunc`: `require.Eventually` waits for it to reach 5 invocations, then asserts the log count once. No sleep left in the test.

elitegreg:
Addressed in f189ca1c:

- Blocking: `EndpointLogger` now dedupes on the `(target, remote)` pair, not target alone. Dropped the `endpoint changed`/`previousRemote` branch since it can't be made meaningful against a rotating DNS record; added a test alternating between two remotes 20 times asserting exactly 2 log lines, and updated the existing "different address" test for the new semantics.
- Deleted the nil-logger fallback in `CachingFetcher`.
- `enterStale` now logs repeated stale reads at Debug, mirroring `markEpochStale`; added a test varying the failure cause across three stale reads.
- Deleted `syncBuffer`, added `reset()` and an `attrs()` helper to `recordingHandler`, and switched `peers_test.go` to use it.
- Replaced the sleep-based steady-state assertion with one that waits on an atomic count of `GetProgramDataFunc` invocations.

The PR description's pinger bullet already correctly attributes the transition logging and Warn→Debug demotion to #4143 — didn't see anything to change there, but flag if I'm missing something.

ben-dz (approved):
All five findings from the previous round are fixed in `f189ca1c`, each verified in the code and pinned by a test.

- The endpoint dedupe is now keyed on `(target, remote)` with the `endpoint changed`/`previousRemote` branch removed tree-wide, and the new alternating-remote subtest asserts 2 lines over 20 rotating dials where the old code emitted 21.
- The nil-logger fallback in `NewCachingFetcher` is deleted, so no path reaches package-global `slog`.
- `enterStale` has the `else` Debug branch mirroring `markEpochStale`, with a test that varies the error across three stale reads and asserts one Warn, two Debug lines, and the changed cause on both.
- `syncBuffer` is gone and `recordingHandler` gained `reset()`/`attrs()`; `peers_test.go` asserts exact level+message counts and typed attributes instead of scraping formatted text.
- The `time.Sleep` steady-state assertion is replaced by `require.Eventually` on a fetch counter, so it now proves the ticker fired repeatedly.

`go test -race -count=2` on `internal/rpclog`, `internal/serviceability`, `internal/telemetry`, and `tools/solana/pkg/rpc` passes.

PR #4150, opened 2026-08-05
cli: show multicast group codes in feed list

bgm-malbeclabs:
Resolves: malbeclabs/infra#2172

## Summary of Changes

- `doublezero feed list` gains a `group_codes` column naming the multicast groups the feed holds, so the list answers "what's in this feed?" without a follow-up `feed get` plus manual pubkey lookups.
- Groups are resolved to codes via one `list_multicastgroup` call; a group the ledger no longer carries falls back to its raw pubkey, matching how `accesspass list` renders its allowlists.
- The existing `groups` count column is unchanged, and the JSON output gains the new field (additive — existing consumers that decode a field subset, including the `feed_crud` e2e test, are unaffected).

## Diff Breakdown

| Core logic   |     1 | +17 / -1    |  +16 |
| Tests        |     1 | +75 / -0    |  +75 |
| Docs         |     1 | +3  / -0    |   +3 |
| **Total**    |     2 | +95 / -1    |  +94 |

A one-column display change; most of the diff is the new list test.

<details>

- `smartcontract/cli/src/feed/list.rs` — adds the `group_codes` field to `FeedDisplay`, resolves group pubkeys to codes with a pubkey fallback, and adds `test_cli_feed_list` covering both the resolved and unknown-group paths
- `CHANGELOG.md` — Unreleased / Changes entry under CLI

</details>

## Testing Verification

- New `test_cli_feed_list` asserts the rendered table carries the column and that a feed holding one known and one unknown group renders `mg01, 11111115q4Ep…` — the code for the resolved group, the raw pubkey for the other.
- Rendered output confirmed by hand: `account | code | name | exchange | groups | group_codes | owner`.
- `cargo test -p doublezero-serviceability-cli feed::` — 34 passing.

### DIFF
--- a/smartcontract/cli/src/feed/list.rs
+++ b/smartcontract/cli/src/feed/list.rs
@@ -2,7 +2,9 @@ use crate::doublezerocommand::CliCommand;
 use clap::Args;
 use doublezero_cli_core::{render_collection, CliContext, OutputFormat};
 use doublezero_program_common::serializer;
-use doublezero_sdk::commands::feed::list::ListFeedCommand;
+use doublezero_sdk::commands::{
+    feed::list::ListFeedCommand, multicastgroup::list::ListMulticastGroupCommand,
+};
 use serde::Serialize;
 use solana_sdk::pubkey::Pubkey;
 use std::io::Write;
@@ -27,6 +29,8 @@ pub struct FeedDisplay {
     #[serde(serialize_with = "serializer::serialize_pubkey_as_string")]
     pub exchange: Pubkey,
     pub groups: usize,
+    /// Codes of the multicast groups joinable in this feed (pubkey if the group is unknown).
+    pub group_codes: String,
     #[serde(serialize_with = "serializer::serialize_pubkey_as_string")]
     pub owner: Pubkey,
 }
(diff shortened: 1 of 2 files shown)


smartcontract/cli/src/feed/list.rs
```diff
@@ -27,6 +29,8 @@ pub struct FeedDisplay {
     #[serde(serialize_with = "serializer::serialize_pubkey_as_string")]
     pub exchange: Pubkey,
     pub groups: usize,
+    /// Codes of the multicast groups joinable in this feed (pubkey if the group is unknown).
```
martinsander00: You will kill me but I think we can remove this comment 😂 should be self explanatory

smartcontract/cli/src/feed/list.rs
```diff
@@ -61,3 +76,79 @@ impl ListFeedCliCommand {
...
+        );
+        assert!(res.is_ok(), "{res:?}");
+
+        let output_str = String::from_utf8(output).unwrap();
+        assert!(
+            output_str.contains("group_codes"),
+            "missing column: {output_str}"
+        );
+        assert!(
+            output_str.contains(&format!("mg01, {unknown_mgroup_pk}")),
```
martinsander00: Should we pin to the exact log and position?

smartcontract/cli/src/feed/list.rs
bgm-malbeclabs: Fair, dropped it. 51bab27

smartcontract/cli/src/feed/list.rs
bgm-malbeclabs: Done — now a single `assert_eq!` against the full rendered table, so column order and position are pinned too (matches `multicastgroup list` / `feed get`). 51bab27

martinsander00 (approved):

PR #4152, opened 2026-08-05
telemetry: surface program errors from finalized transactions

elitegreg:
Resolves: malbeclabs/infra#1703

## Summary of Changes

- The Go telemetry SDK executor no longer reports a transaction the program rejected as a success. Finalization only means the cluster agreed on the transaction: an instruction the program refused finalizes too, carrying the rejection in `err`, which the executor never read. It now returns a `*telemetry.ProgramError` holding the ledger's error and the program's log output.
- The device telemetry submitter treats that rejection as permanent for the tick — it logs the program's own explanation at `Error` and leaves the tick's remaining attempts unspent, rather than spending them on an instruction the ledger has already refused. New `submitter_program_error` type on the existing errors counter.
- `ProgramError.Error()` leads with the program's explanation (`Program log:` lines, minus the runtime's invoke/consumed boilerplate and the instruction-name echo), so a caller that only prints the error still gets the reason. This is the check `smartcontract/sdk/go/serviceability/executor.go` already made; telemetry was the outlier.
- A samples-account-full or missing-account rejection that reaches execution now returns the same `ErrSamplesAccountFull` / `ErrAccountNotFound` its preflight equivalent does, through the new `ProgramError.CustomErrorCode()`. Preflight catches nearly all of these, but a write that simulated cleanly and then failed against the bank it landed on reported its code only through the finalized transaction, so a caller's account-full handling worked on one side of preflight and not the other.
- The internet-latency collector now threads a written count through `SubmitSamples`, the way the device submitter has since #4145. It re-sent a partition from index 0 on every attempt and requeued all of it on failure, so any failure part-way through a multi-batch partition appended the earlier batches a second time. Reachable today from an RPC timeout mid-partition; surfacing program rejections adds another way in.
- The submitter's init→write path now runs the write whether or not the init was accepted, and reports the init failure only if the write still finds nothing there. `AccountAlreadyExists` (1010) was being hidden by the same silence, and the old code survived it by accident: init "succeeded", the write ran, the account was there, samples landed. Surfacing the rejection without this would have turned a self-healing path into a skipped tick plus a "rejected by the program" error on a device that is fine. `submitter_failed_to_initialize_account` likewise waits for the write's verdict instead of firing on a failure the write absorbs.

This is what left chi-dn-dzd4 silent for over an hour. Its `metrics_publisher` had been set to a key the agent did not hold, and the init half of the init→write path skips preflight — so `UnauthorizedAgent` (0x3e9) only arrived on the finalized transaction, which the SDK read as success. The agent looped init → write → `account not found` every few seconds, and the only error it printed named the missing account, not the authorization failure that caused it.

## Diff Breakdown

| Tests        |     6 | +439 / -8   | +431 |
| Core logic   |     4 | +222 / -18  | +204 |
| Docs         |     1 | +6 / -0     |   +6 |
| Scaffolding  |     1 | +6 / -0     |   +6 |
| **Total**    |    12 | +673 / -26  | +647 |

Mostly tests: ~228 non-test lines across the SDK executor and client, the device submitter, and the internet-latency collector.

<details>

- `smartcontract/sdk/go/telemetry/executor.go` — new `ProgramError` type with `CustomErrorCode()` and the program-log filter; `waitForTransactionFinalized` returns it when the finalized signature status or the transaction meta carries an error, fetching the logs best effort so a node that cannot return the transaction costs the logs rather than replacing the rejection with an RPC error
- `controlplane/telemetry/internal/telemetry/submitter.go` — `errors.As` branch in `Tick`'s retry loop (count, log at `Error`, stop the tick's attempts), and the init→write restructure that keeps a rejected-but-harmless init from ending the submission
- `smartcontract/sdk/go/telemetry/client.go` — `sentinelForProgramError` maps finalized-path custom codes onto the sentinels preflight already returns
- `controlplane/internet-latency-collector/internal/exporter/submitter.go` — `SubmitSamples` returns how many samples it wrote, so a retry resumes at the first unwritten one and only the remainder is requeued
- `controlplane/telemetry/internal/metrics/metrics.go` — `ErrorTypeSubmitterProgramError`, which narrows *why* a submission failed and overlaps the write/init types rather than replacing them
- `e2e/sdk_device_telemetry_test.go`, `e2e/sdk_internet_telemetry_test.go` — the "initialize again" cases asserted `NoError` and then read the rejection off `res.Meta.Err` by hand, which is what the SDK's silence forced; they now assert the error it returns

</details>

## Testing Verification

- `TestSDK_Telemetry_Executor_FinalizedWithProgramError` reproduces the chi-dn-dzd4 transaction — finalized, `Custom: 1001`, with the "not authorized for origin device" program log — across three shapes: the rejection on the signature status, on the transaction meta only (a node that returns a clean status), and with the logs unfetchable. All three must return an error rather than a signature; the error message carries the custom code in every case and the program's explanation whenever the logs were available, with the boilerplate stripped.
- `does_not_retry_a_submission_the_program_rejected` drives a full submitter tick with `MaxAttempts: 5` and asserts the init is not re-sent, `submitter_program_error` increments once, `submitter_retries_exhausted` does not, the reason reaches the log, and the samples are requeued for the next tick.
- `writes_anyway_when_the_account_already_exists` covers the 1010 path the e2e run caught: init rejected, write attempted anyway, samples land, nothing requeued, and no "rejected by the program" error logged.
- `retries_resume_at_the_first_unwritten_sample` in the collector: a 262-sample partition whose first batch lands and second fails. Verified it fails without the fix — 262 samples requeued, meaning the next tick would re-send the 245 already onchain — and passes with it at 17.
- `CustomErrorCode` is table-tested across the numeric types a JSON decoder can hand back (`json.Number`, `float64`, `int`, `uint64`) plus the shapes that carry no code, a negative code, and one past `uint32`.
- `ProgramLogMessages` asserts a system-program CPI failure survives the filter: `Transfer: insufficient lamports 0, need 890880` is kept, the invoke/consumed/failed lines and the instruction echo are dropped, and the reason reaches `Error()`.
- `FinalizedCustomErrorsMapToSentinels` drives the client with a finalized status carrying 1006 and 1011 and asserts `ErrSamplesAccountFull` / `ErrAccountNotFound`.
- Both e2e SDK telemetry tests run locally against cEOS containers — `TestE2E_SDK_Telemetry_DeviceLatencySamples` and `TestE2E_SDK_Telemetry_InternetLatencySamples`, including the two `try_to_initialize_..._again` subtests that surfaced the 1010 path.
- The device telemetry package passes under `-race -count=2` including `TestSubmitter_RetainsEverySampleAcrossTheStalenessBound`, which shares the retries-exhausted counter.
- The two pre-existing finalization edge cases (`Meta == nil`, `GetTransaction` returning nil) keep their original error messages — a clean signature status still falls through to them unchanged.
- Full suites pass for `smartcontract/sdk/go/telemetry`, `controlplane/telemetry/...`, and `controlplane/internet-latency-collector/...`. One unrelated pre-existing failure in `controlplane/telemetry/internal/netns` (`TestRunInNamespace_EmptyNameErrors` needs namespace privileges) reproduces identically on a clean `main` checkout.

### Not in scope

- `controlplane/internet-latency-collector/internal/exporter/submitter.go` gets the duplicate-write fix but not the rejection classification: it still spends its attempts retrying a permanent rejection, and returns early on a harmless init rejection rather than writing anyway. Its own retry loop absorbs both within the tick, so the cost is wasted attempts rather than delay or duplication.
- `sdk/geolocation/go/executor.go` is the third copy of this pattern and still has the gap — its `waitForTransactionFinalized` never reads `status.Err` either. Worth its own issue; not touched here.
- A write that passes preflight and then finalizes against a full account now surfaces as a generic `ProgramError` instead of the `ErrSamplesAccountFull` sentinel, so it requeues and takes the drop path on the next tick instead of immediately. Previously that race reported success and lost the samples with no signal at all, so this is strictly better; mapping `meta.err` custom codes back onto the sentinels would need a parser in `client.go` and is a separate change.


smartcontract/sdk/go/telemetry/executor.go
```diff
@@ -19,6 +20,42 @@ var (
...
 )
 
+// ProgramError reports a transaction that finalized onchain while the program rejected the
+// instruction it carried. The instruction did not take effect, and re-sending it unchanged will be
+// rejected the same way, so callers should treat it as a permanent failure for that input rather
+// than retry it.
+type ProgramError struct {
+	// Err is the transaction error the ledger reported, e.g.
+	// map[InstructionError:[0 map[Custom:1001]]] for TelemetryError::UnauthorizedAgent.
+	Err any
```
nikw9944, line 32: `Err any` has no accessor, so custom codes arriving on the finalized path skip the sentinel mapping at `client.go:304-330`. Beyond the account-full case the PR acknowledges, `AccountAlreadyExists` (1010) used to be swallowed so the write at `submitter.go:153` succeeded in the same attempt; now `Tick` breaks and recovery waits a full tick, which also contradicts `metrics.go:46-48`.

smartcontract/sdk/go/telemetry/executor.go
```diff
@@ -19,6 +20,42 @@ var (
...
+	return fmt.Sprintf("transaction finalized with program error: %v", e.Err)
+}
+
+// ProgramLogMessages returns the program's own log lines with the runtime's invoke/consumed/success
+// boilerplate and the instruction-name echo removed. These are the lines that say why the program
+// rejected the instruction, e.g. "Agent <pubkey> is not authorized for origin device <pubkey>".
+func (e *ProgramError) ProgramLogMessages() []string {
+	var msgs []string
+	for _, line := range e.Logs {
+		msg, ok := strings.CutPrefix(line, "Program log: ")
```
nikw9944: This is an allowlist where the doc comment describes a denylist: only `Program log: `-prefixed lines survive, but native programs log unprefixed via `ic_msg!`, so a CPI's reason line (`Transfer: insufficient lamports 0, need 890880`) is dropped — as is the `Program <id> failed: …` line the fixture at `executor_test.go:306` models.

controlplane/telemetry/internal/telemetry/submitter.go
```diff
@@ -247,6 +247,19 @@ func (s *Submitter) Tick(ctx context.Context) {
...
 				}
 
+				// A rejection by the program is not transient: the ledger executed the instruction
+				// and refused it, so every attempt this tick would be refused the same way. Report
+				// it once at Error and leave the rest of the attempts unspent, rather than burying
+				// the reason under a backoff loop. The samples are requeued as with any other
+				// failure, so the next tick retries once the operator fixes what was wrong.
+				var programErr *telemetry.ProgramError
+				if errors.As(err, &programErr) {
+					metrics.Errors.WithLabelValues(metrics.ErrorTypeSubmitterProgramError).Inc()
```
nikw9944, line 269: A rejected init increments `submitter_failed_to_initialize_account` at `:150` and `submitter_program_error` here, but `metrics.go:47` calls the new type "Distinct from the write/init failure types."

controlplane/telemetry/internal/telemetry/submitter_test.go
```diff
@@ -1211,4 +1211,66 @@ func TestAgentTelemetry_Submitter(t *testing.T) {
...
+
+		s.Tick(context.Background())
+
+		assert.Equal(t, int32(1), atomic.LoadInt32(&writes), "the rejection should end the tick, not spend the remaining attempts")
+		assert.Equal(t, int32(1), atomic.LoadInt32(&inits), "an init the program rejected should not be re-sent unchanged")
+
+		programErrs := testutil.ToFloat64(metrics.Errors.WithLabelValues(metrics.ErrorTypeSubmitterProgramError)) - programErrsBefore
+		assert.Equal(t, float64(1), programErrs, "program-error counter should increment once")
+
+		exhausted := testutil.ToFloat64(metrics.Errors.WithLabelValues(metrics.ErrorTypeSubmitterRetriesExhausted)) - exhaustedBefore
```
nikw9944: This delta-0 assertion on `submitter_retries_exhausted` races `retention_test.go:57`, which increments that label concurrently — the comment at `:1213-1215` only rules out sibling subtests, and this parent calls `t.Parallel()` at `:26`.

nikw9944 (changes requested):
The SDK change opens an onchain duplicate-write path in the internet-latency-collector that the "not in scope" note doesn't account for.

**Must fix** — `controlplane/internet-latency-collector/internal/exporter/submitter.go:241` re-sends the whole slice from index 0 each attempt, because its `SubmitSamples` returns no written count, unlike the device submitter (`controlplane/telemetry/internal/telemetry/submitter.go:87-90`).

**Worth addressing** — `ProgramError.Err` has no accessor, so custom codes on the finalized path skip the sentinel mapping at `client.go:304-330`; and `ProgramLogMessages` keeps only `Program log: `-prefixed lines, dropping unprefixed CPI reason lines.

The core executor change is right, and the `written` accounting survives the new early break — I traced both.

- **High — the SDK change turns silent loss into onchain duplicate writes in the internet-latency-collector.** `controlplane/internet-latency-collector/internal/exporter/submitter.go:241`. That retry loop re-sends the whole slice from index 0 each attempt, because its `SubmitSamples` returns no written count — unlike the device submitter, which threads `written` for exactly this reason (`controlplane/telemetry/internal/telemetry/submitter.go:87-90`).

smartcontract/sdk/go/telemetry/executor.go
elitegreg: Accessor added in 911d56de: `ProgramError.CustomErrorCode()`, with `sentinelForProgramError` mapping 1006 → `ErrSamplesAccountFull` and 1011 → `ErrAccountNotFound` in both write methods, so a rejection that reaches execution returns what its preflight twin returns. Table-tested across the numeric types a JSON decoder can hand back (`json.Number`, `float64`, `int`, `uint64`) plus the shapes carrying no code, a negative code, and one past `uint32`.

The init methods stay unmapped deliberately — 1010 is handled by the write-anyway path below rather than a sentinel.

On the `AccountAlreadyExists` half: that was fixed in 3393f70d, pushed shortly before this review, so it may not have been in the diff you were reading. The write now runs whether or not the init was accepted, and only a write that still finds nothing there reports the init failure as the reason. `Tick` no longer breaks on an account that already exists, and recovery is same-attempt again rather than next-tick. Covered by `writes_anyway_when_the_account_already_exists`.

smartcontract/sdk/go/telemetry/executor.go
elitegreg: Correct, and the consequence is worse than the doc mismatch. `initialize_device_latency_samples` CPIs the system program to create the account, so an agent that cannot fund it produces `Transfer: insufficient lamports 0, need 890880` — unprefixed, therefore dropped, leaving an opaque `Custom:1`. That is the same "nothing names the cause" failure this PR exists to remove.

Inverted to the denylist the comment described, in 911d56de: the runtime's own `Program <id> …` lines and the `Program log: Instruction:` echo are dropped, everything else is kept. `TestSDK_Telemetry_ProgramError_ProgramLogMessages` pins it with a system-program CPI failure and asserts the reason survives into `Error()` while the invoke/consumed/failed lines do not.

controlplane/telemetry/internal/telemetry/submitter.go
elitegreg: Fair — the comment claimed exclusivity it does not have. Reworded in 911d56de: the two overlap by design, the write/init types naming the operation that failed and this one narrowing *why* to a rejection that will recur until something changes onchain or in config, as opposed to the transient RPC trouble they also cover. The comment now says a rejected init increments both.

Kept the double count rather than suppressing one: dropping `submitter_failed_to_initialize_account` would lose which operation failed, and dropping the new one would lose the distinction from transient failures. `submitter_failed_to_initialize_account` did move, though — it now fires on the write's verdict rather than on an init failure the write goes on to absorb, so a harmless 1010 no longer trips it.

controlplane/telemetry/internal/telemetry/submitter_test.go
elitegreg: Confirmed and fixed in 911d56de. `TestSubmitter_RetainsEverySampleAcrossTheStalenessBound` drives `refusingProgramClient`, which fails every write at `MaxAttempts: 2`, so it increments that label throughout its run — and the parent here is parallel, so a zero delta was racy exactly as you describe.

Replaced with `assert.NotContains(out, "Submission failed after all retries")`: same guarantee, no shared state. The `submitter_program_error` delta stays, since nothing else in the binary touches that label, and the subtest remains non-parallel for it. Package passes `-race -count=2` with both tests in the same run.

elitegreg:
Thanks — all five addressed in 911d56de. Individual replies are on the threads; the must-fix and one framing correction are worth putting here.

**The duplicate-write path is fixed rather than deferred.** `SubmitSamples` in the collector now returns how many samples it wrote, `Tick` resumes at `tmp[written:]`, and only the remainder is requeued — the same shape the device submitter got in #4145. `retries_resume_at_the_first_unwritten_sample` covers it, and I checked it fails without the fix: a 262-sample partition whose first batch lands and second fails requeues all 262, so the next tick re-sends the 245 already onchain. With the fix, 17.

**One correction on the framing.** That duplicate path is already reachable on `main` without this PR — an RPC timeout mid-partition does the same thing, because both the retry and the requeue restart at index 0 regardless of what landed. Surfacing program rejections adds a trigger to a live bug rather than converting silent loss into duplication. The fix closes both triggers.

**And one stale finding.** The `AccountAlreadyExists` half of the `Err any` comment was fixed in 3393f70d, pushed shortly before the review landed, so it likely was not in the diff being read. The write now runs whether or not the init was accepted, so 1010 recovery is same-attempt again rather than next-tick. The accessor half of that comment was valid and is done.

Still out of scope in the collector, now tracked as malbeclabs/infra#2181: it spends its attempts retrying a permanent rejection, and returns early on a harmless init rejection instead of writing anyway. Its own retry loop absorbs both within the tick, so the cost is wasted attempts, not delay or duplication.

Verification: 228 non-test lines, all affected packages green, lint clean, both e2e SDK telemetry tests run locally against cEOS. The only failing test in the tree is the pre-existing `controlplane/telemetry/internal/netns` privilege case, which fails identically on a clean `main` checkout.

controlplane/telemetry/internal/telemetry/submitter.go
```diff
@@ -146,16 +146,28 @@ func (s *Submitter) SubmitSamples(ctx context.Context, partitionKey PartitionKey
...
+					// write goes on to absorb.
+					log.Warn("Failed to initialize account, attempting the write anyway", "error", initErr)
 				}
 				_, _, err = s.cfg.ProgramClient.WriteDeviceLatencySamples(ctx, writeConfig)
 				if err != nil {
 					if errors.Is(err, telemetry.ErrSamplesAccountFull) {
 						s.handleAccountFull(log, partitionKey, len(samples)-i)
 						return i, nil
 					}
+					if initErr != nil {
```
nikw9944, line 164: The comment says "the account is still not there," but this branch doesn't check that — it fires for any `err`, not just `ErrAccountNotFound`, and `err` is discarded at `:169`. Gate on `errors.Is(err, telemetry.ErrAccountNotFound)`.

`911d56de` narrows it — 1006 and 1011 now map to sentinels, so account-full and missing-account are handled correctly — but a write rejected for any other reason (`EpochMismatch`, `UnauthorizedAgent`) still surfaces as whatever the init failed with. Worse when the init failed transiently: `Tick`'s program-error branch then sees a plain RPC error and spends the tick's remaining attempts on a write the program keeps refusing.

nikw9944 (approved):
`911d56de` addresses all five findings, and I verified each in the code rather than taking the replies at face value: the collector now threads a written count and resumes at `tmp[written:]`, `ProgramError.CustomErrorCode()` plus `sentinelForProgramError` map 1006/1011 on the finalized path, the log filter is inverted to a denylist with a system-program CPI test, the metric comment states the overlap, and the racy counter delta is now a log assertion. Build, the three affected suites, and lint all pass at head.

One thing left, not previously reported: `controlplane/telemetry/internal/telemetry/submitter.go:164` fires for any post-init write failure, not just `ErrAccountNotFound`, so an unrelated write rejection is reported as an init failure and its own reason is dropped.

Also a correction on my side: my "turns silent loss into duplication" framing overstated the novelty. The duplication was already reachable on `main` via an RPC timeout mid-partition, since both the retry and the requeue restarted at index 0. The fix closes both triggers.

Issue #4153, opened 2026-08-05
internet-latency-collector: dead RIPE Atlas source probes are never marked unresponsive


bgm-malbeclabs:
## Summary

The RIPE Atlas collector cannot detect a source probe that stops returning results while the rest of its measurement stays healthy. Circuits whose remote probe has gone dead stay missing indefinitely, and the hourly measurement-creation cycle reports nothing — it has no signal to act on. Recovery today requires restarting the process so that a fresh measurement set is created.

## Evidence: mainnet-beta, 2026-08-05

RIPE Atlas results across all 29 target locations froze simultaneously at 13:04 UTC. Most measurements resumed by 13:32, but ~125 of 435 circuits stayed missing for the next 3.5 hours, at a dead-flat plateau:

```
13:32  missing 125 / actual 310
14:12  missing 131 / actual 304
15:12  missing 122 / actual 313
16:12  missing 133 / actual 302
16:52  missing 134 / actual 301
```

The missing set clusters by **source probe location** — sea (16 circuits), slc (15), pit (14), osl (14), nyc (14), mad (14), sqq (10), hkg (7). `wheresitup` was unaffected for the entire window (435/435).

Measurement-creation cycles ran and completed normally at 14:12, 15:12 and 16:12. Not one `Marking probe as unresponsive` or `Marking source probe as unresponsive` line was emitted across the whole incident. The plateau only broke when a deploy restarted the process at 16:53, which created a new measurement set — and even then coverage recovered only to ~350/435, leaving a residue of the same shape.

## Root cause

Two detectors run in `RunRipeAtlasMeasurementCreation`, and neither can see this failure.

**`collector.go:771-799` — target-level, wrong granularity.** The staleness test reads `meta.LastExportAt`, which advances whenever *any* source probe in that measurement returns a result. The dead probes were a subset of each measurement's sources, so results kept arriving from the others, `UpdateTimestamp` kept firing, and `LastExportAt` never aged past the threshold. A partial failure inside a measurement is invisible here by construction.

**`collector.go:801-841` — source-level, but gated shut.** This is the check that should have caught it. Two guards prevent it:

- `:817` — `meta.CreatedAt == 0 || meta.CreatedAt >= probeTimeout` skips measurements created less than an hour ago, or carrying no `created_at`.
- `:823` — `source.LastResponseAt == 0` skips probes that have never once responded.

[...]

PR #4154, opened 2026-08-05
internet-latency-collector: stop re-exporting the boundary RIPE Atlas result

bgm-malbeclabs:
## Summary of Changes

- RIPE's `?start=` filter is inclusive, and the collector passed it the newest timestamp it had already consumed. Every poll re-fetched that boundary result and re-exported it, appending a duplicate sample to the circuit's onchain account. `?start=` now asks for the following second, and the parameter is documented as exclusive.
- Because a sample's time is derived as `start_timestamp + index × sampling_interval`, the extra samples made derived times outrun wall clock: **403 of 435 mainnet pairs drifted more than an hour within epoch 196, worst case +43.7h**, leaving consecutive epochs overlapping by days and any `event_ts`-keyed view unreliable. `wheresitup` was unaffected (0/435), which is what isolated the defect to the RIPE path.
- The `timestamp not updated (old results?)` warning only fired *because* duplicates were being exported, so this fix would have silenced it — and that warning is what made the 2026-08-05 mainnet stall diagnosable. The stall signal moves to the empty-result path, firing once a measurement returns nothing for longer than 30 minutes (several times the default 600s sampling interval, so ordinary polling jitter stays quiet).

Existing accounts keep their already-inflated sample counts; this stops the drift accumulating from here forward. No migration.

## Diff Breakdown

| Core logic   |     1 | +8 / -1     |   +7 |
| Scaffolding  |     1 | +21 / -4    |  +17 |
| Tests        |     1 | +5 / -3     |   +2 |
| **Total**    |     4 | +36 / -8    |  +28 |

A one-line semantic fix to the fetch cursor, plus an observability branch so the fix does not remove an operator signal.

<details>

- `controlplane/internet-latency-collector/internal/ripeatlas/collector.go` — warn instead of Debug when a measurement returns no new results for over `staleMeasurementWarnAfter` (30m); guarded on having exported at least once, so a fresh measurement does not warn off a zero timestamp
- `controlplane/internet-latency-collector/internal/ripeatlas/client.go` — `?start=` becomes exclusive (`startTimestamp+1`), with a doc comment stating the contract and why
- `controlplane/internet-latency-collector/internal/ripeatlas/client_test.go` — the existing assertion encoded the inclusive behavior; updated to require exclusivity

</details>

## Testing Verification

- The pre-existing test asserted `start == startTimestamp`, i.e. it encoded the bug. Confirmed the updated assertion actually guards the fix: stashing only `client.go` fails `TestInternetLatency_RIPEAtlas_GetMeasurementResultsIncremental` on two subtests with `start must exclude the last consumed timestamp`, and passes with the change restored.
- Quantified the defect against production data before and after diagnosis: per-pair `max(event_ts - ingested_at)` on mainnet showed 403/435 pairs above 1h in epoch 196 versus 0/435 for `wheresitup`, and a worked example (`ymq↔yto` reaching sample_index 527 in 44h against `ams↔fra`'s 252 over the same window) confirmed the ~2x sample rate against the declared 600s interval.
- The new stall warning is not unit-tested — the package has no log-capture harness, and adding one for a single log line was not worth the fixtures. Its condition is a timestamp comparison on a path already covered by the empty-results tests.

## Related

Investigated alongside #4153 (unscheduled RIPE Atlas enlistments silently killing circuits). That is a separate defect in the same export path and is **not** fixed here.


controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -557,10 +562,22 @@ func (c *Collector) exportSingleMeasurementResults(ctx context.Context, measurem
...
 	if len(results) == 0 {
-		c.log.Debug("No new results for measurement",
-			slog.Int("measurement_id", measurement.ID),
-			slog.String("target_location", meta.TargetLocation),
-			slog.Int64("query_start_timestamp", lastTimestampUnix))
+		// An empty page is normal between cycles, but once it persists well past the
+		// sampling interval the measurement has stalled and its circuits are going
+		// missing. This is the only per-measurement signal that the stall is happening;
+		// marking the probes responsible is handled by the measurement creation cycle.
+		if exists && time.Since(lastTimestamp) > staleMeasurementWarnAfter {
```
ben-dz: This warn is guarded on `len(results) == 0`, but a stalled RIPE measurement normally returns a **non-empty** page: when the target probe goes dark the source probes keep pinging and keep uploading results whose `result` array is all `{"x":"*"}` with no `rtt`.

Trace with the fix applied, target dark, cursor at `T`:
- `?start=T+1` returns the timeout results, so `len(results) > 0` and this branch is skipped.
- `parseLatencyFromResult` (collector.go:203) returns `latency == 0` for each, so the `if latency > 0` gate at :617 never fires — `records` stays empty and `maxTimestamp` stays the zero time.
- :635 `maxTimestamp.After(lastTimestamp)` is false, and :645 `len(records) > 0` is false. Nothing above Debug is emitted about the stall.

That is exactly the case the deleted signal covered. Under the old inclusive `?start=T`, RIPE re-returned the boundary result — by construction a *successful* one, since `maxTimestamp` only advances inside `latency > 0` — giving `len(records) == 1` with `maxTimestamp == lastTimestamp`, which fell into :645 and fired `Exported records but timestamp not updated (old results?)`. The old warn required `len(results) > 0` and the new one requires `len(results) == 0`, so the two are complementary rather than overlapping: this moves the signal to the other stall mode instead of preserving it. Per the PR description the 2026-08-05 stall was diagnosed off that warning, so it was this mode.

Minimal fix — reattach the check to "no new samples" instead of "no results", replacing :645:

```go
if maxTimestamp.After(lastTimestamp) {
    ...unchanged...
} else if exists && time.Since(lastTimestamp) > staleMeasurementWarnAfter {
    c.log.Warn("No new samples since last export (measurement stalled?)",
        slog.Int("measurement_id", measurement.ID),
        slog.String("target_location", meta.TargetLocation),
        slog.Time("last_timestamp", lastTimestamp),
        slog.Duration("stale_for", time.Since(lastTimestamp)),
        slog.Int("raw_results", len(results)))
}
```

Then the `len(results) == 0` early return keeps its plain Debug line.

On testability: `MockClient.GetMeasurementResultsIncrementalFunc` already receives `startTimestamp` (collector_test.go:29) and every collector test builds `&Collector{client: mockClient, log: log}` directly (e.g. collector_test.go:260), so a `slog.NewJSONHandler` over a `bytes.Buffer` plus a table over (empty page + stale), (all-timeout results + stale), (fresh results) is a short addition on existing fixtures. The all-timeout row fails today.

controlplane/internet-latency-collector/internal/ripeatlas/client.go
```diff
@@ -338,11 +338,18 @@ func (c *Client) GetAllMeasurements(ctx context.Context, env string) ([]Measurem
...
+// RIPE's ?start= filter is inclusive, so we ask for the following second. Passing the
+// caller's timestamp verbatim returns the boundary result on every poll, and re-exporting
+// it appends a duplicate sample each cycle, which makes the recorded sample count outrun
+// the declared sampling interval.
 func (c *Client) GetMeasurementResultsIncremental(ctx context.Context, measurementID int, startTimestamp int64) ([]any, error) {
 	endpoint := fmt.Sprintf("/measurements/%d/results/", measurementID)
 
 	if startTimestamp > 0 {
-		endpoint = fmt.Sprintf("%s?start=%d", endpoint, startTimestamp)
+		endpoint = fmt.Sprintf("%s?start=%d", endpoint, startTimestamp+1)
```
ben-dz, line 355: `maxTimestamp` is the max across all source probes in the page (collector.go:618), and probes routinely stamp the same whole second. If probe A's result at `T` is visible when the poll runs and probe B's result at `T` uploads afterwards, `?start=T+1` drops B permanently; inclusive `start=T` would have caught it. The window is one second wide and strictly narrower than the pre-existing loss (results below the max were already dropped either way), so the trade against a duplicate on 100% of polls is clearly favorable — not blocking. The doc comment above presents exclusivity as costless though; add a clause naming what it gives up so a later sample-count dip isn't re-diagnosed from scratch.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -557,10 +562,22 @@ func (c *Collector) exportSingleMeasurementResults(ctx context.Context, measurem
 	}
 
 	if len(results) == 0 {
-		c.log.Debug("No new results for measurement",
-			slog.Int("measurement_id", measurement.ID),
-			slog.String("target_location", meta.TargetLocation),
-			slog.Int64("query_start_timestamp", lastTimestampUnix))
+		// An empty page is normal between cycles, but once it persists well past the
+		// sampling interval the measurement has stalled and its circuits are going
+		// missing. This is the only per-measurement signal that the stall is happening;
```
ben-dz: "This is the only per-measurement signal that the stall is happening" isn't accurate — collector.go:806 already emits a per-measurement `Warn` on the same staleness condition (`meta.LastExportAt < probeTimeout`, 1h) during the creation cycle. This warn is a 30-minutes-earlier version of it, not the only one. Drop the clause.

ben-dz (changes requested):
The `+1` cursor fix is correct — `?start=` is inclusive on the result `timestamp` field, which is exactly what `UpdateTimestamp` persists, and one duplicate record does become one permanent extra onchain sample (`exporter/submitter.go:115` writes `rtts []uint32` against a single start timestamp, no dedup downstream). Cold start, the `exists` guard, and the wheresitup path all check out.

What must change is the replacement stall signal. The new warn is inside `if len(results) == 0`, but the stall mode the removed warning actually caught returns a **non-empty** page — when the target probe goes dark, source probes keep uploading all-timeout results, which yield zero records and a zero `maxTimestamp`, so neither the new branch nor the one at collector.go:645 fires. The old inclusive `?start=` re-returned the successful boundary result, which is precisely what made that path log. The two conditions are complementary, not overlapping, so the diff swaps which stall mode is covered rather than preserving coverage — and per the PR's own account, the 2026-08-05 incident was the mode that is now silent. Reattaching the check to "no new samples" instead of "no results" covers both.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
ben-dz: Fixed in 3df2bc343 (tack-on commit, pushed while you were asleep — squash or amend as you like).

Keyed the check off the export cursor failing to advance rather than off the page being empty, which is the condition both stall modes share:

```go
// A stall shows up two ways: an empty page, and a page whose results are all
// timeouts, which parse to zero valid latencies and so export nothing either. Both
// leave the cursor where it was, so key the signal off that rather than off the page.
stalled := exists && time.Since(lastTimestamp) > staleMeasurementWarnAfter
```

`stalled` is computed once at the top and consumed at two sites: the `len(results) == 0` early return, and a new `else if stalled` after the existing `Exported records but timestamp not updated` branch. Both call a `warnStalled` helper, which carries `raw_results` so the modes stay distinguishable in the log — `0` means nothing uploaded, `>0` means uploads with no usable rtt.

Test added along the lines you sketched: `TestInternetLatency_RIPEAtlas_ExportSingleMeasurementResults_StallWarning`, a `slog.NewJSONHandler` over a `bytes.Buffer` at `LevelWarn`, four rows — empty page + stale, all-timeout page + stale, empty page + recent export, and fresh samples after a long gap (the cursor advances, so it must stay quiet). Confirmed the all-timeout row is a real guard rather than decoration: with only `collector.go` reverted to 9d9c118 it fails on an empty log while the other three pass.

CHANGELOG updated too — it described the signal as moving to the empty-result path, which is no longer what it does.

controlplane/internet-latency-collector/internal/ripeatlas/client.go
ben-dz: Added in 3df2bc343:

```go
// The cost is a one-second blind spot: a probe stamping that same second but uploading
// after the caller's last poll is skipped rather than deduplicated.
```

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
ben-dz: Dropped in 3df2bc343. The comment moved onto the new `warnStalled` helper and now names the relationship to the creation-cycle warn instead of claiming exclusivity:

```go
// warnStalled reports a measurement that has produced no new sample for
// staleMeasurementWarnAfter, meaning its circuits are going missing. The measurement
// creation cycle marks the probes responsible, an hour in.
```

ben-dz (approved):
lgtm

No new issues in the follow-up commit: `stalled` is snapshotted before the fetch so the later cursor update can't perturb it, the if/else chain leaves no silent path through a stalled measurement, `raw_results` on the warn distinguishes the two stall modes, and the CHANGELOG was updated to match. One operational note — no check runs have been dispatched against `3df2bc34` yet, so the green checks on record are the parent commit's.

nikw9944 (approved):
One follow-up, no blockers. The exclusive-cursor fix is correct and the stall signal now covers both stall modes.

The "missing samples" alarm has been permanently silent and this fix turns it back on. Worth a CHANGELOG clause so the step change on deploy isn't read as a regression this PR caused.

- **The "missing samples" alarm has been permanently silent and this fix turns it back on.** The duplicate that this PR removes was always a successful result, so every circuit exported at least one record every cycle and the `Missing` counter at `controlplane/internet-latency-collector/internal/ripeatlas/collector.go:512` could never fire. It can now, and with one probe per circuit and the export and sampling intervals both at 600s (`cmd/collector/main.go:32,33,35`), poll jitter will trip it intermittently.

Nothing to change in the code — the counter is finally accurate. Name it in the CHANGELOG so the step change on deploy isn't read as a regression this PR caused.

nikw9944 (approved):
The merge from main leaves the PR's own code unchanged, and the one follow-up from the last round is still open.

The "missing samples" alarm has been permanently silent and this fix turns it back on. Worth a CHANGELOG clause so the step change on deploy isn't read as a regression this PR caused.

- **The "missing samples" alarm has been permanently silent and this fix turns it back on.** The duplicate that this PR removes was always a successful result, so every circuit exported at least one record every cycle and the `Missing` counter at `controlplane/internet-latency-collector/internal/ripeatlas/collector.go:512` could never fire. It can now, and with one probe per circuit and the export and sampling intervals both at 600s (`cmd/collector/main.go:32,33,35`), poll jitter will trip it intermittently.

Nothing to change in the code — the counter is finally accurate. Name it in the CHANGELOG so the step change on deploy isn't read as a regression this PR caused.

PR #4155, opened 2026-08-05
internet-latency-collector: report RIPE Atlas enlistments that never get scheduled

bgm-malbeclabs:
> **Stacked on #4154** — base is `internet-latency-collector-exclusive-start`, not `main`. The dependency is textual (both touch the same const block and CHANGELOG bullet), not logical. Merge #4154 first and this retargets to `main` cleanly.

Refs #4153.

## Summary of Changes

- A source probe can produce no successful sample for hours and nothing detects it. On 2026-08-05, 41 of 435 mainnet enlistments were in that state and were found only by reading the state file by hand. It has since recurred twice more — hkg (6 circuits) and dub (4) — both found the same way.
- The signal already existed and nothing read it: a source's `last_response_at` is zeroed whenever its measurement is created and is only ever advanced by a sample with a latency above zero, so a zero on an older measurement means that probe has contributed nothing to it. Because a circuit is enlisted **exactly once** (a measurement for target `T` enlists only sources sorting after `T`, totalling 29+28+…+0 = 435), each such source darks its circuit with no other trace.
- The hourly measurement cycle now tallies these by source location, exposes `doublezero_internet_latency_collector_ripeatlas_sources_without_samples{source_location}`, and warns with a sample capped at 20 identifiers. Both are silent when the count is zero.

**The 2h grace period is its own constant, not the 1h `probeTimeout`.** RIPE dispatch is not immediate — an accepted enlistment has been observed producing its first result 80–100 minutes after creation, with all of one probe's enlistments starting within seconds of each other — so a 1h window reports probes that are merely warming up. Keeping it separate from `probeTimeout` also means tuning this reporting window cannot perturb the marking path.

**It names no cause, deliberately.** `last_response_at` only advances on a sample with `latency > 0`, and `parseLatencyFromResult` returns 0 for a ping array carrying no `rtt`, so an enlistment RIPE never dispatched and a path at total packet loss are indistinguishable in this state. The metric, log line, and comments say what is measured — "no successful sample since the measurement was created" — not why.

**Observation only.** It changes no state, marks no probes, and cannot trigger recreation. Recreation invalidates every measurement whose target sorts before that metro: rotating two probes on 2026-08-05 recreated 13 measurements covering 299 of 435 circuits and dropped coverage to 136/435 for a warm-up cycle, and every drop observed since has recovered unaided anyway.

This also does **not** seed `last_response_at` at creation, which was the original suggestion in #4153. Seeding would silently enable the existing rotation path at `collector.go:801`, since the zero-guard at `:823` is the only thing holding it back.

Applied to the 08-05 batch created at 13:12, this would have fired at the 15:12 cycle naming `hkg`, roughly 90 minutes before the metros were identified by hand. The 17:54 cycle would have stayed quiet: the sources flagged there (ymq, sjc, sqq) were inside dispatch latency and started reporting on their own at 18:13.

## Diff Breakdown

| Tests        |     1 | +135 / -0   | +135 |
| Core logic   |     1 | +82 / -0    |  +82 |
| Scaffolding  |     1 | +9 / -0     |   +9 |
| **Total**    |     4 | +227 / -0   | +227 |

Nearly two thirds of the diff is tests; the logic is one pure function and its caller.

<details>

- `controlplane/internet-latency-collector/internal/ripeatlas/collector_test.go` — table test over a synthetic `MeasurementState`, plus a sample-cap test
- `controlplane/internet-latency-collector/internal/ripeatlas/collector.go` — new pure `sourcesWithoutSamples` helper, the `sourceSampleGracePeriod` constant, and the call site as step 4c of the measurement cycle
- `controlplane/internet-latency-collector/internal/metrics/metrics.go` — the new `GaugeVec`

</details>

## Testing Verification

- Table test covers the states that matter: a measurement past the grace period with a silent source (counted), all sources responding (silent), a measurement inside the 80–100 minute dispatch window (silent), a measurement 60s old (silent), and `created_at == 0` (skipped rather than assumed old). Plus per-location accumulation across measurements, a measurement with no metadata, and the 20-item sample cap against 30 silent sources.
- The fixtures are anchored on `sourceSampleGracePeriod` rather than a literal, so they straddle the real boundary. Tightening the constant from 2h to 1h fails exactly one subtest — `measurement inside observed RIPE dispatch latency is not reported` — and nothing else, which is the regression that matters here.
- Separately confirmed the age guards are load-bearing: replacing `if !hasMeta || meta.CreatedAt == 0 || meta.CreatedAt >= createdBefore` with `if !hasMeta` fails the two warm-up subtests and the unknown-creation-time subtest.
- The gauge `Reset()` before each cycle's writes means a recovered location goes absent rather than holding its last value. The `WARN` line itself is not asserted — the package has no log-capture harness and one line did not justify building one.
- Extracted the tally as a pure function rather than testing through `RunRipeAtlasMeasurementCreation`, which needs a live client and location fetch.


nikw9944 (commented):
Reviewed against the host logs and the state file rather than the code alone. Detail and method are in https://github.com/malbeclabs/doublezero/issues/4153#issuecomment-5206952243.

The observation-only decision is right, and now more clearly right than the rationale here argues. The scope, the pure helper, the `Reset()` before writing, and aggregating by `source_location` are all good. Aggregating by source location in particular is better than the rationale claims: the drops are per probe and near all-or-nothing, so source location is exactly the right granularity.

One thing should change before this merges, plus two text corrections.

## The 1h window is shorter than dispatch latency

`unscheduledEnlistments` is called with `probeTimeout`, the existing `currentTime - 3600`. Measuring the 2026-08-05 batches against RIPE, an enlistment that RIPE has accepted starts producing results **80 to 100 minutes** after the measurement is created, per probe, with all of that probe's enlistments starting within seconds of each other.

Replaying this code against the 18:04 state snapshot, the cycle at 17:54:11 would have fired on the 41 zero-`last_response_at` sources:

| label | count | what it actually was |
|---|---|---|
| muc | 11 | indeterminate, measurement stopped at 72 min |
| sqq | 10 | 9 started at 18:13, 19 minutes after the warning |
| **ymq** | **10** | **all 10 started at 18:13** |
| hkg | 6 | indeterminate, measurement stopped at 72 min |
| lon | 3 | indeterminate, measurement stopped at 72 min |
| **sjc** | **1** | **started at 18:13** |

Checking every one of the 41 against the RIPE results API: 20 provably started on their own, and the other 21 are all in measurements that were stopped at 18:05 while 72 minutes old, which is before the 80-minute mark at which the other 20 came alive. None of the 41 is confirmed as permanently unscheduled.

Same replay on the 18:05 batch at the 19:05:52 cycle gives 18 flagged. 6982's six started at 19:45:30-19:45:42, exactly 100 minutes after creation, so 6 of 18 were false. 6750's twelve stayed dark past 141 minutes, so those are the genuine signal, and a wider window still catches them.

The cost of this is not a bad state transition, since nothing is marked. It is that the metric exists to drive a human, and on 08-05 the action it drove was the probe rotation plus a hand-edit of the state file that silently broke state persistence for 100 minutes (`sudo mv` left the file root-owned; the service runs as `doublezero`). A gauge that is roughly half false at its threshold will drive more of that.

Suggest giving this its own constant rather than borrowing `probeTimeout`:

```go
// unscheduledEnlistmentAfter is how long after creation an enlistment must stay silent
// before we treat it as unscheduled. RIPE dispatch is not immediate: an accepted
// enlistment has been observed starting 80-100 minutes after the measurement was
// created, so a shorter window reports probes that are merely warming up.
unscheduledEnlistmentAfter = 2 * time.Hour
```

That drops every false positive measured above, keeps the 13:12-batch case (121 enlistments dark across a full 3h41m life) well inside detection range, and decouples this from the marking path so tuning one never perturbs the other.

## It cannot distinguish "never dispatched" from "100% packet loss"

`LastResponseAt` is only set when `latency > 0` (`collector.go:581`), because `parseLatencyFromResult` returns 0 for a result whose ping array holds no `rtt`. A source that RIPE is dispatching normally but whose path is at total loss therefore holds a zero forever and lands in this gauge. The metric help text and the log line, "Enlistments never scheduled by RIPE Atlas, their circuits are dark", assert a cause the code cannot distinguish.

Not theoretical. The same discard is what made the collector mark waw's probe 7447 unresponsive at 12:05:53 today while it was answering on schedule. 108 results, no gap over 30 minutes, last one 16 seconds before the marking, but the last five were `rcvd=0 avg=-1` so the cursor froze. That marking recreated 28 measurements. This gauge would catch the same class of event and label it as a RIPE scheduling failure.

Either land the loss-discard fix first, or reword to what is actually measured, for example "sources with no successful sample since creation". Happy to open the loss fix as a small PR if that is useful, since it is a dependency for both this and the detector work in #4153.

## Two premises are disproved and appear in three places

These are in the commit message, the CHANGELOG entry, and the step 4c comment, so they will read later as settled:

- "a replacement probe inherits the same unscheduled enlistment, verified in production" — 6982 was that replacement. It started reporting in all six of its enlistments at 19:45:30-19:45:42, 100 minutes after creation. 6750 has since recovered as well. Every probe in this thread that looked permanently unscheduled is now reporting.
- "for some source-metro → target combinations" — across the 13:12, 16:53 and 18:05 batches the drops are per probe, near all-or-nothing. The per-pair evidence in #4153 was a single 1-of-22 outlier.

The conclusion they support still stands. Rotation should not be automated. The reason is that it is unnecessary and expensive, not that it is ineffective.

Also worth reconsidering the "Applied to the 2026-08-05 incident, this would have fired at the 14:12 cycle with `hkg=6`" claim in the description. True, and that batch is the genuine case, but with the 1h window it would also have fired at 17:54 naming ymq and sjc, which were healthy.

Not formally requesting changes, since the window is a one-constant fix and the rest is text.

bgm-malbeclabs:
Thanks — the dispatch-latency measurement is the piece I was missing, and it invalidates more of my writeup than the review claims. All three points taken.

**The window.** Added `sourceSampleGracePeriod = 2 * time.Hour` as its own constant, no longer borrowing `probeTimeout`, with the 80–100 minute observation and the decoupling rationale in the comment. The test fixtures now anchor on the constant rather than a literal, and there is a subtest at exactly 100 minutes: tightening the constant back to 1h fails that one subtest and nothing else.

**The causal claim.** Reworded throughout rather than deferring to the loss fix, since the code genuinely cannot distinguish the two:

- metric `..._ripeatlas_unscheduled_enlistments` → `..._ripeatlas_sources_without_samples`
- func `unscheduledEnlistments` → `sourcesWithoutSamples`
- log line → `Sources with no successful sample since their measurement was created, their circuits are dark`, now carrying `grace_period`

The `latency > 0` dependency is named explicitly in the function comment, including that a `rtt`-less ping array parses to zero. Please do open the loss-discard fix as its own PR — waw/7447 makes it a live correctness bug in the marking path, not just a labelling problem here, and it should not be gated behind this.

**The two premises.** Removed from the commit message, the CHANGELOG entry, and the step 4c comment. The remaining argument against rotation is that it is unnecessary and expensive, not that it is ineffective.

One data point that reinforces yours. Coverage as of ~00:20 UTC today, all-time last ingest per source probe:

| source | dark circuits | stopped | |
|---|---|---|---|
| muc | 0 | — | recovered unaided |
| hkg | 6 | 08-06 18:09:35 | all six in the same second |
| dub | 4 | 08-06 14:11:46 | new, not part of the 08-05 incident |

425/435 ripeatlas, 435/435 wheresitup. So it is not a fixed set of permanently-dark circuits, which is how I originally framed it — it is a recurring per-probe stop/restart that moves between metros, and per-probe all-or-nothing is exactly right. Both current cases (3h and 7h dark) clear the 2h threshold, so the constant holds on today's data too.

The runbook is getting the same corrections, plus two defects this turned up: its `unresponsive_probes` reset used a plain `sudo mv` from `/tmp`, which is what left the state file root-owned, and its `journalctl` grep only matched one of the two `Marking ... probe as unresponsive` lines, which is why we concluded nothing had been marked when 6888 had been.

PR #4156, opened 2026-08-06
qa: add retransmit metro test to qa.testnet

martinsander00:
## Summary of Changes
* This PR reworks the `TestQA_MulticastSettlement`.
* Deleted `e2e/qa_shred_settlement_test.go.` since it only contained one setup method that I moved into the `TestQA_MulticastSettlement` file.
* It checks the env we are in, if mainnet, continue as normal, if testnet and program config retransmit flag off, continue as normal too. If testnet and program config retransmit flag on do the following:

1. Assert that when trying to buy a seat in a metro other than frankfurt fails
2. Try to buy a seat in frankfurt and assert that the user accesspass has access to these multicast groups: `edge-solana-retrans`

* The PR deletes `e2e/qa_retransmit_only_settlement_test.go`. `TestQA_RetransmitOnlySettlement` drove the same positive path, so the reworked test covers it now. Its device selector, its metro-code formatter and its group assertion moved into the multicast test file.
* The group codes now come from one flag, `-retransmit-group-codes`, which replaces `-leader-group-code` and `-retransmit-group-code`. The check requires the seat's groups to equal that list, so any extra group fails it.
* `sdk/shreds/go/state.go` gains `IsRetransmitOnlyOnboardingEnforced`, which reads bit 7 of the program config flags.
* `e2e/internal/qa/client_settlement.go` gains that same read, plus `ClosestNonRetransmitOnlyDevice` for the device the rejection step pays on.
* A companion change in `malbeclabs/infra` passes `-retransmit-group-codes=edge-solana-retrans` in `qa.testnet.yml` and deletes `qa.retransmits.testnet.yml`. Merge both together, or the hourly job runs the group check without the codes it needs.


e2e/qa_multicast_settlement_test.go
```diff
@@ -6,36 +6,749 @@ import (
...
+	}
+
+	prices, err := client.SeatPrices(ctx, device.PubKey)
+	require.NoError(t, err, "failed to read the onchain seat price for device %s", device.Code)
+	require.NotZero(t, prices.InstantAllocationDollars, "onchain instant-allocation price is zero for device %s", device.Code)
+	amount := strconv.FormatUint(prices.InstantAllocationDollars, 10)
+
+	log.Info("Paying for a seat outside a retransmit-only metro", "device", device.Code,
+		"metro", device.ExchangeCode, "amount", amount)
+	err = client.FeedSeatPay(ctx, device.PubKey, amount)
```
bgm-malbeclabs, line 59: This pays before `wait_for_open_phase`: `runShredSettlement` runs the preflight hook at line 367, and the open-for-requests wait only happens at line 488. For the grace-period slots at the end of every epoch the program is closed by design, so this `FeedSeatPay` fails for the phase reason, the `ErrorContains("Retransmit-only onboarding enforced")` below fails, and the hourly run pages — the exact false alarm the `EpochTailClosedWindow` classification in `wait_for_open_phase` exists to suppress, which never gets a chance to run because we already failed.

Either move the preflight to just after `wait_for_open_phase`, or classify a closed-phase rejection here and skip instead of asserting on the message.

sdk/shreds/go/state.go
```diff
@@ -69,9 +69,10 @@ type ProgramConfig struct {
 // Flag bits in ProgramConfig.Flags. Mirrors the onchain bit indices from the
 // shred-subscription program.
 const (
-	programConfigFlagIsPausedBit                 = 0
-	programConfigFlagIsMigratedBit               = 1
-	programConfigFlagIsProratedServiceEnabledBit = 2
+	programConfigFlagIsPausedBit                         = 0
+	programConfigFlagIsMigratedBit                       = 1
+	programConfigFlagIsProratedServiceEnabledBit         = 2
+	programConfigFlagRetransmitOnlyOnboardingEnforcedBit = 7
```
bgm-malbeclabs, line 75: Bit 7 has no source of truth in this repo (the shred-subscription program lives elsewhere, and no fixture covers `ProgramConfig.Flags` bit semantics), and the failure mode is silent: if the index is wrong, `IsRetransmitOnlyOnboardingEnforced` returns false, the preflight skips, the test settles on the closest device as before, and the run goes green having exercised none of the new coverage.

Could you confirm the index against the program's `FLAG_RETRANSMIT_ONLY_ONBOARDING_ENFORCED_BIT`, and add a table-driven test next to `TestMetroHistoryFlags` (state_test.go:289)? None of the other three bits are covered either, so one table over all four is the cheap version.

e2e/qa_multicast_settlement_test.go
```diff
@@ -6,36 +6,749 @@ import (
...
+	// preflight runs after the reconciler is enabled and before device selection.
+	preflightSubtestName string
+	preflight            func(t *testing.T, ctx context.Context, log *slog.Logger, test *qa.Test, client *qa.Client)
+
+	selectSubtestName string
+	// selectDevice may call t.Skip when the feature is not configured. It logs
+	// its own selection detail.
+	selectDevice func(t *testing.T, ctx context.Context, log *slog.Logger, test *qa.Test, client *qa.Client) *qa.Device
+
+	priceLogMsg string
```
bgm-malbeclabs: Deleting `TestQA_RetransmitOnlySettlement` also dropped the `assertPrice` hook and `--retransmit-price`, so nothing asserts any more that a retransmit-only seat is charged the discounted price. That is real coverage loss the PR description does not mention. Keeping the `assertPrice` field and wiring it when `onboardingEnforced` is about six lines.

e2e/qa_multicast_settlement_test.go
```diff
@@ -6,36 +6,749 @@ import (
...
 
-var enableSettlementTests = flag.Bool("enable-multicast-settlement-tests", false, "enable multicast settlement tests")
+const (
+	retransmitSubscribeTimeout = 90 * time.Second
+	balanceSettleTimeout       = 30 * time.Second
+)
+
+var (
+	enableSettlementTests    = flag.Bool("enable-multicast-settlement-tests", false, "enable multicast settlement tests")
+	retransmitOnlyDeviceFlag = flag.String("retransmit-only-device", "", "device code or pubkey in a retransmit-only metro (overrides auto-discovery)")
```
bgm-malbeclabs, line 28: `--enable-retransmit-only-settlement-tests`, `--leader-group-code` and `--retransmit-group-code` are gone/renamed. An undefined flag is a hard parse failure of the whole test binary, not a skip, so whatever invokes the QA suite in malbeclabs/infra needs a paired PR landing at the same time.

e2e/qa_multicast_settlement_test.go
```diff
@@ -6,36 +6,749 @@ import (
...
 
-// TestQA_MulticastSettlement pays for a full (leader) multicast seat on the
-// closest device and verifies pricing, tunnel-up, and the withdraw/refund
-// accounting. It is a thin wrapper over the shared settlement flow in
-// runShredSettlement; TestQA_RetransmitOnlySettlement mirrors it with
-// retransmit-only device selection and a group-subscription assertion. Any fix
-// to the settlement machinery belongs in runShredSettlement so both stay in
-// sync.
 func TestQA_MulticastSettlement(t *testing.T) {
+	var onboardingEnforced bool
```
bgm-malbeclabs: `shredSettlementParams` existed so two tests could share the machinery; with the second one deleted it has exactly one caller. The hook indirection plus this bool — assigned inside the preflight subtest and read from two other closures (lines 55 and 65) — is harder to follow than straight-line `if onboardingEnforced` in an inlined harness. Either inline it or keep the second test; the middle is the worst of the two.

Related edge: `-run 'TestQA_MulticastSettlement/select_device'` skips the preflight, leaving this false and silently flipping the run to the closest-device path.

e2e/qa_multicast_settlement_test.go
```diff
@@ -6,36 +6,749 @@ import (
...
+	selectDevice func(t *testing.T, ctx context.Context, log *slog.Logger, test *qa.Test, client *qa.Client) *qa.Device
+
+	priceLogMsg string
+
+	// extraAssertion runs after the tunnel is up and before the seat is withdrawn.
+	extraSubtestName string
+	extraAssertion   func(t *testing.T, ctx context.Context, log *slog.Logger, client *qa.Client, device *qa.Device)
+}
+
+// runShredSettlement picks a device, pays its seat price, checks the debit and
```
bgm-malbeclabs: The move stripped the rationale comments: `runShredSettlement`'s contract, the `assertPrice`/`extraAssertion` field docs, and the doc comments on `selectClosestDevice`, `assertNewSeatRejected`, `flaggedMetroCodes` and `assertSubscribedGroups`. Those carried the why (e.g. why the price assertion reads chain instead of the CLI quote), which is the part the code cannot say for itself.

e2e/internal/qa/client_settlement.go
```diff
@@ -190,6 +190,38 @@ func (c *Client) ClosestRetransmitOnlyDevice(ctx context.Context) (*Device, map[
 	return bestDevice, retransmitOnly, nil
 }
 
+func (c *Client) ClosestNonRetransmitOnlyDevice(ctx context.Context) (*Device, error) {
```
bgm-malbeclabs, line 177: This is `ClosestRetransmitOnlyDevice` with one negated predicate. One unexported helper taking a `want bool` covers both and drops ~40 lines.

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ All notable changes to this project will be documented in this file.
...
   - A failed submission now retries from the first unwritten sample rather than restarting at the beginning of the flushed partition, and only the unwritten remainder is requeued. Previously a mid-partition error made every subsequent attempt re-send batches that were already onchain, appending those samples a second time and pushing the account toward the sample cap it is measured against. (#4145)
   - Agent logs now identify the ledger RPC endpoint in use, and the peer count and the stale-program-data warning report state transitions instead of firing on every refresh. New lines name the resolved remote address of each connection (a bad load balancer address behind a hostname was invisible before), and a refresh that finds no peers is now a warning rather than a Debug line. The stale-cache warning also moves off the package-global `slog` onto the agent's own logger, so it is formatted and leveled with everything else. New: `doublezero_device_telemetry_agent_peers` gauge, and `pinger_epoch_fetch` on the errors counter for every exhausted epoch fetch. (#4147)
+QA
```
bgm-malbeclabs: Missing the `- ` prefix its siblings (`- CLI`, `- SDK`, `- Collector`) have, so this renders as a paragraph inside the previous list rather than a group heading.

bgm-malbeclabs (changes requested):

e2e/qa_multicast_settlement_test.go
```diff
@@ -6,36 +6,749 @@ import (
...
+			subscribed[group] = true
+			subs = append(subs, group.String())
+		}
+		var missing, extra []string
+		for group, code := range required {
+			if !subscribed[group] {
+				missing = append(missing, code)
+			}
+		}
+		for group := range subscribed {
```
nikw9944, line 186: This treats any pubkey in `user.Subscribers` outside `--retransmit-group-codes` as a violation, but `User.Subscribers` (`smartcontract/sdk/go/serviceability/state.go:974`) is the user account's complete multicast subscription list — not the set of groups the shred feed seat granted. A QA client that also subscribes to unrelated groups for other multicast tests will always carry extras, which is the `rocksteady` / `rebop-eu` failure in the PR description. Retransmit-only means the leader group is excluded, not that the user holds no other multicast memberships.

Compounding it: `GetServiceabilityUser` (`e2e/internal/qa/client.go:495-507`) returns the *first* user matching the client IP, so on a host with more than one user account this may not even be the seat's user.

The deleted `TestQA_RetransmitOnlySettlement` had the right shape. Either restore it — require each `--retransmit-group-codes` entry present and a `--leader-group-code` entry absent — or scope the "nothing else" check to the groups the seat's feed actually covers (`User.FeedPks`).

nikw9944 (changes requested):
One finding beyond what's already in review.

The "and nothing else" subscriber assertion compares against `User.Subscribers`, which is the user account's *complete* multicast subscription list — not the groups the shred feed seat granted. Any unrelated group the QA client holds trips it, which is exactly the `rocksteady` / `rebop-eu` failure in the PR description. The deleted test had this right: retransmit group present, leader group absent.

e2e/qa_multicast_settlement_test.go
martinsander00: https://github.com/malbeclabs/doublezero/commit/cb3ec334b63de285e61d1f2c9e3933164ea8c3c3

sdk/shreds/go/state.go
martinsander00: https://github.com/malbeclabs/doublezero/commit/cb3ec334b63de285e61d1f2c9e3933164ea8c3c3

e2e/qa_multicast_settlement_test.go
martinsander00: https://github.com/malbeclabs/doublezero/commit/cb3ec334b63de285e61d1f2c9e3933164ea8c3c3

e2e/qa_multicast_settlement_test.go
martinsander00: https://github.com/malbeclabs/doublezero/commit/cb3ec334b63de285e61d1f2c9e3933164ea8c3c3

e2e/qa_multicast_settlement_test.go
martinsander00: https://github.com/malbeclabs/doublezero/commit/cb3ec334b63de285e61d1f2c9e3933164ea8c3c3

e2e/internal/qa/client_settlement.go
martinsander00: https://github.com/malbeclabs/doublezero/commit/cb3ec334b63de285e61d1f2c9e3933164ea8c3c3

CHANGELOG.md
martinsander00: https://github.com/malbeclabs/doublezero/commit/cb3ec334b63de285e61d1f2c9e3933164ea8c3c3

e2e/qa_multicast_settlement_test.go
martinsander00: checking, ty!

martinsander00: good catch on the user type, fixed in b32b81511735cf7584df4bb5f783130c053429cf. the check reads the multicast user now.

the extras came from the seat though, not from another test. `rocksteady` is the priced feed in the testnet feed-config, and `exchange-multicast-groups.yaml` gives `rebop-eu` to `xfra`. the oracle reads the retransmit-only metros once per cycle and still held a snapshot from before we flagged frankfurt, so it never narrowed the set. after a restart the same assertion passed with `edge-solana-retrans` alone.

that's why i'd keep the exact set: naming `rocksteady` as the leader group would still have let `rebop-eu` through.

e2e/qa_multicast_settlement_test.go
```diff
@@ -6,36 +6,714 @@ import (
...
```
nikw9944, line 186: The test still fails if the seat carries any group beyond the ones named on the command line. That check is here: every subscriber outside `--retransmit-group-codes` is collected as `extra` on lines 171-174.

Thanks for the multicast-user fix — verified. On keeping the exact set: the rationale isn't checkable from the repo (the feed-config and `exchange-multicast-groups.yaml` live elsewhere, and "passed after a restart" is a live-network observation), so flagging it as unverified rather than disputing it. Your point that naming one leader group would still have let `rebop-eu` through is fair.

Worth fixing either way: `lastExtra` is built from `group.String()` on line 173, so a failure prints raw pubkeys — what made the original report unreadable. Resolve extras back to group codes from the program data the poll already fetches.

e2e/qa_multicast_settlement_test.go
```diff
@@ -6,36 +6,714 @@ import (
...
-		selectSubtestName: "find_closest_device",
-		selectDevice:      selectClosestDevice,
-		priceLogMsg:       "Found epoch price",
-	})
-}
+var (
+	enableSettlementTests    = flag.Bool("enable-multicast-settlement-tests", false, "enable multicast settlement tests")
+	retransmitOnlyDeviceFlag = flag.String("retransmit-only-device", "", "device code or pubkey in a retransmit-only metro (overrides auto-discovery)")
+	retransmitGroupCodesFlag = flag.String("retransmit-group-codes", "", "comma-separated multicast group codes a seat in a retransmit-only metro must subscribe to, and nothing else")
+	retransmitPriceFlag      = flag.Uint64("retransmit-price", 0, "expected seat price in whole USDC dollars in a retransmit-only metro; 0 asserts nothing")
```
nikw9944, line 30: The reinstated price check is off by default, so the run passes without ever checking the discounted price. This flag defaults to `0` and the assertion at line 373 is gated on `*retransmitPriceFlag != 0`; the deleted `-retransmit-price` defaulted to `10` and always asserted.

Log at Info when the assertion is skipped for a `0` flag, and pass `--retransmit-price` from the infra side in the same landing as the `--retransmit-group-codes` rename.

nikw9944 (commented):
Re-reviewed at `b32b8151`. The user-type half of the previous finding is verified fixed — `GetMulticastServiceabilityUser` filters on `UserTypeMulticast`, so an IBRL user at the same client IP can no longer be read. The other reviewer's findings are all addressed in `cb3ec334`: the rejection payment moved after `wait_for_open_phase`, the harness is inlined with the flag read hoisted out of every subtest, the selectors share one predicate, `TestProgramConfigFlags` passes (its `0x7c` case discriminates bit 7), and the CHANGELOG prefix is fixed.

Two things left, neither blocking:

1. The exact-set subscriber assertion is unchanged. The rationale for keeping it is reasonable and it is your call, but it rests on config outside this repo and a live-network observation, so it stands as unverified here. Failures should name group codes rather than raw pubkeys, since the strict form will fire again on any oracle lag.
2. `--retransmit-price` defaults to `0`, which asserts nothing — the old flag defaulted to `10`. Until the paired infra change passes it, the enforced path goes green with the discounted price unchecked, silently.

elitegreg (approved):
Approving. The review feedback is addressed: the rejection step now runs after `wait_for_open_phase`, the flag is read outside every subtest, the selector dedup and the flag-bit table test are in, and `GetMulticastServiceabilityUser` fixes the multi-tunnel user ambiguity. `go vet -tags qa ./e2e/...` and the `sdk/shreds/go` tests are clean.

Two follow-ups filed rather than held against this PR:
- #4157 — the retransmit-only price and group assertions are gated on the ProgramConfig enforcement bit, but they only depend on a metro being flagged in MetroHistory, so an environment with flagged metros and enforcement off gets no coverage.
- #4158 — the stale oracle snapshot you traced the `rebop-eu` extra to sounds like a product bug worth confirming.

Remember this needs the malbeclabs/infra change landing alongside it: `-leader-group-code` / `-retransmit-group-code` are gone and an undefined flag fails the whole test binary.

e2e/qa_multicast_settlement_test.go
martinsander00: both fixed in 72e5802bf505d818026b90fc8c80debd22d214df. the check reads the group codes once and labels the extras and the subscribed list with them, so a failure names `rocksteady` rather than a pubkey. that read also replaces the per-code lookups, which each refetched the whole program data.

e2e/qa_multicast_settlement_test.go
martinsander00: fair, an unset flag looked like a passing price check. 72e5802bf505d818026b90fc8c80debd22d214df logs at Info when `--retransmit-price` is unset. the infra side passes `-retransmit-price=36` alongside the group codes, in the same landing. i kept the default at 0 rather than 10, since frankfurt costs 36 and the old default would fail there.

e2e/qa_multicast_settlement_test.go
```diff
@@ -6,36 +6,735 @@ import (
...
+		for _, sub := range user.Subscribers {
+			group := solana.PublicKeyFromBytes(sub[:])
+			subscribed[group] = true
+			subs = append(subs, label(group))
+		}
+		var missing, extra []string
+		for group, code := range required {
+			if !subscribed[group] {
+				missing = append(missing, code)
+			}
```
nikw9944, line 184: Carried forward for the record, nothing to do. The check still fails on any group beyond those named on the command line, and the case for keeping it rests on the testnet feed-config and `exchange-multicast-groups.yaml`, both outside this repo, so it stays unverified here.

It holds as long as every group on the multicast user arrives through the seat. `doublezero connect multicast --subscribe-feed` (v0.33.0) can put one there independently, so if the QA client ever gains a non-seat subscription, this is where it will surface.

nikw9944 (approved):
Both findings from the last round are verified fixed in `72e5802b`.

Failures now name groups by code: `MulticastGroupCodes` reads the program data once and the `label` closure is applied to both the extras and the subscribed list, falling back to the pubkey for an unknown group. That single read also replaces the per-code lookups, each of which refetched the whole program data. And the enforced path logs at Info with the observed price when `--retransmit-price` is unset, so a run can no longer look like it checked the price. Keeping the default at `0` rather than `10` is the right call given Frankfurt costs 36.

One note carried forward, no action needed: the exact-set assertion still rests on config outside this repo, so it stays unverified here rather than disputed.

bgm-malbeclabs (approved):

PR #4160, opened 2026-08-06
internet-latency-collector: count a lost ping as a probe response

nikw9944:
## Summary of Changes

- `parseLatencyFromResult` returns zero for a result whose ping array holds no `rtt`, and the caller gated the whole block on `if latency > 0`. So a source probe at 100% packet loss never advanced its `last_response_at`. The field only ever moved on a successful ping, which made a lossy path indistinguishable from a probe that had stopped answering.
- The measurement cycle marks a source unresponsive once `last_response_at` ages past an hour. That is a rotation, and a metro's probe is a source for every measurement whose target sorts before it, so the blast radius is large.
- A result the probe uploaded is now recorded as a response before the latency check, whether or not anything came back.

**What is exported does not change.** A total-loss result still writes no sample. That is deliberate: it keeps the export cursor stalling when a whole measurement goes quiet, so #4154's `measurement stalled?` warning and the target-level staleness check both keep working. Exporting loss as `sample == 0` is a separate question, since it changes what lands on chain; it is not in this PR.

## Testing Verification

- New table test covers the three states: a total-loss result records a response and exports nothing, a successful result records a response and exports one sample, and a result carrying no `prb_id` or timestamp records nothing.
- Confirmed the test guards the change. Moving `UpdateSourceProbeResponse` back inside the `latency > 0` block fails `a_total-loss_result_counts_as_a_response`, and only that subtest.

## Rollout

**Expect one last round of churn on deploy, and do not read it as the fix failing.**

`Run()` calls `RunRipeAtlasMeasurementCreation` synchronously at startup (`collector.go:1420-1423`), while the export goroutine waits a full `exportInterval` first (10 min default, `cmd/collector/main.go:35`). So the first marking pass after deploy reads the `last_response_at` values loaded from the state file, which were frozen by the pre-fix behavior. Anything already more than an hour stale and not already blacklisted gets marked, and each marking recreates every measurement whose target sorts before that metro.

Self-limiting and no code change needed: once the first export cycle runs, loss results start refreshing `last_response_at` and the marking stops. Worth watching `Marking source probe as unresponsive` for the first ~20 minutes after deploy and confirming it goes quiet.


ben-dz (commented):
Fix is correct and correctly scoped — nothing blocks. Three things worth changing:

1. It retires the only per-probe detector for a source stuck at sustained total loss while its siblings keep the measurement's cursor advancing. The trade is right, but the CHANGELOG says only what's unchanged.
2. `Run()` runs the creation cycle synchronously at startup, before the export goroutine's first tick, so the first marking pass still reads the pre-fix frozen values — one last round of churn on deploy.
3. The new `probeID != 0 && !timestamp.IsZero()` guard can't change behavior (`UpdateSourceProbeResponse` enforces both halves), yet a seven-line comment says the exclusion happens at the call site and a test case claims to cover it.

Plus a changelog ref pointing at the issue instead of the PR, a test comment this change falsifies, and a missing one-line assertion.

- **Deploy ordering.** `Run()` calls `RunRipeAtlasMeasurementCreation` synchronously at startup (`collector.go:1420-1423`), while the export goroutine waits a full `exportInterval` (10 min default, `cmd/collector/main.go:35`). So the first marking pass reads the pre-fix frozen `LastResponseAt` values and marks whatever is >1h stale and not already blacklisted — one last round of the churn this fixes. Self-limiting, so no code change; worth a line in the rollout note so the churn isn't read as the fix failing.
- `collector_test.go:1477` — "because latency=0 when target doesn't reply, `UpdateSourceProbeResponse` doesn't fire" is exactly what this PR removes. The test still passes because it seeds state directly, but it now misdescribes why the target-unresponsive skip at `collector.go:845` exists. Reword to say the state is seeded to the pre-fix shape.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
...
+  - A RIPE Atlas result carrying no successful ping no longer looks like a dead source probe. `parseLatencyFromResult` returns zero for a result whose ping array holds no `rtt`, and the caller skipped everything behind an `if latency > 0` gate, so a path at 100% packet loss never advanced that source's `last_response_at`. It aged past the 1h threshold and the measurement cycle marked a probe that was uploading results on schedule as unresponsive — observed on mainnet-beta on 2026-08-06, where probe 7447 was marked 16 seconds after its most recent upload and the rotation recreated 28 measurements. A result the probe uploaded now counts as a response regardless of whether anything came back. What gets exported is unchanged: a total-loss result still writes no sample. (#4153)
```
ben-dz: This says what's unchanged but not what stopped being detected.

One source at 100% loss to one target while its siblings keep returning RTTs: loss rows now keep `LastResponseAt` fresh forever, so Step 4b (`collector.go:864`) never fires, and the siblings keep `LastExportAt` advancing, so neither Step 4 (`collector.go:818`) nor #4154's stall warning fires. `Marking source probe as unresponsive` (`collector.go:866`) was the only log naming the probe and there's no unresponsive-probe metric, so what's left is `LatencySamplesPerCollectionIntervalMissing{circuit=...}` with no rotation.

The trade is right — a wrongly-marked probe recreates every measurement it appears in, re-arming the `LastResponseAt == 0` blind spot at `collector.go:860` for all their sources — and the all-sources-lossy case is still covered via `LastExportAt`. One sentence here: a sustained-loss source is no longer rotated, and the missing-samples counter is the remaining signal. Or say that on #4153 so the follow-up accounts for it.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -592,6 +592,18 @@ func (c *Collector) exportSingleMeasurementResults(ctx context.Context, measurem
 	for _, result := range results {
 		// Parse latency from result (now also returns probe ID)
 		latency, timestamp, probeID := c.parseLatencyFromResult(result)
+
+		// A result the probe uploaded is proof the probe ran the measurement, whether or
```
ben-dz: This guard can't change behavior, and the comment says it can. `UpdateSourceProbeResponse` (`state.go:213-220`) already matches `probeID` against `meta.Sources` (RIPE IDs are never 0) and writes only when `timestamp > LastResponseAt`, which `time.Time{}.Unix() == -62135596800` never satisfies. Verified: replacing the guard with an unconditional call leaves all three new subtests passing, including `an unparseable result does not`.

So the comment's last sentence points at the wrong file for the exclusion, `collector_test.go:1768` covers a branch that can't fail, and seven lines re-argue an incident the CHANGELOG already carries. Either drop the guard and cut the comment to why the call moved above the latency gate:

```suggestion
		// A result the probe uploaded proves it ran the measurement even if nothing came
		// back, and LastResponseAt aging out rotates the probe (Step 4b) and recreates
		// measurements.
		measurementState.UpdateSourceProbeResponse(measurement.ID, probeID, timestamp.Unix())
```

or keep it as a hedge and drop the claim that the exclusion happens here.

CHANGELOG.md
ben-dz: Entry ends `(#4153)`, the tracking issue, not this PR. Neighbors put the PR last (`(malbeclabs/infra#2172, #4150)`). Make it `(#4153, #4160)`.

controlplane/internet-latency-collector/internal/ripeatlas/collector_test.go
```diff
@@ -1720,3 +1720,91 @@ func TestInitializeCreditBalance(t *testing.T) {
...
+			measurementState.SetMetadata(1, MeasurementMeta{
+				TargetLocation: "lax",
+				TargetProbeID:  200,
+				Sources:        []SourceProbeMeta{{LocationCode: "nyc", ProbeID: 100}},
+				CreatedAt:      time.Now().Add(-2 * time.Hour).Unix(),
+			})
+
+			count, _, err := c.exportSingleMeasurementResults(t.Context(), Measurement{ID: 1}, measurementState)
+			require.NoError(t, err)
+			require.Equal(t, tt.wantExportedNum, count, "exported record count should not change")
```
ben-dz, line 1937: The cursor not advancing on loss is what keeps Step 4 and #4154's stall warning alive, and it's pinned only transitively (`StallWarning/results_uploaded_but_all_timed_out`). One line puts it in the test that makes the claim:

```suggestion
			require.Equal(t, tt.wantExportedNum, count, "exported record count should not change")

			_, cursorSet := measurementState.GetLastTimestamp(1)
			require.Equal(t, tt.wantExportedNum > 0, cursorSet, "a total-loss result must not advance the export cursor")
```

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
nikw9944: Verified and dropped. `state.go:213-220` matches `source.ProbeID == probeID` (no enlisted source is 0) and writes only when `timestamp > LastResponseAt`, which `time.Time{}.Unix() == -62135596800` never satisfies. Replacing the guard with an unconditional call leaves all three subtests passing, so it was dead code and the comment was claiming an exclusion that happens elsewhere. Took your suggested wording.

Kept the `an unparseable result does not` subtest, with its comment rewritten to attribute the exclusion to `UpdateSourceProbeResponse`. It no longer pins anything in `collector.go`, but it does pin those two guards in `state.go` against a later change there. Say the word if you would rather it go.

CHANGELOG.md
nikw9944: Now `(#4153, #4160)`.

controlplane/internet-latency-collector/internal/ripeatlas/collector_test.go
nikw9944: Added, as suggested. Confirmed it guards rather than decorates: making loss advance `maxTimestamp` fails `a_total-loss_result_counts_as_a_response`. Previously only `StallWarning/results_uploaded_but_all_timed_out` would have caught that, and only transitively.

CHANGELOG.md
nikw9944: Added. The entry now states the trade explicitly: a single source stuck at sustained total loss to one target, while its siblings keep returning RTTs, is no longer marked or rotated, because its `last_response_at` stays fresh and the siblings keep the export cursor advancing. `..._latency_samples_per_collection_interval_missing` for that circuit is named as the remaining signal, and the entry says this is deliberate rather than incidental.

Confirmed your point that nothing else covers it: `grep -ci unresponsive internal/metrics/metrics.go` returns 0, so there is no unresponsive-probe metric and that counter really is all that is left.

nikw9944:
Pushed `af662017`. All five points verified before implementing; all five held up.

**Deploy ordering.** Confirmed. `Run()` runs the creation cycle synchronously at startup (`collector.go:1420-1423`, "Run measurement creation immediately on startup") and `defaultRipeAtlasExportInterval` is 10 minutes (`cmd/collector/main.go:35`), so the first marking pass reads the pre-fix frozen values loaded from the state file. Agreed no code change: it stops once the first export cycle refreshes `last_response_at`. Added a **Rollout** section to the description saying to expect it and to watch `Marking source probe as unresponsive` go quiet within ~20 minutes rather than reading the churn as the fix failing.

**`collector_test.go:1477`.** Reworded. It now says the state is seeded to the target-unresponsive shape directly and that results are not replayed through the export path, instead of asserting the `latency=0` mechanism this PR removes.

The other three are answered in their threads. Summary of what changed:

- guard dropped, comment cut to your suggested three lines
- cursor assertion added, verified to fail if loss advances `maxTimestamp`
- CHANGELOG now names the detector that goes away and ends `(#4153, #4160)`

`golangci-lint` clean, package tests pass.

One note on CI: `rust-build` and `release-validation-client` are red, but this branch touches four Go files under `controlplane/internet-latency-collector/` and a CHANGELOG line. Reads as the Actions outage rather than anything here.

ben-dz (approved):
All six findings verified fixed in `af662017` — checked against the code, not the replies.

PR #4167, opened 2026-08-07
cli: show the groups a feed grants in access-pass and user views

martinsander00:
Resolves: malbeclabs/infra#2178

## Summary of Changes

- `doublezero access-pass list` renders the groups a feed grants in the `multicast` column with an `F:` prefix, next to the existing `P:` and `S:` entries. An EdgeSeat pass showed an empty column and read as granting no multicast access; it now reads `F:qa-payments-group`. A feed grants subscribe rights only, so one prefix covers it. `--narrow` gets the same entry.
- `doublezero access-pass get` gains a `feeds` row with the feed codes, and a `feed_groups` row with the group codes those feeds carry. The plain table showed no feed at all before this.
- The feed seats in `access-pass get --json` now carry `feed_code` in place of `feed_key`, because an operator cannot read a key. No in-repo consumer reads that field; the SDK deserializers read the onchain account, which does not change.
- `doublezero user get` gains a `feeds` row naming the feeds whose seats this user holds. That is the field which answers why a feed reads 2 of 2 seats consumed.
- The two `get` commands read only the feeds they name, in one `getMultipleAccounts` call, so each adds one read. `access-pass list` scans the feed accounts once, and only when a listed pass carries a feed.

For reference on how each output will look like with this change: https://github.com/malbeclabs/infra/issues/2178#issuecomment-5222028900


smartcontract/cli/src/accesspass/get.rs
```diff
@@ -24,19 +23,32 @@ pub struct GetAccessPassCliCommand {
     pub json: bool,
 }
 
+#[derive(Serialize)]
+struct FeedSeatDisplay {
+    pub feed_code: String,
```
nikw9944, line 28: The `access-pass get --json` feed seats no longer carry the feed's pubkey, so a seat can no longer be mapped back to a specific feed. Feed accounts are keyed by `(code, exchange)` (`pda.rs:109`) and `feed get` carries an `--exchange` flag to disambiguate a code that exists in multiple metros, so two seats on one pass can print the same `feed_code` with different caps and billing windows and be indistinguishable. Keep `feed_key` and add `feed_code` alongside it; if the removal is intentional, the CHANGELOG entry belongs under `### Breaking`, not `### Changes`.

smartcontract/cli/src/accesspass/list.rs
```diff
@@ -293,71 +319,92 @@ impl ListAccessPassCliCommand {
             });
         }
 
+        // Only an EdgeSeat pass carries feeds, so a run without one skips the account scan.
+        let feeds = if access_passes
```
nikw9944, line 323: The subscriber filters read only `mgroup_sub_allowlist`, so they now disagree with the column next to them. `--not-multicast-group-subscriber foo` prints an EdgeSeat pass whose `multicast` cell reads `F:foo`, and `--multicast-group-subscriber foo` omits it. Extend the four `--multicast-group-*` filters to consider `groups_from_feeds`, or say "allowlist only" in their help text.

smartcontract/cli/src/accesspass/list.rs
```diff
@@ -293,71 +319,92 @@ impl ListAccessPassCliCommand {
...
-                multicast_users: format!(
-                    "{} / {}",
-                    access_pass.multicast_user_count, access_pass.max_multicast_users
-                ),
-                status: access_pass.status,
-                owner: access_pass.owner,
-                accesspass_type_value: access_pass.accesspass_type.clone(),
-                mgroup_pub_allowlist: access_pass.mgroup_pub_allowlist.clone(),
-                mgroup_sub_allowlist: access_pass.mgroup_sub_allowlist.clone(),
+                        for group in &feed_groups {
```
nikw9944, line 368: The `F:` entries and the `feed_groups` row in `accesspass/get.rs:123` union every seat's groups with no metro, so they claim access the pass does not grant. A feed's groups are joinable only on a device in that feed's exchange (`check_feed_metro_coverage` returns `MetroMismatch`), so a pass seated on `feed-a@ams` and `feed-b@fra` renders `F:x, F:y, F:z` while a user on an ams device joining `z` is rejected. Qualify each entry with the feed's exchange, the way `feed list` already does.

nikw9944 (dismissed):
Three things to settle before merge. `access-pass get --json` drops `feed_key` entirely, leaving no way to map a seat back to a feed — feed codes are unique only per metro, so two seats on one pass can print identically. The new `F:` entries also disagree with the subscriber filters right beside them, and they flatten away the per-metro scoping that actually decides whether a group is joinable. Re-checked at e8ebc8f7: all three findings remain open.

smartcontract/cli/src/accesspass/list.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4167/changes/5cd18c5bddac5f00f7ee96661a910758284752cc

smartcontract/cli/src/accesspass/list.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4167/changes/5cd18c5bddac5f00f7ee96661a910758284752cc

smartcontract/cli/src/accesspass/get.rs
martinsander00: https://github.com/malbeclabs/doublezero/pull/4167/changes/5cd18c5bddac5f00f7ee96661a910758284752cc

bgm-malbeclabs (approved):

PR #4173, opened 2026-08-10
cli: join purchased feeds on a bare multicast connect

ben-malbeclabs:
## Summary

- A bare `doublezero connect multicast` on an EdgeSeat pass carrying purchased feeds subscribed the host to nothing and reported success. The no-flags path reads the pass's `mgroup_pub_allowlist` / `mgroup_sub_allowlist`, which the feed purchase path never writes — only the administrative subscriber-allowlist instruction does — so a feed pass has both empty. It printed `The AccessPass has no authorized multicast groups; nothing to connect to.` and exited **0**. `doublezero-edge-connect`'s installers run exactly this form, so their `|| warn` guard could not fire: the install printed its success banner while the host held no subscription, and the idle WebSocket sink then looked like a publisher problem.
- It now selects the purchased feeds on the pass that still have a free seat in the connected device's metro, joins them, and prints every feed it skipped with the reason.
- Joining nothing is an error when this machine holds no feed and could take none — whether the seats are full or serve another metro. Exiting 0 with nothing subscribed is indistinguishable from success to an unattended installer.
- A machine that already holds a feed still succeeds when it joins nothing, and activates as usual, so repeated installs stay safe. Keying the failure on "joined nothing" instead would fail a correctly-provisioned host whose remaining feed is full.
- Headroom is measured against `FeedSeat.max_users`, the field `try_add_feed_user` enforces. `max_future_users` is written and validated by `SetAccessPassFeeds` and documented to flip the cap at `window_end`, but nothing reads it, so trusting it would propose feeds the program rejects. Worth its own issue.
- Candidate devices are restricted to the metros the purchased feeds serve, matching `resolve_feed_join`, so a host that bought an Amsterdam feed while sitting nearest Frankfurt connects through Amsterdam instead of failing. Only metros holding a feed with a free seat are candidates, so latency cannot land the host where everything it bought is full. One host can still only ever hold one metro's feeds, since a user is keyed on `(client_ip, user_type)` with no device, and feeds in the other metros are named as skipped.
- When a nearer device is passed over because the pass holds nothing in its metro, an informational line names it and the latency difference. It is advisory only: fetched after device selection succeeds so a failing connect never waits on it, silent when latency data is missing, and suppressed below a 1ms threshold so jitter cannot trigger it.
- When no eligible device serves any candidate metro, the failure names those feeds and any feed skipped as full, instead of reporting a metro problem for a cluster that simply has no activated devices.
- `--device` naming a device other than the existing Multicast user's now fails instead of being silently ignored, matching the guard `resolve_feed_join` already applies.
- Selection is bounded by `MAX_USER_FEEDS` only. It deliberately adds no transaction-size budget: `SubscribeFeedCommand::execute` already splits a join past `MAX_FEED_TX_ACCOUNTS` across transactions, and a second budget here would restate that bound and get it wrong, since the SDK's 25 counts variable feed and group accounts alone.

## Testing

- 22 new tests, 85 passing in `doublezero-daemon-cli` (`cargo test -p doublezero-daemon-cli connect`), `make rust-lint` clean.
- Table-driven unit coverage of the selection function: headroom, full seats, already-held feeds, other-metro feeds, unreadable feed accounts, the per-user feed cap, held-plus-cap together, and a guard that group-heavy feeds are all still selected rather than dropped for transaction size.
- Integration coverage through the mock ledger: joins both feeds, joins only the one with a free seat, fails when every seat is full, no-ops and activates on a re-run that already holds a feed, fails when feeds serve only another metro, rejects a conflicting `--device`, and a regression guard that a non-EdgeSeat pass still takes the allowlist path unchanged.
- The held-plus-cap test is a real discriminator: against the previous arithmetic it selects 5 feeds, against the fix 2 with 3 reported over the limit.
- End-to-end coverage of the three-machine seat-filling walk is `malbeclabs/doublezero-shreds#663`. It is verified locally against a pin bumped to this branch, and its CI stays red until this merges and that pin advances — the e2e image builds the client from the committed pin, so it cannot exercise an unmerged branch.


crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -534,6 +544,197 @@ impl Connect {
...
+                (FeedJoinUser::Existing(*pk), device, user.feed_pks.clone())
+            }
+            None => {
+                let mut devices = ledger.list_device()?;
+                if self.device.is_none() {
+                    devices.retain(|_, d| {
+                        d.is_device_eligible_for_provisioning()
+                            && d.check_user_type_capacity(UserType::Multicast, false)
+                                .is_none()
+                    });
```
martinsander00, line 632: This `retain` keeps every eligible device, so `find_or_create_device` picks the nearest one in any metro. A customer who bought an Amsterdam feed but sits nearest to Frankfurt then fails with `no purchased feed serves the metro of device fra`, rather than connecting through an Amsterdam device. Can this also require `d.exchange_pk` to be one of the exchanges the pass's feeds serve, the way `resolve_feed_join` filters at line 827?

When the purchased feeds span two metros, the latency pick then still lands on the closest metro. The feeds in the other metro keep the existing `Skipped, another metro` line instead of failing the command.

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -534,6 +544,197 @@ impl Connect {
...
+                "    Skipped, no free seat: {}",
+                describe_full(&selection.full)
+            )?;
+        }
+        if !selection.other_metro.is_empty() {
+            let described: Vec<String> = describe_other_metro(&selection.other_metro, &feeds)
+                .into_iter()
+                .map(|(code, exchange)| format!("{code} (exchange {exchange})"))
+                .collect();
+            writeln!(out, "    Skipped, another metro: {}", described.join(", "))?;
```
martinsander00: `describe_other_metro` renders `feed.exchange`, which is a pubkey, so this prints `Skipped, another metro: kalshi (exchange 7dHb...)`. The operator cannot tell that the pubkey means Amsterdam, and the same pubkey shows up again in the bail at line 707. Would `kalshi does not serve the metro of device fra01` work here instead, since `device.code` is already in hand and needs no exchange lookup?

Printing `ams` reads better still, but it costs a `list_exchange` on the ledger client trait plus the mock.

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -534,6 +544,197 @@ impl Connect {
...
+            .map(|(pk, u)| (*pk, u.clone()));
+
+        let (join_user, device, held_feed_pks) = match &mcast_user {
+            Some((pk, user)) => {
+                let device = ledger
+                    .get_device(user.device_pk.to_string())
+                    .map_err(|err| {
+                        eyre::eyre!("failed to fetch device {}: {err}", user.device_pk)
+                    })?;
+                writeln!(out, "    An account already exists with Pubkey: {pk}")?;
```
martinsander00, line 609: This branch never names the device, while the branch that creates a user prints `Device selected: fra01` at line 606. On a re-run the operator sees `Skipped, another metro: kalshi (exchange 7dHb...)` with no device anywhere in the output, so the line has nothing to compare against. Can this also print the device code, given `device` is already fetched two lines above?

crates/doublezero-daemon-cli/src/connect.rs
ben-malbeclabs: Adopted — candidate devices are now filtered to the metros the purchased feeds serve, matching `resolve_feed_join`. The Amsterdam-feed/nearest-Frankfurt case now connects through Amsterdam instead of failing. One refinement: the candidate set is built only from seats that still have a free seat, so latency can't land you in a metro where everything you bought is already full.

Your second paragraph holds — multi-metro passes still land on the nearest purchased metro and the rest keep the skip line.

Worth flagging that adopting this exposed two bugs in my implementation of it, both now fixed. The new "no eligible device serves the metro" bail also fired when there were no eligible devices *at all*, masking `best_latency`'s accurate "No activated devices found". And where one purchased feed was full and another had a free seat in a metro with no eligible device, the failure named only the metro problem — so you'd go find a machine in that metro and still not get the full feed. It now names both. `92387db4c`, `49f70f98f`.

crates/doublezero-daemon-cli/src/connect.rs
ben-malbeclabs: Fixed. First pass took your cheap suggestion — compare against `device.code`, no pubkeys anywhere in operator output. Then I went the further step you priced: `list_exchange` is now on the `LedgerClient` trait, so the metro is named properly — `Lower latency is available from fra01 in Frankfurt (8.20ms vs 21.40ms)`.

What tipped the cost/benefit was the failure path rather than this line: the bail ended with "Connect from a machine in that metro" while never naming the metro, so the one message whose job is to tell an operator where to go didn't say where. It now names it. Resolution failures degrade to the previous wording rather than erroring, since all of this is cosmetic. `92387db4c`, `2c062a43a`.

crates/doublezero-daemon-cli/src/connect.rs
ben-malbeclabs: Fixed — the existing-user branch now prints `Device selected: <code>` with the same wording as the create branch, so a re-run's skip lines have something to compare against. `92387db4c`.

martinsander00 (approved):

PR #4177, opened 2026-08-11
cli: name a feed by its metro in access-pass and user views

elitegreg:
## Summary of Changes

- The `feeds` row of `doublezero access-pass get` and `doublezero user get` now names each feed `code:metro`, e.g. `lashay1-feed:xams`. A feed is keyed by `(code, exchange)`, so one code is a different feed in every metro that sells it; the row printed the code alone, so a pass holding one code in three metros rendered the same name three times and gave no way to tell which metros it covered.
- Feed seats in `access-pass get --json` carry a new `exchange_code` beside the existing `feed_code`. Both stay unqualified, so nothing has to be split back out of a joined string.
- A feed whose exchange account cannot be read falls back to the exchange key rather than dropping the suffix — uglier, but it keeps same-coded feeds apart, which is the point of qualifying them. A feed that cannot be read at all still renders as its own key, as before.
- Exchanges are read with `getMultipleAccounts` for the keys the feeds name, not a program scan: one extra read of a handful of accounts, and none at all for a record that names no feed.
- The `feed_groups` row is deliberately unchanged. Its metro TODO stays open, because qualifying groups changes which entitlements are claimed and not just how they are named.

## Diff Breakdown

| Core logic   |     2 | +108 / -6   | +102 |
| Scaffolding  |     3 | +19 / -12   |   +7 |
| Tests        |     4 | +574 / -16  | +558 |
| Docs         |     1 | +1 / -1     |   +0 |
| **Total**    |     6 | +702 / -35  | +667 |

127 lines of implementation behind a resolver both views share; the bulk of the diff is the tests that pin the rendering and its fallbacks.

<details>

- `smartcontract/cli/src/feed/resolve.rs` — `get_feeds` becomes private behind `FeedLabels` / `resolve_feed_labels`, which reads the feeds and then the metros they serve. `label` and `exchange_code` share one lookup, so the table and the JSON cannot disagree. Exchange keys are collected by walking the caller's ordered keys rather than the feed map, so a shared metro is read once and the read is deterministic.
- `smartcontract/cli/src/exchange/resolve.rs` — new `get_exchanges`, a mirror of `get_feeds`: one `getMultipleAccounts`, keys that do not resolve left out of the map, no RPC for an empty key list.
- `smartcontract/cli/src/accesspass/get.rs` — `feeds` is built from `FeedLabels::label`; `FeedSeatDisplay` gains `exchange_code`.
- `smartcontract/cli/src/user/get.rs` — same helper, so both views name a feed identically.

</details>

## Testing Verification

- New unit tests cover an EdgeSeat pass holding one code in three metros (the reported case), two feeds sharing one metro — where the mock's exact key vector is what asserts the metro is read once — an unreadable exchange, an unreadable feed, and a pass naming no feed at all, which must issue no read on either leg.
- Against testnet, a two-feed pass renders `feeds | lashay1-feed:xlax, qa-payments:xfra`, and `--json` carries `exchange_code` on each seat. The pass from the report now holds a single seat onchain and renders `lashay1-feed:xlax`.
- `doublezero user get` is covered by unit tests only: testnet currently lists no users.


smartcontract/cli/src/accesspass/get.rs
```diff
@@ -379,13 +383,40 @@ mod tests {
         );
     }
 
-    /// An EdgeSeat pass with one feed, that feed carrying one group. Both the feed account and the
-    /// group are resolvable, so the display shows codes rather than keys.
+    fn test_exchange(code: &str) -> Exchange {
```
martinsander00, line 386: `test_exchange` appears three times. It is here, at `exchange/resolve.rs:38`, and at `feed/resolve.rs:124`. `user/get.rs:481` writes the same `Exchange` literal inline.

`create_test_client` lives in `smartcontract/cli/src/tests.rs`. One builder there serves every call site.

martinsander00 (approved):

PR #4181, opened 2026-08-12
ci: wait for cloudsmith to list the nightly before deploying devnet

martinsander00:
## Summary

- The devnet daily deploy now waits for CloudSmith to list the nightly it just built before it runs ansible.
  - On 2026-08-12 the client package finished uploading at 14:37:00 UTC, and ansible started 11 seconds later.
  - Four devnet hosts ran `apt-get update` between 14:37:21 and 14:37:24 and stayed on the 2026-08-11 nightly, while the four hosts ansible reached 7 seconds later upgraded.
  - The new step polls the jammy and noble package indexes for up to 10 minutes, then fails the deploy, so a stale fleet is loud rather than silent.
  - A component that builds no deb, such as the agent, skips the step.
- `TestQA_DeviceProvisioning` now passes `--no-version-warning` to the CLI, and it warns when the client is behind the program.
  - The CLI writes its upgrade banner to stderr, and `runCLI` merges stderr into stdout.
  - That banner reached the JSON parser in the [failing devnet QA run](https://github.com/malbeclabs/infra/actions/runs/31621977149/job/94198789716) as `invalid character 'A' looking for beginning of value`, when chi-dn-bm4 ran 0.34.0 against a program at 0.35.0.
  - The test reads the client and program versions at the start, so the next skew shows up as a warning instead of a parse error.

## Testing Verification

- I re-ran the deploy job of run 31606641653 once CloudSmith had indexed the package, and all eight devnet hosts upgraded, where the 14:37 deploy had upgraded four.


.github/workflows/release.daily.yml
```diff
@@ -75,11 +77,46 @@ jobs:
...
+            missing=""
+            for suite in jammy noble; do
+              url="https://dl.cloudsmith.io/public/malbeclabs/doublezero-devnet/deb/ubuntu/dists/$suite/main/binary-amd64/Packages"
+              if ! curl -sfL --max-time 30 "$url" | grep -qF "Version: $VERSION"; then
+                missing="$missing $suite"
+              fi
+            done
+            if [ -z "$missing" ]; then
+              echo "CloudSmith lists $VERSION on attempt $attempt"
+              exit 0
```
bgm-malbeclabs, line 125: The wait can pass while a devnet host still reads the old index. `InRelease` is served through CloudFront with `cache-control: max-age=30, must-revalidate`, so a POP that fetched the index 29 s before the reindex keeps serving the stale copy for another 29 s. The runner's `curl` and each host's `apt-get update` hit different POPs.

Scenario: CloudSmith reindexes at T; the runner's POP misses cache and this loop exits at T+3 s; ansible reaches a host at T+8 s whose POP cached the pre-reindex object at T-25 s, so that host reads the old list and stays on yesterday's nightly. That is the same ~30 s window as the 2026-08-12 failure (upload 14:37:00, `InRelease` Date 14:37:27, stale hosts at 14:37:21-24).

Fix: `sleep 35` after the loop succeeds, or require two consecutive passes 30 s apart.

.github/workflows/release.daily.yml
```diff
@@ -75,11 +77,46 @@ jobs:
...
+      - name: Wait for CloudSmith to list the new package
+        if: needs.release.outputs.deb_version != ''
+        env:
+          VERSION: ${{ needs.release.outputs.deb_version }}
+        run: |
+          for attempt in $(seq 1 40); do
+            missing=""
+            for suite in jammy noble; do
+              url="https://dl.cloudsmith.io/public/malbeclabs/doublezero-devnet/deb/ubuntu/dists/$suite/main/binary-amd64/Packages"
+              if ! curl -sfL --max-time 30 "$url" | grep -qF "Version: $VERSION"; then
```
bgm-malbeclabs: `curl … | grep -qF` is correct only because the implicit step shell is `bash -e {0}`, without `pipefail`. `grep -q` exits at the first match near the top of a 428 KB index, curl then dies on the broken pipe, and under `pipefail` the pipeline status becomes curl's. Confirmed against the live URL: with `set -o pipefail` a *successful* match returns 56; without it, 0.

Scenario: someone adds `shell: bash` to this step, a routine hardening edit, and this same job already sets `shell: bash -leo pipefail {0}` on the ansible step. Every success then reads as "missing", so the deploy burns 10 minutes and fails, every night.

Fix: `grep -qF "Version: $VERSION" <<<"$(curl -sfL --max-time 30 "$url")"`.

e2e/qa_provisioning_test.go
```diff
@@ -56,6 +56,13 @@ func TestQA_DeviceProvisioning(t *testing.T) {
 	prov, err := qa.NewProvisioningTest(ctx, log, networkConfig, envArg, bmHost)
 	require.NoError(t, err, "failed to create provisioning test")
 
+	versions, err := prov.GetVersions(ctx)
+	require.NoError(t, err, "failed to read the CLI and program versions")
```
bgm-malbeclabs: `require.NoError` turns a purely informational probe into a hard abort. `GetVersions` fails whenever the `program version` line is absent, and `smartcontract/cli/src/version.rs:20` swallows the RPC error and prints only `client version` when `get_program_config` fails.

Scenario: a transient ledger RPC hiccup in the first second kills `TestQA_DeviceProvisioning` with `could not read both versions from output: client version: …`, before it touches a device.

Fix: log the error and carry on instead of failing the test.

e2e/internal/qa/provisioning.go
```diff
@@ -105,7 +105,9 @@ func NewProvisioningTest(ctx context.Context, log *slog.Logger, networkConfig *c
 
 func (p *ProvisioningTest) runCLI(ctx context.Context, args ...string) ([]byte, error) {
 	// Build SSH command: ssh <bm-host> doublezero <args...>
-	sshArgs := []string{p.bmHost, "doublezero"}
+	// The CLI writes its upgrade banner to stderr, and CombinedOutput folds that
+	// banner into the JSON the callers below parse.
+	sshArgs := []string{p.bmHost, "doublezero", "--no-version-warning"}
```
bgm-malbeclabs: This treats one symptom, not the root cause. `runCLI` still uses `CombinedOutput`, so every remaining stderr writer lands in the bytes `json.Unmarshal` parses at line 417 and in `device list --json` further down. `--no-version-warning` silences only the upgrade banner; the CLI still routes diagnostics through `tracing` at the default `warn` level to stderr (RFC-20), and `ssh` writes its own messages there (`Warning: Permanently added '…' to the list of known hosts.`). Either one reproduces the same failure as `invalid character 'W' looking for beginning of value`.

Fix: `cmd.Output()` with a separate `bytes.Buffer` on `cmd.Stderr` removes the whole class of failure and keeps stderr for the error message.

e2e/internal/qa/provisioning.go
```diff
@@ -117,6 +119,61 @@ func (p *ProvisioningTest) runCLI(ctx context.Context, args ...string) ([]byte,
...
+			versions.Program = strings.TrimSpace(value)
+		}
+	}
+	if versions.Client == "" || versions.Program == "" {
+		return Versions{}, fmt.Errorf("could not read both versions from output: %s", string(output))
+	}
+	return versions, nil
+}
+
+func (v Versions) ClientIsOlder() bool {
```
bgm-malbeclabs: `ClientIsOlder` will not fire for the staleness this PR is about. It compares only the semver triple, dropping the `~git<BUILD_DATE>.<sha>` suffix that is the only thing separating two nightlies inside one release cycle, and the nightly template (`{{ incpatch .Version }}~git…`, `release/.goreleaser.base.client.yaml:192`) puts every nightly one patch *above* the released program version.

Scenario: a host stuck on `0.35.1~git20260813…` while today's is `0.35.1~git20260814…` and the program is at `0.35.0`. The client parses to `[0 35 1]`, the program to `[0 35 0]`, so the client reads as newer and no warning prints. The warning fires only across a release bump, about one day in 30.

Fix: compare the build date in the suffix, or compare against the latest nightly rather than the program version.

e2e/internal/qa/provisioning.go
```diff
@@ -105,7 +105,9 @@ func NewProvisioningTest(ctx context.Context, log *slog.Logger, networkConfig *c
 
 func (p *ProvisioningTest) runCLI(ctx context.Context, args ...string) ([]byte, error) {
 	// Build SSH command: ssh <bm-host> doublezero <args...>
-	sshArgs := []string{p.bmHost, "doublezero"}
+	// The CLI writes its upgrade banner to stderr, and CombinedOutput folds that
+	// banner into the JSON the callers below parse.
+	sshArgs := []string{p.bmHost, "doublezero", "--no-version-warning"}
 	sshArgs = append(sshArgs, args...)
```
bgm-malbeclabs, line 112: `--no-version-warning` also suppresses the hard incompatibility error, not just the warning: `client/doublezero/src/main.rs:436` gates the whole `check_version` call, and `checkversion.rs:16` bails when the client is below `min_compatible_version`.

Scenario: the bare metal host runs a client the program declares incompatible. The run used to stop at once with "Your client version is no longer up to date"; it now proceeds and fails later in some unrelated assertion. The new `t.Logf` warning does not cover this, since it usually will not print (see the `ClientIsOlder` comment).

Fix: keep the check and drop only the banner, or assert compatibility from the versions the test already reads.

bgm-malbeclabs (commented):
Review notes on the CloudSmith wait and the QA version probe.

e2e/internal/qa/provisioning.go
martinsander00: https://github.com/malbeclabs/doublezero/pull/4181/changes/b00d22d30132d08bdbe6b4ee6f8e671912bd3e8a

e2e/internal/qa/provisioning.go
martinsander00: https://github.com/malbeclabs/doublezero/pull/4181/changes/b00d22d30132d08bdbe6b4ee6f8e671912bd3e8a

e2e/internal/qa/provisioning.go
martinsander00: https://github.com/malbeclabs/doublezero/pull/4181/changes/b00d22d30132d08bdbe6b4ee6f8e671912bd3e8a

e2e/qa_provisioning_test.go
martinsander00: https://github.com/malbeclabs/doublezero/pull/4181/changes/b00d22d30132d08bdbe6b4ee6f8e671912bd3e8a

.github/workflows/release.daily.yml
martinsander00: https://github.com/malbeclabs/doublezero/pull/4181/changes/b00d22d30132d08bdbe6b4ee6f8e671912bd3e8a

.github/workflows/release.daily.yml
```diff
@@ -75,11 +77,51 @@ jobs:
...
+      - name: Wait for CloudSmith to list the new package
+        if: needs.release.outputs.deb_version != ''
+        env:
+          VERSION: ${{ needs.release.outputs.deb_version }}
+        run: |
+          for attempt in $(seq 1 40); do
+            missing=""
+            for suite in jammy noble; do
+              url="https://dl.cloudsmith.io/public/malbeclabs/doublezero-devnet/deb/ubuntu/dists/$suite/main/binary-amd64/Packages"
+              index=$(curl -sfL --max-time 30 "$url" || true)
```
ben-dz: **Optional** — `|| true` discards curl's exit status, so a 404, DNS failure or 5xx is indistinguishable from "not indexed yet": the step burns the whole loop and then blames CloudSmith's index, for every component. The stated bound is also optimistic — 40 attempts × (two 30s curls + 15s) is up to ~50 minutes. Fix: keep curl's status and log it, and loop on a deadline (`end=$((SECONDS+600))`).

ben-dz (approved):
LGTM, Two non-blocking items: `|| true` now discards curl's status, so a 404 or DNS failure spends the full loop and reports "CloudSmith did not list"; and the caller matrix has no `fail-fast: false`, so this new failure can cancel sibling deploys mid-play.

- **Optional** — `.github/workflows/release.devnet.all.daily.yml:22` — the caller matrix has no `fail-fast: false`, so this new hard failure cancels sibling legs' deploy jobs, some of them mid-`ansible-playbook`. A partially applied fleet is worse than the stale package the gate replaces. Add `fail-fast: false`.

.github/workflows/release.daily.yml
martinsander00: https://github.com/malbeclabs/doublezero/pull/4181/commits/ea0089dceef2fbf612b0d9ae5194eafe885ade5d

Issue #4200, opened 2026-08-17
serviceability-instruction/sdk-rs: carry the IP ownership proof and Ed25519 instruction


elitegreg:
Part of RFC-27 (tracker: #4194). Depends on #4195 and #4197.

## What

- `crates/doublezero-serviceability-instruction/src/user.rs`: extend `create_user` and `create_subscribe_user` to take `Option<IpOwnershipProof>`, write it into the args, and add the Instructions sysvar to the account list at the position #4197 settled on. Update the account-layout doc comments — they are the reference for these instructions.
- Add a helper that builds the native `Ed25519SigVerify` instruction from a proof, so callers can't get the precompile offset layout wrong.
- Add a helper that assembles the pair in the right order (Ed25519 instruction before the create instruction) for callers that just want a working transaction.
- `smartcontract/sdk/rs`: thread the proof through `CreateUserCommand` and `CreateSubscribeUserCommand`.

## Notes

Both instructions already carry a long account list — resource extensions plus one `dz_prefix_block` per prefix — and now gain the sysvar plus a second instruction in the transaction. Check the transaction still fits at realistic `dz_prefix_count` values and note the headroom in the PR.

## Acceptance

- A test building the full transaction and running it against the program in the existing test harness, proving the assembled Ed25519 instruction validates.
- Passing `None` produces the pre-RFC-27 instruction shape.

Issue #4198, opened 2026-08-17
ip-verifier: add the IP ownership verification service


elitegreg:
Part of RFC-27 (tracker: #4194). Depends on #4195.

A stateless HTTP service that signs the source IP it observes. It is the only party that can attest an IP, so its correctness about *which* IP it saw is the whole security property.

## API

```
POST /v1/proof
{ "payer": "<Pubkey>", "user_pubkey": "<Pubkey>" }

200 { "payer": "...", "client_ip": "a.b.c.d", "epoch": 1234,
      "user_pubkey": "...", "signature": "<base58>" }
```

`client_ip` is always the observed source address of the request; it is never accepted from the body. `user_pubkey` is the User PDA the client intends to create, so the caller derives it before verifying. Also expose `GET /health` and a Prometheus `/metrics` endpoint consistent with the other DoubleZero services.

## Requirements

- Reject a source address that isn't globally routable, with the same `is_global` semantics the program applies, so the service never issues a proof the program would refuse.
- **Proxy handling.** Behind a proxy or CDN the peer address is the proxy. Take a configured list of trusted proxy CIDRs and only then honor `X-Forwarded-For` / `Forwarded`, taking the last untrusted hop. With no trusted proxies configured, use the peer address and nothing else. Getting this wrong signs the proxy's IP for everyone, so it needs its own tests.
- Epoch comes from the DoubleZero Ledger RPC, cached with a short TTL and refreshed in the background. Fail closed if the epoch is stale beyond a threshold rather than signing an old epoch.
- Signing key loaded from a keypair file path (custody options are the deployment issue). Never logged.
- Per-source-IP rate limiting, request size caps, and a request timeout.
- IPv4 only. Reject IPv6 sources with a clear error until the proof layout gains an address family.
- Structured logging through `tracing`, config through clap + env vars, matching `crates/` conventions.

## Implementation notes

The workspace already depends on `hyper` 1 and `hyper-util` but has no HTTP server framework. Either build on hyper directly or add `axum` — call the choice out in the PR since it sets a precedent for future Rust services in the repo.

## Acceptance

- Unit tests for proxy header handling: no trusted proxies, trusted proxy with one hop, trusted proxy with a spoofed extra hop, untrusted client sending `X-Forwarded-For`.
- Signature output verifies against the committed test vectors from #4195.
- Non-routable and IPv6 sources rejected.
- Integration test that runs the service, requests a proof, and verifies it with the same code path the program uses.

Issue #4197, opened 2026-08-17
serviceability: validate IpOwnershipProof via the Ed25519 precompile during user creation


elitegreg:
Part of RFC-27 (tracker: #4194). Depends on #4195 and #4196.

This is the enforcement point. BPF can't verify Ed25519 cheaply, so the transaction carries a native `Ed25519SigVerify` instruction and the program introspects the Instructions sysvar to confirm it covers the message we expect with the key we trust.

## What

**Feature flag.** Add `FeatureFlag::RequireIpOwnershipProof = 2` to `state/feature_flags.rs` with the string `require-ip-ownership-proof`, so it parses in the existing `SetFeatureFlags` CLI path. Off means the proof is validated when supplied and ignored when absent; on means user creation without a valid proof is rejected — uniformly, for wildcard and specific-IP passes alike.

**Instruction surface.** Add an optional proof to `UserCreateArgs` and `UserCreateSubscribeArgs` via `borsh_incremental` (`#[incremental(default = None)]`), and accept the Instructions sysvar account. Document exactly where the sysvar sits in the account layout — the account ordering in `create.rs` and `create_subscribe.rs` is already load-bearing, with optional tenant/feed and a trailing peeled Permission account, so pick a position that can't collide and state it in the builder docs.

**Validation**, in a shared helper called from `create_user_core` so `CreateUser` and `CreateSubscribeUser` cannot drift:

1. Reconstruct `signed_message()` from `(payer, client_ip, epoch, user_pda)`.
2. Load the Ed25519 instruction from the Instructions sysvar and confirm its pubkey equals `globalstate.ip_verifier_authority_pk`, its message equals the reconstructed bytes, and its signature equals `proof.signature`. Parse the precompile's offset layout properly and reject any instruction whose offsets point outside the instruction, or that carries more than one signature.
3. `proof.payer == payer_account.key`.
4. `proof.client_ip == client_ip`.
5. `proof.user_pubkey == the derived User PDA`.
6. `proof.epoch == clock.epoch || proof.epoch == clock.epoch - 1`.
7. Reject if `ip_verifier_authority_pk` is `Pubkey::default()`.

New `DoubleZeroError` variants for each rejection class, distinguishable enough that the CLI can print something useful.

## Notes

- The idempotent path matters: `create_user_core` returns early when the user already exists and matches. Decide and document whether a proof is required on that path — it should be, otherwise a squatter's existing account absorbs later calls without ever proving anything. Cover it with a test either way.
[...]

Issue #4196, opened 2026-08-17
serviceability: add ip_verifier_authority_pk to GlobalState and rotate it via SetAuthority


elitegreg:
Part of RFC-27 (tracker: #4194).

The verifier public key is the trust root for proof validation. RFC-27 puts it in onchain global state so it rotates without a program upgrade.

## What

- Append `pub ip_verifier_authority_pk: Pubkey` to `GlobalState` (`state/globalstate.rs`), after `feed_authority_pk`. Append only — existing accounts deserialize with `Pubkey::default()` through the `unwrap_or_default()` path, same as `feed_authority_pk` did.
- Add it to `Default`, `Display`, and the `TryFrom<&[u8]>` chain.
- Add `ip_verifier_authority_pk: Option<Pubkey>` to `SetAuthorityArgs` (`processors/globalstate/setauthority.rs`) and its `Debug`. `BorshDeserializeIncremental` keeps old-arg transactions decoding.
- Surface it in the CLI: `smartcontract/cli/src/init.rs` and the `global-config`/authority set + get paths so operators can read and rotate it.
- Rust SDK (`smartcontract/sdk/rs`) `GlobalState` mirror and any `SetAuthority` command builder.

## Behavior when unset

`Pubkey::default()` means "no verifier configured". Enforcement (the onchain validation issue) must treat that as a hard reject when the feature flag is on, never as "any signature passes". Add a test for exactly that.

## Acceptance

- `test_state_compatibility_globalstate` still passes against the existing base64 vectors, plus a new vector that includes the field.
- Serialization test updated.
- `SetAuthority` test rotating the key and leaving the other authorities untouched, and one confirming a `None` leaves it unchanged.
- Authorization unchanged: GLOBALSTATE_ADMIN permission or the legacy foundation path.
- SDK deserializer work for Go/TS/Python is tracked separately.

PR #4206, opened 2026-08-17
ip-proof: add a crate defining the IpOwnershipProof layout and signed message

### LINKED ISSUE #4195: ip-proof: add a crate defining the IpOwnershipProof layout and signed message
Part of RFC-27 (tracker: #4194).

The proof format is consumed by three independent places: the serviceability program (BPF), the CLI, and the verification service. They must agree byte for byte or every proof fails validation. Put the definition in one crate rather than duplicating it.

## What

New crate `crates/doublezero-ip-proof` (no Solana runtime dependency beyond `solana-program`'s `Pubkey`, buildable for BPF):

```rust
pub struct IpOwnershipProof {
    pub payer: Pubkey,       // 32
    pub client_ip: Ipv4Addr, // 4, network order
    pub epoch: u64,          // 8, little-endian
    pub user_pubkey: Pubkey, // 32 — the User PDA this proof authorizes
    pub signature: [u8; 64],
}
```

- `signed_message()` returns the exact bytes the verifier signs: a domain-separation prefix (`b"DZ_IP_PROOF"`), a `u8` version byte set to `1`, then `payer || client_ip || epoch || user_pubkey`. Fixed length, no length prefixes, no Borsh on the signed portion — the program reconstructs it and compares against what the Ed25519 instruction covers.
- Borsh serialization for the struct itself, for carrying it in instruction data.
- Sign and verify helpers behind a non-BPF feature so the service and tests can use them without pulling `ed25519-dalek` into the program.

## Why the extra fields

`user_pubkey` binds a proof to one target account, so a proof obtained for a routine connect can't be replayed into another operation on a different account within the same epoch (#4192, item 2). The prefix and version byte make the signed bytes unambiguous and give a clean path to a v2 layout later — including IPv6, which is out of scope now.

## Acceptance

- Round-trip Borsh tests and a `signed_message()` byte-layout test with a hardcoded expected vector, so a future refactor that reorders fields fails loudly.
- Sign/verify round trip against a fixed test keypair.
- Committed test vectors (proof JSON + expected signed message hex + signature) that the program tests, CLI tests, and service tests all assert against.
- Crate builds under `cargo build-sbf` as a dependency of the program.

elitegreg:
Resolves: #4195

Part of [RFC-27](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc27-ip-verification.md) (`rfcs/rfc27-ip-verification.md`), tracker #4194.

## Summary of Changes

- Adds `crates/doublezero-ip-proof`, defining `IpOwnershipProof` and the exact bytes the IP verification service signs. Three independent consumers — the serviceability program (BPF), the CLI, and the verification service — must agree byte for byte or every proof fails validation, so the format is defined once rather than duplicated three times.
- The signed message is a fixed 57-byte layout: a `DZ_IP_PROOF` domain-separation prefix, a version byte, then `payer || client_ip || epoch || user_type`. No length prefixes and no Borsh on the signed portion, so the program can reconstruct it on the stack and compare against what the Ed25519 precompile instruction covers.
- The layout version travels **in** the proof, and `signed_message_for()` takes it as an argument, so a verifier reconstructs the bytes for the version a proof declares rather than for whatever version it was compiled against. `SUPPORTED_IP_PROOF_VERSIONS` gives the program one place to accept a set. A v2 rollout can therefore teach verifiers both versions before switching issuers, instead of needing an atomic cutover across the program, the service, and every deployed CLI inside one freshness window.
- `user_type` is bound rather than the derived User pubkey. The User account is a PDA over `(client_ip, user_type)`, so `client_ip` alone does not pin the account being created — a proof for an IP would otherwise be reusable for a different connection type on that IP. Binding the User pubkey instead would not work: the service is the party that discovers `client_ip`, so the client would have to supply a pubkey derived from its own autodetected IP, and on any NATed or multi-homed host that pubkey disagrees with the IP the service observed, producing a validly signed proof that can never verify.
- Off-chain sign/verify helpers sit behind a non-default `signer` feature. The default build's dependency surface is exactly borsh and `solana-program` — `thiserror` and `IpProofError` are gated with the helpers that need them.
- Committed test vectors behind a `test-vectors` feature, for the program, CLI, and service tests to assert against as those land.
- Updates RFC-27's proof specification to this layout and names the crate as normative, so the verification service (#4198) cannot be written from prose that disagrees with the code. The request body gains `user_type`; onchain validation gains a supported-version check and a `user_type` match.
- Adds a per-crate `cargo hack clippy --each-feature` pass to `rust-lint`. The workspace pass runs with default features only, so a crate whose default feature set is empty would otherwise have most of its code exempt from clippy.

Nothing consumes the crate yet; this is a foundation change with no behavior impact. Onchain validation (#4197), the `GlobalState` verifier key (#4196), the service (#4198), and the client path (#4200/#4201) build on it.

### Note on the issue text

The issue named `ed25519-dalek` for the sign/verify helpers. This uses `solana-keypair` / `solana-signer` / `solana-signature` instead — already resolved in `Cargo.lock`, and the verification service will hold a Solana keypair JSON, so `sign_message()` / `Signature::verify()` are the natural fit with no new crypto dependency. Same intent: the helpers stay behind a non-default feature so nothing extra reaches the program. Test vectors are Rust constants rather than a JSON file, since all three consumers named in the acceptance criteria are Rust.

## Diff Breakdown

| Core logic   |     2 | +198 / -0   | +198 |
| Tests        |     3 | +372 / -0   | +372 |
| Docs         |     2 | +73 / -18   |  +55 |
| Config/build |     3 | +53 / -0    |  +53 |
| Generated    |     1 | +13 / -0    |  +13 |
| **Total**    |     9 | +709 / -18  | +691 |

Two thirds of the diff is tests and committed vectors; the format itself is ~200 lines. (Core and test counts split within `lib.rs` and `signer.rs`, which carry inline `#[cfg(test)]` modules.)

<details>

- [`crates/doublezero-ip-proof/src/lib.rs`](https://github.com/malbeclabs/doublezero/pull/4206/files#diff-d4d650ceadb602f65a20d6dfa0d7bd44f4aee33475970e0c54a3825b28dff3d9) — the `IpOwnershipProof` struct, layout constants derived from the field offsets, version support set, and `signed_message_for()`, plus byte-layout and Borsh round-trip tests
- [`crates/doublezero-ip-proof/src/test_vectors.rs`](https://github.com/malbeclabs/doublezero/pull/4206/files#diff-bf9a7fc69db7c8d337777ec98425c31a89ea67d80ef2c14509b257f62efd1ccb) — two committed vectors (specific IP at the default `user_type`, and a wildcard-pass connect with a high epoch and non-zero `user_type`) signed by a fixed test keypair
- [`crates/doublezero-ip-proof/src/signer.rs`](https://github.com/malbeclabs/doublezero/pull/4206/files#diff-e66097b68664b56cac53fc76073d633c44f1f205d2a28e544cdc3dce80688788) — `sign()` / `sign_version()` / `verify()` behind the `signer` feature, off-chain only
- [`rfcs/rfc27-ip-verification.md`](https://github.com/malbeclabs/doublezero/pull/4206/files#diff-9f9110b47a1525a093f13a2a4644490e1ee50f8284596e1b83f2b25e0b228386) — proof spec updated to the 57-byte layout, with the versioning-rollout and `user_type` rationale
- [`crates/doublezero-ip-proof/Cargo.toml`](https://github.com/malbeclabs/doublezero/pull/4206/files#diff-d72436b57edc750568012bc8f948ada8042107f0573109c151de44a526516854) — feature split keeping the default build BPF-clean
- [`Makefile`](https://github.com/malbeclabs/doublezero/pull/4206/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52) — per-crate `--each-feature` clippy pass

</details>

## Testing Verification

- Byte-layout test asserts `signed_message()` against a hardcoded expected vector, with offset assertions on the prefix, version byte, little-endian epoch, and trailing `user_type`, so a future field reorder fails loudly rather than silently invalidating every proof.
- A proof carrying `version = 2` reconstructs v2 bytes, not `IP_PROOF_VERSION` bytes — the property that makes a staged version rollout possible.
- Sensitivity test confirms each of `version`, `payer`, `client_ip`, `epoch`, and `user_type` changes the signed message, and that `signature` does not.
- Borsh round trip pins the serialized length at 110 bytes and confirms `client_ip` serializes in network order.
- Sign/verify round trip against a fixed keypair; verify rejects a different verifier key and each individually tampered field, including a bumped version.
- Every committed vector re-derives its own signed message and re-produces its own signature, so the vectors cannot drift from the implementation unnoticed.
- `cargo hack clippy -p doublezero-ip-proof --each-feature --all-targets` is clean across all five feature permutations, which is the pass that keeps `signer.rs` and `test_vectors.rs` linted at all.
- Default-feature dependency surface verified with `cargo tree -e normal`: exactly borsh and `solana-program`.
- BPF buildability: temporarily added the crate as a dependency of `doublezero-serviceability` with a function constructing an `IpOwnershipProof`, calling `signed_message()`, Borsh-serializing it, and calling `is_supported_version()`; `cargo build-sbf --tools-version v1.54` produced `doublezero_serviceability.so` successfully, then reverted. The program does not consume the crate until #4197.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ members = [
     "crates/doublezero-serviceability-instruction",
     "crates/doublezero-daemon-cli",
     "crates/doublezero-geolocation-cli",
+    "crates/doublezero-ip-proof",
     "crates/sentinel",
 ]
 default-members = []
@@ -59,6 +60,7 @@ env_logger = "0"
 eyre = "0"
 futures = "0"
 futures-util = "0"
+hex = "0.4"
 http = "1"
 http-body-util = "0"
 hyper = "1"
@@ -85,6 +87,7 @@ solana-client = "3.0"
 solana-bincode = "3.0"
 solana-commitment-config = "3.0"
 solana-compute-budget-interface = "3.0"
+solana-keypair = "3.1"
 solana-loader-v3-interface = { version = "6", features = ["serde"] }
 solana-program = "3.0"
 solana-program-test = "=3.0.12"
@@ -92,6 +95,8 @@ solana-pubsub-client = "3.0"
 solana-rpc-client-api = "3.0"
 solana-sdk = "=3.0"
 solana-sdk-ids = "3.0"
+solana-signature = "3.4"
+solana-signer = "3.0"
 solana-system-interface = { version = "3", features = ["bincode"] }
 solana-transaction-status = "3.0"
 bitflags = "2"
@@ -113,6 +118,7 @@ doublezero-cli-core = { path = "crates/doublezero-cli-core" }
 doublezero-config = { path = "config" }
 doublezero-daemon-cli = { path = "crates/doublezero-daemon-cli" }
 doublezero-geolocation-cli = { path = "crates/doublezero-geolocation-cli" }
+doublezero-ip-proof = { path = "crates/doublezero-ip-proof" }
 # default-features = false MUST be set here, not on the consumer edge: an inheriting
 # member's `default-features = false` is ignored unless the workspace dep declares it.
 # This keeps sentinel's server-only deps (Prometheus exporter → rustls/aws-lc-sys) out of
--- a/crates/doublezero-ip-proof/Cargo.toml
+++ b/crates/doublezero-ip-proof/Cargo.toml
@@ -16,23 +16,28 @@ name = "doublezero_ip_proof"
 
 # The RFC-27 IpOwnershipProof layout, shared by three independent consumers that
 # must agree byte for byte: the serviceability program (BPF), the CLI, and the
-# verification service. The default build stays BPF-clean — only borsh,
-# solana-program's Pubkey, and thiserror. Signing/verification lives behind the
-# `signer` feature so no crypto crate reaches the program.
+# verification service. The default build stays BPF-clean — only borsh and
+# solana-program's Pubkey. Signing/verification lives behind the `signer`
+# feature so no crypto crate reaches the program.
 [dependencies]
 borsh.workspace = true
 solana-program.workspace = true
-thiserror.workspace = true
 
 hex = { workspace = true, optional = true }
 solana-keypair = { workspace = true, optional = true }
 solana-signature = { workspace = true, optional = true }
 solana-signer = { workspace = true, optional = true }
+thiserror = { workspace = true, optional = true }
 
 [features]
 default = []
 # Off for the BPF build; the program only reconstructs the signed bytes and
 # compares them against what the Ed25519 precompile instruction covers.
-signer = ["dep:solana-keypair", "dep:solana-signature", "dep:solana-signer"]
+signer = [
+  "dep:solana-keypair",
+  "dep:solana-signature",
+  "dep:solana-signer",
+  "dep:thiserror",
+]
 # Committed cross-consumer test vectors. Enabled by consumers in dev-dependencies.
 test-vectors = ["dep:hex"]
--- a/crates/doublezero-ip-proof/Cargo.toml
+++ b/crates/doublezero-ip-proof/Cargo.toml
@@ -1,0 +1,38 @@
+[package]
+name = "doublezero-ip-proof"
+
+# clippy ignores rust-version.toml
+rust-version = "1.84.1" # cargo build-sbf --version
+
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+repository.workspace = true
+
+[lib]
+name = "doublezero_ip_proof"
+
+# The RFC-27 IpOwnershipProof layout, shared by three independent consumers that
+# must agree byte for byte: the serviceability program (BPF), the CLI, and the
+# verification service. The default build stays BPF-clean — only borsh,
+# solana-program's Pubkey, and thiserror. Signing/verification lives behind the
+# `signer` feature so no crypto crate reaches the program.
+[dependencies]
+borsh.workspace = true
+solana-program.workspace = true
+thiserror.workspace = true
+
+hex = { workspace = true, optional = true }
+solana-keypair = { workspace = true, optional = true }
+solana-signature = { workspace = true, optional = true }
+solana-signer = { workspace = true, optional = true }
+
+[features]
+default = []
+# Off for the BPF build; the program only reconstructs the signed bytes and
+# compares them against what the Ed25519 precompile instruction covers.
+signer = ["dep:solana-keypair", "dep:solana-signature", "dep:solana-signer"]
+# Committed cross-consumer test vectors. Enabled by consumers in dev-dependencies.
+test-vectors = ["dep:hex"]
--- a/crates/doublezero-ip-proof/src/lib.rs
+++ b/crates/doublezero-ip-proof/src/lib.rs
@@ -1,6 +1,6 @@
 //! The RFC-27 `IpOwnershipProof`: a signed attestation from the DoubleZero IP verification
 //! service that a given payer originated a request from a given client IP around a given epoch,
-//! for a specific User account.
+//! for a given user type.
 //!
 //! The signed bytes are consumed by three independent places — the serviceability program (BPF),
 //! the CLI, and the verification service — which must agree byte for byte or every proof fails
@@ -16,7 +16,7 @@ use std::net::Ipv4Addr;
 #[cfg(feature = "signer")]
 mod signer;
 #[cfg(feature = "signer")]
-pub use signer::{sign, verify};
+pub use signer::{sign, sign_version, verify};
 
 #[cfg(feature = "test-vectors")]
 pub mod test_vectors;
@@ -25,34 +25,51 @@ pub mod test_vectors;
 /// message a verifier key might ever be asked to sign.
 pub const IP_PROOF_DOMAIN: &[u8; 11] = b"DZ_IP_PROOF";
 
-/// Layout version of the signed message. Bump for any change to the field set or ordering — for
-/// example a v2 that carries an IPv6 client address.
+/// Layout version an issuer writes into [`IpOwnershipProof::version`]. Bump for any change to the
+/// field set or ordering — for example a v2 that carries an IPv6 client address.
 pub const IP_PROOF_VERSION: u8 = 1;
 
-/// Length of the signed message: prefix(11) + version(1) + payer(32) + client_ip(4) + epoch(8) +
-/// user_pubkey(32).
-pub const SIGNED_MESSAGE_LEN: usize = 88;
+/// Versions a verifier accepts. The version travels in the proof, so a new layout can be rolled
+/// out by teaching verifiers to accept both before issuers start writing the new one; drop the old
+/// entry once no outstanding proof can still be inside its freshness window.
+pub const SUPPORTED_IP_PROOF_VERSIONS: &[u8] = &[IP_PROOF_VERSION];
+
+/// Whether a verifier accepts proofs at this layout version.
+pub fn is_supported_version(version: u8) -> bool {
+    SUPPORTED_IP_PROOF_VERSIONS.contains(&version)
+}
 
 const DOMAIN_END: usize = IP_PROOF_DOMAIN.len();
 const VERSION_END: usize = DOMAIN_END + 1;
 const PAYER_END: usize = VERSION_END + 32;
 const CLIENT_IP_END: usize = PAYER_END + 4;
 const EPOCH_END: usize = CLIENT_IP_END + 8;
+const USER_TYPE_END: usize = EPOCH_END + 1;
+
+/// Length of the signed message: prefix(11) + version(1) + payer(32) + client_ip(4) + epoch(8) +
+/// user_type(1). Derived from the field offsets so inserting a field cannot leave the length
+/// behind.
+pub const SIGNED_MESSAGE_LEN: usize = USER_TYPE_END;
 
 /// A signed attestation of IP ownership, carried in instruction data alongside the Ed25519
 /// precompile instruction that actually verifies [`IpOwnershipProof::signature`].
 #[derive(BorshSerialize, BorshDeserialize, Debug, Clone, Copy, PartialEq, Eq)]
 pub struct IpOwnershipProof {
+    /// Layout version of the signed message, written by the issuer. Verifiers reconstruct the
+    /// bytes for *this* version and reject a version they do not support.
+    pub version: u8,
     /// The account paying for / owning the user being created.
     pub payer: Pubkey,
     /// The address the verification service observed the request originate from.
     pub client_ip: Ipv4Addr,
     /// The DoubleZero ledger epoch the proof was issued in.
     pub epoch: u64,
-    /// The User account this proof authorizes. Binding it stops a proof obtained for a routine
-    /// connect from being replayed into another operation on a different account in the same
-    /// epoch.
-    pub user_pubkey: Pubkey,
+    /// Discriminant of the serviceability `UserType` this proof authorizes, kept as a `u8` so this
+    /// crate does not depend on the program it is consumed by. Binding it stops a proof obtained
+    /// for one connection type from being replayed into a different one on the same IP in the same
+    /// epoch — the User PDA is `f(client_ip, user_type)`, so `client_ip` alone does not pin the
+    /// account being created.
+    pub user_type: u8,
     /// Ed25519 signature by the verifier key over [`IpOwnershipProof::signed_message`].
     pub signature: [u8; 64],
 }
@@ -60,7 +77,13 @@ pub struct IpOwnershipProof {
 impl IpOwnershipProof {
     /// The exact bytes the verifier signed. See [`signed_message_for`].
     pub fn signed_message(&self) -> [u8; SIGNED_MESSAGE_LEN] {
-        signed_message_for(&self.payer, &self.client_ip, self.epoch, &self.user_pubkey)
+        signed_message_for(
+            self.version,
+            &self.payer,
+            &self.client_ip,
+            self.epoch,
+            self.user_type,
+        )
     }
 }
 
@@ -73,28 +96,30 @@ impl IpOwnershipProof {
 /// ```text
 /// offset  len  field
 ///      0   11  b"DZ_IP_PROOF"
-///     11    1  version (1)
+///     11    1  version
 ///     12   32  payer
 ///     44    4  client_ip, network order
 ///     48    8  epoch, little-endian
-///     56   32  user_pubkey
+///     56    1  user_type
 /// ```
 pub fn signed_message_for(
+    version: u8,
     payer: &Pubkey,
     client_ip: &Ipv4Addr,
     epoch: u64,
-    user_pubkey: &Pubkey,
+    user_type: u8,
 ) -> [u8; SIGNED_MESSAGE_LEN] {
     let mut message = [0u8; SIGNED_MESSAGE_LEN];
     message[..DOMAIN_END].copy_from_slice(IP_PROOF_DOMAIN);
-    message[DOMAIN_END] = IP_PROOF_VERSION;
+    message[DOMAIN_END] = version;
     message[VERSION_END..PAYER_END].copy_from_slice(payer.as_ref());
     message[PAYER_END..CLIENT_IP_END].copy_from_slice(&client_ip.octets());
     message[CLIENT_IP_END..EPOCH_END].copy_from_slice(&epoch.to_le_bytes());
-    message[EPOCH_END..].copy_from_slice(user_pubkey.as_ref());
+    message[EPOCH_END] = user_type;
     message
 }
 
+#[cfg(feature = "signer")]
 #[derive(thiserror::Error, Debug, PartialEq, Eq)]
 pub enum IpProofError {
     #[error("ip ownership proof signature does not verify against the verifier key")]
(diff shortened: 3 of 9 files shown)


crates/doublezero-ip-proof/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+    /// The account paying for / owning the user being created.
+    pub payer: Pubkey,
+    /// The address the verification service observed the request originate from.
+    pub client_ip: Ipv4Addr,
+    /// The DoubleZero ledger epoch the proof was issued in.
+    pub epoch: u64,
+    /// The User account this proof authorizes. Binding it stops a proof obtained for a routine
+    /// connect from being replayed into another operation on a different account in the same
+    /// epoch.
+    pub user_pubkey: Pubkey,
```
juan-malbeclabs: `user_pubkey` is the User PDA, and the User PDA is `f(program_id, client_ip, user_type)` (`smartcontract/programs/doublezero-serviceability/src/pda.rs:58` — seeds are `[SEED_PREFIX, SEED_USER, ip.octets(), user_type]`). But the service is the party that *discovers* `client_ip`: RFC-27's request body is `{ "payer": ... }` only (`rfcs/rfc27-ip-verification.md:123`), so the CLI has to send a PDA it derived from its own autodetected IP.

On a NATed or multi-homed host where the autodetected IP differs from the IP the service observes, the service signs `client_ip = observed` next to a `user_pubkey` derived from the *other* IP. The program re-derives the User PDA from `proof.client_ip`, they don't match, and the proof fails 100% of the time — indistinguishable from a bad signature. Neither `sign()` nor `signed_message_for()` checks the two fields agree, so the crate happily mints permanently-invalid proofs.

Two ways out:

1. Add `user_type` to the RFC-27 request body and have the service derive the PDA itself from the IP it observed. The mismatch becomes structurally impossible.
2. Drop `user_pubkey`. `client_ip` is already in the signed message, so the only *new* binding this field adds is `user_type` (plus the program id) — narrower than the replay protection the doc comment and the PR body claim. If `user_type` is the thing worth binding, bind `user_type` directly.

Either way, please don't leave a field whose only valid value is one the issuer cannot compute.

crates/doublezero-ip-proof/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+mod signer;
+#[cfg(feature = "signer")]
+pub use signer::{sign, verify};
+
+#[cfg(feature = "test-vectors")]
+pub mod test_vectors;
+
+/// Domain-separation prefix. Keeps the signed bytes from colliding with any other DoubleZero
+/// message a verifier key might ever be asked to sign.
+pub const IP_PROOF_DOMAIN: &[u8; 11] = b"DZ_IP_PROOF";
```
juan-malbeclabs, line 26: This message layout diverges from RFC-27 and the RFC isn't updated in this PR. RFC-27 §Proof Specification (`rfcs/rfc27-ip-verification.md:149-172`) specifies `{payer, client_ip, epoch, signature}` and tells the program to "reconstruct `message = payer || client_ip || epoch`" — no domain prefix, no version byte, no `user_pubkey`. That's 44 bytes against the 88 here.

The crate's whole premise is that three independent consumers agree byte for byte, and the verification service (#4198) is the consumer most likely to be written from the RFC rather than from this crate. A service implementing the RFC as written signs 44 bytes while the program reconstructs 88, and every proof is rejected.

The prefix and version byte are genuine improvements — keep them, and update RFC-27's proof spec in this PR to the 88-byte layout, pointing at `signed_message_for` as the normative definition.

crates/doublezero-ip-proof/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+
+/// Layout version of the signed message. Bump for any change to the field set or ordering — for
+/// example a v2 that carries an IPv6 client address.
+pub const IP_PROOF_VERSION: u8 = 1;
+
+/// Length of the signed message: prefix(11) + version(1) + payer(32) + client_ip(4) + epoch(8) +
+/// user_pubkey(32).
+pub const SIGNED_MESSAGE_LEN: usize = 88;
```
juan-malbeclabs: `88` is a literal while every other offset below is derived. A future v2 that changes the prefix or inserts a field, without also editing this literal, turns `message[EPOCH_END..].copy_from_slice(user_pubkey.as_ref())` into a length-mismatch panic — onchain, a program abort rather than a compile error. The test pinning `88` doesn't help: the literal and the offsets move together in the author's head, not in the code.

```suggestion
pub const SIGNED_MESSAGE_LEN: usize = EPOCH_END + 32;
```

(Module-level consts are order-independent, so the forward reference to `EPOCH_END` is fine. If you'd rather keep the literal as documentation, `const _: () = assert!(EPOCH_END + 32 == SIGNED_MESSAGE_LEN);` gets the same build-time guarantee.)

crates/doublezero-ip-proof/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+#[cfg(feature = "test-vectors")]
+pub mod test_vectors;
+
+/// Domain-separation prefix. Keeps the signed bytes from colliding with any other DoubleZero
+/// message a verifier key might ever be asked to sign.
+pub const IP_PROOF_DOMAIN: &[u8; 11] = b"DZ_IP_PROOF";
+
+/// Layout version of the signed message. Bump for any change to the field set or ordering — for
+/// example a v2 that carries an IPv6 client address.
+pub const IP_PROOF_VERSION: u8 = 1;
```
juan-malbeclabs, line 30: The version byte doesn't travel with the proof, so it can't do what this doc comment says. It isn't a field of `IpOwnershipProof` and `signed_message_for()` takes no version argument, so `signed_message()` always reconstructs the *current* version's bytes.

The moment this constant becomes `2`, every outstanding v1 proof stops verifying, and neither the program nor the service has any way to accept v1 and v2 at once. The rollout becomes an atomic cutover across the program upgrade, the service, and every deployed CLI, inside one epoch freshness window — which is the situation a version byte is supposed to prevent.

Fix option — carry it in the proof and let callers reconstruct any version:

```rust
pub struct IpOwnershipProof {
    pub version: u8,
    ...
}

pub fn signed_message_for(
    version: u8,
    payer: &Pubkey,
    ...
)
```

with `IP_PROOF_VERSION` as the value the *issuer* writes, and the program accepting a set of supported versions. Alternatively, if concurrent versions are explicitly not a goal, say so in the doc comment so nobody plans a migration around a capability that isn't there.

crates/doublezero-ip-proof/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+    message[PAYER_END..CLIENT_IP_END].copy_from_slice(&client_ip.octets());
+    message[CLIENT_IP_END..EPOCH_END].copy_from_slice(&epoch.to_le_bytes());
+    message[EPOCH_END..].copy_from_slice(user_pubkey.as_ref());
+    message
+}
+
+#[derive(thiserror::Error, Debug, PartialEq, Eq)]
+pub enum IpProofError {
+    #[error("ip ownership proof signature does not verify against the verifier key")]
+    InvalidSignature,
```
juan-malbeclabs, line 126: `InvalidSignature` is produced solely by `signer::verify()`. With `default = []` the BPF build gets an error enum nothing can construct, plus the `thiserror` dependency the Cargo.toml comment is trying to keep out of the program.

```suggestion
#[cfg(feature = "signer")]
#[derive(thiserror::Error, Debug, PartialEq, Eq)]
pub enum IpProofError {
    #[error("ip ownership proof signature does not verify against the verifier key")]
    InvalidSignature,
}
```

Pair it with `thiserror = { workspace = true, optional = true }` and `signer = ["dep:thiserror", ...]` so the default surface really is just borsh + `solana-program`.

crates/doublezero-ip-proof/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+solana-program.workspace = true
+thiserror.workspace = true
+
+hex = { workspace = true, optional = true }
+solana-keypair = { workspace = true, optional = true }
+solana-signature = { workspace = true, optional = true }
+solana-signer = { workspace = true, optional = true }
+
+[features]
+default = []
```
juan-malbeclabs, line 33: With `default = []`, nothing behind the non-default features is ever linted. `make rust-lint` runs `cargo hack clippy --workspace --all-targets ... -Dclippy::all -Dwarnings` (`Makefile:97`) with no feature flags, and `cargo hack` without `--each-feature`/`--feature-powerset` behaves like plain `cargo` — default features only. So `signer.rs` and `test_vectors.rs`, roughly two thirds of this crate, are exempt from clippy; a violation there lands green and only ever surfaces as a compile error via `cargo test --all-features`.

Fix option — add a per-crate pass to the `rust-lint` target:

```make
cargo hack clippy -p doublezero-ip-proof --each-feature --all-targets -- -Dclippy::all -Dwarnings
```

(`--workspace --each-feature` over the whole tree would be far too slow; scoping it to this crate is cheap.)

juan-malbeclabs (commented):
Reviewed at high effort. Verified locally before commenting: `cargo build -p doublezero-ip-proof --features signer` succeeds standalone (`solana-signature`'s `verify` feature arrives transitively via `solana-keypair`), `cargo test -p doublezero-ip-proof --features signer,test-vectors` passes all 11 tests, and `make rust-test` uses `--all-features` so the feature-gated tests do run in CI. The byte layout and offsets check out (11+1+32+4+8+32 = 88), both committed vectors decode and verify, borsh's `Ipv4Addr` impl is `core::net`-based so it's BPF-safe, and `Signature::verify` uses `verify_strict`, matching the Ed25519 precompile.

Two findings worth resolving before merge — the `user_pubkey` derivation and the un-updated RFC-27 proof spec. The other four are smaller. Each comment carries a concrete fix option.

crates/doublezero-ip-proof/src/lib.rs
elitegreg: Good catch, and the causality argument is the convincing part: the correct value depends on data that does not exist until after the request it has to be part of. Went with option 2.

`user_pubkey` is replaced by `user_type: u8` (the serviceability `UserType` discriminant, kept as a raw `u8` so this crate does not depend on the program that will depend on it). Signed message is 57 bytes now.

You are also right that the doc comment overclaimed. Since the User PDA is `f(client_ip, user_type)` and `client_ip` was already signed, `user_type` is the entire marginal binding — a different account with the same IP and the same `user_type` cannot exist. The comment now says what it actually buys: it stops a proof obtained for one connection type from being replayed into a different one on the same IP in the same epoch.

Option 1 would also have closed the mismatch, but it puts the program id and the PDA derivation into the verification service. `user_type` is client-supplied, which is fine — it is not a trust boundary, and the program checks `proof.user_type` against the user it is creating (added to the RFC's required checks).

crates/doublezero-ip-proof/src/lib.rs
elitegreg: Agreed, and the RFC is updated in this PR rather than left to drift. The prefix and version byte stay.

RFC-27's Proof Specification now carries the byte-offset table for the 57-byte layout and states that `crates/doublezero-ip-proof` is normative — `signed_message_for()` is the source of truth and the crate wins where the two disagree. Also updated: the request body (`{ payer, user_type }`), the response shape, the flow diagram, the required onchain checks, the rejection conditions, the Replay security bullet, and the open question that asked whether to bind the user account. The versioning-rollout and why-not-the-User-pubkey rationale are written down there too, so #4198 cannot be built from prose that contradicts the code.

crates/doublezero-ip-proof/src/lib.rs
elitegreg: Fixed. There is now a `USER_TYPE_END` in the offset chain and `SIGNED_MESSAGE_LEN = USER_TYPE_END`, so the length moves with the fields. The sizes stay in the doc comment as documentation.

crates/doublezero-ip-proof/src/lib.rs
elitegreg: Took the first option — concurrent versions are worth having.

`version` is now a field of `IpOwnershipProof`, and `signed_message_for(version, payer, client_ip, epoch, user_type)` takes it as an argument, so `signed_message()` reconstructs the version the proof declares rather than whatever the verifier was compiled against. `IP_PROOF_VERSION` is the value an issuer writes; `SUPPORTED_IP_PROOF_VERSIONS` plus `is_supported_version()` give the program one place to accept a set, and the supported-version check is now in RFC-27's required checks and rejection conditions.

That makes the rollout staged: teach the program and the service to accept both, switch issuers over, then drop the old entry once no outstanding proof can still be inside its freshness window. `sign()` issues at `IP_PROOF_VERSION` and `sign_version()` covers the window where two are in flight. There is a test asserting a `version = 2` proof reconstructs v2 bytes, since that is the whole property.

crates/doublezero-ip-proof/src/lib.rs
elitegreg: Fixed as suggested. `IpProofError` is behind `#[cfg(feature = "signer")]`, `thiserror` is `{ workspace = true, optional = true }`, and `signer` gained `"dep:thiserror"`. Verified rather than assumed — `cargo tree -p doublezero-ip-proof --depth 1 -e normal` on the default build is now exactly `borsh` and `solana-program`.

crates/doublezero-ip-proof/Cargo.toml
elitegreg: You are right, and this is the finding that would have kept hiding the others. Added the scoped pass to `rust-lint`, with a comment explaining why the workspace pass above it does not cover this crate:

```make
cargo hack clippy -p doublezero-ip-proof --each-feature --all-targets -- -Dclippy::all -Dclippy::unreadable_literal -Dwarnings
```

Kept `-Dclippy::unreadable_literal` to match the workspace pass. It runs five permutations (all-features, no-default-features, default, signer, test-vectors) and all five are clean.

Cargo.toml
```diff
@@ -85,13 +87,16 @@ solana-client = "3.0"
...
 solana-compute-budget-interface = "3.0"
+solana-keypair = "3.1"
 solana-loader-v3-interface = { version = "6", features = ["serde"] }
 solana-program = "3.0"
 solana-program-test = "=3.0.12"
 solana-pubsub-client = "3.0"
 solana-rpc-client-api = "3.0"
 solana-sdk = "=3.0"
 solana-sdk-ids = "3.0"
+solana-signature = "3.4"
```
juan-malbeclabs, line 98: Minor, non-blocking: this needs the `verify` feature declared explicitly.

`signer.rs:64` calls `Signature::verify`, which solana-signature 3.4 gates behind `#[cfg(any(test, feature = "verify"))]` (`verify = ["dep:ed25519-dalek"]` in its manifest). We declare `solana-signature = "3.4"` with no features, so the build only works because solana-keypair 3.1.2 happens to declare it as `features = ["std", "verify"]`. That is a transitive dependency's private choice holding up our compile — a keypair patch release that drops or reshuffles it breaks the `signer` feature with a `no method named 'verify'` error pointing at our code rather than at the cause.

```suggestion
solana-signature = { version = "3.4", features = ["verify"] }
```

No resolution change today; feature unification already gives us the same graph. It just stops depending on someone else's feature list.

rfcs/rfc27-ip-verification.md
```diff
@@ -235,9 +287,10 @@ may bind:
 - **Centralized trust root.** The verifier keypair is a DoubleZero‑operated authority. The benefit
   delivered is automated, non‑bypassable per‑IP verification — not decentralization. Key rotation is
   supported via global state.
-- **Replay.** The epoch window bounds proof reuse; because the proof binds `payer` and `client_ip`,
-  reuse only re‑asserts the same binding. A nonce or binding to the specific user account can further
-  constrain reuse if needed (see Open Questions).
+- **Replay.** The epoch window bounds proof reuse; because the proof binds `payer`, `client_ip`, and
```
juan-malbeclabs, line 290: Minor, non-blocking, docs only: worth adding a bullet here that the verifier keypair MUST be distinct per environment.

The `DZ_IP_PROOF` prefix separates these bytes from other DoubleZero messages, but nothing in the signed message binds the cluster — no program id, no environment tag. So a proof minted against DZ testnet is byte-identical to a valid one on mainnet, and vice versa, for the same `payer`/`client_ip`/`epoch`/`user_type`.

Distinct verifier keypairs per environment make that structurally impossible, and since the pubkey lives in per-environment global state that is almost certainly what will happen anyway. Right now it is an implicit operational requirement, and it is the only thing standing between the two environments — a shared key during, say, a staging shortcut would silently make proofs portable. One sentence in this section is enough; no code change.

juan-malbeclabs (approved):
Thanks for these — all six findings are resolved, and the two design calls are the right ones rather than just workable ones.

Option 2 on `user_type` is correct for the reason you gave: the User PDA is `f(client_ip, user_type)`, so `user_type` is exactly the marginal binding the User pubkey would have added, minus the program id — and it keeps PDA derivation out of the verification service, which is the part that would have been awkward to own. Carrying `version` in the proof is what actually makes a staged rollout possible, and the test asserting a `version = 2` proof reconstructs v2 bytes pins the property that matters. The `--each-feature` pass is the one that keeps the rest honest.

Verified locally rather than assumed: `cargo test -p doublezero-ip-proof --all-features` passes 14/14; `cargo hack clippy -p doublezero-ip-proof --each-feature --all-targets` is clean across all five permutations; `cargo tree -e normal --depth 1` on the default build is exactly borsh + `solana-program`, so the `thiserror` gating did what it was meant to. Grepped RFC-27 for leftovers of the old 44-byte layout and `user_pubkey` — none, and the crate is not in `rust-test`'s exclude list, so the feature-gated tests do run in CI.

Two minors below, neither blocking. The `solana-signature` one is worth taking before it turns into a confusing build break for someone else; the RFC one is a single sentence.

Issue #4194, opened 2026-08-17
RFC-27: IP ownership verification for user connection (tracker)


elitegreg:
Tracking issue for implementing [RFC-27](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc27-ip-verification.md) (`rfcs/rfc27-ip-verification.md`, added in #2304).

## Why

`create_user_core` accepts any globally-routable `client_ip` when the AccessPass is a wildcard pass — stored at `0.0.0.0` or flagged `allow_multiple_ip`, which includes the `EdgeSeat` passes the shred-oracle issues. `is_global()` is the only check applied. That leaves two holes: squatting on the `(client_ip, user_type)` User PDA to lock out the legitimate operator, and pointing device tunnel/route provisioning at a third party's address.

RFC-27 closes them with a signed `IpOwnershipProof` from a DoubleZero-operated verification service, validated onchain through the Ed25519 precompile so it cannot be bypassed by calling the program directly.

## Decisions made ahead of implementation

These resolve RFC open questions and the review points in #4192. The RFC text is amended to match as part of the onchain validation issue.

| Question | Decision |
| --- | --- |
| Replay binding | Signed message is `payer \|\| client_ip \|\| epoch \|\| user_pubkey`, with a domain-separation prefix and version byte. Binding the target User PDA stops a routine connect proof from being replayed into a destructive operation (see #4192 item 2 and the future ReclaimUser work in #4190). |
| Freshness window | Accept `clock.epoch` and `clock.epoch - 1`. Fixed constant, not configurable. |
| Rollout | `FeatureFlag::RequireIpOwnershipProof` (bit 2). Proof optional until the flag is set per environment, then required for every user creation — wildcard and specific-IP alike. |
| Persist proofs onchain | No. The bound `client_ip` plus transaction history is the audit record. |
| IPv6 | Out of scope. The program's `client_ip` surface is `Ipv4Addr` and the User PDA derives from it. |
| Periodic re-verification | Out of scope here. Re-proving on a schedule and releasing a user whose IP changed hands is tracked in #4193 / #4190. |
| Key rotation | Verifier pubkey lives in `GlobalState` and rotates through the existing `SetAuthority` instruction. No cadence set; rotation invalidates outstanding proofs and clients re-verify on next connect. |

## Work

Foundations:
- [x] #4195 — `crates/doublezero-ip-proof`: canonical proof type and signed-message layout
[...]

PR #4207, opened 2026-08-17
serviceability: add ip_verifier_authority_pk to GlobalState and rotate it via SetAuthority

### LINKED ISSUE #4196: serviceability: add ip_verifier_authority_pk to GlobalState and rotate it via SetAuthority

elitegreg:
Closes #4196. Part of RFC-27 ([rfcs/rfc27-ip-verification.md](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc27-ip-verification.md)); tracker #4194.

## Summary

- `GlobalState` carries `ip_verifier_authority_pk`, the trust root for RFC-27 IP ownership proof validation, so the verifier key rotates without a program upgrade.
- `SetAuthority` gains an optional `ip_verifier_authority_pk`; authorization is unchanged (GLOBALSTATE_ADMIN permission or the legacy foundation path).
- `doublezero global-config authority set --ip-verifier-authority <pubkey|me>` rotates the key, and `authority get` reports it in both the table and JSON views.
- The Rust SDK `SetAuthorityCommand` forwards the new field.
- `Pubkey::default()` means "no verifier configured". Enforcement, which lands with the onchain validation issue, must treat that as a hard reject rather than as "any signature passes".

## Details

The field is append-only, so accounts written before it existed deserialize with `Pubkey::default()` through the same `unwrap_or_default()` path `feed_authority_pk` uses. On the instruction side, `BorshDeserializeIncremental` keeps transactions built against the old argument layout decoding, with the new field defaulting to `None`.

SDK deserializer work for Go, TypeScript, and Python is tracked separately. The `global_state` fixture is regenerated here because CI's `make check-fixtures` diffs the goldens against the Rust producer; all three readers ignore trailing bytes and skip meta fields they do not map, so they stay green until that work lands. Running the generator also refreshed its `Cargo.lock` from `0.31.0` to `0.36.0`.

`smartcontract/cli/src/init.rs` needed no change: it sends `InitGlobalState` and carries no authority arguments, and the program's initializer seeds the new field to `Pubkey::default()`.

## Testing Verification

- New `tests/setauthority_test.rs`: rotating the verifier key leaves the other authorities, the foundation allowlist, and the feature flags untouched; a second rotation replaces the previous key; and a later `SetAuthority` passing `None` does not clear it. It also asserts a freshly initialized `GlobalState` reads `Pubkey::default()`, the state enforcement must reject.
- Backwards compatibility against the committed base64 vector: it reads back `Pubkey::default()`, and re-serialized with the field set it round-trips with every other field identical. Note that this stored vector is old enough that its bytes end at `user_airdrop_lamports`, so appending 32 bytes to it would not have exercised the new field.
- `SetAuthorityArgs` unit tests: a pre-field four-option encoding still decodes with `ip_verifier_authority_pk: None`, and the new field round-trips.
- Go, TypeScript, and Python SDK suites pass against the regenerated fixture.


smartcontract/programs/doublezero-serviceability/src/processors/globalstate/setauthority.rs
```diff
@@ -95,6 +100,9 @@ pub fn process_set_authority(
     if let Some(feed_authority_pk) = value.feed_authority_pk {
         globalstate.feed_authority_pk = feed_authority_pk;
     }
+    if let Some(ip_verifier_authority_pk) = value.ip_verifier_authority_pk {
+        globalstate.ip_verifier_authority_pk = ip_verifier_authority_pk;
+    }
```
juan-malbeclabs, line 105: `Some(Pubkey::default())` is accepted here and written straight through, silently clearing the trust root.

The field's own doc comment defines `Pubkey::default()` as the "no verifier configured, treat as hard reject" sentinel — so an operator who templates an empty or zero pubkey into `--ip-verifier-authority` disables every IP-verified user connection with no error and no distinguishable onchain signal. `Pubkey::from_str("11111111111111111111111111111111")` parses cleanly, so the CLI won't catch it either.

Non-blocking, but the guard is cheap and matches the semantics the field already documents:

```rust
if let Some(pk) = value.ip_verifier_authority_pk {
    if pk == Pubkey::default() {
        return Err(DoubleZeroError::InvalidAccountData.into()); // or the closest existing variant
    }
    globalstate.ip_verifier_authority_pk = pk;
}
```

I'd only guard the new field — retrofitting the other four is a behavior change that doesn't belong in this PR.

smartcontract/cli/src/globalconfig/authority/get.rs
```diff
@@ -41,6 +43,7 @@ impl GetAuthorityCliCommand {
             access_authority: gstate.sentinel_authority_pk,
             feed_authority: gstate.feed_authority_pk,
             health_oracle: gstate.health_oracle_pk,
+            ip_verifier_authority: gstate.ip_verifier_authority_pk,
```
juan-malbeclabs: An unset verifier renders as the literal `11111111111111111111111111111111`, which is indistinguishable from a deliberately configured key.

Concrete case: right after the program upgrade but before the first rotation, an operator runs `doublezero global-config authority get` to confirm "is the verifier set?", sees a valid-looking base58 pubkey, and concludes yes.

Rendering `Pubkey::default()` as `not set` (and `null` under `--json`) would match the sentinel the field documents. Only this field needs it — the other four have no documented sentinel.

sdk/serviceability/testdata/fixtures/global_state.json
```diff
@@ -81,6 +81,11 @@
       "name": "FeedAuthorityPk",
       "value": "QRSsyMWN1yHT9ir42bgNZUNZ4PdEhcSWCrL2AryKpy5",
       "typ": "pubkey"
+    },
+    {
+      "name": "IpVerifierAuthorityPk",
```
juan-malbeclabs, line 86: This golden now advertises `IpVerifierAuthorityPk`, but nothing validates it:

- Go's `GlobalState` struct (`smartcontract/sdk/go/serviceability/state.go:74-88`) stops at `FeedAuthorityPK`, and `fixture_test.go` has no GlobalState case at all.
- The TS harness skips unknown names (`fixtures.test.ts`: `if (!(f.name in got)) continue;`) and Python does the same (`test_fixtures.py`: `if name not in got: continue`).

So these bytes can drift or regress undetected. I know the SDK deserializer work is tracked separately — the point here is just that the golden currently has zero coverage for the field it advertises, which is a slightly worse state than not adding it yet. Adding `IpVerifierAuthorityPK` to the Go struct plus a GlobalState fixture case would close it.

smartcontract/programs/doublezero-serviceability/tests/setauthority_test.rs
```diff
@@ -0,0 +1,151 @@
...
+    instruction::AccountMeta,
+    pubkey::Pubkey,
+    signature::{Keypair, Signer},
+};
+
+mod test_helpers;
+use test_helpers::*;
+
+/// Bring up a program instance with an initialized GlobalState.
+async fn init_globalstate() -> (
```
juan-malbeclabs: Two small things on this file, neither blocking:

1. No test writes `GlobalState` starting from a **legacy-sized** account, so the 32-byte grow + rent top-up that mainnet will actually take on the first post-upgrade write is never exercised. `setup_program_with_globalconfig_context` exposes `set_account`, so seeding the `LEGACY_GLOBALSTATE_VECTOR` bytes (already defined in `state/globalstate.rs`'s test module) and then running `SetAuthority` would cover that path directly. That's the one migration step with real blast radius, and right now only the pure-deserialization half of it is tested.

2. `init_globalstate()` here duplicates the InitGlobalState half of `test_helpers::init_globalstate_and_config` — worth extracting a shared helper instead.

smartcontract/programs/doublezero-serviceability/src/authorize.rs
```diff
@@ -498,6 +498,7 @@ mod tests {
     fn gs_with_feed(authority: &Pubkey) -> GlobalState {
         GlobalState {
             feed_authority_pk: *authority,
+            ip_verifier_authority_pk: Pubkey::default(),
```
juan-malbeclabs: Nit: redundant next to `..GlobalState::default()` on the following line, which already produces `Pubkey::default()` for this field. Same at line 527.

juan-malbeclabs (approved):
Approving — the change is sound and safe to land.

I read every hunk and traced the surrounding code: the `GlobalState` layout and its `TryFrom`, `process_set_authority`, `try_acc_write` → `resize_account_if_needed`, the `borsh-incremental` derive, the SDK fixture readers, and the remaining `GlobalState { .. }` literals in the tree.

The thing I specifically went looking for — the account growing 32 bytes on the first `GlobalState` write after the upgrade — is safe. Every processor that calls `try_acc_write(&globalstate, ..)` (setauthority, setairdrop, setfeatureflags, the four allowlist add/remove, and the five `create_*` index bumps) takes `system_program` as a fixed account, so `resize_account_if_needed`'s `invoke(transfer)` always has it available. The append-only field, the `unwrap_or_default()` read path, and the `BorshDeserializeIncremental` instruction path all behave as advertised, and the legacy-vector test is a good way to pin that down.

No correctness bugs. The inline comments are all non-blocking — the first one is the only one I'd actually push for, and it's a three-line guard. They cluster around one theme: the field documents `Pubkey::default()` as a meaningful sentinel ("no verifier configured → enforcement must hard-reject"), but nothing in the write path, the read path, or the SDK goldens treats it as one yet. Worth closing before RFC-27 enforcement starts depending on that sentinel.

Ran locally: `cargo test -p doublezero-serviceability --lib globalstate` (9 passed) and `--test setauthority_test` (2 passed).

smartcontract/programs/doublezero-serviceability/src/processors/globalstate/setauthority.rs
elitegreg: I understand this, but don't think it represents an actual issue. Setting the authority is something DZF would very rarely and very purposefully.

smartcontract/cli/src/globalconfig/authority/get.rs
elitegreg: Fixed in 3be0b91. `AuthorityDisplay.ip_verifier_authority` is now `Option<Pubkey>`, mapped from the onchain field with `Pubkey::default()` → `None`: the table prints `not set` and `--json` prints `null`. Left the other four alone since they have no documented sentinel.

Added `test_cli_globalconfig_authority_get_unset_ip_verifier`, which asserts the row reads `not set`, does not contain the all-zero base58 key, and that the JSON value is null.

sdk/serviceability/testdata/fixtures/global_state.json
elitegreg: Fixed in 3be0b91. Added `IpVerifierAuthorityPK` to the Go `GlobalState` struct and the read in `DeserializeGlobalState` — `ReadPubkey` returns the all-zero key at EOF, so accounts predating the field decode to the same `Pubkey::default()` sentinel the program uses.

Also added the missing GlobalState fixture cases: `TestFixtureGlobalState` asserts every field against `global_state.json` (so the trailing verifier key pins field order as well as value), and `TestFixtureGlobalStateLegacyNoIpVerifier` drops the last 32 bytes and asserts the zero sentinel.

Left TS/Python to the tracked SDK work as you suggested.

smartcontract/programs/doublezero-serviceability/tests/setauthority_test.rs
elitegreg: Both done in 3be0b91.

1. New `test_setauthority_grows_legacy_sized_globalstate`: uses `setup_program_with_globalconfig_context`, truncates the live GlobalState to the pre-field layout, funds it via `set_account` to exactly the *legacy* rent minimum so the grow needs a top-up, then runs `SetAuthority`. Asserts the account grew by 32 bytes, is rent-exempt at the new size, reads back the rotated key, and that every other field survived.

   I built the legacy bytes by truncation rather than reusing `LEGACY_GLOBALSTATE_VECTOR` — that const lives in a `#[cfg(test)]` module of the lib, so it is not visible to integration tests, and truncation exercises the same grow path without a second hardcoded blob.

2. Extracted `init_globalstate` into `test_helpers`; `init_globalstate_and_config` now calls it, and this file’s local helper (renamed `setup` to avoid shadowing it) does too.

smartcontract/programs/doublezero-serviceability/src/authorize.rs
elitegreg: Removed both in 3be0b91.

PR #4211, opened 2026-08-19
serviceability: validate IpOwnershipProof via the Ed25519 precompile

### LINKED ISSUE #4197: serviceability: validate IpOwnershipProof via the Ed25519 precompile during user creation

elitegreg:
Resolves #4197. Part of RFC-27 ([rfcs/rfc27-ip-verification.md](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc27-ip-verification.md)); tracker #4194.

## Summary

- `CreateUser` and `CreateSubscribeUser` validate an optional RFC-27 `IpOwnershipProof`, so a caller can no longer bind a `client_ip` it cannot originate traffic from. A wildcard access pass — one stored at `0.0.0.0` or flagged `allow_multiple_ip`, which includes the `EdgeSeat` passes the shred-oracle issues — accepts any globally-routable address today, letting a registrant squat the `(client_ip, user_type)` User PDA of an address someone else operates and point that device's tunnel and route provisioning at a third party.
- BPF cannot verify Ed25519 cheaply, so the transaction carries a native `Ed25519SigVerify` instruction and the program introspects the Instructions sysvar to confirm it covers the message this creation implies, signed by `globalstate.ip_verifier_authority_pk`.
- The program **scans** for that instruction rather than reading a fixed index, so a client may place it anywhere and interleave compute-budget instructions. It rejects any instruction whose precompile offsets name a different instruction, run past the end of its data, or that carries other than exactly one signature — an offset pointing elsewhere means the precompile verified bytes the program never reads, which would turn validation into theatre.
- The proof binds the **owner** of the user (`effective_owner`), not the transaction payer. On the ordinary path they are the same account; on the owner-override path they differ, and it is the owner who operates `client_ip` and whom the AccessPass is keyed on.
- Enforcement is gated on the new `require-ip-ownership-proof` feature flag (bit 2). While clear, a missing proof is accepted so pre-RFC-27 clients keep working, but a supplied proof is validated in full either way — a client attaching a broken proof is broken now, not at rollout. With the flag set a proof is required for every creation, wildcard and specific-IP passes alike.
- The proof is also required on the **idempotent rerun path**, which returns early when the user already exists; otherwise an account squatted before the flag was set would absorb every later call without its holder ever demonstrating control of the address.
- **The sentinel authority is exempt from the requirement** (only the requirement — a proof it supplies is still validated in full). See the reviewer note below.
- An unconfigured verifier key (`Pubkey::default()`) is a hard reject, never "any signature passes". A proof whose layout version the program does not support is rejected rather than reconstructed against the version it does know.
- 14 new `DoubleZeroError` variants (105–118), one per rejection class, so an operator can tell a stale proof from a rotated verifier key from a client that never attached the Ed25519 instruction.
- The Instructions sysvar rides as an optional account appended **last** and peeled off the tail by its fixed key, leaving `CreateUser`'s length-detected tenant slot and `CreateSubscribeUser`'s PDA-matched trailing Permission account untouched. The instruction builders append it whenever the args carry a proof, so a caller cannot assemble a transaction that is certain to fail.
- RFC-27 amended: the `user_type` binding and why not the User pubkey, the epoch window, the flag-gated rollout replacing the blanket version-compatibility window, the sentinel exemption, and a Non-Goals section covering lapsed IP control.
- Nothing carries the proof yet — the CLI and SDK work is #4200.

## Notes for reviewers

**The proof binds `user_type`, not the derived User pubkey.** Issue #4197 item 5 asks for `proof.user_pubkey == the derived User PDA`. The already-merged `doublezero-ip-proof` crate (#4195, #4206) binds `user_type` instead, which pins the same account since the PDA is `f(client_ip, user_type)`. Binding the pubkey would require the client to derive it from its own autodetected IP, which disagrees with the IP the service observed on any NATed or multi-homed host — producing a validly signed proof that can never verify. The RFC records this under the proof specification and in Open Questions.

**The sentinel exemption, and why it exists.** `crates/sentinel` builds `CreateSubscribeUser` with `owner` set to the validator and the sentinel as payer. Since the proof binds the owner, it would have to name the validator for an address the verification service never sees a request from — there is no proof the oracle could obtain, and without an exemption, setting the flag would break every oracle-provisioned multicast publisher with `IpOwnershipProofRequired`. Unlike the wildcard-pass gap this RFC closes, the exemption is not reachable by a registrant: it requires a DoubleZero-operated key. It waives the requirement only, so the oracle can start carrying real proofs without a program change. Two caveats worth knowing: a compromised sentinel key can bind any IP (as it already could), and `InitGlobalState` seeds `sentinel_authority_pk` to whoever initialized global state, so a fresh environment grants this to the deployer until the key is rotated. Replacing it is tracked in **#4215**.

**Sysvar scanning does not allocate.** `load_instruction_at_checked` materializes an `Instruction` per record — a `Vec<AccountMeta>` sized to the account count plus a copy of the data — and the BPF heap is a 32 KiB bump allocator that never frees, so a scan would allocate for every instruction in the transaction, including account metas of instructions unrelated to the proof, and a large enough batch would abort on allocation failure rather than return a named error. `ip_proof::instruction_at` reads the program ID and data straight out of the buffer, skipping account metas by arithmetic. This costs 662 CU instead of 1,336.

**Size.** 567 lines of non-test code, above the repo's 500-line guidance. Issue #4197 anticipated this: the natural seam is "flag + arg plumbing" then "validation logic", but that lands a half-wired instruction on main, so it is one PR.

## Diff Breakdown

| Tests        |    18 | +2425 / -2   | +2423 |
| Core logic   |     4 | +445  / -15  |  +430 |
| Scaffolding  |     9 | +122  / -5   |  +117 |
| Docs         |     2 | +102  / -22  |   +80 |
| Fixtures     |     8 | +24   / -3   |   +21 |
| Config/build |     3 | +18   / -0   |   +18 |
| Generated    |     1 | +7    / -0   |    +7 |
| **Total**    |    45 | +3143 / -47  | +3096 |

Four fifths of the diff is tests (the Tests row folds in the inline `#[cfg(test)]` modules, which the file-level count would otherwise attribute to the source files they live in); the logic lands in one new module plus a single call site shared by both user-creation instructions.

<details>

- `smartcontract/programs/doublezero-serviceability/src/ip_proof.rs` — new module: proof validation, the allocation-free Instructions sysvar scan, precompile offset parsing and bounds checks, and the trailing-sysvar split helper
- `smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs` — the single validation call site, placed after the PDA check and before the idempotent early return; binds `effective_owner` and passes the sentinel exemption; `Clock::get()` now read once and shared with the epoch gate
- `smartcontract/programs/doublezero-serviceability/src/error.rs` — 14 rejection-class error variants with `Custom(105..=118)` round-trip mappings
- `crates/doublezero-serviceability-instruction/src/user.rs` — both builders append the Instructions sysvar when the args carry a proof, plus the docs pinning its position at the very end of each account layout
- `smartcontract/programs/doublezero-serviceability/src/processors/user/create.rs` — optional `ip_proof` arg; peels the sysvar before the tenant length arithmetic
- `smartcontract/programs/doublezero-serviceability/src/processors/user/create_subscribe.rs` — optional `ip_proof` arg; peels the sysvar before `split_trailing_permission`
- `smartcontract/programs/doublezero-serviceability/src/state/feature_flags.rs` — `RequireIpOwnershipProof` (bit 2, `require-ip-ownership-proof`), parseable through the existing `SetFeatureFlags` path
- `crates/doublezero-serviceability-instruction/src/common.rs` — `append_instructions_sysvar`, so "sysvar last, Permission second-to-last" lives with the other trailing-account conventions

</details>

## Testing Verification

- 34 integration tests in `tests/user_ip_proof_test.rs` covering the full acceptance list: valid proof on both `CreateUser` and `CreateSubscribeUser`; missing Ed25519 instruction; missing sysvar account; wrong verifier key; unset verifier key; mismatched payer, client_ip, and user_type; unsupported version; signature over a different message; signature disagreeing with the proof; previous epoch accepted, epoch−2 rejected after a slot warp, future epoch rejected; Ed25519 instruction found at any index and after the program instruction.
- Owner-override coverage: a proof naming the effective owner is accepted, and one naming the transaction payer is rejected — the case that would pass silently if the binding used the payer.
- Sentinel coverage: exempt when the payer is the sentinel authority, still rejected when a supplied proof is bad, and not exempt when the sentinel authority is somebody else, so the exemption cannot pass by being vacuous. `setup()` rotates `sentinel_authority_pk` away from the fixture payer, which `InitGlobalState` seeds it to — otherwise every test would take the exempt path.
- Precompile-abuse cases are asserted twice over: as unit tests against the program's own checks (offsets naming another instruction, offsets past the end, multi-signature, zero-signature, truncated data, wrong-length message), and as integration tests asserting the runtime's precompile refuses to land those transactions at all — the program must not depend on that outer layer staying strict.
- The sysvar scan is unit-tested against buffers built by the runtime's own serializer (`construct_instructions_data`) rather than a hand-rolled imitation: the Ed25519 instruction found at each of four positions behind instructions carrying 1/5/9/13 accounts (a wrong account-meta stride is what this catches), no Ed25519 instruction present, first-rejection propagation instead of a generic miss, every truncated prefix of a valid buffer, and offsets pointing outside it.
- Builder tests pin the sysvar's position: `[.., tenant, payer, system, sysvar]` for `CreateUser`, and for `CreateSubscribeUser` everything ahead of the sysvar is byte-identical to the no-proof layout.
- Rerun path: a rerun without a proof is rejected while the flag is set; a rerun with a valid proof is still a no-op.
- Flag-off path: creation with no proof succeeds, a valid proof is accepted, and an invalid or Ed25519-less proof still rejects.
- CU measured under SBF with `tests/user_create_cu_benchmark.rs` (`--ignored`), same PDAs and bump seeds on both runs so the delta isn't buried in `find_program_address` noise: **`CreateUser` 35,550 → 36,212 CU, a delta of 662** against the 1,400,000 budget.


smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs
```diff
@@ -154,6 +166,23 @@ pub fn create_user_core(
...
     };
 
+    // Read once and reuse for the epoch gate further down; Clock::get() is a syscall.
+    let clock = Clock::get()?;
+
+    // RFC-27. Placed here because the PDA check above has just established that user_account.key
+    // is a legitimate User PDA — under either derivation — so the (client_ip, user_type) pair the
+    // proof binds is the pair this account is derived from. Placed before the already_exists
+    // return below so a rerun is covered too.
+    validate_ip_ownership_proof(
```
juan-malbeclabs, line 186: The proof is bound to `core.payer_account.key`, but on the owner-override path the account that owns the user (and that the AccessPass is keyed on) is `effective_owner`, not the transaction payer. RFC-27's own required check 4 in this PR says `proof.payer == user_payer` ("the account paying / owning the user").

Concrete breakage: `crates/sentinel/src/dz_ledger_writer.rs:141` builds `CreateSubscribeUser` with `owner: *owner` (the validator) and the sentinel as payer. Today it passes `ip_proof: None`; the moment `require-ip-ownership-proof` is set, every oracle-provisioned multicast publisher fails with `IpOwnershipProofRequired` (105). And it cannot be fixed by #4200 alone: the proof would have to name the *sentinel* as payer for an address the *validator* operates, which the verification service can never observe. Either bind `effective_owner` here, or exempt/handle the sentinel + USER_ADMIN override explicitly — and add a test (the existing `test_create_subscribe_user_sentinel_owner_override` / `..._user_admin_owner_override` cases all run with the flag off, so nothing currently covers this combination).

crates/doublezero-serviceability-instruction/src/user.rs
```diff
@@ -129,6 +136,13 @@ pub fn create_subscribe_user(
 /// tenant                      (writable)  — OPTIONAL, only when Some and non-default
 /// ```
 ///
+/// A further optional account, the **Instructions sysvar**, is appended *after*
```
juan-malbeclabs, line 149: This documents the sysvar's position but nothing enforces it, while `UserCreateArgs.ip_proof` is now a public field the builder happily serializes. Any caller that sets `ip_proof: Some(..)` and calls `create_user(..)` / `create_subscribe_user(..)` gets a transaction that is guaranteed to fail with `IpProofInstructionsSysvarMissing` (111), because the builder never appends the sysvar meta.

Since the builder already owns `args`, closing the trap here is cheap and removes the coupling with #4200:

```rust
if args.ip_proof.is_some() {
    accounts.push(AccountMeta::new_readonly(solana_program::sysvar::instructions::ID, false));
}
```

(appended last, after `[payer, system]` / after the Permission slot, exactly as the doc above says).

smartcontract/programs/doublezero-serviceability/src/ip_proof.rs
```diff
@@ -0,0 +1,485 @@
...
+        }
+        u16::from_le_bytes([data[0], data[1]])
+    };
+
+    // Remember why the first Ed25519 instruction was rejected. Returning that beats a generic
+    // "none found" when the client attached one proof and got a detail wrong — which is the
+    // common case, since a transaction carries at most one of these in practice.
+    let mut first_rejection: Option<DoubleZeroError> = None;
+
+    for index in 0..num_instructions {
```
juan-malbeclabs, line 218: The scan calls `load_instruction_at_checked` for *every* instruction in the transaction just to read its `program_id`. That helper fully deserializes each instruction — `Vec::with_capacity(num_accounts)` of `AccountMeta` plus a `to_vec()` of the data (solana-instructions-sysvar 3.0.1, `deserialize_instruction`) — and the BPF heap is a 32 KiB bump allocator that never frees.

Two consequences worth being explicit about:

- The measured 1,336 CU is a floor for a 2-instruction transaction. Cost is O(instructions in the tx) per `CreateUser`, so it grows with any compute-budget / memo / batched instructions the client adds, and is O(n^2) if several `CreateUser`s are batched (each one re-deserializes the whole list, including the other `CreateUser`s' ~11 account metas each).
- Heap, not just CU, is what runs out first: allocations from earlier iterations are never reclaimed, so a large enough batch aborts with an allocation failure rather than one of the 14 new named errors.

Either peek the program id straight out of the sysvar bytes (the per-instruction offset table at `data[2 + 2*i]` gives you the record; program_id sits after the account metas) or cap the scan, so the cost stops scaling with instructions that have nothing to do with the proof.

rfcs/rfc27-ip-verification.md
```diff
@@ -21,7 +21,7 @@ Today the serviceability program gates user creation with an **AccessPass** keye
 `(client_ip, user_payer)`. For a pass bound to a specific IP, the program enforces that the user's
 `client_ip` matches the pass, so the issuing authority effectively chose the IP. But for **wildcard
 passes** the program skips that check entirely
-(`smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs:173-183`):
+(`smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs:186-201`):
```
juan-malbeclabs: Both refreshed line references land on the wrong code after this PR's own changes to `create_core.rs`:

- `create_core.rs:186-201` is the device account-type check plus the AccessPass PDA derivation. The wildcard `client_ip` skip quoted below now lives at **215-228**.
- Line 41's `create_core.rs:141` is inside the owner-override `NotAllowed` branch; `get_user_pda(program_id, &client_ip, user_type)` is now at **153**.

The old numbers (173-183 / 125) were presumably re-derived before the `Clock::get()` hoist and the `validate_ip_ownership_proof` call were inserted above them.

smartcontract/programs/doublezero-serviceability/src/error.rs
```diff
@@ -471,7 +530,7 @@ mod tests {
         }
 
         // EnumIter generates Custom(0) by default, so we explicitly test values
-        // outside the known variant range (currently 0-100) to ensure the conversion
+        // outside the known variant range (currently 0-117) to ensure the conversion
```
juan-malbeclabs: Off by one: the last variant added in this PR is `IpProofVersionUnsupported` = 118, so the known range is 0-118, not 0-117.

smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs
elitegreg: Good catch, and the binding was wrong on both counts. Fixed in 0e0aa7ae: the proof now binds `effective_owner`, which matches the RFC's own check 4 and the identity `accesspass.user_payer` is keyed on. On the ordinary path the two are the same account, so only the override path changes.

That alone does not unbreak the sentinel, though — you are right that the oracle has no proof it could obtain, since the proof would have to name the validator for an address the verification service never sees a request from. So the sentinel authority is now exempt from the *requirement*: a creation paid for by `globalstate.sentinel_authority_pk` may omit the proof. It requires a DoubleZero-operated key, so it is not the registrant-reachable path RFC-27 exists to close, and it waives the requirement only — a proof the sentinel does attach is still validated in full, which is what lets the oracle start carrying real proofs without a program change.

Two things worth flagging about the exemption:

- `InitGlobalState` seeds `sentinel_authority_pk` to whoever initialized global state, so in a fresh environment it belongs to the deployer until the key is rotated.
- That is also why every existing test in `user_ip_proof_test.rs` was silently taking the exempt path once I added it — the fixture's payer *is* the sentinel. `setup()` now rotates it away, so the default tests exercise an ordinary payer.

Replacing the exemption is filed as #4215, with the options laid out (validator-supplied proof relayed by the oracle, narrowing it to the multicast-publisher shape, or verifying at access-pass issuance).

New tests: `test_owner_override_proof_binds_the_owner_not_the_payer`, `test_owner_override_proof_naming_the_payer_is_rejected` (the case that would have passed silently before), `test_sentinel_payer_may_create_without_a_proof_while_the_flag_is_set`, `test_sentinel_payer_still_has_a_supplied_proof_validated`, and `test_non_sentinel_payer_is_not_exempt` so the exemption cannot pass by being vacuous.

crates/doublezero-serviceability-instruction/src/user.rs
elitegreg: Agreed — a public field that serializes into a guaranteed failure is a trap. Done in 0e0aa7ae, for both `create_user` and `create_subscribe_user`.

One adjustment to your snippet: it has to be appended *after* `build` / `build_with_permission` returns rather than inside the account vec, because the sysvar must land after `[payer, system]` and, once the Permission rollout activates, after the Permission PDA too. `split_trailing_instructions_sysvar` peels by fixed key and runs first; `split_trailing_permission` matches by PDA and runs second — so "sysvar last, Permission second-to-last" is the only order both can parse. That is now a small `common::append_instructions_sysvar` helper, so the rule lives in one place next to the other trailing-account conventions.

Tests pin the ordering: `test_create_user_appends_the_instructions_sysvar_after_payer_and_system` builds with the optional tenant present too and asserts `[.., tenant, payer, system, sysvar]`, and the `create_subscribe_user` test asserts everything ahead of the sysvar is byte-identical to the no-proof layout.

smartcontract/programs/doublezero-serviceability/src/ip_proof.rs
elitegreg: You are right, and the heap point is the one that mattered — a bump allocator that never frees turns a per-instruction allocation into a batch-size limit, and hitting it aborts instead of returning one of the named errors.

Took the first option in 0e0aa7ae: `instruction_at` reads the program ID and data straight out of the sysvar buffer, skipping the account metas by arithmetic rather than decoding them. Nothing is allocated, so cost no longer scales with the account counts of unrelated instructions, and the batched-`CreateUser` case stops multiplying by ~11 account metas each. Every field is read through `get()` / `checked_add`, so a malformed record is an error rather than a wrapped offset.

Cost went from 1,336 to 662 CU (35,550 → 36,212 under SBF), and the RFC's figure is updated. The per-instruction scan is still O(instructions in the tx), but the constant is now a couple of bounds-checked reads.

Testing: the scan is split out as `scan_for_ed25519_instruction(&[u8], ..)` so it can be driven directly, and the unit tests build their buffers with `solana_instructions_sysvar::construct_instructions_data` — the runtime's own serializer — rather than an imitation of the layout. Cases: the Ed25519 instruction found at each of four positions with preceding instructions carrying 1/5/9/13 accounts (the stride is what a wrong `ACCOUNT_META_LEN` would break), no Ed25519 instruction at all, first-rejection propagation instead of a generic miss, every truncated prefix of a valid buffer, and offsets pointing outside it.

rfcs/rfc27-ip-verification.md
elitegreg: Both fixed in 0e0aa7ae: 215-228 and 153. Your diagnosis is right — they were re-derived before the `Clock::get()` hoist and the `validate_ip_ownership_proof` call went in above them, so the branch invalidated its own references.

smartcontract/programs/doublezero-serviceability/src/error.rs
elitegreg: Fixed in 0e0aa7ae — 0-118. The comment was updated when the variants were added and then `IpProofVersionUnsupported` arrived later in the branch without it being bumped again.

juan-malbeclabs (approved):

PR #4213, opened 2026-08-20
ip-verifier: add the IP ownership verification service

### LINKED ISSUE #4198: ip-verifier: add the IP ownership verification service

elitegreg:
Resolves #4198. Part of RFC-27 ([rfcs/rfc27-ip-verification.md](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc27-ip-verification.md)); tracker #4194. #4211 has landed, so this is now rebased onto `main` and the diff is this crate alone.

## Summary

- New `doublezero-ip-verifier` crate: a stateless HTTP service that signs the source address it observes as an RFC-27 `IpOwnershipProof`, over `POST /v1/proof`. It is the only party that can attest an address, so its answer to "which address did I actually see?" is the whole security property — everything else here is bookkeeping around that one question.
- **Proxy handling.** With no `--trusted-proxy` CIDR configured, forwarded headers are ignored entirely and the connection peer address is the only thing signed. For a peer inside a trusted CIDR the chain (`X-Forwarded-For`, else RFC 7239 `Forwarded`) is walked from the **right** — the end our own proxies append to — and the first hop that is not itself a trusted proxy is the client, so hops a client prepended are ignored. A trusted proxy with no forwarded header, a chain of only trusted hops, or an entry that is not an address signs nothing: the alternative is attesting the proxy's own address for everyone behind it.
- **The service never issues a proof the program would refuse.** It calls the program's own `helper::is_global` rather than a second copy of the predicate, and refuses IPv6 sources outright — a v1 proof carries a 4-byte address, and mapping an IPv6 client onto something IPv4-shaped would attest an address it does not have.
- **Epoch fails closed.** The epoch comes from the DoubleZero Ledger RPC, refreshed in the background and cached with a hard age limit. Past that limit requests are refused rather than signed with an epoch the service can no longer vouch for; the program's window is `clock.epoch` and `clock.epoch - 1`, so a stale number silently burns the one epoch of slack. A failed refresh keeps the last value and retries, so a brief RPC outage costs nothing and a refresh loop that dies quietly still cannot leave the service signing forever.
- **The client cannot name its own address.** `client_ip` is absent from the request body and `deny_unknown_fields` makes a caller that sends one get a `400`, rather than having the field silently ignored. The response deliberately omits the verifier public key: a client reads that from `GlobalState`, the same place the program reads it, so an impersonated service cannot talk one into building an Ed25519 instruction naming a key the program will not accept.
- Per-source-address token-bucket rate limiting, a request body cap, and a per-request timeout. The rate limit is charged before the body is parsed.
- `GET /health` returns `503` while the cached epoch is unusable, so an instance that cannot issue a usable proof drops out of a load balancer instead of answering requests. Prometheus metrics ride on a separate `--metrics-addr` listener, not the public one.
- Config through clap with `DZ_IP_VERIFIER_`-prefixed env vars, structured logging through `tracing`. The keypair is never logged and never serialized; even the read failure names only the path.

## Notes for reviewers

**axum, not hyper by hand.** #4198 asked for this call to be made explicitly since it sets a precedent for future Rust services. axum 0.8 (plus `tower-http` for the body cap and timeout) is now a workspace dependency. The deciding factor was that the connection peer address, the body cap, and the timeout all come from reviewed middleware rather than from this crate — the peer address *is* the security property, and it is not somewhere to be inventing plumbing.

**`user_type`, not `user_pubkey`.** #4198's API sketch has the client send the derived User PDA. The merged `doublezero-ip-proof` crate (#4195, #4206) and RFC-27 as amended in #4211 both bind `user_type` instead, which pins the same account since the PDA is `f(client_ip, user_type)` — and unlike the pubkey it does not require the client to derive an address from its own autodetected IP, which disagrees with the observed IP on any NATed host. The request body is `{ payer, user_type }`.

**`helper` is now `pub` in the serviceability program.** Three lines, so the service can apply `is_global` rather than reimplement it. #4198 requires the same semantics on both sides; sharing the function is the only way that stays true.

**Out of scope, per #4198.** Key custody, the Dockerfile, and the release workflow belong to the deployment issue. Nothing calls this service yet — the CLI work is #4200.

## Diff Breakdown

| Core logic   |     4 | +1060 / -0   | +1060 |
| Tests        |     1 | +387  / -0   |  +387 |
| Scaffolding  |     4 | +357  / -1   |  +356 |
| Generated    |     1 | +99   / -0   |   +99 |
| Docs         |     3 | +74   / -2   |   +72 |
| Config/build |     2 | +57   / -0   |   +57 |
| **Total**    |    15 | +2034 / -3   | +2031 |

One new crate, no changes to existing behavior anywhere. The core-logic files carry their unit tests inline: about 590 lines are logic and 470 are tests, which together with `tests/proof_test.rs` puts the split at roughly 860 lines of tests against 950 of code.

<details>

- `crates/doublezero-ip-verifier/src/client_ip.rs` — peer address vs. forwarded chain: trust check, right-to-left walk, RFC 7239 and `X-Forwarded-For` parsing, and every fail-closed case
- `crates/doublezero-ip-verifier/tests/proof_test.rs` — end-to-end tests against the real router on a loopback port
- `crates/doublezero-ip-verifier/src/server.rs` — the two endpoints, request/response shapes, and the refusal taxonomy behind the `reason` metric label
- `crates/doublezero-ip-verifier/src/epoch.rs` — `EpochSource` trait, the age-limited cache, and the background refresher
- `crates/doublezero-ip-verifier/src/settings.rs` — flags and `DZ_IP_VERIFIER_*` env vars
- `crates/doublezero-ip-verifier/src/rate_limit.rs` — per-address token bucket with idle-bucket eviction
- `crates/doublezero-ip-verifier/README.md` — API, proxy semantics, and how to get them wrong
- `smartcontract/programs/doublezero-serviceability/src/lib.rs` — `helper` made public for `is_global`

</details>

## Testing Verification

- 44 tests, all passing. 16 unit tests cover proxy header handling on its own: no trusted proxies, no trusted proxies with a client-sent `X-Forwarded-For`, an untrusted peer sending one, a trusted proxy with one hop, a spoofed extra hop ahead of the real one, two of our own proxies in series, repeated header lines as one chain, entries with ports, bracketed and bare IPv6 hops, an IPv6 peer as the trusted proxy, `Forwarded` honored only when `X-Forwarded-For` is absent, and the three fail-closed cases.
- 12 integration tests run the real router on an ephemeral loopback port and talk to it over HTTP, so the peer-address plumbing is exercised rather than mocked: a proof issued for the observed address and verified with the same `signed_message_for` bytes the program reconstructs; a body-supplied `client_ip` rejected as `invalid_request`; loopback, private, CGNAT, and documentation sources refused; an IPv6 source refused with a message naming the layout limit; a forwarded header from an untrusted peer leaving the address where it was; a spoofed hop ignored; an unfetched epoch returning `503`; `/health` flipping on epoch readiness; the burst limit returning `429` while a different source address is unaffected; and an oversized body returning `413`.
- Signature output is asserted against the committed test vectors from #4195 — both vectors, signature and signed-message bytes. They use documentation addresses, which `is_global` refuses, so they are driven through the signing layer the handler calls rather than over HTTP; the note is in the test.
- Epoch behavior tested at the boundary: fresh, exactly at the age limit, one second past it, a stale cache made usable again by a refresh, and a refresh loop whose source keeps failing leaving the previous epoch in place.


crates/doublezero-ip-verifier/src/client_ip.rs
```diff
@@ -0,0 +1,389 @@
...
+    let peer = peer.ip();
+
+    if !is_trusted(peer, trusted_proxies) {
+        return Ok(ResolvedClientIp {
+            addr: peer,
+            source: AddressSource::Peer,
+        });
+    }
+
+    let (source, hops) = match collect_hops(headers, X_FORWARDED_FOR, parse_forwarded_for_entry)? {
```
juan-malbeclabs: **Required before any `--trusted-proxy` deployment.** The header choice is hardcoded: `X-Forwarded-For` wins whenever it yields at least one hop, and `Forwarded` is only consulted as a fallback. `x_forwarded_for_wins_over_forwarded` argues this stops a client adding the other header, but the protection is one-directional.

If the trusted proxy writes RFC 7239 `Forwarded` (HAProxy `option forwarded`, or any RFC-7239-configured proxy) and does not also set or strip `X-Forwarded-For`, the client-supplied XFF is forwarded untouched — nginx and HAProxy pass unknown request headers through by default — and it *wins*. The client then names its own `client_ip` and gets it signed, which is the exact failure this module exists to prevent.

Suggest making the trusted header explicit in config (`--forwarded-header`), or failing closed when both headers are present and disagree.

crates/doublezero-ip-verifier/src/client_ip.rs
```diff
@@ -0,0 +1,389 @@
...
+/// end our own proxies append to. The first hop that is not itself a trusted proxy is the client. A
+/// chain the client prepended extra hops to is therefore ignored: those entries sit to the left of
+/// the address the nearest trusted proxy observed. If the walk runs out of hops without finding an
+/// untrusted one, or an entry does not parse, nothing is signed.
+pub fn resolve_client_ip(
+    peer: SocketAddr,
+    headers: &axum::http::HeaderMap,
+    trusted_proxies: &[IpNetwork],
+) -> Result<ResolvedClientIp, ClientIpError> {
+    let peer = peer.ip();
```
juan-malbeclabs: **Worth fixing in this PR — it's a few lines.** `peer.ip()` is used verbatim and `issue_proof` rejects any `IpAddr::V6`, so IPv4-mapped IPv6 addresses are never normalized.

On Linux with the default `IPV6_V6ONLY=0`, binding `--listen-addr [::]:8080` — the natural dual-stack choice, and a common container/k8s default — makes every IPv4 client's peer address arrive as `::ffff:a.b.c.d`, so every request returns `400 ipv6_unsupported`. The same address also fails `is_trusted` against IPv4 `--trusted-proxy` CIDRs, so a v4 proxy behind a `[::]` listener silently stops being trusted and its forwarded headers get dropped. Some proxies also emit `X-Forwarded-For: ::ffff:1.2.3.4`, which hits the same path.

`Ipv6Addr::to_ipv4_mapped()` on the peer address and on parsed hops, before the trust check and the family check, turns a silent total outage into a non-event.

crates/doublezero-ip-verifier/src/client_ip.rs
```diff
@@ -0,0 +1,389 @@
...
+
+    for value in headers.get_all(name) {
+        // A header that is not valid UTF-8, or an entry that is not an address, is a malformed
+        // chain. Skipping the bad entry would silently shift which hop is treated as the client.
+        let value = value
+            .to_str()
+            .map_err(|_| ClientIpError::MalformedForwardedEntry(format!("{value:?}")))?;
+
+        for entry in value.split(',').map(str::trim).filter(|e| !e.is_empty()) {
+            hops.push(parse(entry)?);
```
juan-malbeclabs: **Required before any `--trusted-proxy` deployment.** `collect_hops` parses every entry left-to-right and propagates the first error, before the right-to-left walk ever runs.

With the standard nginx `proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for`, the client's header is concatenated verbatim, so a client sending `X-Forwarded-For: unknown` (a real, widely emitted convention — as is RFC 7239 `for=unknown`) produces `unknown, <observed-client>`. The rightmost hop is valid and trustworthy, but the request dies as `MalformedForwardedEntry`.

The fail-closed rationale in the comment ("skipping the bad entry would silently shift which hop is treated as the client") only holds for entries to the *right* of the resolved hop — walking from the right and parsing lazily is strictly safer and doesn't turn a legitimate client into a hard failure. Compounding it, `status()` maps this to **500**, so client-controllable input produces 5xx noise and alerting on the assumption it can't be the client's fault.

crates/doublezero-ip-verifier/src/server.rs
```diff
@@ -0,0 +1,272 @@
...
+
+/// The body is taken as bytes rather than through the `Json` extractor so that the rate limit is
+/// charged before any parsing work, and so a bad body produces this module's error shape.
+async fn issue_proof(
+    State(state): State<AppState>,
+    ConnectInfo(peer): ConnectInfo<SocketAddr>,
+    headers: HeaderMap,
+    body: Bytes,
+) -> Result<Json<ProofResponse>, ProofError> {
+    let ResolvedClientIp { addr, source } =
```
juan-malbeclabs: **Worth fixing in this PR.** `resolve_client_ip` runs before `state.limiter.check(addr)`, so every request that fails resolution is entirely unmetered — the malformed-header path above, plus `NoForwardedAddress` and `AllHopsTrusted`.

Once a proxy is configured, one peer can drive unlimited 500 responses, unlimited `warn!` lines, and unlimited `proofs_refused_total` increments. There's no client address to charge in that case, so charge the peer address as a fallback bucket. Closing it now means it isn't waiting behind the proxy rollout.

crates/doublezero-ip-verifier/src/rate_limit.rs
```diff
@@ -0,0 +1,163 @@
...
+        self.check_at(addr, Instant::now())
+    }
+
+    fn check_at(&self, addr: IpAddr, now: Instant) -> bool {
+        let mut buckets = self
+            .buckets
+            .lock()
+            .expect("rate limiter lock is never held across a panic");
+
+        if buckets.len() >= self.max_entries && !buckets.contains_key(&addr) {
```
juan-malbeclabs, line 51: Low — non-blocking, noting it for the record. `evict_idle` only drops buckets that have refilled to full capacity, so `max_entries` isn't strictly a cap, and past the threshold each new-address request does a `retain` pass over the whole map while holding the `Mutex`.

In practice the defaults make this benign: `burst=5` at `30/min` means a one-request-per-address spray leaves a bucket at 4 tokens, full again after 2s idle, so `retain` *does* reclaim it, and each scan drops the map well below the threshold (amortized, not per-request). Holding the map above 100k entries means re-hitting every tracked address inside its 2s refill window — roughly 50k rps, where raw volume is the problem, not the map.

So no action needed at these defaults. It does mean `--rate-limit-max-entries` is a soft target rather than a bound, which is worth a word in the flag's doc comment if someone later tunes `burst` and `per-minute` toward a longer refill time.

crates/doublezero-ip-verifier/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+        .with(tracing_subscriber::fmt::layer())
+        .init();
+
+    PrometheusBuilder::new()
+        .with_http_listener(args.metrics_addr)
+        .install()
+        .context("failed to install the Prometheus exporter")?;
+    export_build_info();
+
+    let verifier = args.keypair()?;
```
juan-malbeclabs, line 32: Low, ops-facing. `--keypair`'s doc comment notes that a mismatch means every proof this service issues is rejected onchain, but the pubkey is only logged — nothing checks it against `GlobalState.ip_verifier_authority_pk`.

If the wrong keypair is mounted, or the authority is rotated without redeploying, the service reports healthy, `/health` returns 200, `proofs_issued_total` climbs, and every user creation fails onchain with no signal from the service side. It already has a ledger RPC client here — reading `GlobalState` once at startup and refusing to start (or flipping `/health`) on a mismatch is cheap, and turns a silent, hard-to-diagnose outage into a startup error.

juan-malbeclabs (approved):
Approving. Nice, well-documented crate — the doc comments carry the *why*, the epoch age limit lines up with the program's `[epoch-1, epoch]` window, and the `is_global` check is strictly stricter than the onchain one.

None of my findings are reachable in the configuration this PR ships: `trusted_proxies` defaults to empty, so `resolve_client_ip` returns the peer address without ever touching a forwarded header, and `--listen-addr` defaults to `0.0.0.0`. That's what makes this approvable as-is.

Two inline comments are small enough that I'd fold them into this PR (IPv4-mapped normalization, rate-limiter fallback). Two more are **hard prerequisites before anything sets `--trusted-proxy`** — they're the difference between "ignores headers" and "signs a client-chosen address" — but they can land separately if the initial rollout has clients reaching the service directly. Flagged which is which on each comment.

crates/doublezero-ip-verifier/src/client_ip.rs
elitegreg: Fixed, in the strict direction. The header is now named in configuration — `--forwarded-header {x-forwarded-for|forwarded}`, default `x-forwarded-for` — and **only** that header is read. The other is ignored even when present, so the HAProxy-`option forwarded`-plus-passed-through-XFF case cannot produce a client-chosen address, and neither can its mirror image. Two tests pin both directions: `only_the_configured_header_is_read` (unit) and `only_the_configured_forwarded_header_is_read` (integration, both headers set, configured one wins). I went with an explicit header over failing closed on disagreement because a proxy that rewrites one header and passes the other through produces a legitimate disagreement on every request — that deployment would be permanently down rather than misconfigured-but-loud.

crates/doublezero-ip-verifier/src/client_ip.rs
elitegreg: Fixed. `normalize()` collapses IPv4-mapped addresses via `to_ipv4_mapped()`, applied to the peer address and to every parsed hop before the trust check and before the family check. Three tests: a mapped peer resolving to its IPv4 form, a mapped peer still matching an IPv4 `--trusted-proxy` CIDR, and a mapped hop both resolving and being recognized as a trusted proxy when it is one.

crates/doublezero-ip-verifier/src/client_ip.rs
elitegreg: Fixed. `raw_entries` now collects unparsed `&str` entries and the right-to-left walk parses lazily, so an entry only has to parse if it could be the client. `unknown, <observed-client>` resolves to the observed client and never touches the junk. A malformed entry *at or right of* the first untrusted hop still fails closed, which is the case the original rationale actually covered — `a_malformed_entry_right_of_a_trusted_hop_fails_closed` pins it.

Also took the status code: `MalformedForwardedEntry` now maps to **400** with its own reason `invalid_forwarded_header`, while `NoForwardedAddress` and `AllHopsTrusted` stay 500. Those two cannot come from a client, so they still mean our proxy config is wrong; the client-controllable one no longer pages anyone or pollutes that signal.

crates/doublezero-ip-verifier/src/server.rs
elitegreg: Fixed. On a resolution error the peer address is charged as a fallback bucket, so the unresolvable-header path is metered too, and a peer over its limit gets 429 instead of an unbounded stream of errors, `warn!` lines, and `proofs_refused_total` increments. `an_unresolvable_chain_is_a_client_error_and_is_rate_limited` covers it end to end: two 400s, then 429.

crates/doublezero-ip-verifier/src/rate_limit.rs
elitegreg: Agreed on all of it, and thanks for working the arithmetic through — no behavior change. Documented on the flag: `--rate-limit-max-entries` now says it is a soft target rather than a hard bound, that only buckets refilled to full capacity are dropped, and that tuning `--rate-limit-burst` and `--rate-limit-per-minute` toward a long refill time is what lets the map grow past it.

crates/doublezero-ip-verifier/src/main.rs
elitegreg: Implemented. New `authority.rs` reads `GlobalState.ip_verifier_authority_pk` through an `AuthoritySource` trait (backed by a new `ledger.rs` that carries the one `RpcClient` and serves both this and the epoch):

- **Startup**: mismatch is fatal, and the error names both keys. An unreadable `GlobalState` is a warning only — an RPC problem already surfaces as a stale epoch, and failing to start on a read timeout would be its own outage.
- **After startup**: re-read every `--authority-refresh-secs` (default 60), so a rotation without a redeploy is caught too. A mismatch makes `/health` return 503 and the proof endpoint refuse with `verifier_key_mismatch`, rather than handing back a proof that dies onchain where the client cannot see why.
- `/health` now reports `verifier_key` as `matches`/`mismatch`/`unknown`, and there is a `doublezero_ip_verifier_authority_matches` gauge.

Five tests on the module plus an integration test for the rotated-authority path.

PR #4217, opened 2026-08-20
cli: take a pubkey in balance and add transfer

juan-malbeclabs:
## Summary of Changes

- `doublezero balance` takes an optional address: `doublezero balance <pubkey>` reports that account's balance, while the bare form keeps reporting the configured keypair's. Querying another account needs no local keypair, and an address that was never funded prints `0 Credits` instead of failing the account lookup.
- New `doublezero transfer <RECIPIENT> <AMOUNT>` sends credits from the configured keypair to another account on the DoubleZero Ledger. `AMOUNT` is a credit amount, or `ALL` to send the whole balance minus the transaction fee.
- A recipient that does not exist yet is created by the transfer, with the amount raised to the rent-exempt minimum when it falls below it — the runtime rejects a transfer that would leave a new account under rent exemption, so `solana transfer`'s `--allow-unfunded-recipient` opt-in is unnecessary here.
- Both verbs mirror the equivalent Solana CLI commands in shape (`solana balance [ADDRESS]`, `solana transfer <RECIPIENT> <AMOUNT>` with `ALL`), and both skip the program version check since they only touch the system program.
- No new trait or SDK surface: the balance of an arbitrary account comes from the existing `get_multiple_accounts`, and the transfer from the existing `transfer_sol`.

## Diff Breakdown

| Core logic   |     2 | +169 / -6   | +163 |
| Tests        |     3 | +329 / -0   | +329 |
| Scaffolding  |     3 | +7 / -1     |   +6 |
| Docs         |     1 | +3 / -0     |   +3 |
| **Total**    |     6 | +508 / -7   | +501 |

Two CLI verbs (~176 lines of new non-test code) with the bulk of the diff in their unit tests.

<details>

- `smartcontract/cli/src/transfer.rs` — new `transfer` verb: amount parsing (credits or `ALL`), fee reservation, rent-minimum top-up for a new recipient, balance checks
- `smartcontract/cli/src/balance.rs` — optional address argument, keypair preflight only on the own-balance form, shared `get_account_balance`/`credits` helpers
- `smartcontract/cli/src/cli/command.rs` — `Transfer` variant, dispatch arm, parse-routing tests
- `client/doublezero/src/main.rs` — `transfer` joins the verbs that skip the program version check

</details>

## Testing Verification

- 10 new unit tests: rent-minimum top-up for an unfunded recipient (and no top-up when the amount already clears it), `ALL` reserving the signature fee, rejection when the amount plus fee exceeds the balance, rejection when `ALL` cannot cover the fee, amount parsing (`ALL` casing, sub-lamport rounding, negative/`inf`/overflow rejection), zero balance for a nonexistent account, and parse routing for `balance <pubkey>` / `transfer`.
- Full `doublezero-serviceability-cli` suite green (432 tests).
- `--help` and top-level command list checked on the built `doublezero` binary for both verbs.
- Not exercised against a live ledger: no `dz-local-*` devnet was running, so the rent-exemption top-up is covered by unit tests only.


smartcontract/cli/src/transfer.rs
```diff
@@ -0,0 +1,353 @@
...
+                    "Recipient {} is not funded yet; sending the rent-exempt minimum of {} Credits to create it",
+                    self.to,
+                    credits(rent_min)
+                )?;
+                lamports = rent_min;
+            }
+        }
+
+        let required = lamports.saturating_add(SIGNATURE_FEE_LAMPORTS);
+        if from_balance < required {
```
martinsander00, line 112: we should refuse unless the leftover is 0 or at least 0.00089088 Credits, and to say that in the same “Insufficient funds” voice.

If I send to Bob 0.000001 and then he makes a transfer for 0.00001, he would keep 0 < amount < 0.00089088 credits which the chain would reject

juan-malbeclabs: Good catch — fixed in 95ea45d6.

The rent-exempt minimum is now fetched once and applied to both ends, not just to a new recipient. After the existing funds check, the sender's leftover (`balance - amount - fee`) is validated: it has to be either exactly 0 or at least the minimum, otherwise the transfer is refused before it ever reaches the runtime, in the same "Insufficient funds" voice:

```
Insufficient funds: <pubkey> holds 0.00089088 Credits, and sending 0.00001 Credits would leave 0.00087588 Credits behind, under the 0.00089088 Credits an account has to keep to stay rent-exempt; send 'ALL' to empty the account
```

When a smaller amount would still clear the minimum, the message names it instead: `send at most 0.00088588 Credits, or 'ALL' to empty the account`. `ALL` always passes since it leaves exactly 0, and so does an explicit amount that drains to 0 exactly.

Added tests for your exact scenario (an account created at the minimum trying to pass part of it on), for the largest-payable-amount hint, and for the exact drain that must keep working.

martinsander00 (approved):

PR #4223, opened 2026-08-21
ci: catch release-bump dependency rebinds on the PR that causes them

### LINKED ISSUE #4220: ci: catch release-bump dependency rebinds on the PR that causes them
#4219 fixes the immediate cause of the v0.37.0 testnet release failure, but nothing would have caught it on the PR that introduced it.

To be fair to #4213: it wasn't malformed. `reqwest = "0"` already existed — #4213 just became the first crate to want reqwest 0.13 alongside Solana's 0.12, which gave `cargo update --workspace` something to rebind onto. The breakage then sat dormant until release time, a week later.

Proposal — run the guard on every PR:

1. **Dry-run the release bump in PR CI.** Run `scripts/release/bump-version.sh` against a throwaway version, assert the `Cargo.lock` diff is only member version lines, discard the result. This is the check that would actually have failed #4213, because it exercises the same re-resolution path the release does.
2. **Lint `[workspace.dependencies]` for bare `"0"` requirements.** Cheap, but only stops *new* instances of the declaration pattern, and needs the remaining 27 pinned first (see the audit in #4219).

(1) is the one worth having.

nikw9944:
## Summary of Changes

* New `release-bump-dry-run` job runs the real `scripts/release/bump-version.sh` against a throwaway next-minor version on every PR and push to `main`, then discards the result.
* The release bump's `cargo update --workspace` can silently rebind workspace members onto a different locked minor. `bump-version.sh` already fails on that, but only ran at release time, so the rebind that broke v0.37.0 sat dormant for a week after #4213 (pinned in #4219). This puts the failure on the PR that causes it.
* Fixes #4220

Two things the diff cannot cover:

* **The job is advisory until you add it to the required checks.** Ruleset 3777660 lists individual job contexts, so living in `rust.yml` confers nothing (`rust-validator-test` and `rust-cli-static` are in this workflow and not required either). Once it is green on `main`, append `{"context":"release-bump-dry-run"}` to `required_status_checks`. Noted in the docs paragraph and CHANGELOG entry.
* **It can go red for reasons unrelated to a PR.** Re-resolution hits the live crates.io index, so an upstream publish can trip the guard on an unrelated PR — and on `main` too. This is the release-day flakiness moved somewhere cheap to fix; the remedy is always to pin the requirement, which the job's failure output says.

The bare-`"0"` lint (item 2 in #4220) is out: 27 such requirements remain and would all need pinning first, and this guard catches the consequence however the requirement is spelled.

## Testing Verification

* Ran the job's step verbatim on `95336a36` (#4213 as merged, pre-#4219): fails, naming the added `reqwest 0.12.28` / `tokio-util 0.6.10` lines and the new `hyper-tls` / `native-tls` / `tokio-native-tls` packages — the same diagnosis the v0.37.0 release produced a week later. Passes on the current tree.
* Traced the release branch case (`Cargo.toml` already bumped, CHANGELOG already promoted): next-minor keeps `PREV != NEW`, the second bump adds only member version lines, job passes. No special-casing needed.


packethog (approved):

.github/workflows/rust.yml
```diff
@@ -74,3 +74,22 @@ jobs:
...
+          set -euo pipefail
+          PREV=$(grep -m1 '^version = ' Cargo.toml | sed 's/version = "\(.*\)"/\1/')
+          NEXT=$(echo "$PREV" | awk -F. '{print $1"."$2+1".0"}')
+          echo "Dry-running the release bump $PREV -> $NEXT; the result is discarded."
+          if ! ./scripts/release/bump-version.sh "$NEXT"; then
+            echo "Fix: if the failure lists Cargo.lock lines, cargo update --workspace rebound a workspace dependency onto a different version. Pin the offending requirement in [workspace.dependencies] to its locked minor (0.13, not a bare 0). The guard lives in scripts/release/bump-version.sh."
+            echo "If the rebind looks unrelated to your changes, check this job on main: re-resolution runs against the live crates.io index, so a rebind can pre-date your PR or arrive with an upstream publish."
+            echo "::error::The release version bump fails on this tree, so the next release would fail the same way. The cause and the fix are in this job's log."
+            exit 1
+          fi
```
ben-dz, line 96: **Minor** — The hint blames a dependency rebind, but the likelier trigger on an ordinary PR is a `Cargo.toml` dependency edit whose regenerated `Cargo.lock` was never committed — nothing here builds with `--locked`, so that state merges green today and `cargo update --workspace` syncs it, tripping the same guard. Verified: adding `unicode-xid` as a dev-dependency without touching the lock fails the job with the whole diagnostic being `+ "unicode-xid",`, after which "pin the offending requirement" points at a requirement that does not exist. Name the second cause: `…, or a Cargo.toml dependency change whose Cargo.lock was not committed: run cargo update --workspace and commit the lock.`

.github/workflows/rust.yml
```diff
@@ -74,3 +74,22 @@ jobs:
             echo "::error::doublezero CLI is not statically linked"
             exit 1
           fi
+  # The release bump runs `cargo update --workspace`, which can silently rebind
+  # members onto a different locked minor.
+  release-bump-dry-run:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@1.97.1
```
ben-dz, line 84: **Optional** — `release-bump-dry-run` has no `timeout-minutes`, so a hung step runs to GitHub's six-hour default. Once this context is in the required set — the PR's stated end state — that holds up every merge for six hours. Add `timeout-minutes: 10` to the job.

ben-dz (approved):
The guard works — verified it fails on `95336a36` naming the reqwest/tokio-util rebind, passes on HEAD, and passes on an already-bumped release tree. The findings are all about the failure path: the `Fix:` hint attributes any Cargo.lock failure to a rebind, but an uncommitted lock trips the same guard and gets sent to pin a requirement that does not exist. Two smaller items on the second hint's stated mechanism and a missing `timeout-minutes`.

- `.github/workflows/rust.yml:96` — An upstream publish cannot trip this guard: `cargo update --workspace` keeps already-locked versions rather than chasing the index. Three runs of this step (pre-#4219 `95336a36`, the v0.37.0 bump commit, and HEAD) each left ~261-263 dependencies "behind latest" while touching only the 16 member entries; a rebind needs a tree change (loose requirement plus a second locked version, as in #4213) or a yank. "Check this job on main" is still the right action — attribute it to a PR merged since your base and drop the publish clause.

.github/workflows/rust.yml
nikw9944: Named it in 947f476c — the hint now reads "either `cargo update --workspace` rebound a workspace dependency … or a `Cargo.toml` dependency change landed without its regenerated lock, so run `cargo update --workspace` and commit the `Cargo.lock`."

Reproduced your case first (added `unicode-xid` to `config/Cargo.toml` dev-deps without touching the lock): the entire diagnostic is `+ "unicode-xid",`, which the old single-cause hint would have misdirected.

.github/workflows/rust.yml
nikw9944: Added `timeout-minutes: 10` in 947f476c. The job runs in ~14s on this PR, so the cap is far off the real duration and only bounds a hang.

nikw9944:
Rebased onto `main` (`c15a7935`) and addressed all three items in 947f476c.

- The `Fix:` hint now names both causes — a rebind, and a `Cargo.toml` dependency edit whose regenerated `Cargo.lock` was never committed. Reproduced your `unicode-xid` case before changing the text.
- Dropped the upstream-publish clause. You're right that `cargo update --workspace` keeps already-locked third-party versions; the second hint now points at a dependency arriving on another branch since the base. Same correction applied to the paragraph in `docs/testnet-release.md`, which carried the identical claim.
- `timeout-minutes: 10` on the job.

Nothing skipped.

Issue #4204, opened 2026-08-17
dev/dzctl: run an ip-verifier in the local devnet


elitegreg:
Part of RFC-27 (tracker: #4194). Depends on #4198.

Once the feature flag is on, local `connect` needs a reachable verifier or every client flow breaks.

## What

- Add a `dz-local-ip-verifier` container to `dev/dzctl`, with a devnet-only keypair.
- Set `ip_verifier_authority_pk` in the local `GlobalState` during setup, using that keypair's pubkey.
- Point local clients at the container.
- The observed source IP inside Docker is the container's address on the CYOA network, which is what the local client actually uses — confirm the proof it signs matches the IP `connect` binds, since this is the same class of problem as the proxy handling in production.
- Feature flag off by default locally, with a documented way to turn it on for testing.

## Acceptance

- `dev/dzctl` brings up the verifier and a client connects successfully with the flag both off and on.

Issue #4199, opened 2026-08-17
ip-verifier: deployment, key custody, and monitoring


elitegreg:
Part of RFC-27 (tracker: #4194). Depends on #4198.

Ship the service built in #4198.

## What

- Dockerfile and release packaging following the pattern in `controlplane/*` and `release/`.
- Deploy to devnet and testnet, behind TLS at a stable hostname per environment. The hostnames become the client defaults in the CLI issue, so pick them here.
- Load balancer and proxy configuration that preserves the real client source address, matched to the trusted-proxy CIDRs the service is configured with. Verify end to end from a known external IP that the signed `client_ip` equals that IP — a misconfiguration here silently signs the wrong address for every user.
- Verifier keypair custody: where it lives, who can read it, and the rotation procedure via `SetAuthority` (#4196). Note that rotation invalidates outstanding proofs and clients re-verify on their next connect.
- Alerts: service down, signing failures, epoch source stale, and an unusual rate of proof issuance for a single source IP.

## Acceptance

- Service reachable in devnet and testnet, `/health` green, dashboards and alerts wired.
- Documented verification that the observed source IP matches the true client IP through the full ingress path.
- Rotation runbook written and dry-run once in devnet.

PR #4227, opened 2026-08-21
cli: obtain an IP ownership proof during connect and attach it

### LINKED ISSUE #4201: cli: obtain an IP ownership proof during connect and attach it to user creation

elitegreg:
Resolves #4201. Part of RFC-27 ([rfcs/rfc27-ip-verification.md](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc27-ip-verification.md)); tracker #4194.

#4224 has merged, so this now sits directly on `main`. Rebased onto #4229, whose bare `doublezero connect` interacts with the proof — see the per-leg point below.

## Summary of Changes

- `doublezero connect` obtains an RFC-27 `IpOwnershipProof` from the verification service and attaches it to user creation. #4200 gave the SDK somewhere to put a proof; nothing obtained one until now.
- **The service decides the address, not the host.** The verifier signs the source address it observes and refuses a caller-supplied one — its request body has no `client_ip` field at all. So the returned address is authoritative, and the daemon's own discovery (`ifconfig.me`, inside `doublezerod`) stays a convenience for display and the AccessPass pre-flight.
- **Multi-homed hosts.** The request is bound to the address the tunnel will use, so an operator who set `--client-ip` on the daemon proves that address rather than whichever one the default route prefers. Whether the bind is possible is probed first, with a UDP bind and route lookup that send nothing: reqwest only stores `local_address` in its connector config, so a client built with an unassignable one succeeds and fails later as an ordinary connect error. A NATed host, whose `client_ip` is on no local interface, therefore falls back to the default egress where the service observes the NAT's public address — the same one the daemon discovered. A host that owns the address but cannot reach the verifier from it warns, because that request and the tunnel would take different paths.
- **One hard failure, and only where it applies.** When the service's observed address disagrees with the one being provisioned, `connect` stops and names both. Attaching the proof would guarantee an onchain `IpProofClientIpMismatch`; dropping it would bind an address nobody proved. Guessing between them is worse than stopping. But that only holds where a user is actually created, so nothing is requested until a creation path asks: a host whose user already exists attaches nothing and binds nothing, and an `--unsubscribe-feed`-only run creates nothing at all. Neither can be taken offline by a verifier disagreement it would never have acted on.
- **Everything else is reported and non-fatal.** Unconfigured, unreachable, and declined (`not_globally_routable` for a CGNAT source, `rate_limited`, …) each print their own reason and continue without a proof. The program is the enforcement point, so this succeeds while `require-ip-ownership-proof` is clear and fails with a named error once it is set — the behavior wanted during rollout. Breaking `connect` here would take every host in a not-yet-enforcing environment offline the moment a verifier went down.
- **One proof per `user_type`, not per invocation.** The proof binds `user_type`, and the bare `doublezero connect` that #4229 added provisions a unicast leg and a multicast leg in one run — so a single proof would be refused onchain by whichever leg it did not name. Each leg builds its own fetcher; every other mode creates one kind of user and so needs only one.
- **Two attempts, one second of connect budget.** A blackholed address is abandoned after 1s so a second DNS answer is actually reached, and a transport error or a 5xx gets one retry 250ms later. A decline is not retried: `rate_limited` and `not_globally_routable` are answers, not accidents. Whole budget 5.25s, from a 2.5s per-attempt timeout. Note hyper divides the connect budget across the resolved addresses, so it is a total, not a per-address value.
- The service is only ever reached directly, never through a proxy. `HTTP_PROXY` and friends would terminate the connection at the proxy, hand the verifier the proxy's address, and turn the agreement check into a hard failure on a host where `connect` used to work.
- The wire response is validated against what was asked for — layout version, payer, and `user_type` — so a verifier rolled forward to a layout this client does not know produces a clean "continuing without a proof" rather than a user creation that is paid for and then refused onchain.
- `NetworkConfig` gains `ip_verifier_url`, overridable by `DZ_IP_VERIFIER_URL` and by a new `--ip-verifier-url` on `connect`.

## Notes for reviewers

**`--client-ip` disagreement.** The issue asks to fail when an explicit `--client-ip` disagrees with what the service observed. That flag is already deprecated and ignored on the CLI — `connect` warns and reads the address from the daemon instead. So the disagreement that can actually occur is daemon-discovered versus service-observed, and that is what is checked. The remedy the error suggests is `--client-ip` **on the daemon**, which is the flag that still does something.

**Per-environment URLs.** Every environment's `ip_verifier_url` is `None`. Mainnet-beta, testnet, and devnet have no verifier deployed yet — that is #4199. Localnet had a default of the verifier's own listen address, but that was wrong for the case that matters: in the dockerized devnet the client container's `localhost` is not the verifier host, and #4204 is what will know whether the CLI reaches it by container name or on the host. A guess here would be the placeholder-wrong-on-arrival this note argues against for the other three, and the service signs the address it observes a request arrive from, so pointing at the wrong host is worse than pointing at nothing. `--ip-verifier-url` and `DZ_IP_VERIFIER_URL` cover the gap. **#4199 and #4204 should fill these in.**

**Redundancy behind the verifier's name.** hyper tries the resolved addresses in sequence, but its happy-eyeballs timer only crosses address *families*, not two A records — so with no connect timeout a blackholed address (SYN dropped, how a down or firewalled host usually fails) burns the whole request budget and its siblings are never tried. `CONNECT_TIMEOUT` is what makes multiple A records redundancy rather than a coin flip.

It is a *total* budget, though, not a per-address one: hyper-util divides it by the number of resolved addresses, so one A record gets the full second and eight would get 125ms each, under a single cross-ocean RTT. The value is sized for the single address that holds today.

The retry is deliberately not failover, and does not claim to be. A transport error leaves no pooled connection, so that retry does redial and can land elsewhere; a 5xx came from a host that is up, whose keep-alive connection the pool hands straight back, and which hyper would try first even on a fresh dial. What two attempts buy is a transient blip. Reaching a *different* backend is a load balancer's or an anycast VIP's job in front of the fleet — worth settling in #4199 rather than deepening the retry here. `POST /v1/proof` is idempotent, signing the address it observes and holding no per-request state, so a retry cannot double anything.

**A new dependency.** `doublezero-daemon-cli` gains `reqwest` (blocking, already a workspace dep used by two other crates). Blocking is deliberate: it sits beside `LedgerClient`'s blocking RPC in the same code path, and one short request per invocation does not justify a second async HTTP stack in this crate. `local_address` is what makes the source binding above possible.

**Two changelog lines from #4224 move.** That branch predated the v0.37.0 release bump, so its entries landed under the released `v0.37.0` heading and would never have appeared in the next release's notes. They move to `Unreleased` here, along with this PR's own entry, and the two sentences that said nothing set or consumed the new surfaces yet are gone now that this changeset does. Every RFC-27 entry is trimmed to two sentences.

**One line outside this PR's subject.** `DZClient::send_transaction_inner` now rejects an empty instruction vec, which would otherwise submit the compute-budget prelude on its own and pay a fee for a no-op. It is the `send_instructions` path #4224 added, flagged in review there after it merged.

## Diff Breakdown

| Core logic   |     3 | +1360 / -14  | +1346 |
| Config/build |     2 | +28  / -0    |   +28 |
| Docs         |     1 | +9   / -5    |    +4 |
| Generated    |     1 | +5   / -0    |    +5 |
| **Total**    |     8 | +1393 / -19  | +1374 |

Roughly 850 of the core-logic lines are inline `#[cfg(test)]` tests in those same files, leaving about 510 lines of new logic — one new module and the threading of a single value through `connect`.

<details>

- [`crates/doublezero-daemon-cli/src/ip_proof.rs`](https://github.com/malbeclabs/doublezero/pull/4227/files#diff-66aea1da6470a039c72db9f21be40058dc08fdc3ac9d48c3de41b25dbdc3a47f) — new module: the `IpProofClient` trait, its blocking HTTP implementation with the source-address binding probe, and the classified refusal reasons
- [`crates/doublezero-daemon-cli/src/connect.rs`](https://github.com/malbeclabs/doublezero/pull/4227/files#diff-59e335d4527da129d614de48572856bacb13b173e1248afe4313f32ceca553e2) — `IpProofFetcher` (the lazy fetch, the address-agreement check, the warn-and-continue paths), `--ip-verifier-url`, and the fetcher threaded to all four creation call sites
- [`config/src/env.rs`](https://github.com/malbeclabs/doublezero/pull/4227/files#diff-2ce57d9594541a2d12be9a07b25a724f4d094eac9c6ca52335752c883be41323) — `NetworkConfig::ip_verifier_url`, its `DZ_IP_VERIFIER_URL` override, and why no environment has a default yet

</details>

## Testing Verification

- Five `connect` tests drive every branch against a mocked verifier: a proof attached and asserted field for field on the command that reaches the SDK; the proof bound to `Multicast` on the multicast path and to `IBRL` on the unicast one, so a proof for the wrong `user_type` would fail the test rather than the chain; an unreachable service continuing without a proof; a `not_globally_routable` refusal surfacing the service's own reason; and the address disagreement failing with both addresses in the message and no ledger call at all.
- The fallback test reuses the existing `expect_create_user_with_tenant` helper, which pins `ip_proof: None`, so if the warn-and-continue path ever starts attaching something the test fails.
- The other 76 `connect` tests run against a verifier that reports "not configured", which is the pre-RFC-27 shape they were written to assert — they are unchanged apart from the new field in their `Connect` literals.
- Two tests drive the retry over a real socket against a listener serving canned responses: a 503 then a 200 produces the proof and the listener records exactly two connections, and a `rate_limited` 429 is reported from the first attempt with only one connection served — a spurious retry would find the listener gone and report `Unreachable` instead, failing the assertion. Three more cover the retry classification directly: a 5xx retryable, `rate_limited` and `not_globally_routable` not.
- Wire-form tests for the proof response: a round trip, a signature that is not base58, a payer that is not a pubkey, and the unconfigured-client case.
- Four tests drive the source-address binding probe over real sockets: a local address toward a reachable destination binds, an address on no local interface does not, and a local address that is not a legal source for the destination does not — the last being the pairing that made the previous localnet default unusable.
- One test asserts that a `connect` whose user already exists never asks for a proof at all: its mock verifier is set up with `expect_request_proof().never()`, so the lazy fetch failing back to eager would fail the test rather than only showing up as an offline host.
- Three tests cover the response validation: an unsupported layout version, a proof naming another payer, and one naming another `user_type` are each refused before the transaction is built.
- A bare `connect` test asserts the verifier is asked twice, once per leg, and that each leg's creation carries a proof bound to the `user_type` it creates — so a single shared proof, or two proofs of the same type, fails the test rather than the chain.
- 221 tests pass in `doublezero-daemon-cli`.

Still outstanding from the issue's acceptance list: **manual verification against local devnet end to end**, which needs #4204 to run a verifier in `dev/dzctl`. Worth doing before this merges.


CHANGELOG.md
```diff
@@ -21,8 +21,14 @@ All notable changes to this project will be documented in this file.
...
+- Rust SDK
+  - `CreateUserCommand` and `CreateSubscribeUserCommand` take an optional RFC-27 `ip_proof`. Supplying one attaches the native `Ed25519SigVerify` instruction the program looks for and sends both as one transaction; the verifier key comes from `GlobalState.ip_verifier_authority_pk`, the same place the program reads it, so a caller cannot pair a proof with the wrong key. A proof naming a different owner, address, or user type is refused before the transaction is paid for. On the `--owner` override path the proof must name that owner, because the program binds it to the user's effective owner. Omitting it produces the pre-RFC-27 transaction unchanged. Nothing sets it yet; the CLI is #4201. (#4200)
+  - `DoubleZeroClient` gains `send_instructions`, for a transaction that needs more than one instruction. `send_transaction` is unchanged. (#4200)
 - Utility crates
   - New `doublezero-ip-proof` crate defines the RFC-27 `IpOwnershipProof` and the exact bytes the verifier signs, in one place the serviceability program, the CLI, and the verification service all share. Nothing consumes it yet. (#4195, #4206)
```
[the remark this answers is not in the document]
elitegreg: Fixed. "Nothing consumes it yet" is gone; it was already inaccurate at release time, since #4197 consumes the crate.

CHANGELOG.md
```diff
@@ -21,8 +21,14 @@ All notable changes to this project will be documented in this file.
...
+- Rust SDK
+  - `CreateUserCommand` and `CreateSubscribeUserCommand` take an optional RFC-27 `ip_proof`. Supplying one attaches the native `Ed25519SigVerify` instruction the program looks for and sends both as one transaction; the verifier key comes from `GlobalState.ip_verifier_authority_pk`, the same place the program reads it, so a caller cannot pair a proof with the wrong key. A proof naming a different owner, address, or user type is refused before the transaction is paid for. On the `--owner` override path the proof must name that owner, because the program binds it to the user's effective owner. Omitting it produces the pre-RFC-27 transaction unchanged. Nothing sets it yet; the CLI is #4201. (#4200)
```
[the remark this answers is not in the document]
elitegreg: Fixed. That sentence is gone, and the entry moved to `Unreleased` — it had landed under the already-released v0.37.0 heading.

config/src/env.rs
```diff
@@ -106,6 +110,11 @@ impl Environment {
         if std::env::var("DZ_SOLANA_RPC_URL").is_ok() {
             config.solana_l1_rpc_url = std::env::var("DZ_SOLANA_RPC_URL").unwrap();
         }
+        // Set this to point at a verifier in an environment that has no deployed one yet, or at
+        // a local one while developing. `--ip-verifier-url` on `connect` overrides it in turn.
+        if std::env::var("DZ_IP_VERIFIER_URL").is_ok() {
+            config.ip_verifier_url = Some(std::env::var("DZ_IP_VERIFIER_URL").unwrap());
+        }
```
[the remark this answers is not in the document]
elitegreg: Fixed, though not with `if let Ok(..)` — this file has `use eyre::Ok;` at the top, so that form does not parse, which is why the neighboring overrides use `is_ok()` + `unwrap()`. It is now `std::env::var(..).ok().or(config.ip_verifier_url)`: one lookup, no unwrap, with a comment saying why.

smartcontract/sdk/rs/src/client.rs
```diff
@@ -171,20 +171,24 @@ impl DZClient {
...
-    fn send_transaction_inner(&self, ix: Instruction) -> eyre::Result<Signature> {
+    ///
+    /// Almost every caller passes one instruction. RFC-27 user creation passes two:
+    /// the native `Ed25519SigVerify` instruction and the creation it authorizes,
+    /// which have to land together or not at all.
+    fn send_transaction_inner(&self, ixs: Vec<Instruction>) -> eyre::Result<Signature> {
         let payer = self
             .payer
             .as_ref()
             .ok_or_eyre("No default signer found, run \"doublezero keygen\" to create a new one")?;
```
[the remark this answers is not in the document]
elitegreg: Fixed. `send_transaction_inner` now bails on an empty instruction list rather than paying for a compute-budget-only transaction.

crates/doublezero-daemon-cli/src/ip_proof.rs
```diff
@@ -0,0 +1,279 @@
...
+                    %source_addr,
+                    error = %bind_err,
+                    "could not bind the verification request to the tunnel source address; \
+                     falling back to the host's default egress"
+                );
+                let client = reqwest::blocking::Client::builder()
+                    .timeout(REQUEST_TIMEOUT)
+                    .build()
+                    .map_err(|e| IpProofError::Malformed(e.to_string()))?;
+                Ok((client, false))
```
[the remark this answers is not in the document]
elitegreg: Fixed. It maps to `Unreachable` now, and the URL is threaded into `build` so the message names it instead of rendering an empty one.

crates/doublezero-daemon-cli/src/ip_proof.rs
```diff
@@ -0,0 +1,477 @@
...
+        source_addr: Ipv4Addr,
+        url: &str,
+    ) -> Result<(reqwest::blocking::Client, bool), IpProofError> {
+        // Bind the request to the address the tunnel will use, so a multi-homed host proves the
+        // right one. On a NATed host that address is not assigned to any local interface and the
+        // bind fails, which is expected, not an error: fall back to an unbound request, where
+        // the service observes the NAT's public address — the same address the daemon
+        // discovered. Both paths are reported, because a silent fallback on a multi-homed host
+        // would prove the wrong address.
+        match reqwest::blocking::Client::builder()
```
juan-malbeclabs: **high** — the bind fallback is dead code, so NATed hosts never get a proof.

`Client::builder().local_address(Some(addr)).build()` does not bind anything: reqwest only stores the address in the connector config (`reqwest-0.13.4/src/connect.rs:193` → `HttpConnector::set_local_address`), and the actual `bind()` happens per connection in `hyper_util`'s `bind_local_address`. `build()` therefore essentially cannot fail because `source_addr` is unassignable, so the `Err(bind_err)` arm — and the whole documented "on a NATed host the bind fails; fall back to the default egress" behaviour — never executes.

Concretely, on a host behind NAT whose daemon-discovered `client_ip` is the NAT's public address: the bind is attempted at connect time, fails with `EADDRNOTAVAIL`, is classified as `Unreachable`, retried once, and `connect` prints "could not reach the verification service" and proceeds with no proof. Every NATed host silently ends up unprovable, and will fail user creation the moment `require-ip-ownership-proof` is set.

Suggestion: probe the address first (e.g. `std::net::UdpSocket::bind((source_addr, 0))`) before deciding to bind, or retry unbound when a connect error indicates an address-assignment failure.

crates/doublezero-daemon-cli/src/ip_proof.rs
```diff
@@ -0,0 +1,477 @@
...
+
+impl HttpIpProofClient {
+    pub fn new(base_url: Option<String>) -> Self {
+        Self { base_url }
+    }
+
+    /// Blocking on purpose: it sits alongside `LedgerClient`'s blocking RPC calls in the same
+    /// `connect` code path, and one short request per invocation does not justify a second
+    /// async HTTP stack in this crate.
+    fn build(
```
juan-malbeclabs, line 157: **medium** — proxy env vars silently defeat the source binding and turn into the hard failure.

reqwest's builder defaults to `auto_sys_proxy: true` (`async_impl/client.rs:309`), so `HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY` are honoured. On a host with a proxy configured, the `local_address` binding applies to the connection *to the proxy*, the verifier observes the proxy's egress address, and `obtain_ip_proof` then takes the hard-failure branch — `doublezero connect` now aborts on a host where it previously worked, with an error message ("set `--client-ip` on the daemon", "point `--ip-verifier-url` at a verifier reachable over that path") that names neither the cause nor the remedy.

Consider `.no_proxy()` on this client, or at least detecting a configured proxy and saying so in the error.

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -159,12 +166,112 @@ fn check_accesspass<L: LedgerClient>(
...
+            writeln!(
+                out,
+                "    Continuing without an IP ownership proof. This will be rejected onchain \
+                 once IP ownership verification is required for this environment."
+            )?;
+            return Ok(None);
+        }
+    };
+
+    if proof.client_ip != client_ip {
```
juan-malbeclabs: **medium** — the address-mismatch hard failure fires on paths that create no user.

`obtain_ip_proof` is called unconditionally (line 324) before `parsed_mode` is dispatched, and its `Err` propagates with `?`. So a mismatch aborts `connect` even when (a) the user account already exists and `find_or_create_user` would only re-activate it, or (b) the invocation is `--unsubscribe-feed`-only and never reaches a create path at all. In both cases neither justification for the hard failure applies — nothing would be attached and nothing bound — yet an already-provisioned host can no longer run `connect`.

Restricting the hard failure to the creation paths (or to "a user is about to be created") would avoid taking working hosts offline.

config/src/constants.rs
```diff
@@ -83,3 +83,15 @@ pub const ENV_LOCAL_INTERNET_LATENCY_COLLECTOR_PUBKEY: Pubkey =
...
+// source address it observes a request originate from, so this URL has to be reachable from the
+// connecting host — and reachable over the same path the tunnel will use.
+//
+// Only localnet has a default, the verifier's own default listen address, which is what
+// `dev/dzctl` runs (#4204). Deployed URLs for mainnet-beta, testnet, and devnet land with the
+// deployment work (#4199); until then those environments have no verifier, and `doublezero
+// connect` creates users without a proof — which the program accepts while
+// `require-ip-ownership-proof` is clear. `--ip-verifier-url` or `DZ_IP_VERIFIER_URL` points at
+// one in the meantime.
+pub const ENV_LOCAL_IP_VERIFIER_URL: &str = "http://localhost:8080";
```
juan-malbeclabs: **medium** — the localnet default URL cannot work with the source binding.

`ENV_LOCAL_IP_VERIFIER_URL = "http://localhost:8080"` resolves to `127.0.0.1`, but `source_addr` is the daemon-discovered CYOA address. Binding a non-loopback source and connecting to `127.0.0.1` fails at `connect()` (`EINVAL` on Linux, since the loopback route's source does not match the bound address), so the only environment that ships a default will always take the "unreachable" path.

Separately, in the dockerized local devnet the client container's `localhost` is not the verifier host in any case — the URL needs to name the verifier container.

crates/doublezero-daemon-cli/src/ip_proof.rs
```diff
@@ -0,0 +1,477 @@
...
+            reason: format!("http_{}", status.as_u16()),
+            message: body.trim().chars().take(200).collect(),
+        },
+    };
+    (err, status.is_server_error())
+}
+
+/// Converts the wire form into the struct the instruction carries, rejecting anything the
+/// program would reject anyway.
+fn proof_from_response(parsed: ProofResponse) -> Result<IpOwnershipProof, IpProofError> {
```
juan-malbeclabs: **low** — `proof_from_response` does not validate `version`, `payer` or `user_type`.

The module doc says "a field the program does not understand must fail here rather than be silently coerced", but `version` is copied verbatim and `doublezero_ip_proof::is_supported_version` / `IP_PROOF_VERSION` are never consulted; the returned `payer` / `user_type` are likewise never compared against what was requested. A verifier rolled forward to layout version 2 (or a stale one) returns a proof the CLI happily attaches, and the SDK or the program then rejects it — converting a clean client-side "no proof, continuing" into a failed, paid-for user creation.

crates/doublezero-daemon-cli/src/ip_proof.rs
```diff
@@ -0,0 +1,477 @@
...
+/// purpose: `connect` is interactive, and the fallback is a working connection while the feature
+/// flag is clear. [`MAX_ATTEMPTS`] of these plus the delay between them is the whole budget.
+const REQUEST_TIMEOUT: Duration = Duration::from_millis(2500);
+
+/// Abandon one address this fast so a second DNS answer is actually reached. hyper tries the
+/// resolved addresses in sequence, but its happy-eyeballs timer only crosses address families,
+/// not two A records — so without a connect timeout a blackholed address (SYN dropped, which is
+/// how a host that is down or firewalled usually fails) burns the entire request budget and its
+/// siblings are never tried. Redundant A records are then a coin flip rather than redundancy.
+const CONNECT_TIMEOUT: Duration = Duration::from_secs(1);
```
juan-malbeclabs, line 51: **low** — `CONNECT_TIMEOUT` is a total connect budget divided across resolved addresses, not a per-address timeout.

`hyper_util`'s `ConnectingTcpRemote::new` does `connect_timeout.checked_div(addrs.len())`, so with N A records each address gets `1s / N`. Once #4199 puts a fleet behind one name — the scenario this constant exists for — N=8 gives each host 125 ms, shorter than a single cross-ocean RTT, and every *healthy* host is abandoned. Worth correcting the comment ("abandon one address this fast") and picking the value with the division in mind.

crates/doublezero-daemon-cli/src/ip_proof.rs
```diff
@@ -0,0 +1,477 @@
...
+        &self,
+        payer: Pubkey,
+        user_type: UserType,
+        source_addr: Ipv4Addr,
+    ) -> Result<IpOwnershipProof, IpProofError> {
+        let base_url = self
+            .base_url
+            .as_deref()
+            .ok_or(IpProofError::NotConfigured)?;
+        let url = format!("{}/v1/proof", base_url.trim_end_matches('/'));
```
juan-malbeclabs, line 268: **low** — the retry reuses the pooled connection, so it cannot reach a different backend.

The stated rationale is that "a second attempt may resolve to a healthy host behind the same name", but the same `reqwest::blocking::Client` is reused and its connection pool will hand the retry the same keep-alive connection to the same backend. The wire test only observes two connections because it forces `Connection: close` on every canned response; a real server with keep-alive will serve both attempts from one host.

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -159,12 +166,112 @@ fn check_accesspass<L: LedgerClient>(
...
+    client_ip: Ipv4Addr,
+    out: &mut W,
+) -> eyre::Result<Option<IpOwnershipProof>> {
+    let proof = match proof_client.request_proof(payer, user_type, client_ip) {
+        Ok(proof) => proof,
+        Err(IpProofError::NotConfigured) => {
+            tracing::debug!("no IP ownership verification service configured; continuing");
+            return Ok(None);
+        }
+        Err(err @ IpProofError::Declined { .. }) => {
```
juan-malbeclabs: **low** — duplicate match arms.

`Err(err @ IpProofError::Declined { .. })` and the following catch-all `Err(err)` have byte-identical bodies, so the `Declined` arm is dead. Either drop it, or give the decline the distinct message it was presumably added for.

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -159,12 +166,112 @@ fn check_accesspass<L: LedgerClient>(
...
+        ctx: &CliContext,
+        daemon: &D,
+        ledger: &L,
+        out: &mut W,
+    ) -> eyre::Result<()> {
+        // The flag wins over the environment default, which in turn is overridable by
+        // DZ_IP_VERIFIER_URL inside `config()`.
+        let verifier_url = match &self.ip_verifier_url {
+            Some(url) => Some(url.clone()),
+            None => ctx.env.config().ok().and_then(|c| c.ip_verifier_url),
```
juan-malbeclabs: **low** — config failure and "not configured" are both completely silent.

`ctx.env.config().ok().and_then(...)` discards any config-load error, and `IpProofError::NotConfigured` is reported only via `tracing::debug!`, which is invisible without `--log-verbose`. An operator who exported `DZ_IP_VERIFIER_URL` but whose config load failed sees no proof, no warning, and no way to tell that apart from a correctly-unconfigured environment — which matters once enforcement is on and the onchain rejection is the first signal.

juan-malbeclabs (commented):
Review of the IP-ownership-proof path. One high finding (the bind fallback never executes, so NATed hosts silently end up without a proof), three mediums around how the proof failure interacts with proxies, unrelated `connect` invocations and the localnet default, plus a handful of smaller ones. Details inline.

crates/doublezero-daemon-cli/src/ip_proof.rs
elitegreg: Fixed, and thanks — the mechanism is exactly as you describe: `local_address` only reaches `HttpConnector::set_local_address`, so `build()` had no way to fail and the `Err(bind_err)` arm was unreachable.

`probe_source_binding` now decides before the client is built. A `UdpSocket::bind((source_addr, 0))` followed by a `connect()` to the resolved destination sends nothing but answers both halves: `bind` asks whether the address is assigned to this host, `connect` is a pure route lookup asking whether it is a legal source for that destination. `local_address` is only set when both succeed.

The two failures now mean different things and are reported differently. Bind failure means the address is on no local interface — the ordinary NATed host, where the service observes the NAT address over the default egress anyway, so that stays at `debug`. Bind success with a failed route lookup means the host owns the address but cannot reach the verifier from it, which is worth saying out loud, so it is a `warn` (the default log level, so it is visible without `--log-verbose`).

Four tests over real sockets cover it.

crates/doublezero-daemon-cli/src/ip_proof.rs
elitegreg: Fixed with `.no_proxy()` on the builder.

Agreed on the analysis, and the verifier side confirms it: `server.rs` reads `ConnectInfo<SocketAddr>` and only substitutes a forwarded header for a peer in its own `--trusted-proxy` list, so a client-side proxy gets no relief there either. Going through a proxy would make the proof describe the proxy rather than this host, which defeats the binding entirely — so the request now always goes direct. A proxy-only host lands on `Unreachable` and continues without a proof, which is the non-fatal path.

crates/doublezero-daemon-cli/src/connect.rs
elitegreg: Fixed. Both cases you name were real.

`obtain_ip_proof` is now `IpProofFetcher`, which fetches nothing at construction. `get()` is called at the four points that actually call `create_user` / `create_subscribe_user`, and the outcome is cached in a `OnceLock` so the one-proof-per-invocation property is preserved across the several creation paths a multicast connect can take. A mismatch is stored as data rather than raised, so it can be re-raised at each creation site.

So an existing user being re-activated and an `--unsubscribe-feed`-only run now make no request at all. There is a test for the first case whose mock is `expect_request_proof().never()`, which fails if anything ever asks on that path.

config/src/constants.rs
elitegreg: Fixed, and confirmed empirically — one of the new binding-probe tests asserts exactly the pairing you describe (a local non-loopback source toward a loopback destination is rejected at the route lookup), so the constant really could never have worked.

Both halves are addressed. The probe now skips the bind on that pairing rather than failing the request. And the default is gone: `ENV_LOCAL_IP_VERIFIER_URL` is removed and localnet is `None` like the other three. Your second point is the deciding one — the client container's `localhost` is not the verifier, and #4204 is what will know whether the CLI reaches it by container name or on the host. Guessing here would have been the placeholder-that-is-wrong-on-arrival this PR argues against for the other environments. The reasoning is now recorded on the field itself, and `DZ_IP_VERIFIER_URL` / `--ip-verifier-url` cover the gap.

crates/doublezero-daemon-cli/src/ip_proof.rs
elitegreg: Fixed — the module doc was writing a cheque the function did not honour.

`proof_from_response` now takes the payer and user type that were requested and checks all three: `is_supported_version(parsed.version)`, the payer against the one asked for, and the user type against the one asked for. Any disagreement is `Malformed`, so it takes the reported, non-fatal path and the connect continues without a proof — which is the point, as you say: it turns a paid-for onchain rejection back into a clean skip. Three tests added.

crates/doublezero-daemon-cli/src/ip_proof.rs
elitegreg: Confirmed and comment corrected — `ConnectingTcpRemote::new` does `connect_timeout.and_then(|t| t.checked_div(addrs.len() as u32))` at `hyper-util-0.1.17/src/client/legacy/connect/http.rs:755`, exactly as you say.

The doc comment now states it is a total rather than a per-address value, names the division, and spells out the consequence (one A record gets the full second, eight get 125ms — under a single cross-ocean RTT, so healthy hosts would be abandoned). I left the value at 1s: it is correct for the one-address case that holds today, and whether a fleet behind this name should raise it or sit behind a single load-balancer or anycast VIP so it stays one address is really the same question #4199 has to answer. Flagged there rather than guessed at here.

crates/doublezero-daemon-cli/src/ip_proof.rs
elitegreg: Fair, and the rationale was the wrong part. Rewritten rather than the retry changed.

Worth adding to your analysis: forcing a fresh connection would not have rescued the claim either, since hyper tries the resolved addresses in order and a healthy-but-5xx first host would be picked again. The two cases do differ though — a transport error leaves no pooled connection, so that retry does redial and can land elsewhere; a 5xx came from a host that is up, whose keep-alive connection the pool hands straight back.

So `MAX_ATTEMPTS` now claims a transient blip rather than failover, and points failover at a load balancer or anycast VIP in front of the fleet (#4199), which is what the PR description already argued anyway. The canned server's doc comment now says that its `Connection: close` is the test making the attempt count legible, not what a real server does.

crates/doublezero-daemon-cli/src/connect.rs
elitegreg: Fixed — collapsed. It fell out of the item-3 rewrite, since `IpProofFetcher::request` now has only the `NotConfigured` arm and a catch-all.

I went with dropping it rather than giving the decline its own message: the `Declined` error already carries the service's own reason and message, so the operator gets the specific cause, and the follow-up line about onchain enforcement is genuinely identical for every reason.

crates/doublezero-daemon-cli/src/connect.rs
elitegreg: Half fixed, half deliberately not — worth explaining the split.

The swallowed config-load error was a real failure hiding, and now warns with the error named, so the case you describe (exported `DZ_IP_VERIFIER_URL`, no proof, no explanation) is distinguishable.

`NotConfigured` stays at `debug`, and item 4 above is why: dropping the localnet default means no environment carries a verifier URL now, so `NotConfigured` is every operator's expected state until #4199 and #4204 deploy something. A warning would fire on literally every `connect` in every environment. Once a verifier is deployed, `NotConfigured` becomes genuinely anomalous and worth raising — I think that belongs with the change that makes it so, rather than shipping the noise ahead of it.

crates/doublezero-daemon-cli/src/ip_proof.rs
```diff
@@ -0,0 +1,665 @@
...
+        .ok()
+        .and_then(|u| u.socket_addrs(|| None).ok())
+    else {
+        return SourceBinding::Unresolved;
+    };
+    if addrs.iter().any(|addr| socket.connect(addr).is_ok()) {
+        SourceBinding::Bind
+    } else {
+        SourceBinding::Unroutable
+    }
```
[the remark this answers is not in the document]
elitegreg: Confirmed and fixed in 3ddf0472.

The cause is that the binding decision has to be made before the client is built — reqwest only stores `local_address` in its connector config — so the probe's `getaddrinfo` runs *ahead* of `REQUEST_TIMEOUT` rather than inside it. Before the probe existed, resolution happened during the request and was covered. So this was a regression the probe introduced, and the 5.25s figure in the description really did not cover it.

Resolution now runs on a worker thread with a one-second deadline, `RESOLVE_TIMEOUT`, matching `CONNECT_TIMEOUT`. Exceeding it returns `Unresolved`, which the caller already reads as "do not bind" — the request then proceeds unbound and resolves the name itself, where `REQUEST_TIMEOUT` covers it and the failure carries a better message than the probe could produce. The abandoned thread ends when the lookup returns and its send into a dropped receiver just fails, which the comment says so it does not read as a leak.

`std` has no async resolver, and one short lookup per invocation does not justify pulling one in, so a thread is the only way to bound a call that cannot be cancelled. The deadline logic is split into `resolve_within` so it is testable with a slow closure rather than a resolver that has to be made to hang; two tests cover it.

crates/doublezero-daemon-cli/src/ip_proof.rs
```diff
@@ -0,0 +1,665 @@
...
+    if !status.is_success() {
+        return Err(classify_failure(status, &body));
+    }
+
+    let parsed: ProofResponse = serde_json::from_str(&body).map_err(|e| {
+        (
+            IpProofError::Malformed(format!("{e} (body: {})", body.trim())),
+            false,
+        )
+    })?;
```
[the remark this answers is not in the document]
elitegreg: Not changing this one — the asymmetry with `classify_failure` is deliberate rather than an oversight.

That path truncates junk arriving alongside a *non-success* status, where a snippet is all there is to learn from it. This path has a 2xx body that was supposed to *be a proof*, and the entire diagnostic is what arrived instead. A 200-character cap would sever exactly the malformed field on a body that is large but proof-shaped — a verifier emitting a slightly wrong field is the case this error exists to explain, and it is the one the cap would blind.

The captive-portal case is real, but it is rare, non-fatal, printed once, and `REQUEST_TIMEOUT` bounds what can be buffered in the first place. Trading the common diagnostic for the rare cosmetic one is the wrong way round here.

juan-malbeclabs (approved):

PR #4229, opened 2026-08-25
cli: connect with no mode provisions everything the access pass allows

juan-malbeclabs:
## Summary of Changes

- `doublezero connect` with no mode now provisions everything the server's AccessPass authorizes in one run: the IBRL tunnel plus a multicast tunnel joined to the groups or purchased feeds the pass grants. Previously the mode was required, so an operator had to know their entitlements and issue `connect ibrl` and `connect multicast` separately, and an unattended installer had to branch on the shape of the pass — branching the AccessPass already encodes onchain.
- The provisioning itself reuses the existing legs unchanged. `execute_multicast` called with empty group slices already meant "join every group this pass allows" and already covered the EdgeSeat purchased-feed path, so the new code is only entitlement gating, enabling the reconciler, and per-leg reporting.
- Each mode reports on its own line. One the pass does not cover is skipped with the reason instead of failing the run; the two are attempted independently, so a failure in one keeps the other's work and names the command that finishes the rest. The run exits non-zero if an attempted mode failed, or if the pass authorized nothing at all — exiting 0 having provisioned nothing is indistinguishable from success to an unattended installer.
- Epoch expiry still gates unicast only, so an expired pass connects multicast and skips IBRL. EdgeSeat seat caps are checked up front rather than letting `try_add_user` reject the transaction onchain, and are guarded by an "already exists" check because the program charges a seat only on create — a machine already holding a user re-runs cleanly even when its pass reports every seat in use.
- The bare form enables the reconciler up front, through a helper now shared with `doublezero enable`, so a run that provisions nothing still leaves the daemon managing tunnels. Previously enabling was only a side effect of a leg reaching `user_activated`.
- IBRL runs before multicast so the multicast leg's fresh user list places the multicast user on a different device, per [RFC-13](../blob/main/rfcs/rfc13-simultaneous_tunnels.md).
- `connect ibrl` and `connect multicast` are unchanged. The bare form's `--tenant`/`--allocate-addr` are deliberately not global, so they cannot leak into `connect multicast` or shadow `connect ibrl`'s own `-a`.

## Diff Breakdown

| Category     | Files | Lines (+/-)   | Net    |
|--------------|-------|---------------|--------|
| Core logic   |     2 | +270  / -12   |   +258 |
| Scaffolding  |     1 | +5    / -0    |     +5 |
| Tests        |     3 | +724  / -115  |   +609 |
| Docs         |     2 | +31   / -4    |    +27 |
| **Total**    |     7 | +1030 / -131  |   +899 |

Mostly tests: the core change is ~270 lines, while the test delta is inflated by a mechanical sweep converting 55 existing `Connect { .. }` literals to `Some(..)` + `..Default::default()`, which the field's type change forced.

<details>

- `crates/doublezero-daemon-cli/src/connect.rs` — `dz_mode` becomes `Option<DzMode>`; adds `execute_both` (entitlement gating + leg orchestration), `report_both` (per-leg summary and exit status), the `LegOutcome` enum, and `require_accesspass`. Both leg handlers are called unchanged.
- `crates/doublezero-daemon-cli/src/enable.rs` — extracts `ensure_reconciler_enabled` so `Enable::execute` and the bare connect share one implementation instead of duplicating it.
- `crates/doublezero-daemon-cli/src/cli.rs` — long help on the `Connect` variant explaining the bare form.

</details>

## Testing Verification

- 12 new unit tests in `doublezero-daemon-cli` covering: both modes provisioned on separate devices; IBRL-only when the pass grants no groups; multicast-only on an epoch-expired pass; the expired-and-no-groups case exiting non-zero; EdgeSeat unicast seats exhausted; an EdgeSeat re-run when both users already exist; each leg failing while the other succeeds; `--tenant`/`--allocate-addr` reaching the IBRL leg; and those flags being rejected alongside a subcommand. 197 tests pass in the crate (185 pre-existing, unchanged in behaviour).
- The four existing `enable` tests pass untouched, guarding the helper extraction.
- A clap test asserts all three forms parse (`connect`, `connect ibrl foo -a`, `connect multicast`), that the parent flags stay unset when a subcommand is used rather than aliasing its arguments, and that `connect multicast --allocate-addr` and `connect ibrl --tenant foo` are rejected rather than silently ignored.
- Verified against clap 4.6's `_propagate_global_args` that a `global = true` parent flag would have leaked into `multicast`; the flags are non-global for that reason.
- Ran all three forms against the real binary: each reaches the daemon preflight and fails there identically (no `doublezerod` on the dev box), confirming they parse and dispatch.

- New E2E test `TestE2E_Connect_AllModesFromAccessPass` (`e2e/connect_all_modes_test.go`) covers the single-invocation path end to end: on a devnet with two devices and a pass granting unicast plus `mg01`, a bare `doublezero connect` must bring up both tunnels on different endpoints, and the onchain Multicast user must hold exactly the publisher and subscriber roles the pass allowlists. The reconciler is disabled beforehand so that enabling it is attributable to the connect. It reuses `setupMultiTunnelDevnet`, and is picked up automatically by the e2e workflow's test discovery.

The E2E test has not been run locally — it is left to CI, where the e2e workflow shards and runs it on this PR.


e2e/connect_all_modes_test.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+		// The pass allowlists the client as both publisher and subscriber on mg01,
+		// so the auto-join must claim both roles and nothing beyond them.
+		return len(mcast.Publishers) == 1 && holds(mcast.Publishers, mg01) &&
+			len(mcast.Subscribers) == 1 && holds(mcast.Subscribers, mg01)
+	}, 60*time.Second, 2*time.Second,
+		"expected an activated IBRL user and an activated Multicast user publishing and subscribing to mg01")
+
+	log.Info("--> Bare connect provisioned both modes from the access pass")
+}
```
martinsander00, line 157: I would like a test that tests the following scenario too:
1. User's closest dzd is in nyc
2. User has access to 1 feed group in nyc
3. User runs `doublezero connect` and the closest dzd (let's call it `nyc-dzd1`) only has one interface available so the IBRL interface takes it.
4. Then what should happen IMO is that a second User would be created and connect to a second dzd in nyc (call it `nyc-dzd2`) and it should connect to it for the feed side of things.

I think this should work, the user would then have 2 "users" on the same machine each connected to a different device.

5. Then he should run `doublezero disconnect ibrl` and just disconnect from nyc-dzd1
6. Then run `doublezero disconnect multicast` and disconnect from dzd1

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -159,6 +183,87 @@ fn check_accesspass<L: LedgerClient>(
...
+        _ if provisioned => Ok(()),
+        _ => eyre::bail!(
+            "nothing to connect: the AccessPass authorizes no unicast connection and no \
+             multicast groups or feeds"
+        ),
+    }
+}
+
 impl Connect {
     pub async fn execute<D: DaemonClient, L: LedgerClient, W: Write>(
```
martinsander00, line 289: - On `doublezero connect --device ny5-dz01`, an `EdgeSeat` machine with shreds in New York and only `198.51.100.8` free on that device creates an `IBRL` connection on `.8`, then still pins multicast to `ny5-dz01`. `select_tunnel_endpoint` returns `0.0.0.0` and `execute_feed_join` writes that address (latency.rs:87, connect.rs:1311).
- The controller treats unspecified as the device public IP (server.go:618), so both GRE tunnels share `198.51.100.8`. Auto-select fails instead of writing. `--device` was already global on each subcommand. The new command applies it to both halves after `IBRL` just took the last address.
- Fix: Reject `0.0.0.0` before create, the same way `best_latency` already fails when no endpoint is left.

martinsander00: this would address this issue I think: https://github.com/malbeclabs/infra/issues/2128

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -257,6 +372,122 @@ impl Connect {
...
+    /// The provisioning itself is entirely the existing legs — `execute_ibrl`, and
+    /// `execute_multicast` with empty group slices, which already means "join every
+    /// group this pass allows" and covers the EdgeSeat purchased-feed path. What is new
+    /// here is the entitlement gating, enabling the reconciler, and per-leg reporting.
+    ///
+    /// IBRL runs first on purpose: the multicast leg re-reads `list_user()`, so it sees
+    /// the fresh unicast user and places the multicast user on a *different* device
+    /// (RFC-13 concurrent tunnels). Both legs are attempted even when one fails, so a
+    /// second-leg failure never discards the first leg's work.
+    async fn execute_both<D: DaemonClient, L: LedgerClient, W: Write>(
```
martinsander00, line 407: - After `IBRL` takes the only free endpoint, the command prints `Rerun 'doublezero connect multicast' to finish` (connect.rs:237). That rerun hits the same exclude list.
- The `IBRL` user is already on chain. `Multicast` cannot land until the operator runs `doublezero disconnect ibrl` or another endpoint is free. This is the old connect ibrl then connect multicast order, now with a hint that cannot finish the job.
- Fix: Name `doublezero disconnect ibrl` when the miss is a taken endpoint, or drop the rerun line in that case.

crates/doublezero-daemon-cli/src/connect.rs
juan-malbeclabs: Confirmed, and fixed in 2310174.

Traced the whole chain: `select_tunnel_endpoint` returns `UNSPECIFIED` once every endpoint is excluded (latency.rs:87), `find_or_create_device`'s `Some(device)` branch returned it unchanged, and nothing downstream rejects it — `User::validate` deliberately permits a zero `tunnel_endpoint` (user.rs:451, it only constrains the value when non-zero), so the create succeeds and server.go:621-626 maps it to the device public IP. Two GRE tunnels on one underlay source address, which is the exact collision that comment says per-endpoint addressing exists to avoid.

One thing worth adding: it isn't only `--device`. In `auto_join_purchased_feeds` the `if self.device.is_none()` guard at connect.rs:869 skips both the eligibility filter and the feed-metro filter, so `--device` bypasses `best_latency` entirely and goes straight to that fallback. That makes your `ny5-dz01` example deterministic rather than incidental.

Also worth being explicit that this predates the bare form — `connect ibrl --device X` followed by `connect multicast --device X` already does it today. The bare form just makes it reachable in one command with no second chance to notice, so the guard belongs here.

Fix is the one you suggested, hoisted so it covers every caller: `find_or_create_device` now rejects `UNSPECIFIED` before returning, and both it and `best_latency` fail with a typed `NoAvailableEndpoint` instead of an ad-hoc message. New unit test asserts the create never happens (the mock has no `create_user` expectation, so a write panics the test).

crates/doublezero-daemon-cli/src/connect.rs
juan-malbeclabs: Agreed, and worse than described: the PR shipped a test that pinned the bad advice. `test_connect_bare_multicast_failure_after_ibrl_success` was written for exactly this scenario — its doc comment says `exclude_ips` starves the multicast leg — and it asserted the `Rerun 'doublezero connect multicast' to finish.` line. So the unhelpful hint was locked in by a test, which is the part that most needed fixing.

Fixed in 2310174. `report_both` now checks the error chain for `NoAvailableEndpoint` and prints, for a starved multicast leg:

```
❌  Multicast failed: no tunnel endpoint is free: every endpoint on the eligible device(s) is already carrying a tunnel from this machine
    Rerunning 'doublezero connect multicast' will fail the same way. Free an endpoint with 'doublezero disconnect ibrl', or connect from a machine that can reach another device.
```

I did not make `doublezero disconnect ibrl` the headline advice, because taking down working unicast to get multicast is a trade the operator should choose rather than be told to make. It is named as the way to free an endpoint, alongside the other way out, and the line leads with the fact that a rerun is pointless — which was your actual point. Happy to shorten it to just the disconnect if you'd rather.

The existing test now asserts the new wording and asserts the old line is absent.

crates/doublezero-daemon-cli/src/connect.rs
juan-malbeclabs: I don't think it does — different field, and neither fix reaches the other.

malbeclabs/infra#2128 is about a user's `client_ip` being `0.0.0.0`. That one matters because the user PDA is seeded on `(client_ip, user_type)` with no owner in the seed, so a zero `client_ip` yields a single globally-shared account. Its fix is program-side, in `create_user_core`.

What you found above is `tunnel_endpoint` being `0.0.0.0`. That value is intentionally legal there — `User::validate` only constrains `tunnel_endpoint` when it's non-zero (user.rs:451), because the controller reads zero as "use the device's public IP" (server.go:621). Nothing about it is a PDA problem; it's an underlay-address collision. The fix is CLI-side, refusing to write a value that means "pick for me" when the picking already failed.

So #2128 still wants its own program-side change. Only the literal `0.0.0.0` is shared.

e2e/connect_all_modes_test.go
juan-malbeclabs: Added in cf276e79 as `TestE2E_Connect_SameMetroThenDisconnectPerMode`, with one substitution I want to flag.

What your steps 1-4 add over this test is **same metro**, and that turned out to be the important part. `setupMultiTunnelDevnet` puts its two devices in `ewr` and `pit`, so latency alone can separate the legs. Your scenario is two devices in one metro, where endpoint exclusion is the only thing that can move the second leg — and that's the shape a feed customer always sees, since a purchased feed only admits devices in its own metro. I parameterized the second device's metro and the new test runs both devices in `ewr`/`xewr`, asserting each leg landed on a different one of the two.

Steps 5-6 are now covered too: `doublezero disconnect ibrl` must drop only the IBRL tunnel and leave multicast on the device it started on (asserted both in daemon status and onchain), then `doublezero disconnect multicast` clears the rest.

**The substitution:** the pass grants a multicast group, not a purchased feed. Feed seats are written by the `SetAccessPassFeeds` instruction, which has no CLI surface — `doublezero access-pass set --accesspass-type edge-seat` creates `EdgeSeat(vec![])` with no seats — so the devnet can't provision a feed customer today without either adding that command or running the Feed Oracle. Both are bigger than this PR. The device-selection path under test is identical either way (both reach `find_or_create_device` with the same `exclude_ips`); what the substitution doesn't cover is the feed-metro filter itself, which stays on unit tests. Say the word if you'd rather I add an `access-pass set-feeds` command in a follow-up so this can be a real EdgeSeat test.

Your step 4 prediction was right, incidentally — that's what the pre-existing `TestE2E_MultiTunnel_FallbackToSecondDevice` and the bare-connect test both show. The bug your other comment found is what happens when there is no second device to fall back to.

martinsander00 (approved):

PR #4230, opened 2026-08-25
cli: drop the unreachable AccessPass pre-flight copy

### LINKED ISSUE #4202: cli: probe the dynamic (0.0.0.0) access pass PDA in check_accesspass
Called out in RFC-27's Impact section (tracker: #4194). Independent of the rest of the work.

`check_accesspass` looks up the AccessPass PDA at `(client_ip, payer)` only. A holder of a wildcard pass — stored at `0.0.0.0`, including the `EdgeSeat` passes from the shred-oracle — has no pass at their specific IP, so the pre-flight check reports no valid AccessPass and `connect` bails before it ever reaches the program, which would have accepted them.

Two copies exist and the comment on each says to keep them in sync:
- `smartcontract/cli/src/requirements.rs:87`
- `crates/doublezero-daemon-cli/src/connect.rs:147`

## What

Fall back to the `Ipv4Addr::UNSPECIFIED` PDA when the specific-IP lookup misses, applying the same epoch validity check. Consider collapsing the two copies into one shared helper while touching both.

## Acceptance

- Tests: specific-IP pass found; no specific pass but a valid dynamic pass; neither; dynamic pass present but epoch-expired.

elitegreg:
Closes #4202. Part of RFC-27 ([`rfcs/rfc27-ip-verification.md`](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc27-ip-verification.md), tracker #4194), though independent of the rest of that work.

## Summary

- Delete `smartcontract/cli/src/requirements.rs::check_accesspass`. It has no callers — `check_requirements` dispatches only `CHECK_ID_JSON`, `CHECK_BALANCE` and `CHECK_FOUNDATION_ALLOWLIST`, and nothing else in the repo references the function. That leaves one live copy of the pre-flight, in `doublezero-daemon-cli`, so the "keep the two in sync" comment each copy carried goes with it.
- No behavior change. `GetAccessPassCommand` has probed the dynamic (`0.0.0.0`) PDA before the exact-IP one since #3860, so the live pre-flight already accepts a wildcard-seat holder — including the `EdgeSeat` passes the Feed Oracle issues — without a fallback of its own. Its doc comment now records where that resolution order lives and that `CreateUserCommand` picks its PDA the same way, so the check accepts exactly what the transaction will send.

## What changed since the first review

The original version of this PR added a `None → lookup(0.0.0.0)` fallback to the pre-flight and hoisted both copies into a new SDK helper. Reviewers were right on both counts:

- The fallback could never fire. Both lookups routed through `GetAccessPassCommand`, which already reads the `0.0.0.0` PDA first, so a miss already proved that account absent; the fallback bought one extra RPC on the no-pass path. (@juan-malbeclabs, @martinsander00)
- There were not two live copies to deduplicate — one was dead code. (@juan-malbeclabs)

So the change is now the deletion, and the CHANGELOG entry is dropped: nothing user-visible moves.

The remaining gap @juan-malbeclabs identified — an epoch-stale `0.0.0.0` pass masking a valid exact-IP pass, which needs `GetAccessPassCommand` to prefer a *usable* pass so `create_user` sends the one the check blessed — is pre-existing on `main` and tracked separately in #4244.

## Testing Verification

- `cargo test -p doublezero_sdk -p doublezero-daemon-cli -p doublezero-serviceability-cli` green (860 tests).
- The pre-flight's body is byte-identical to `main`; only its doc comment changed. The resolution order it depends on is already pinned by the `GetAccessPassCommand` tests in `smartcontract/sdk/rs/src/commands/accesspass/get.rs` (`test_get_accesspass_prefers_dynamic_pass`, `_falls_back_to_exact_ip`, `_returns_none_when_neither_present`, `_unspecified_does_single_lookup`).
- Confirmed the deleted function is unreachable: `grep -rn check_accesspass --include=*.rs` returns only the daemon-cli copy and its call site.

### DIFF
--- a/crates/doublezero-daemon-cli/src/connect.rs
+++ b/crates/doublezero-daemon-cli/src/connect.rs
@@ -168,12 +168,16 @@ enum FeedJoinUser {
     },
 }
 
-/// AccessPass pre-flight: `Ok(false)` when no pass exists for
-/// `(client_ip, payer)` so the caller can render its own diagnostic before
-/// bailing. With `enforce_epoch`, the pass must also cover the current epoch.
+/// AccessPass pre-flight: `Ok(false)` when no pass exists for `(client_ip, payer)` so the caller
+/// can render its own diagnostic before bailing. With `enforce_epoch`, the pass must also cover
+/// the current epoch.
 ///
-/// Mirrors `check_accesspass` in `smartcontract/cli/src/requirements.rs` —
-/// keep the two in sync if AccessPass validity semantics change.
+/// [`LedgerClient::get_accesspass`] resolves through `GetAccessPassCommand`, which probes the
+/// dynamic (0.0.0.0) PDA before the exact-IP one, so a wildcard-seat holder — including the
+/// `EdgeSeat` passes the Feed Oracle issues — is found here without a second lookup. That is the
+/// same ordering `CreateUserCommand` uses to pick the PDA it sends, and the program accepts either
+/// (`smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs`), so
+/// this pre-flight accepts exactly what the transaction will.
 fn check_accesspass<L: LedgerClient>(
     ledger: &L,
     client_ip: Ipv4Addr,
--- a/smartcontract/cli/src/requirements.rs
+++ b/smartcontract/cli/src/requirements.rs
@@ -1,12 +1,6 @@
-use std::net::Ipv4Addr;
-
 use crate::doublezerocommand::CliCommand;
 use doublezero_sdk::{
-    commands::{
-        accesspass::get::GetAccessPassCommand,
-        allowlist::foundation::list::ListFoundationAllowlistCommand,
-    },
-    get_doublezero_pubkey,
+    commands::allowlist::foundation::list::ListFoundationAllowlistCommand, get_doublezero_pubkey,
 };
 use indicatif::ProgressBar;
 
@@ -79,32 +73,6 @@ pub fn check_balance(client: &dyn CliCommand, spinner: Option<&ProgressBar>) -> 
         }
         Err(e) => Err(eyre::eyre!("Unable to get balance: {:?}", e)),
     }
-}
-
-/// Mirrors `check_accesspass` in `doublezero-daemon-cli`
-/// (`crates/doublezero-daemon-cli/src/connect.rs`) — keep the two in sync if
-/// AccessPass validity semantics change.
-pub fn check_accesspass(
-    client: &dyn CliCommand,
-    client_ip: Ipv4Addr,
-    enforce_epoch: bool,
-) -> eyre::Result<bool> {
-    // A missing AccessPass returns Ok(false) (not an error) so the caller can
-    // render its own diagnostic (e.g. the client IP and UserPayer) before
-    // bailing, rather than surfacing a generic "not found" message.
-    let Some((_, accesspass)) = client.get_accesspass(GetAccessPassCommand {
-        client_ip,
-        user_payer: client.get_payer(),
-    })?
-    else {
-        return Ok(false);
-    };
-
-    if !enforce_epoch {
-        return Ok(true);
-    }
-    let epoch = client.get_epoch()?;
-    Ok(accesspass.last_access_epoch >= epoch)
 }
 
 pub fn check_allowlist(


smartcontract/sdk/rs/src/commands/accesspass/check.rs
```diff
@@ -0,0 +1,188 @@
...
+    epoch: E,
+) -> eyre::Result<bool>
+where
+    L: Fn(Ipv4Addr) -> eyre::Result<Option<AccessPass>>,
+    E: FnOnce() -> eyre::Result<u64>,
+{
+    let accesspass = match lookup(client_ip)? {
+        Some(accesspass) => Some(accesspass),
+        // Already the dynamic PDA — a second identical lookup would tell us nothing.
+        None if client_ip == Ipv4Addr::UNSPECIFIED => None,
```
martinsander00: I wonder if we should look for both, instead of falling back. Thoughts?

smartcontract/sdk/rs/src/commands/accesspass/check.rs
```diff
@@ -0,0 +1,188 @@
...
+        None => lookup(Ipv4Addr::UNSPECIFIED)?,
```
juan-malbeclabs: **The UNSPECIFIED fallback never fires in production.** Both `lookup` closures added by this PR go through `GetAccessPassCommand::execute`, which (since #3860) probes the UNSPECIFIED PDA *first* and only falls back to the exact-IP PDA — see `smartcontract/sdk/rs/src/commands/accesspass/get.rs:26-38`. So `lookup(client_ip) == None` already proves the `0.0.0.0` account is absent, and `lookup(Ipv4Addr::UNSPECIFIED)` re-reads the very same PDA and returns `None` again.

Concrete effect: on the no-pass path, every `doublezero connect` now pays one extra RPC round-trip before printing the same `Unable to find a valid AccessPass` message it printed before; the wildcard-seat case this doc comment describes was already handled one layer down. Nothing else changes.

Either push the resolution rule up here (make the closures do a plain exact-IP read, so this fallback is the only place the ordering lives) or drop the fallback and keep the helper as pure deduplication of the two copies.

smartcontract/sdk/rs/src/commands/accesspass/check.rs
```diff
@@ -0,0 +1,188 @@
...
+    };
+
+    let Some(accesspass) = accesspass else {
+        return Ok(false);
+    };
+
+    if !enforce_epoch {
+        return Ok(true);
+    }
+    Ok(accesspass.last_access_epoch >= epoch()?)
```
juan-malbeclabs: **The fallback is one-directional: it triggers on `None`, but not on "found but unusable"** — which leaves the mirror image of the bug this PR targets open.

With the dynamic-first `GetAccessPassCommand` both callers use, consider a payer who holds an epoch-stale `0.0.0.0` pass *and* a valid prepaid pass at their exact IP. (An `EdgeSeat` pass is a realistic source of the stale one: multicast is not epoch-gated — `epoch_allows_connection` / `UserType::is_epoch_gated` — so nothing keeps its `last_access_epoch` current.) `lookup(client_ip)` returns the stale dynamic pass, `last_access_epoch >= epoch()` fails, and `doublezero connect ibrl` bails with "Unable to find a valid AccessPass" even though `create_core.rs:207-231` would have accepted the exact-IP PDA.

The pre-flight is at least *faithful* today, because `CreateUserCommand` resolves its accesspass account through the same `GetAccessPassCommand` and would send the stale dynamic PDA and get rejected onchain. But that means the "the program accepts either PDA" premise this doc comment argues from is only half implemented: neither the check nor the create path picks a *usable* pass among the two. If the intent is to honor that premise, the resolution should evaluate both candidates and prefer one that passes the epoch check, in `GetAccessPassCommand` (so `create_user` sends the same one the check blessed) rather than only here.

smartcontract/sdk/rs/src/commands/accesspass/check.rs
```diff
@@ -0,0 +1,188 @@
...
+        let found = check_accesspass(
+            CLIENT_IP,
+            true,
+            recording_lookup(&seen, |ip| (ip == CLIENT_IP).then(|| pass(ip, EPOCH))),
+            || Ok(EPOCH),
+        )
+        .unwrap();
+
+        assert!(found);
+        assert_eq!(*seen.borrow(), vec![CLIENT_IP]);
```
juan-malbeclabs: This test pins an **exact-IP-first ordering that neither production caller has**: both `check_accesspass` wrappers resolve through `GetAccessPassCommand`, which queries the UNSPECIFIED PDA first and the exact IP second. So `assert_eq!(*seen.borrow(), vec![CLIENT_IP])` certifies a contract the shipped lookups violate, and the suite gives false confidence — a regression in the ordering that actually reaches the ledger would not fail anything here.

Same applies to `falls_back_to_a_valid_dynamic_pass`: its `[CLIENT_IP, UNSPECIFIED]` sequence is not the sequence a real `connect` issues.

Worth either scoping the names/asserts to the helper's own contract explicitly, or adding a test at the `GetAccessPassCommand` level so the two orderings cannot drift apart silently.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ All notable changes to this project will be documented in this file.
...
   - `doublezero connect` obtains an RFC-27 IP ownership proof from the verification service and attaches it to user creation; the address the service observes is authoritative, and where it disagrees with the daemon's own discovery `connect` stops and names both. A verifier that is unreachable, unconfigured, or that declines is reported and the connect continues without a proof, which the program accepts until `require-ip-ownership-proof` is set; `--ip-verifier-url` or `DZ_IP_VERIFIER_URL` points at one, and no environment has a built-in default yet. (#4201)
+  - The AccessPass pre-flight that `doublezero connect` and the serviceability CLI's requirement check run falls back to the dynamic (`0.0.0.0`) AccessPass PDA when the caller holds no pass at their specific IP, applying the same epoch check. A wildcard-seat holder — including the `EdgeSeat` passes the Feed Oracle issues — holds no pass at their exact IP, and the program accepts either PDA, so probing only the exact IP could report "no valid AccessPass" and bail before reaching the program that would have accepted them. (#4230)
```
juan-malbeclabs: Two inaccuracies in this entry, both user-facing:

1. It describes a behavior fix that does not land. `GetAccessPassCommand` has probed the `0.0.0.0` PDA first since #3860, so no wildcard-seat holder was being turned away on `main` — the PR body says as much. As written, the changelog tells operators that a released `doublezero connect` was rejecting `EdgeSeat` holders, which never shipped.

2. "the serviceability CLI's requirement check" does not run this. `smartcontract/cli/src/requirements.rs::check_accesspass` has no callers anywhere in the repo — `check_requirements` only dispatches `CHECK_ID_JSON`, `CHECK_BALANCE`, and `CHECK_FOUNDATION_ALLOWLIST`, and nothing else references the function. Nothing user-visible changes on that surface at all.

Suggest recasting as an internal dedup note (or dropping it from the user-facing section).

smartcontract/sdk/rs/src/commands/accesspass/check.rs
elitegreg: The code is redundant. See Juan's comment below.

smartcontract/sdk/rs/src/commands/accesspass/check.rs
elitegreg: Confirmed, and this is the comment that decided the PR. You are right that `lookup(client_ip) == None` already proves the `0.0.0.0` account is absent, so the fallback bought one extra RPC on the no-pass path and nothing else.

Taking your second option: the fallback is gone. And your CHANGELOG comment turned up the other half of it — there were never two live copies to deduplicate, since `requirements.rs::check_accesspass` has no callers. So instead of a shared helper the PR is now just the deletion of the dead copy. The live pre-flight in `connect.rs` is back to its `main` body byte for byte; only its doc comment changed, to say that `GetAccessPassCommand` owns the dynamic-first ordering and that `CreateUserCommand` picks its PDA the same way.

smartcontract/sdk/rs/src/commands/accesspass/check.rs
elitegreg: Agreed, and agreed it belongs in `GetAccessPassCommand` rather than in a pre-flight. The point about `create_user` needing to send the same pass the check blessed is what makes this a resolution problem rather than a check problem — a fix only here would just move the disagreement.

Out of scope for what this PR has shrunk to (a deletion, no behavior change), so I filed it as #4244 with your `EdgeSeat` / `is_epoch_gated` reasoning.

smartcontract/sdk/rs/src/commands/accesspass/check.rs
elitegreg: Right — and the tests went with the helper, so the false contract is gone.

On the second half: coverage at the `GetAccessPassCommand` level already exists, in `smartcontract/sdk/rs/src/commands/accesspass/get.rs` — `test_get_accesspass_prefers_dynamic_pass`, `_falls_back_to_exact_ip`, `_returns_none_when_neither_present`, `_unspecified_does_single_lookup`. That is where the ordering actually reaches the ledger, so the two orderings cannot drift: there is now only one.

CHANGELOG.md
elitegreg: Both correct, and (2) is what reshaped the PR — I had not checked that `requirements.rs::check_accesspass` is uncalled, and it is the reason the "two copies to keep in sync" premise did not hold.

Entry dropped rather than recast: deleting an unreachable function is not a notable change, and there is nothing user-facing left to describe.

smartcontract/sdk/rs/src/commands/accesspass/check.rs
elitegreg: Two answers, depending on which "both" you mean.

For the pre-flight: neither look-for-both nor fall back. `GetAccessPassCommand` already reads both PDAs, dynamic first, so anything here was redundant — that is Juan's comment below, and the fallback is now removed along with the helper.

For "evaluate both and pick the one that is actually usable": that is the right idea, and it is a real gap today — a stale `0.0.0.0` pass currently masks a valid exact-IP pass. But it has to happen inside `GetAccessPassCommand` so `create_user` sends the same pass the check approved. Filed as #4244.

smartcontract/cli/src/requirements.rs
```diff
@@ -81,30 +83,27 @@ pub fn check_balance(client: &dyn CliCommand, spinner: Option<&ProgressBar>) ->
...
-    Ok(accesspass.last_access_epoch >= epoch)
+        enforce_epoch,
+        |client_ip| {
+            Ok(client
+                .get_accesspass(GetAccessPassCommand {
+                    client_ip,
+                    user_payer: client.get_payer(),
+                })?
+                .map(|(_, accesspass)| accesspass))
+        },
```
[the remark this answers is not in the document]
elitegreg: Going the other way on this one. `requirements.rs::check_accesspass` is deleted in this revision — it had no callers, so there was no second copy to route through a helper.

On the proposed fix in general: `GetAccessPassCommand` is deliberately the single place the dynamic / exact-IP ordering lives, and `CreateUserCommand` resolves through it to pick the PDA it sends. Giving the pre-flight its own exact-IP-only read plus its own fallback would put a second copy of that ordering in the tree and let the check and the transaction drift apart. The correct place to improve the resolution is the command itself — tracked in #4244.

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -168,26 +169,21 @@ enum FeedJoinUser {
...
-        return Ok(true);
-    }
-    let epoch = ledger.get_epoch()?;
-    Ok(accesspass.last_access_epoch >= epoch)
+    check_accesspass_shared(
+        client_ip,
+        enforce_epoch,
+        |client_ip| ledger.get_accesspass(client_ip, ledger.get_payer()),
+        || ledger.get_epoch(),
+    )
```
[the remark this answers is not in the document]
elitegreg: The shared helper is removed in this revision, so this closure is gone and `check_accesspass` is back to a single `ledger.get_accesspass(client_ip, payer)` read.

Declining the proposed fix on the merits too: `GetAccessPassCommand` is the one place that owns the dynamic / exact-IP ordering, and `CreateUserCommand` uses it to choose the PDA it actually sends. Moving a copy of the fallback into the pre-flight would let the check approve one pass while the transaction sends another. The gap worth fixing — preferring a *usable* pass, not just the dynamic one — belongs in that command, and is tracked in #4244.

smartcontract/sdk/rs/src/commands/accesspass/check.rs
```diff
@@ -0,0 +1,188 @@
...
+
+use doublezero_serviceability::state::accesspass::AccessPass;
+use std::net::Ipv4Addr;
+
+/// Reports whether the caller holds a usable AccessPass for `client_ip`.
+///
+/// `lookup` resolves the pass stored at a given `client_ip` for the caller's payer; `epoch` reads
+/// the current ledger epoch, and is only called when `enforce_epoch` is set so the common
+/// "no pass at all" answer costs one RPC.
+///
```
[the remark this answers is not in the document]
elitegreg: Moot: the helper and its doc comment are removed in this revision. The remaining pre-flight in `connect.rs` does one pass lookup and reads the epoch only when `enforce_epoch` is set, and its comment no longer makes any RPC-count claim.

smartcontract/sdk/rs/src/commands/accesspass/check.rs
```diff
@@ -0,0 +1,188 @@
...
+/// A pass stored at [`Ipv4Addr::UNSPECIFIED`] (0.0.0.0) is valid for any client IP — that is how
+/// dynamic seats, including the `EdgeSeat` passes issued by the Feed Oracle, are held. The
+/// program accepts either the exact-IP PDA or the UNSPECIFIED one (see `create_core.rs`), so
+/// probing only the exact IP would report "no valid AccessPass" for a wildcard holder and bail
+/// before ever reaching the program, which would have accepted them. Hence the fallback.
```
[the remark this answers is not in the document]
elitegreg: Fair point, and kept. The file is deleted, but the surviving doc comment in `crates/doublezero-daemon-cli/src/connect.rs` now spells the path out in full: `smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs`.

smartcontract/sdk/rs/src/commands/accesspass/check.rs
```diff
@@ -0,0 +1,188 @@
...
+            true,
+            recording_lookup(&seen, |ip| {
+                (ip == Ipv4Addr::UNSPECIFIED).then(|| pass(ip, EPOCH - 1))
+            }),
+            || Ok(EPOCH),
+        )
+        .unwrap();
+
+        assert!(!found);
+    }
```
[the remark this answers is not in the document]
elitegreg: Correct on both lines, and the tests are removed with the helper. Worth noting they were asserting the wrong order anyway — exact-IP first, where every real caller queries `0.0.0.0` first through `GetAccessPassCommand`. The ordering that ships is pinned by the tests in `smartcontract/sdk/rs/src/commands/accesspass/get.rs`.

martinsander00 (approved):

PR #4232, opened 2026-08-25
e2e: run an ip-verifier in every devnet and cover the RFC-27 proof outcomes

### LINKED ISSUE #4204: dev/dzctl: run an ip-verifier in the local devnet

elitegreg:
Closes #4204. Part of RFC-27 ([`rfcs/rfc27-ip-verification.md`](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc27-ip-verification.md), tracker #4194). Depends on #4198.

## Summary of Changes

- **Every devnet runs an IP verifier**, `dzctl` and the Go e2e suite alike, so `connect` obtains a real RFC-27 proof and attaches it to user creation instead of diverging from production. `IPVerifierSpec.Disabled` is the opt-out; the zero value runs one.
- **Existing e2e tests now exercise the proof path for free.** No test file needed changing — the ordinary `connect` in every test obtains a proof and the program validates it.
- **The container attaches to the CYOA network** as well as the default network, and clients get `DZ_IP_VERIFIER_URL` pointing at its CYOA address. This is the substance of the change, not a detail: the service signs the source address it observes, and `connect` hard-fails on a proof for any address other than the one it is provisioning. A local client provisions its CYOA address, so the request has to arrive over the CYOA network for the two to agree — reached over the default network, the observed address would be the client's default-network address and every connect would fail on the mismatch. Same class of problem as the proxy handling in production.
- **The verifier keypair is generated per deploy** and its pubkey written to `GlobalState.ip_verifier_authority_pk` **before** the container starts. The service reads the authority at startup and exits if it does not name its own key, so the ordering is load-bearing.
- **Enforcement stays off**: `require-ip-ownership-proof` is clear, so a proof is attached but not demanded. `devnet.SetIPOwnershipProofFeatureFlag(ctx, bool)` turns it on.
- **New `e2e/ip_ownership_proof_test.go`** covers the three outcomes the shared connect path can have and that no other test distinguishes: a proof the program accepts, no proof at all, and a proof signed by a key the program does not trust.
- Two knobs support that last case: `ClientSpec.NoIPVerifier` leaves `DZ_IP_VERIFIER_URL` unset for one client, and `IPVerifierSpec.AuthorityRefreshSecs` pins how often the service re-reads the onchain authority — set it past the test, rotate the authority, and the service keeps signing with a key `GlobalState` no longer names.
- New doc: `e2e/docs/IP_VERIFIER_LOCAL_DEVNET.md`, covering the topology, the flag, and how to poke at the service.

## Diff Breakdown

| Core logic        |     6 | +385   / -0     |   +385 |
| Tests             |     2 | +260   / -0     |   +260 |
| Docs              |     2 | +92    / -0     |    +92 |
| Config/build      |     4 | +17    / -0     |    +17 |
| **Total**         |    14 | +754   / -0     |   +754 |

Entirely test-infrastructure: the "core logic" here is the devnet harness that brings the verifier up and wires clients to it. No production code is touched.

<details>

- [`e2e/internal/devnet/ip_verifier.go`](https://github.com/malbeclabs/doublezero/pull/4232/files#diff-9e77be4cf9074edc175a28527357296e095ced76c316839a9f2a19e7c9616aec) — the container: spec, CYOA addressing, keypair, `/health` gate, and the `AuthorityRefreshSecs` override
- [`e2e/ip_ownership_proof_test.go`](https://github.com/malbeclabs/doublezero/pull/4232/files#diff-b417d80663a6dc5cceb1a7e04f10e2e3f72e7108ea9a0c9a1056001dbfa5412e) — the three proof outcomes, each on its own single-device devnet
- [`e2e/internal/devnet/devnet.go`](https://github.com/malbeclabs/doublezero/pull/4232/files#diff-ca8d50a1f0d9f53e2b49a398c54da12626b332eb92ea6cb302eaea9621ce22e3) — start ordering: CYOA network, then authority onchain, then the container
- [`e2e/internal/devnet/smartcontract_init.go`](https://github.com/malbeclabs/doublezero/pull/4232/files#diff-b2b355699dc974cae32c03422bd3d0837bb3870fd70dd84aa76313b2d6f18edf) — `SetIPVerifierAuthority` and `SetIPOwnershipProofFeatureFlag`
- [`e2e/internal/devnet/client.go`](https://github.com/malbeclabs/doublezero/pull/4232/files#diff-c4df4bc9f768cbd09e772a6f59b2832b5dce2ff9370cc66504f2fd97466f8a6b) — `DZ_IP_VERIFIER_URL` wiring and the `NoIPVerifier` opt-out
- [`e2e/internal/devnet/ip_verifier_test.go`](https://github.com/malbeclabs/doublezero/pull/4232/files#diff-2ee94962d8643f2246b35a06fb7e4f92a91bd455e2f18b5650398d12ad73cea6) — `IPVerifierSpec.Validate`: host-ID defaulting and range, relative keypair path

</details>

## Testing Verification

**New e2e tests, all passing against real devnets:**

- `TestE2E_IPOwnershipProof_ValidProof` — connect reports `IP ownership verified for 9.236.198.100` (the client's CYOA address, not its default-network address), user provisioned, tunnel up. Asserts the run did *not* silently fall back to no proof.
- `TestE2E_IPOwnershipProof_NoProof` — a client with `NoIPVerifier` reports no proof and is provisioned anyway, which is the path an environment takes before its verifier exists.
- `TestE2E_IPOwnershipProof_UntrustedSigner` — with `AuthorityRefreshSecs: 3600` the service never observes the rotation, so it still issues a proof (asserted, so the test cannot pass vacuously by the service simply refusing), and the create is then refused: `IP ownership proof does not verify against the onchain verifier dKD6f2cN...`. No user is left onchain.

**Where the untrusted-signer refusal lands:** the SDK checks the proof against the onchain verifier key *before* building the transaction (`smartcontract/sdk/rs/src/commands/user/mod.rs:86`), so it fails client-side. That is the intended "refused before the transaction is paid for" pre-flight, but it does mean this test does not reach the program's own Ed25519 precompile check — reaching that needs a client that skips the pre-flight, which `connect` gives no way to do. Noted in the test's doc comment so it is not later mistaken for onchain coverage.

**Regression check for turning the verifier on everywhere:** `TestE2E_IBRL` and `TestE2E_Connect_AllModesFromAccessPass` both pass. The second is the one that mattered — a bare `connect` runs two legs and brings up two tunnels, so it is where a proof/tunnel address disagreement would have surfaced.

**Manual devnet run** (`dzctl build` → `start` → `add-device` → `add-client`): `/health` returns `{"epoch":0,"status":"ok","verifier_key":"matches"}`; with `require-ip-ownership-proof` **on** and the verifier stopped, connect fails onchain with `Program log: IP ownership proof required but none supplied` / `custom program error: 0x69`, and restarting the verifier restores it. That negative control is what shows the proof is load-bearing rather than the flag being inert.

## Notes for the reviewer

- **Resource cost.** Every e2e devnet now carries one more container — small next to the cEOS devices, but it is ~45 deploys across the shards. Worth watching the first full e2e run for memory pressure. A devnet opts out with `IPVerifier: devnet.IPVerifierSpec{Disabled: true}`.
- **Rate limits.** The devnet runs burst 1000 / 6000 per minute rather than the production 5 / 30. A devnet has one source address per client and a test can reconnect in a tight loop, which the production values would turn into `rate_limited` refusals unrelated to what is being tested. If you would rather tests confront the limiter, it is a one-line change.
- **CYOA host ID 250** is reserved for the verifier. Devices use single digits and clients 100–131 in the existing tests, so it is clear today, but it is a convention rather than an enforced allocation.
- **Still uncovered by e2e:** the enforcement path with the flag *on*. It is verified manually above but has no automated test. That is the natural follow-up.


e2e/internal/devnet/ip_verifier.go
```diff
@@ -0,0 +1,244 @@
...
+	if s.ContainerImage == "" {
+		s.ContainerImage = os.Getenv("DZ_IP_VERIFIER_IMAGE")
+	}
+	if s.CYOANetworkIPHostID == 0 {
+		s.CYOANetworkIPHostID = defaultIPVerifierCYOANetworkIPHostID
+	}
+	maxHostID := uint32(1) << (32 - cyoaNetworkSpec.CIDRPrefix)
+	if s.CYOANetworkIPHostID >= maxHostID {
+		return fmt.Errorf("hostID %d is out of valid range (1 to %d)", s.CYOANetworkIPHostID, maxHostID-1)
+	}
```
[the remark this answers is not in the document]
elitegreg: Changed in 8792223 — but the reason is not the one given here, and the difference matters for anyone reading this later.

255 would not have failed IPAM. The CYOA Docker network is allocated with a **/23** mask (`subnetAllocatorMask = 23`, in both `e2e/main_test.go` and `e2e/internal/devnet/cmd/devnet.go`), while `CYOANetworkSpec.CIDRPrefix` is 24. The address is derived from the real `SubnetCIDR`, so the broadcast host ID of the network the container actually joins is 511, not 255. Bounding by `CIDRPrefix` was already stricter than the network, and `DeriveIPFromCIDR` range-checks against the real prefix independently.

The genuine defect is that this check diverged from `ClientSpec.Validate` (client.go:88-93) and `DeviceSpec.Validate` (device.go:158-163), which are byte-for-byte identical to each other and use `2^hostBits - 1`. This one used `2^hostBits`, so it accepted the broadcast host ID **while its own error message said the valid range was `1 to 254`** — untrue on both ends, since it also never enforced the lower bound.

Now identical to the other two. The `0 -> default 250` sentinel stays ahead of the check, so 0 still means "use the default" rather than being rejected. Added tests pinning 255 rejected and 254 accepted; the existing test only covered 256, which passed under the old check too.

e2e/internal/devnet/ip_verifier.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		return false, fmt.Errorf("failed to inspect container: %w", err)
+	}
+	if !container.State.Running {
+		if err := v.dn.dockerClient.ContainerStart(ctx, container.ID, dockercontainer.StartOptions{}); err != nil {
+			return false, fmt.Errorf("failed to start ip-verifier: %w", err)
+		}
+	} else {
+		v.log.Debug("--> IPVerifier already running", "container", shortContainerID(container.ID))
+	}
```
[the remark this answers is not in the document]
elitegreg: Agreed, fixed in 8792223.

`Start` waits on `/health` through testcontainers; `StartIfNotRunning` called `ContainerStart` and returned, so the restart path silently dropped the gate. `StartIfNotRunning` now applies the same 60s `/health` wait when it actually started the container (an already-running one is untouched), and re-inspects on each tick so an exited container fails immediately with its exit code rather than burning the full budget.

**Verified manually**, since no automated test covers `dzctl`:

- Made the container die on startup, then `dzctl start`:
  ```
  Error: failed to start ip-verifier: ip-verifier did not become healthy: ip-verifier exited
  with code 1 (check `docker logs dz-local-ip-verifier`); the usual cause is
  GlobalState.ip_verifier_authority_pk no longer naming its key 2euZn2ZLiEC2fKfmKoj9zArcxe1tyZe1CjhAQBkUg4RA
  ```
  Before the change this reported success and left clients pointed at a dead service.
- Normal stop/restart still passes the gate and comes back healthy in ~4s.

Worth recording one thing the manual run turned up: the *specific* scenario of rotating the authority while the container is stopped **self-heals**, because `Devnet.Start` re-asserts the authority before restarting the verifier. So the exit-on-startup case is reachable through other failures, not that one.

**Scope, deliberately:** this fixes the verifier only. 8 of 10 components have the same gap — `Controller`, `Manager`, `InfluxDB` and `Prometheus` all define a `WaitingFor` strategy on the create path and drop it on restart. `Ledger` is the sole component that re-verifies (`ledger.go:156-171`, reusing its own `waitForSolanaReady`), and it is the precedent followed here. Fixing the rest is a real cleanup but belongs outside an RFC-27 PR; flagging so the verifier fix is not mistaken for the established convention.

e2e/internal/devnet/devnet.go
```diff
@@ -295,6 +302,24 @@ func New(spec DevnetSpec, log *slog.Logger, dockerClient *client.Client, subnetA
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to generate ip-verifier keypair: %w", err)
+		}
+		spec.IPVerifier.KeypairPath = ipVerifierKeypairPath
+		if generated {
+			log.Debug("--> Generated ip-verifier keypair", "path", ipVerifierKeypairPath)
+		} else {
+			log.Debug("--> Using existing ip-verifier keypair", "path", ipVerifierKeypairPath)
+		}
+	}
```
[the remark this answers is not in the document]
elitegreg: Agreed on the failure mode, fixed in 8792223 — with one correction and a different fix than proposed.

**Correction:** this is latent, not an active break. Both `SkipProgramDeploy` callers (`compatibility_test.go:562` and `:720`) start components by hand — `DefaultNetwork.CreateIfNotExists`, `Ledger`, `Manager`, `Funder` — and never call `Devnet.Start`, which is where the verifier block lives. They also never create a CYOA network. So the 60s `/health` timeout you describe is real but unreachable today. `TestE2E_BackwardCompatibility` passes on this branch both before and after the fix.

**The fix**, in `New` before validation:

```go
if spec.SkipProgramDeploy && spec.IPVerifier.KeypairPath == "" {
    spec.IPVerifier.Disabled = true
}
```

Requiring an explicit keypair path instead, as suggested, would make both existing compatibility devnets fail at `devnet.New` even though neither ever starts a verifier — turning a latent problem into an immediate one. Disabling by default keeps them working, still lets a stack that *does* have a verifier its cloned state trusts opt back in by naming the keypair, and has the side benefit that no unused keypair gets generated into the deploy directory.

It also lets the comment on the authority write stop describing a half-measure: reaching that line under `SkipProgramDeploy` now means the caller named a keypair explicitly, so the authority is already correct and needs no write.

.github/workflows/e2e.yml
```diff
@@ -291,6 +292,7 @@ jobs:
           docker push ${{ env.DZ_IMAGE_REPO }}/device-health-oracle:${{ env.DZ_IMAGE_TAG }}
           docker push ${{ env.DZ_IMAGE_REPO }}/geoprobe:${{ env.DZ_IMAGE_TAG }}
           docker push ${{ env.DZ_IMAGE_REPO }}/sentinel:${{ env.DZ_IMAGE_TAG }}
+          docker push ${{ env.DZ_IMAGE_REPO }}/ip-verifier:${{ env.DZ_IMAGE_TAG }}
```
juan-malbeclabs, line 295: The image is built and pushed here, but the shard job's `Pull pre-built images` step (~line 439) was not updated -- `ip-verifier` ends up the only e2e image missing from that list. Since `IPVerifierSpec.Disabled` defaults to false, every shard devnet now needs `.../ip-verifier:${DZ_IMAGE_TAG}`, and will instead fall back to testcontainers' implicit pull: one unretried pull per devnet, racing at `-parallel=12`, outside the `pull_with_retry` wrapper every other image gets. Worth adding `pull_with_retry ${{ env.DZ_IMAGE_REPO }}/ip-verifier:${{ env.DZ_IMAGE_TAG }}` next to the sentinel line.

e2e/internal/devnet/ip_verifier.go
```diff
@@ -0,0 +1,327 @@
...
+
+func (v *IPVerifier) Start(ctx context.Context) error {
+	v.log.Debug("==> Starting ip-verifier", "image", v.dn.Spec.IPVerifier.ContainerImage)
+
+	if err := v.Prepare(); err != nil {
+		return err
+	}
+
+	env := map[string]string{
+		"DZ_IP_VERIFIER_ENV":         "local",
```
juan-malbeclabs: `DZ_IP_VERIFIER_ENV=local` makes the service resolve its serviceability program ID from `ENV_LOCAL_SERVICEABILITY_PUBKEY` (`7CTniUa8...`), not from the program this devnet actually deployed -- the client container passes `DZ_SERVICEABILITY_PROGRAM_ID` explicitly for exactly this reason, and the service has no equivalent flag.

`devnet.New` generates a random serviceability program keypair whenever `ManagerSpec.ServiceabilityProgramKeypairPath` is unset, and `e2e/sentinel_multicast_publisher_test.go:30` does that. In that devnet `Ledger::ip_verifier_authority()` calls `get_account_data` on a GlobalState PDA that does not exist, so `check_at_startup` only warns and the status stays `AuthorityStatus::Unknown` -- which `health()` treats as servable (only `Mismatch` returns 503).

Net effect: the container comes up, `/health` returns 200, `StartIfNotRunning` reports success, and the invariant this file is built around ("the service reads GlobalState at startup and exits if the authority is not its own key", and "waiting on /health proves both") silently does not hold there -- it just logs a warning every 60s forever. Proofs still validate only because `SetIPVerifierAuthority` wrote the same key into the real program, i.e. by coincidence rather than by the check. Either pin that devnet to the fixed program keypair or give the service an explicit program-id override.

e2e/internal/devnet/ip_verifier.go
```diff
@@ -0,0 +1,327 @@
...
+		}
+	}
+	return started, nil
+}
+
+// waitForHealthy blocks until the container reports healthy, or fails as soon as it is clear it
+// never will. /health is 200 only once the cached ledger epoch is fresh and the ledger names this
+// container's key, so it proves the same two things the create path's wait strategy does.
+func (v *IPVerifier) waitForHealthy(ctx context.Context, containerID string) error {
+	port, err := v.dn.waitForContainerPortExposed(ctx, containerID, ipVerifierInternalPort, 10*time.Second)
```
juan-malbeclabs, line 179: The port wait runs before the container-state check, and Docker drops `NetworkSettings.Ports` entries for a container that is no longer running. So when the restarted verifier exits quickly -- the case this whole path was added for, where the authority rotated while the container was stopped and `check_at_startup` bails after a single local-ledger RPC round trip (tens of ms) -- this can lose the race, burn the full 10s, and return `failed to wait for ip-verifier port: ...`, never reaching the carefully worded `ip-verifier exited with code %d ... the usual cause is GlobalState.ip_verifier_authority_pk no longer naming its key` branch below. Diagnostics only, but it defeats the message where it matters most; inspecting for `!State.Running` before waiting on the port would keep it.

e2e/docs/IP_VERIFIER_LOCAL_DEVNET.md
```diff
@@ -0,0 +1,91 @@
...
+```bash
+# Health: 200 once the cached ledger epoch is fresh and the ledger names this key.
+docker exec dz-local-ip-verifier curl -sS localhost:8080/health
+
+# What the ledger thinks the authority is.
+docker exec dz-local-manager doublezero global-config authority get
+
+# A proof, as a client would ask for it.
+docker exec dz-local-client-<pubkey> \
+  curl -sS -X POST "$DZ_IP_VERIFIER_URL/v1/proof" \
```
juan-malbeclabs: `docker exec` runs `curl` directly, with no shell in the container, so `$DZ_IP_VERIFIER_URL` is expanded by the *host* shell where it is unset. Copy-pasted, this becomes `curl -sS -X POST /v1/proof ...` and fails with a protocol error rather than hitting the verifier. It needs a shell inside the container:

```bash
docker exec dz-local-client-<pubkey> bash -c \
  'curl -sS -X POST "$DZ_IP_VERIFIER_URL/v1/proof" \
   -H "content-type: application/json" \
   -d "{\"payer\":\"<pubkey>\",\"user_type\":0}"'
```

e2e/internal/devnet/ip_verifier.go
```diff
@@ -0,0 +1,351 @@
...
+		// than spending the whole budget waiting on it.
+		if exitErr := v.exitedError(ctx, containerID); exitErr != nil {
+			return false, exitErr
+		}
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+		if err != nil {
+			return false, err
+		}
+		resp, err := http.DefaultClient.Do(req)
```
juan-malbeclabs: `http.DefaultClient` has no timeout, and `poll.Until` calls the condition synchronously — so the `ipVerifierStartupTimeout` budget only bounds this loop if every request returns. The request carries the caller's `ctx`, not the poll deadline's derived one, so a health check that *hangs* rather than being refused (an accepting docker-proxy port with nothing answering behind it) blocks past 60s until the outer test context dies, which is the opposite of the bounded failure this function exists to give.

One line fixes it:

```go
var healthClient = &http.Client{Timeout: 5 * time.Second}
```

and use it instead of `http.DefaultClient`. Non-blocking, but cheap.

e2e/internal/devnet/client.go
```diff
@@ -234,6 +238,17 @@ func (c *Client) Start(ctx context.Context) error {
...
 	}
+	// Point `connect` at the devnet verifier. The `--env local` config carries no verifier URL,
+	// so without this a client obtains no proof at all. The URL is the verifier's CYOA address,
+	// so the source address it observes is the same one the client binds its tunnel to —
+	// `connect` hard-fails on a proof for any other address.
+	//
+	// NoIPVerifier leaves it unset, which is how a test covers the no-proof path: the CLI reports
+	// nothing to reach and creates the user without a proof, which the program accepts while
+	// require-ip-ownership-proof is clear.
+	if c.dn.IPVerifier != nil && c.dn.IPVerifier.InternalURL != "" && !c.Spec.NoIPVerifier {
```
juan-malbeclabs: The `InternalURL != ""` clause makes a real bug look like a supported mode. A non-nil `IPVerifier` whose `InternalURL` is empty means `Prepare` never ran — that is a programming error, not a configuration. Today it silently degrades the client to the no-proof path.

A test that asserts on the proof would still fail loudly, but every *other* test would quietly stop covering the proof path, which is exactly the "existing e2e tests exercise this for free" property the PR is built on — and it would fail no test while doing it. Since `Devnet.Start` always prepares before any client is added, I would drop the clause (`NoIPVerifier` already covers the intentional opt-out) or return an error there, so the invariant is enforced rather than papered over.

juan-malbeclabs (approved):
Approving. The design holds up and the parts that are easy to get wrong are right.

- **Start ordering is load-bearing and correct**: ledger → manager → CYOA network → `SetIPVerifierAuthority` → container. The service `?`s on `check_at_startup` (`main.rs:59`), so writing the authority first genuinely matters, and `Devnet.Start` sequences it that way.
- **The CYOA-attachment argument is real, not decorative**: `connect` hard-fails on an observed/provisioned address mismatch (`connect.rs:430`), so reaching the verifier over the default network would break every connect.
- **`/health` is readiness, not liveness** (`server.rs:228`) — 503 on a stale epoch or an authority mismatch — so gating on it proves what `ip_verifier.go` claims it does. The new `--serviceability-program-id` flag is what closes the hole where `AuthorityStatus::Unknown` (no `GlobalState` under the environment's constant program ID) would answer 200 and make the gate vacuous. Good catch.
- Verified the wiring end to end: `--ip-verifier-authority` (`globalconfig/authority/set.rs:30`), `feature-flags set --enable/--disable`, `solana.GenerateKeypairJSON`, `poll.Until`, `waitForContainerPortExposed` all exist; `e2e.yml` got all three image lines (env / push / pull); teardown is label-based so the container is cleaned up with no change needed. `go vet -tags e2e ./e2e/...` is clean, and the new `IPVerifierSpecValidate` subtests plus the two `settings.rs` tests pass locally.

Two non-blocking inline notes below, plus one thing worth fixing in the description:

**The PR body is stale.** It says *"Entirely test-infrastructure … No production code is touched"* with a 14-file / +754 table, but the branch is now **15 files, +920 / −7** and includes `crates/doublezero-ip-verifier/src/settings.rs` (+52 / −7) — a new production CLI flag on the verifier binary. The change is additive, defaults to the previous behavior, and has two unit tests, so it is fine on the merits; it just should not be invisible to the next reader.

**What I would want as follow-up** (agreeing with your own note): the enforcement path with `require-ip-ownership-proof` **on**. It is the path the whole RFC exists for and it is currently only manually verified. Worth watching the first full e2e run for memory pressure too, given the start-time gate is fatal — a flake in this one container turns every e2e test red, not just the new file.

e2e/internal/devnet/ip_verifier.go
elitegreg: @-

e2e/internal/devnet/client.go
elitegreg: @-

PR #4233, opened 2026-08-26
controlplane/telemetry: keep newest samples on partial buffer overcapacity

### LINKED ISSUE #4149: device/telemetry: over-capacity submitter path discards the whole backlog, not the oldest slice
When submission fails all attempts, `Tick` prepends the batch back only while there is room:

```go
overCapacity := s.cfg.Buffer.Len(partitionKey)+len(tmp) >= s.cfg.Buffer.Capacity(partitionKey)
if !success && !overCapacity {
    s.cfg.Buffer.PriorityPrepend(partitionKey, tmp)
}
```

(`controlplane/telemetry/internal/telemetry/submitter.go:246-252`)

The first time that check trips, the **entire** accumulated backlog goes rather than the oldest slice, and accumulation restarts from zero. Two problems follow:

- The samples that survive are worse than the ones that were lost. The account stores one `start_timestamp_microseconds` and one `sampling_interval_microseconds` with a flat `samples` array (`smartcontract/programs/doublezero-telemetry/src/state/device_latency_samples.rs:117-123`) and the reader synthesizes each timestamp as `start + i*interval` (`controlplane/telemetry/internal/data/device/latencies.go:205`). A dropped run leaves no marker, so every sample appended afterwards is backdated by the full length of the gap.
- Dropping everything loses far more than it has to. Keeping the newest `Capacity - Len` samples would bound the loss to what genuinely does not fit.

#4143 bounds `-max-epoch-staleness` to `partitionBufferCapacity × -probe-interval` with headroom, so the probe loop stops adding before this path is reached. That closes the route in but leaves the path itself sharp for any other caller or any future capacity change.

Fix: keep the newest `Capacity - Len` samples instead of discarding all of `tmp`, and log what was dropped. Note the slice handed to `Recycle` must stay the original one — reslicing from the middle would return a shifted array to the pool.

Reported by @ben-dz in review of #4143.

ayushsingh82:
## Summary

Fixes #4149.

- `Tick()`'s failed-retry path (`controlplane/telemetry/internal/telemetry/submitter.go`) replaced the binary "does the whole unwritten remainder fit, yes/no" check with `room := capacity - bufLen`.
- When the unwritten backlog doesn't fully fit, keeps the newest `room` samples (the tail — batches submit oldest-first, so `unwritten` is already chronological) and drops only the oldest excess, instead of discarding the entire backlog.
- Fixes an off-by-one where an exact-fit backlog (`bufLen + len(unwritten) == capacity`) was treated as over capacity and dropped in full; it now requeues in full.

## Test coverage

Rewrote 4 pre-existing tests whose assertions baked in the old boundary bug, rather than just adding new ones:

- `failed_retries_drop_when_over_capacity` → `failed_retries_keep_newest_when_over_capacity`: was asserting full drop at a boundary that should now partially keep.
- `no_backpressure_when_drop_on_overcapacity` → `kept_samples_never_exceed_capacity_after_partial_drop`: retargeted at the invariant that actually matters post-fix (buffer never exceeds capacity), since "producer never blocks" isn't true in general once partial-keep can fill a buffer to exactly capacity.
- `drops_failed_samples_when_requeue_would_meet_capacity` → `requeues_failed_samples_when_they_exactly_meet_capacity`: exact-fit (`bufLen + len(unwritten) == capacity`) now requeues in full instead of dropping — that was the off-by-one in the old `>=` check.
- `logs_and_counts_dropped_samples_when_over_capacity`: updated to a genuine 3-samples-into-capacity-2 case so it exercises partial drop and asserts the new `keptSamples` log field.

## Test plan

- [x] `go test ./internal/telemetry/...` — 8/8 top-level tests pass, 0 failures (includes all 3 renamed subtests)
- [x] `go vet ./internal/telemetry/...`
- [x] `gofmt -l` clean
- [x] `golangci-lint run ./internal/telemetry/...` — 0 issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
+- Device Telemetry
+  - A submission that fails every attempt while the partition buffer is at capacity now drops only the oldest unwritten samples that do not fit and requeues the newest that do, instead of discarding the whole unwritten batch the moment it stops fitting entirely. Also corrects an off-by-one that treated an exact-fit backlog (`bufLen + len(unwritten) == capacity`) as over capacity and dropped it in full. (#4233)
+
 ## [v0.40.0](https://github.com/malbeclabs/doublezero/compare/client/v0.39.0...client/v0.40.0) - 2026-09-11
 
 ### Breaking
--- a/controlplane/telemetry/internal/telemetry/submitter.go
+++ b/controlplane/telemetry/internal/telemetry/submitter.go
@@ -302,16 +302,30 @@ func (s *Submitter) Tick(ctx context.Context) {
 			// attempts, they are discarded; log and count the loss.
 			if !success {
 				unwritten := tmp[written:]
-				overCapacity := s.cfg.Buffer.Len(partitionKey)+len(unwritten) >= s.cfg.Buffer.Capacity(partitionKey)
-				if overCapacity {
+				capacity := s.cfg.Buffer.Capacity(partitionKey)
+				bufLen := s.cfg.Buffer.Len(partitionKey)
+
+				// room is how many of the unwritten samples still fit without exceeding capacity.
+				// Keep the newest `room` of them (the tail: batches are submitted oldest-first, so
+				// unwritten is already in chronological order) and drop only the rest, rather than
+				// discarding the whole slice the moment it stops fitting entirely.
+				room := max(capacity-bufLen, 0)
+				kept := unwritten
+				if len(unwritten) > room {
+					kept = unwritten[len(unwritten)-room:]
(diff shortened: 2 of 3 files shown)


ayushsingh82:
@elitegreg PR is up, ready for CI.

ayushsingh82:
@elitegreg following up on this one — all checks green, mergeable, fixes #4149 (submitter drops the whole backlog rather than the oldest slice on partial overcapacity). Happy to adjust anything; otherwise it's ready for review.

ayushsingh82:
Rebased onto main (v0.40.0 cut) and re-pushed, changelog entry moved back under Unreleased. `go test ./controlplane/telemetry/internal/telemetry/...` green.

Issue #4084, opened 2026-07-24
serviceability tests: prune User test literals to only the fields each test asserts on


nikw9944:
## Context

PR #4080 added a manual `impl Default for User` (`smartcontract/programs/doublezero-serviceability/src/state/user.rs`) and switched ~30 test literals across `smartcontract/cli`, `smartcontract/sdk/rs`, and `crates/doublezero-daemon-cli` to end with `..Default::default()`.

As merged, the struct-update tail absorbs exactly one field (`feed_pks`) — every literal still names the other 22 fields explicitly. That gets the churn-reduction benefit only for *future* `User` fields while keeping all of the current verbosity.

## Proposal

Lean into the pattern: prune each `User` test literal down to only the fields that test actually asserts on or that drive its behavior, letting `..Default::default()` fill the rest. This is where the manual `Default` impl pays for its existence — literals become a readable statement of what each test cares about, and future field additions touch no test files.

Same treatment could apply to `Device`/`GlobalState` literals, which already have manual `Default` impls following the same pattern.

## Caveat

`User::default()` picks non-neutral values (`status: UserStatus::Activated`, `user_type: UserType::IBRL`, matching the `Device::default()` precedent). When pruning, keep any field whose value the test's assertions depend on — dropping a field a test implicitly relies on would silently change what's under test. Worth doing file-by-file with tests passing at each step rather than as one mechanical sweep.

## References

- PR #4080 (introduced `impl Default for User` in commit ffca0c16)
- Related discussion: #2076 in malbeclabs/infra (the feed_pks change that prompted this)

ayushsingh82:
@nikw9944 — I'd like to pick this up. I traced through the ~50 `User { .. }` literals across the 23 files in `smartcontract/cli`, `smartcontract/sdk/rs`, and `crates/doublezero-daemon-cli` and confirmed the pattern still applies (`impl Default for User` at `smartcontract/programs/doublezero-serviceability/src/state/user.rs:299` hasn't changed since #4080).

Given the scope, I'd rather send this as a few small file-by-file PRs than one big sweep, per your note above. First one I have ready covers three CLI tests that turned out to need no fields at all — `execute()` in each of `user/delete.rs`, `user/request_ban.rs`, and `user/update.rs` never reads the mocked `get_user` return value, so the whole literal collapses to `User::default()`:

```
smartcontract/cli/src/user/delete.rs      | 32 +++++--------------------------
[...]

PR #4235, opened 2026-08-26
serviceability-cli: prune User test literals to only asserted fields

ayushsingh82:
Progress on #4084

## Summary of Changes
- `test_cli_user_delete`, `test_cli_user_request_ban`, and `test_cli_user_update` each build a full 22-field `User` literal only to hand it back through the mocked `get_user`. None of the three tests assert on any field of that value — `execute()` only cares that `get_user` returns *a* user so the command can proceed to `delete_user`/`update_user`/`request_ban_user`.
- Replaced each with `User::default()` (the manual `impl Default for User` added in #4080), with a comment noting why no field is load-bearing.

This is a first, narrow slice of #4084 — scoped to `smartcontract/cli/src/user/*`. The issue also names `smartcontract/sdk/rs` and `crates/doublezero-daemon-cli`, which have their own `User` literals (some with fields that likely *are* asserted on and need per-test judgment, per the issue's caveat about `User::default()`'s non-neutral values). Left those for follow-up PRs rather than one large mechanical sweep, per the issue's own guidance.

## Diff Breakdown
| Tests      |     3 | +14 / -80   | -66  |
| **Total**  |     3 | +14 / -80   | -66  |

Pure test cleanup — no production code touched.

## Testing Verification
- `cargo test --lib user::` in `smartcontract/cli`: all 29 user-module tests pass, including the 3 changed.
- `cargo clippy --lib -- -Dclippy::all -D warnings`: clean.
- Manually verified each of the 3 tests never reads a field of the `user` value it constructs — only its presence (via `get_user` returning `Ok`) drives the command under test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
+- CLI
+  - The `user delete`, `user request-ban` and `user update` tests build their expected `User` with `User::default()` instead of a 22-field literal whose non-asserted fields were noise. First slice of #4084. (#4235)
+
 ## [v0.40.0](https://github.com/malbeclabs/doublezero/compare/client/v0.39.0...client/v0.40.0) - 2026-09-11
 
 ### Breaking
--- a/smartcontract/cli/src/user/delete.rs
+++ b/smartcontract/cli/src/user/delete.rs
@@ -57,7 +57,7 @@ mod tests {
     };
     use doublezero_sdk::{
         commands::user::{delete::DeleteUserCommand, get::GetUserCommand},
-        AccountType, User, UserCYOA, UserStatus, UserType,
+        User,
     };
     use doublezero_serviceability::pda::get_user_old_pda;
     use mockall::predicate;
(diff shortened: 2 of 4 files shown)


ayushsingh82:
@nikw9944 PR is up: prunes the `User` test literals in `smartcontract/cli/src/user/*` down to only the fields each test asserts on, per #4084. Ready for CI.

ayushsingh82:
The only red check here is `changelog / check-changelog`; all 30 others pass. This PR is a test-only cleanup (pruning `User` literals in `smartcontract/cli/src/user/*` tests down to the asserted fields), no behavior change, so it seems like a `skip-changelog` case in line with #4094 / #4089. @nikw9944 could you add the label? Happy to add a CHANGELOG line instead if you'd prefer.

ayushsingh82:
@nikw9944 gentle follow-up — test-only cleanup for #4084, all checks green now. Ready for review whenever you have a moment; happy to rebase or adjust if anything's drifted.

ayushsingh82:
Rebased onto main and re-pushed. The auto-merge dropped the `Pubkey` import in `user/delete.rs`'s test module (main added an `accesspass_pk` field/test since this branched) — restored it; `cargo test -p doublezero-serviceability-cli user::` is 29/29 green.

PR #4243, opened 2026-08-28
e2e: cover RFC-27 proof enforcement with the feature flag set

### LINKED ISSUE #4205: e2e: cover IP ownership proof enforcement in user creation
Part of RFC-27 (tracker: #4194). Depends on #4197, #4201, and #4204.

## What

Tests in `e2e/`:
- Flag off, no proof: user creation succeeds (the pre-RFC-27 path still works).
- Flag on, valid proof: user created, tunnel provisioned, BGP established — the full connect path, not just the instruction.
- Flag on, no proof: rejected with the expected error.
- Flag on, proof for a different IP: rejected.
- Flag on, stale-epoch proof: rejected.
- Wildcard/`EdgeSeat` pass with the flag on: creation requires a proof and binds the observed IP, which is the case RFC-27 exists for.
- Verifier key rotated via `SetAuthority`: proofs signed by the old key are rejected.

## Acceptance

- Tests run under `make e2e-test RUN=...` individually, per the repo's guidance on not running the full suite concurrently.

elitegreg:
Closes #4205. Part of RFC-27 ([`rfcs/rfc27-ip-verification.md`](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc27-ip-verification.md), tracker #4194).

## Summary of Changes

- **`require-ip-ownership-proof` had no e2e coverage at all.** It was never set in any test, and `devnet.SetIPOwnershipProofFeatureFlag` had zero callers. One test now runs with the flag set, as four subtests over a single devnet.
- **Wildcard access passes get their first e2e coverage of any kind.** All 72 `access-pass set` call sites under `e2e/` name a `--client-ip`, so the case RFC-27 actually exists for was untested. A pass at the `0.0.0.0` PDA — the shape the shred-oracle issues — authorizes its payer for any routable address, which makes the proof the only thing binding `client_ip`. Covered both ways: with a proof the user binds the observed address and reaches BGP, without one the create is rejected onchain.
- **The sentinel exemption is covered**, because enforcement must not break the oracle path. The manager is the sentinel authority in a local devnet, so a manager-side `user create` still succeeds under enforcement while a client-paid one does not — the difference being the transaction payer, which is what `is_sentinel` compares.
- **An address mismatch is asserted client-side**, where the guard actually lives: `connect` binds its proof request to the address it provisions and refuses a proof for any other.
- One harness addition: `ClientSpec.DaemonClientIP` overrides the daemon's `-client-ip`, validated in `ClientSpec.Validate` alongside the other address-shaped fields. Needed because `doublezero connect --client-ip` is deprecated and explicitly ignored — the address comes from the daemon.

## Scope: what is *not* here, and why

Two facts bound what e2e can usefully assert, and they cut several of the issue's bullets:

**Most bad-proof cases never reach the chain.** The Rust SDK pre-flights version, payer, `client_ip`, `user_type` and the signature before it builds a transaction, and the runtime's Ed25519 precompile rejects bad signatures before the program runs. Of the program's proof errors only `IpOwnershipProofRequired` (105) and `IpProofEpochOutOfWindow` (110) are reachable end to end. Testing the others here would assert CLI strings while duplicating program tests.

**The program already covers them.** `smartcontract/programs/doublezero-serviceability/tests/user_ip_proof_test.rs` has 36 tests spanning every rejection condition, epoch window included, using `warp_to_epoch` — a `solana-program-test` facility with no equivalent against a real validator.

So this PR asserts 105 onchain and puts the rest of its weight on integration: a real verifier, a real ledger, a real tunnel, real BGP.

**Not covered — a valid proof on a specific-IP pass.** The flag is read at exactly one site, inside the `proof == None` arm (`ip_proof.rs`). A supplied valid proof takes a byte-identical program path with the flag set or clear, so such a subtest would duplicate `TestE2E_IPOwnershipProof_ValidProof` for the price of a client and a 90s `WaitForTunnelUp`. The "enforcement disturbs nothing downstream" assertion lives on `wildcard_pass_with_a_proof` instead, which is not a duplicate of anything.

**Deferred — stale epoch (issue bullet 5).** Feasible, but it needs two harness additions and is better as its own change: `LedgerSpec.SlotsPerEpoch` plumbed through the ledger entrypoint as `--slots-per-epoch 32` (silently ignored if the ledger volume already exists), plus `IPVerifierSpec` knobs for `epoch_refresh_secs`/`max_epoch_age_secs` so the service keeps signing a stale epoch while the ledger moves on. Today the devnet validator uses the default 432,000-slot epoch, so the epoch never advances — every existing proof test lives inside epoch 0, where the `{epoch-1, epoch}` window degenerates to `{0}`. Error 110 is the only other reachable onchain rejection, so this is the one remaining case with real e2e value; happy to file a follow-up issue.

**Deferred — rotated verifier key (bullet 7).** Already covered by `TestE2E_IPOwnershipProof_UntrustedSigner` from #4232, and it can only ever be a client-side assertion: the SDK's `verify` catches it before the transaction is built.

## Diff Breakdown

| Tests             |     1 | +221   / -0     |   +221 |
| Docs              |     2 | +27    / -2     |    +25 |
| Core logic        |     1 | +19    / -1     |    +18 |
| **Total**         |     4 | +267   / -3     |   +264 |

Almost entirely tests; the harness change is a spec field, a validation check and a two-line branch.

<details>

- `e2e/ip_ownership_proof_enforcement_test.go` — the enforcement test and its four subtests
- `e2e/docs/IP_VERIFIER_LOCAL_DEVNET.md` — a "testing enforcement" section, including the two traps below
- `e2e/internal/devnet/client.go` — `ClientSpec.DaemonClientIP` and its validation

</details>

## Testing Verification

`TestE2E_IPOwnershipProof_Enforced` passes in CI (e2e shard 3, 142.52s) and locally.

- **`wildcard_pass_with_a_proof`** — the one that matters. The pass names no address, so the proof is the only thing that could bind `client_ip`: `IP ownership verified for <addr>`, `✅  User Provisioned`, the user appears in `user list` at the observed address, and `WaitForTunnelUp` reaches `BGP Session Up`. Enforcement disturbs nothing downstream of the proof.
- **`wildcard_pass_without_a_proof`** — the same wildcard pass does not rescue a client with no verifier to reach. The program rejects it:
  ```
  Program log: Instruction: CreateUser(... client_ip: 9.210.238.101, ... ip_proof: false)
  Program log: IP ownership proof required but none supplied
  Program 7CTniUa88iJKUHTrCkB4TjAoG6TD7AMivhQeuqN2LPtX failed: custom program error: 0x69
  ```
- **`sentinel_authority_is_exempt`** — manager-side `user create` with no proof succeeds under enforcement.
- **`proof_for_a_different_address_is_refused`** — daemon set to an address the container does not own, so the verifier observes the real CYOA source:
  ```
  ❌  The verification service observed this host at 9.130.200.100, but the daemon is provisioning 9.0.0.7.
  ```

Also re-ran `TestE2E_IPOwnershipProof_ValidProof` locally, since `DaemonClientIP` touches shared client startup — passes (173s).

## Notes for the reviewer

- One devnet for the whole file. The two things a client cannot change after it starts — whether it has a verifier to reach, and what address its daemon provisions — are per-`ClientSpec`, so each costs one small container rather than a second ledger + manager + controller + cEOS device.
- Every rejection assertion keys on the *specific* error string rather than just a non-zero exit. A test that passes because connect broke for an unrelated reason was the main risk in this set.
- Two traps I hit and wrote into the docs, in case they save someone else the time: the manager is the sentinel authority locally, so "flag on rejects everything" is false for manager-driven creates; and `doublezero connect --client-ip` is deprecated and ignored, so the daemon flag is the only way to move the provisioned address.


e2e/ip_ownership_proof_enforcement_test.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+// A proof for an address other than the one being provisioned is refused.
+//
+// The daemon is told to provision an address this container does not own, so the CLI cannot bind
+// its proof request to it (`probe_source_binding`) and the request leaves from the real CYOA
+// address. The verifier signs what it observes, and `connect` refuses rather than binding an
+// address it has no proof for.
+//
+// This needs its own devnet: DaemonClientIP is fixed when the client container starts, so the
```
juan-malbeclabs: The stated reason for the separate devnet doesn't hold. `DaemonClientIP` is a per-`ClientSpec` field and `dn.AddClient` adds a client to an already-running devnet — this file does exactly that twice (lines 49 and 55) with different specs. A fourth client with `CYOANetworkIPHostID: 103, DaemonClientIP: unownedIP` on the shared devnet would behave identically, and the whole test could fold into `TestE2E_IPOwnershipProof_Enforced` as another subtest.

As written every CI e2e run pays for an extra ledger + manager + controller + cEOS device to cover one client-side guard.

Related: the flag enabled at line 194 is inert here. The refusal happens client-side in `crates/doublezero-daemon-cli/src/connect.rs:389` (`IpProofOutcome::Mismatch` → `mismatch_error`), before any transaction is built, so this test asserts the same output with the flag clear. `Enforced` in the name and the `SetIPOwnershipProofFeatureFlag` call imply enforcement coverage that isn't there.

e2e/ip_ownership_proof_enforcement_test.go
```diff
@@ -0,0 +1,232 @@
...
+	// A client picks its device from its own latency measurements; connecting before they exist
+	// fails on endpoint selection rather than on anything these tests are about.
+	for _, c := range []*devnet.Client{wildcard, unverified} {
+		require.NoError(t, c.WaitForLatencyResults(t.Context(), device.ID, 75*time.Second))
+		log.Info("--> Client added", "clientIP", c.CYOANetworkIP, "pubkey", c.Pubkey)
+	}
+
+	// The working path still works with enforcement on: proof obtained, attached, accepted,
+	// tunnel up.
+	t.Run("specific_pass_with_a_proof", func(t *testing.T) {
```
juan-malbeclabs: This subtest can't fail unless `TestE2E_IPOwnershipProof_ValidProof` already does. As the file header says, `RequireIpOwnershipProof` is read at exactly one site, inside the `proof == None` arm (`smartcontract/programs/doublezero-serviceability/src/ip_proof.rs:99-110`). A supplied valid proof takes a byte-identical program path with the flag set or clear, so this is a duplicate of the sibling test — and it costs a devnet client plus a 90s `WaitForTunnelUp`.

If it's deliberately a guard against a future flag that *does* gate the supplied-proof path, that intent is worth stating in the comment; otherwise it's wall clock for no marginal coverage.

e2e/ip_ownership_proof_enforcement_test.go
```diff
@@ -0,0 +1,232 @@
...
+// disagreement cannot be arranged on a client any other case can use.
+//
+// This asserts the client-side guard on purpose. The equivalent onchain error
+// (`IpProofClientIpMismatch`, 108) is unreachable through the SDK, which pre-flights the same
+// comparison before building a transaction; the program-level case is covered by
+// `test_proof_for_a_different_client_ip_is_rejected`.
+func TestE2E_IPOwnershipProof_EnforcedClientIpMismatchRefused(t *testing.T) {
+	t.Parallel()
+
+	// Routable, in the CYOA range, and not an address this container holds.
```
juan-malbeclabs: "in the CYOA range" is backwards. Devnet CYOA subnets are allocated from `9.128.0.0/9` (`e2e/main_test.go:101`), so `9.0.0.7` is deliberately *outside* every CYOA subnet — which is exactly what makes it unowned and makes the test work. Same for `sentinelUserIP = "9.0.0.9"` at line 149.

The property that matters is "publicly routable and not assigned to this host". As phrased, the comment invites someone to move the constant into the container's own /24, where it stops being safe from collision with an allocated container address.

e2e/internal/devnet/client.go
```diff
@@ -65,6 +65,12 @@ type ClientSpec struct {
 	// RFC-27 proof even in a devnet running a verifier.
 	NoIPVerifier bool
 
+	// DaemonClientIP overrides the address the daemon provisions, which is otherwise this client's
```
juan-malbeclabs, line 68: `DaemonClientIP` ends up the only address-shaped `ClientSpec` field with no validation — `ClientSpec.Validate` range-checks `CYOANetworkIPHostID` (lines 92-99) but never looks at this one.

A typo'd or IPv6 value goes straight to `doublezerod -client-ip`, where `DiscoverClientIP` rejects it (`client/doublezerod/internal/runtime/clientip.go:88-95`) and the daemon exits at startup. The test then fails ~30s later inside `AddClient` as an opaque `failed to wait for client daemon` timeout instead of a spec error at the call site. A `net.ParseIP(...).To4() == nil` check in `Validate` would fail fast like every neighbouring field.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file.
...
   - `.cursor/BUGBOT.md` and `.github/copilot-instructions.md` now tell Bugbot and Copilot to read the nearest sibling, flag a path that skips a zero or a duplicate, and assert a specific error and the exact log line at the expected index. Onchain checks apply only when the repository has onchain code. The eight path-scoped files under `.github/instructions/` are removed so Copilot reads only the repo-wide file. (#4247)
 - E2E/QA
+  - New e2e coverage for RFC-27 proof enforcement with `require-ip-ownership-proof` set: the working path still reaches BGP, a client with no verifier to reach is rejected with `IpOwnershipProofRequired`, a wildcard (`0.0.0.0`) access pass binds `client_ip` only when a proof is attached, the sentinel authority stays exempt so the oracle path keeps working, and `connect` refuses a proof whose address disagrees with the one it provisions.
```
juan-malbeclabs: Missing the PR reference. Every other bullet in the E2E/QA and CI sections ends with `(#NNNN)` — add `(#4243)`.

juan-malbeclabs (commented):
Reviewed the full diff and traced each behavioral claim back through the code it depends on.

Things I verified as correct, so they don't come up below: the wildcard `0.0.0.0` PDA path is real and is preferred over the exact-IP PDA (`smartcontract/sdk/rs/src/commands/accesspass/get.rs:27`, `create_core.rs:207-231`), so both wildcard subtests exercise what they claim; `access-pass set` with no `--client-ip` does land at `Ipv4Addr::UNSPECIFIED`; `IpOwnershipProofRequired` is custom error 105 and its `Display` string matches the asserted substring; the manager really is `sentinel_authority_pk` and `is_sentinel` compares the tx payer; `feature-flags set` is a read-modify-write of the mask only, so it can't clobber `ip_verifier_authority_pk`; the mismatch mechanism holds because `probe_source_binding` returns `NotLocal` for an unassigned address and falls back to default egress; the `DaemonClientIP` override touches only the daemon flag, leaving `Client.CYOANetworkIP` from docker inspect intact; and both new `TestE2E_` names are picked up by the shard matrix.

The findings are all about cost and comment accuracy rather than broken tests — the two big ones are that the second devnet doesn't look necessary, and that two of the three flag-set assertions don't actually depend on the flag.

e2e/ip_ownership_proof_enforcement_test.go
elitegreg: Agreed, and folded in. `TestE2E_IPOwnershipProof_EnforcedClientIpMismatchRefused` is gone; it is now the `proof_for_a_different_address_is_refused` subtest on a third client (`CYOANetworkIPHostID: 102, DaemonClientIP: unownedIP`) of the shared devnet. You are right that `AddClient` on a running devnet is all it needed — the "fixed when the container starts" fact is about the client, not the devnet, and I drew the wrong conclusion from it.

On the flag being inert there: also correct, so the comment now says so outright rather than implying enforcement coverage that isn't there — the guard is `IpProofOutcome::Mismatch` in `connect.rs`, client-side, and the subtest lives on this devnet because a client is cheaper than another ledger + manager + controller + cEOS device, not because the flag matters to it.

e2e/ip_ownership_proof_enforcement_test.go
elitegreg: Dropped. It was not a deliberate guard against a future flag — it was a duplicate, and your reading of the single read site in `ip_proof.rs:99-110` is the right one.

The one thing it did carry that the sibling doesn't is "enforcement disturbs nothing downstream of the proof", so I moved the `WaitForTunnelUp` into `wildcard_pass_with_a_proof`, which is not a duplicate of anything. Net effect is one fewer client and one fewer 90s tunnel wait, with the BGP assertion still made under enforcement. The file header now states the reasoning so the absence doesn't read as an oversight later.

e2e/ip_ownership_proof_enforcement_test.go
elitegreg: You're right, the comment had it exactly backwards. `unownedIP` is now a package-level const with the accurate reason, and it names both properties that matter so it doesn't get "corrected" into the container's own /24 later:

```go
// unownedIP is publicly routable and deliberately *outside* every devnet CYOA subnet — those are
// allocated from 9.128.0.0/9 (`main_test.go`) — so no container can hold it and it cannot collide
// with an allocated container address. Both properties matter: routable, or `connect` rejects it
// before the proof; unowned, or the daemon would bind it and there would be no disagreement.
```

`sentinelUserIP` now points back at that explanation instead of repeating the wrong one.

e2e/internal/devnet/client.go
elitegreg: Added, in `ClientSpec.Validate` next to the `CYOANetworkIPHostID` range check:

```go
// A bad DaemonClientIP is otherwise only caught by the daemon at startup, where it surfaces
// as an opaque "failed to wait for client daemon" timeout ~30s into AddClient.
if s.DaemonClientIP != "" {
	if ip := net.ParseIP(s.DaemonClientIP); ip == nil || ip.To4() == nil {
		return fmt.Errorf("daemon client IP %q is not a valid IPv4 address", s.DaemonClientIP)
	}
}
```

The `To4()` half is what catches the IPv6 case you called out, which `ParseIP` alone would accept.

CHANGELOG.md
elitegreg: Added `(#4243)`.

e2e/docs/IP_VERIFIER_LOCAL_DEVNET.md
```diff
@@ -70,6 +70,30 @@ attaches a real proof. Two knobs cover the cases that need something else:
 
 `e2e/ip_ownership_proof_test.go` uses all three paths.
 
+### Testing enforcement
+
+`devnet.SetIPOwnershipProofFeatureFlag(ctx, true)` sets `require-ip-ownership-proof`, which changes
+exactly one thing in the program: whether a *missing* proof is an error. A supplied proof is
+validated in full either way, so most proof failures are testable with the flag clear.
```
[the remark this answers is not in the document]
elitegreg: Fixed. These are methods on `*devnet.Devnet`, so the examples now read `dn.SetIPOwnershipProofFeatureFlag` and `dn.SetIPVerifierAuthority` — including the two occurrences on lines 57 and 67 that came in with #4232 and had the same mistake.

e2e/internal/devnet/client.go
```diff
@@ -223,7 +229,11 @@ func (c *Client) Start(ctx context.Context) error {
 	if c.Spec.LatencyProbeTunnelEndpoints {
 		extraArgs = append(extraArgs, "-latency-probe-tunnel-endpoints")
 	}
-	extraArgs = append(extraArgs, "-client-ip", clientCYOAIP)
+	daemonClientIP := clientCYOAIP
+	if c.Spec.DaemonClientIP != "" {
+		daemonClientIP = c.Spec.DaemonClientIP
+	}
+	extraArgs = append(extraArgs, "-client-ip", daemonClientIP)
```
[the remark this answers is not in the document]
elitegreg: Done, though in `ClientSpec.Validate` rather than at the call site in `Start`, which is where the other address-shaped spec fields are checked and so fails at the same point they do. It uses `net.ParseIP(...).To4() == nil` so an IPv6 literal is rejected too.

elitegreg:
Pushed as an amend to the single commit on the branch. All six threads addressed; replies are on each one.

The structural changes, since they are easier to see together than thread by thread:

- **One devnet, not two.** The mismatch test folded into `TestE2E_IPOwnershipProof_Enforced` as a fourth subtest on a third client. A separate ledger + manager + controller + cEOS device for one client-side guard was not justified.
- **`specific_pass_with_a_proof` dropped** as a duplicate of `TestE2E_IPOwnershipProof_ValidProof`. Its one non-duplicate assertion, that enforcement disturbs nothing downstream of the proof, moved to `wildcard_pass_with_a_proof`, so BGP is still asserted under enforcement — one fewer client and one fewer 90s tunnel wait.

Net: 2 devnets + 4 clients down to 1 devnet + 3 clients. `DaemonClientIP` is now validated in `ClientSpec.Validate`, the `unownedIP` comment says the true reason the address works, the doc examples name the right receiver, and the CHANGELOG bullet carries `(#4243)`.

Verified: `TestE2E_IPOwnershipProof_Enforced` passes in CI (e2e shard 3, 142.52s, all four subtests) and locally (210s). `TestE2E_IPOwnershipProof_ValidProof` re-run locally as well, since `DaemonClientIP` touches shared client startup.

The PR description is updated to match — the old one described five tests and five devnets, and argued for the separate mismatch devnet.

juan-malbeclabs (approved):
Approving. I re-checked each thread against the amended commit rather than against the replies, and the code matches what they claim.

- **One devnet.** The mismatch test is now the fourth subtest on a third client of the shared devnet, and its comment says outright that the guard is client-side and flag-independent instead of implying enforcement coverage. That is the right trade: a client is one small container, a devnet is a ledger + manager + controller + cEOS.
- **The duplicate is gone,** and the one thing it carried that the sibling doesn't — "enforcement disturbs nothing downstream of the proof" — survives as the `WaitForTunnelUp` inside `wildcard_pass_with_a_proof`. Coverage went up while wall clock went down.
- **`unownedIP`** now names both properties that make the address work (publicly routable, and deliberately outside `9.128.0.0/9`), so the next person doesn't move it into the container's own /24.
- **`DaemonClientIP` validation** landed in `ClientSpec.Validate` next to the `CYOANetworkIPHostID` range check, with the `To4()` half that catches an IPv6 literal — it fails at the same point every neighbouring address-shaped field does.

Two things I confirmed beyond reading the diff:

- The test runs, it doesn't merely compile: `e2e (shard 3)` reports `--- PASS: TestE2E_IPOwnershipProof_Enforced (142.52s)` with all four subtests passing.
- The strings the assertions hang on are the real ones — `error.rs:222` for `IpOwnershipProofRequired` and `connect.rs:418` for the address disagreement — so neither assertion can pass on an incidental failure.

The file header earning its length is worth calling out separately: stating *why* the reachable surface is two errors and only one of them testable here means the next person doesn't re-derive it, or worse, "fix" the gap by adding subtests that take a byte-identical program path.

PR #4247, opened 2026-08-31
repo: tell Bugbot and Copilot what to flag

martinsander00:
## Summary

- This pull request adds review checks to the Bugbot file and the Copilot file so both agents read the nearest sibling before they review a new instruction.
- Both files now tell the reviewer to flag a wrapper that only forwards, a key taken as an argument when the account already holds it, and a path that skips a zero or a duplicate.
- Both files now tell the reviewer to assert a specific error and the exact log line, and to reach the start state through the real earlier instruction.

## Testing

- Open this pull request and wait for the Cursor Bugbot check.
- If the check does not start, comment `cursor review` on this pull request.
- Confirm a finding uses the headings Issue, Context, and Proposed Fix.


.github/copilot-instructions.md
```diff
@@ -1,14 +1,62 @@
...
+
+- When two arguments must both be set or both be empty, such as a validator count and a merkle root, flag a path that accepts one set and one empty. The instruction must revert.
+
+- When a later instruction decides the earlier work is done by reading a token balance or whether an account exists, flag that read. The earlier instruction must set a done flag on its own account.
+
+- When the change ties two accounts, such as a distribution and a shred distribution, check that the instruction proves they share the same epoch or parent.
+
+- When two token accounts share the same owner, flag a close or withdraw that only checks the owner. The instruction must also check the seeds of the account it means to touch.
+
+- When the change checks owner, signer, writable, or overflow after the account loader, the token program, or a CPI already reverts on that case, flag the extra check.
```
nikw9944: **Blocking** — The file tells the reviewer to flag owner, signer, and overflow checks as redundant, so an author can be asked to delete a guard this repository requires. These programs are native `solana-program` with no Anchor account loader and no token CPIs, and `.github/instructions/solana-programs.instructions.md` requires verifying the owner of every PDA read and using checked arithmetic because the deployed SBF release build wraps instead of reverting.

.github/copilot-instructions.md
```diff
@@ -1,14 +1,62 @@
...
+
+**Bad input**
+
+- When the change skips, dedupes, or no-ops a duplicate or a zero, flag that path. The path must revert.
+
+- When the change adds a flag such as `--all` so the command can choose a subset, flag the flag if the command should always process every unset item.
+
+**Tests**
+
+- In tests, assert a specific error and the exact log line. Do not use `.is_err()` or `.contains()`. Those keep passing after the failure or the message changes.
```
nikw9944: **Minor** — `.contains()` and `.is_err()` are banned with no scope, so the reviewer will also flag production code: `(0.01..=1000.0).contains(&delay)` in `crates/doublezero-cli-core/src/validators.rs:47` is a range check, not an assertion. `.github/instructions/testing.instructions.md` already carries this rule scoped to test paths.

.github/copilot-instructions.md
```diff
@@ -1,14 +1,62 @@
...
+
+- When a one-shot test adds a helper to `tests/common`, flag the helper. Keep it in that test file.
+
+**Onchain**
+
+For a repository with onchain code:
```
nikw9944, line 38: **Minor** — Several onchain bullets describe programs this repository does not contain: token vaults, SPL token accounts, merkle roots, and shred distributions (lines 50-62) are only decoded read-only by `sdk/revdist` and `sdk/shreds`. Line 32 also puts "must revert" in the unscoped section, where a Go or TypeScript reviewer cannot apply it.

.github/copilot-instructions.md
```diff
@@ -1,14 +1,62 @@
...
+
+- When the change adds a method that only calls a shared helper and returns, flag the method. Callers can call the shared helper.
+
+- When every caller of a shared helper repeats the same extra check, flag the copies. Put the check in the helper.
+
+**Bad input**
+
+- When the change skips, dedupes, or no-ops a duplicate or a zero, flag that path. The path must revert.
+
+- When the change adds a flag such as `--all` so the command can choose a subset, flag the flag if the command should always process every unset item.
```
nikw9944: **Minor** — "every unset item" names nothing in this repository, so the reviewer cannot decide when an `--all` flag is wrong. Say which command and which items.

nikw9944 (changes requested):
Line 58 tells the reviewer to flag owner, signer, and overflow checks as redundant, which contradicts the onchain rules this repo already ships in `.github/instructions/` — the programs are native Solana with no account loader and no `overflow-checks` profile. Several other new bullets are unscoped or describe programs this repo does not contain, so they will fire on Go and TypeScript reviews. Every finding applies twice: `.cursor/BUGBOT.md` and `.github/copilot-instructions.md` are byte-identical.

.github/copilot-instructions.md
martinsander00: The new rule flags a second owner check only when the account-walk helper already verified the owner.

.github/copilot-instructions.md
martinsander00: The ban now applies only in tests that match a substring. A test must pin the error and the log line at the expected index.

.github/copilot-instructions.md
martinsander00: Onchain bullets no longer name programs this repository does not contain. I also removed the eight path-scoped instruction files so Copilot reads only the repo-wide file.

.github/copilot-instructions.md
martinsander00: I removed that bullet. It named nothing in this repository.

nikw9944 (approved):

PR #4254, opened 2026-09-01
offchain/scheduler: fail the boot when SOLANA_RPC is unset

bgm-malbeclabs:
Three findings raised on #4240 that could not be fixed there. That pull request gates on each imported tree staying byte-identical to its filtered source, so any edit to `offchain/` or `solana/` had to wait for a follow-up. **Stacked on #4240** and based on that branch, so the diff shown here is these fixes alone. It retargets to `main` once #4240 merges.

## Summary

- **The scheduler refuses to boot when `SOLANA_RPC` is unset.** All three workers read the one `:solana_rpc` config key and hand it straight to the NIF, which wants a `String`. `config/runtime.exs` read the variable with `System.get_env`, so an unset variable reached Rust as `nil`. `System.fetch_env!` stops the boot instead. One line in the config covers all three workers, which is where the single cause was, rather than a guard in each of them.
- **The test environment is exempt.** CI runs `mix test` with no `SOLANA_RPC`, and the tests set the key themselves through `Application.put_env`. An unguarded `fetch_env!` would break the release workflow's test step.
- **`DZ_LEDGER_RPC` no longer sets a key nothing reads.** It set `:ledger_rpc`, and no worker, module or NIF callback reads that key. It has the same unset-variable gap, so leaving it in place would invite a fix to dead config.
- **The scheduler README closes its shell code block**, so the Installation heading and everything after it stop rendering inside it.
- **The revenue-distribution journal test is renamed** from `test_lifetime_swept_2z_amount` to `test_lifetime_swapped_2z_amount`, matching the `lifetime_swapped_2z_amount` field and method it covers, so a search for the swapped-amount tests finds it.

Reported in the copilot review on #4240 and tracked as malbeclabs/infra#2395.

## Diff Breakdown

|---|---|---|---|
| Core logic | 1 | +7 / -2 | +5 |
| Docs | 2 | +7 / -0 | +7 |
| Tests | 1 | +1 / -1 | 0 |

One behaviour change, in a config file. The rest is a heading that rendered in the wrong place and a test name.

## Testing Verification

- `Config.Reader.read!("config/runtime.exs", env: ...)` run against each case: the key is absent in the test environment with the variable unset, the read raises `System.EnvError` in prod with it unset, and it carries the URL in prod with it set.
- `cargo test -p doublezero-revenue-distribution --lib test_lifetime_swapped_2z_amount` passes from inside `solana/`.
- `mix format --check-formatted config/runtime.exs` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ All notable changes to this project will be documented in this file.
   - `.cursor/BUGBOT.md` and `.github/copilot-instructions.md` now tell Bugbot and Copilot to read the nearest sibling, flag a path that skips a zero or a duplicate, and assert a specific error and the exact log line at the expected index. Onchain checks apply only when the repository has onchain code. The eight path-scoped files under `.github/instructions/` are removed so Copilot reads only the repo-wide file. (#4247)
 - E2E/QA
   - Remove `TestQA_MulticastSettlement`. It funded a seat through `doublezero-solana shreds pay`, which is going away. The agent seat-pay RPC now returns Unimplemented if something still calls it. Unused settlement helpers go with the test. (#4248)
+- Offchain
+  - The scheduler refuses to boot when `SOLANA_RPC` is unset, rather than handing `nil` to the Rust NIF. All three workers read the one config key, so the check sits in `config/runtime.exs` where the single cause was. The test environment is exempt, since the tests set the key themselves. `DZ_LEDGER_RPC` no longer sets a `ledger_rpc` key that nothing in the application read. (#4240 follow-up)
+  - The scheduler README closes its shell code block, so the Installation heading and everything after it stop rendering inside it.
+- Solana programs
+  - Rename `test_lifetime_swept_2z_amount` to match the `lifetime_swapped_2z_amount` field and method it covers, so a search for the swapped-amount tests finds it.
 
 ## [v0.38.0](https://github.com/malbeclabs/doublezero/compare/client/v0.37.0...client/v0.38.0) - 2026-08-28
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,11 @@ All notable changes to this project will be documented in this file.
   - `.cursor/BUGBOT.md` and `.github/copilot-instructions.md` now tell Bugbot and Copilot to read the nearest sibling, flag a path that skips a zero or a duplicate, and assert a specific error and the exact log line at the expected index. Onchain checks apply only when the repository has onchain code. The eight path-scoped files under `.github/instructions/` are removed so Copilot reads only the repo-wide file. (#4247)
 - E2E/QA
   - Remove `TestQA_MulticastSettlement`. It funded a seat through `doublezero-solana shreds pay`, which is going away. The agent seat-pay RPC now returns Unimplemented if something still calls it. Unused settlement helpers go with the test. (#4248)
+- Offchain
+  - The scheduler refuses to boot when `SOLANA_RPC` is unset, rather than handing `nil` to the Rust NIF. All three workers read the one config key, so the check sits in `config/runtime.exs` where the single cause was. The test environment is exempt, since the tests set the key themselves. `DZ_LEDGER_RPC` no longer sets a `ledger_rpc` key that nothing in the application read. (#4240 follow-up)
```
martinsander00: fetch_env! only fails when the variable is missing. SOLANA_RPC= (empty) still boots and still hands "" to the NIF.

martinsander00 (approved):

CHANGELOG.md
bgm-malbeclabs: Right, and the empty case lands in exactly the same place: `fetch_env!` raises only when the variable is absent, so `SOLANA_RPC=` would hand `""` to the NIF instead of `nil`.

Fixed in 03630ec3c. The value is trimmed and an empty one raises, naming the variable and why it matters. Checked each case with `Config.Reader`: absent in the test environment with the variable unset, `System.EnvError` in prod with it unset, a `RuntimeError` in prod with it empty or whitespace-only, and the URL in prod with a real value.

PR #4268, opened 2026-09-02
sdk,cli: send the per-pass-type access pass removal instructions

bgm-malbeclabs:
Part of malbeclabs/infra#2470. Child of malbeclabs/infra#2385.

#4267 added one removal instruction per `AccessPassType` and is merged. Nothing sent them. This PR moves every caller in the repository onto them — the builders, both SDKs, the CLI, the client daemon, the stress tools and the e2e suite.

**Variants 69 and 42 stay live and keep skipping the pass type check.** Deprecating them was in an earlier version of this PR and has been pulled out into a follow-up. The reason is a cross-repo ordering problem: `shreds-e2e.yml` builds the oracle from `malbeclabs/doublezero-shreds` at its default branch, that oracle still sends variant 42, and merged main carries no builder it could use instead — `delete_user` builds only 42 and `close_access_pass` only 69. So the oracle could not move first, and deprecating here first turns its user removals into `Deprecated` and fails `TestE2E_FeedSubscriptionOracleExpiryTeardown`, which is exactly what CI showed. This PR ships the builders the oracle needs; once it has moved, the follow-up flips the switch.

The only change to the program is two doc comments pointing at the per-type variants.

## Summary of Changes

- `doublezero access-pass close` and `doublezero user delete` both require `--accesspass-type`, with values `prepaid`, `solana-validator`, `solana-rpc`, `others`, `edge-seat`. **Neither has a default, deliberately.** A default would pick a kind for an operator who did not state one, which is the exact failure this change exists to prevent. One spelling across all three verbs: `access-pass set` already ships `--accesspass-type`, so the two new flags match it rather than introducing `--type` and `--access-pass-type` alongside it.
- `close_access_pass` and `delete_user` in the instruction crate pick their variant from an `AccessPassKind` the caller supplies, variants 125 to 134.
- `CloseAccessPassCommand` and `DeleteUserCommand` carry a `kind`. On `DeleteUserCommand` it is an `Option`, which is the distinction that matters (see below).
- The Go SDK's `Executor.DeleteUser` takes a kind. Its zero value is `AccessPassKindUnspecified` and is rejected before the account fetch, rather than silently meaning `Prepaid` — Go cannot make a parameter required, so the zero value has to fail loudly.
- `CliAccessPassType` moves out of `accesspass/set.rs` into `accesspass/types.rs` so `user delete` can share it. No second enum was introduced.

## Declared kind, or derived

The program's refusal only bites when the kind carries the caller's intent. A caller that reads the pass and picks the instruction from what it read has chosen the instruction from the very byte the program then checks, so the two always agree. That is unavoidable on the paths with no operator, and `Option` says which is which instead of leaving it to comments:

- `Some(kind)` — an operator declared it. Both CLI verbs. `DeleteUserCommand` also refuses locally if the declared kind disagrees with the pass it resolved, naming both types, rather than letting it arrive onchain as a bare custom error code after the caller has paid.
- `None` — nobody was there to declare one, so the kind comes from the pass the command itself resolved. `doublezero disconnect` (a self delete; the handler already checks the owner and the client IP) and both tenant cascades (they sweep users across mixed kinds, and forcing one declared kind would turn "delete every user under this tenant" into "delete only users of one kind", stranding the tenant record while it waits for `reference_count` to reach 0).

An earlier version of this PR had those three callers read the kind back themselves, through `get_accesspass(client_ip, owner)`, while `DeleteUserCommand` resolved the pass through `resolve_user_accesspass`. For a user recording an access pass those name different accounts, so a legitimate `disconnect` could be refused for a mismatch that existed only between the two lookups. `None` removes the second lookup entirely, which also drops one account fetch per user in both cascades.

## Breaking

**CLI only.** Old binaries do not accept `--accesspass-type`, so `user delete` and `access-pass close` fail at argument parsing for them. `e2e/compatibility_test.go` records this as four `knownIncompatibilities` entries with `before("0.42.0")` — **if the release lands as something other than 0.42.0 those entries silently stop matching and need one edit.** main is at 0.41.0 as of this writing.

No program behaviour changes, so no deployed caller breaks. The oracle keeps working on variant 42.

## Known coverage loss in the compatibility matrix

`cascadeKnownFail` on `user_delete` sets `phaseFailed`, which skips every remaining phase and not only the ones that need the user gone. For every CLI below the gate that permanently stops running `user_delete_2`, `accesspass_close`, `accesspass_close_2`, `device_interface_delete_1` to `_4`, `iface_wait_removed_1` and `_2`, `exchange_clear_devices`, `device_drain_1` and `_2`, `device_delete_1` and `_2`, `device_wait_removed_1` and `_2`, `exchange_delete`, `contributor_delete` and `location_delete`. Those passed from 0.19.0 up and an old binary can never gain the flag, so they will not pass again. Keeping the teardown alive would mean running that one delete with the current CLI, which stops the phase testing the old binary at all. Accepted rather than fixed, and recorded in the comment at the call site.

## Diff Breakdown

| Category    | Files | Lines (+/-) |
|-------------|-------|-------------|
| Code        |    21 | +961 / -80  |
| Tests       |     6 | +90 / -22   |
| Docs        |     2 | +11 / -4    |
| **Total**   |    29 | +1062 / -106 |

The Code row overstates the production change: most of the growth in the Rust files is inline `#[cfg(test)]` modules, which the categoriser cannot split out.

<details>

- `smartcontract/sdk/rs/src/commands/user/delete.rs` — the `Option<AccessPassKind>`, the local mismatch refusal, and tests for all three outcomes.
- `crates/doublezero-daemon-cli/src/disconnect.rs` — declares no kind, and a test with no `get_accesspass` expectation so the mock fails if it reads the pass to pick one.
- `crates/doublezero-serviceability-instruction/src/{accesspass,user}.rs` — variant selection per kind, pinned by byte.
- `smartcontract/sdk/go/serviceability/executor.go` — the Go `AccessPassKind`, the selector, and the rejected zero value.
- `smartcontract/cli/src/accesspass/close.rs` and `src/user/delete.rs` — the two required flags.
- `smartcontract/cli/src/accesspass/types.rs` — `CliAccessPassType` and its conversion, moved here so both verbs share one enum.
- `e2e/compatibility_test.go` — the four `knownIncompatibilities` entries, the flag on every invocation, and the cascade note above.

</details>

## Testing Verification

- Both new CLI flags are proved required by a clap `try_parse_from` test asserting `MissingRequiredArgument`, not merely that parsing failed.
- `DeleteUserCommand` has a test per outcome: a declared kind reaches the builder, a mismatched one is refused with no `expect_send_transaction` registered so the mock panics if it tries to send, and `None` derives the variant from the stored pass. The derivation test seeds a `SolanaValidator` pass and pins the instruction bytes, so a derivation that always returned `Prepaid` fails it.
- Every command test uses two distinct kinds. With a single kind per file, an implementation that ignored its `kind` field and hardcoded that literal would have passed.
- The builders' byte selection is pinned per kind in two independent places, the Rust instruction crate and the Go SDK, so a transposed pair fails rather than silently routing a removal to the wrong instruction.
- All eleven e2e invocations were traced to the `access-pass set --accesspass-type` call in their own test before choosing a value.
- `cargo test` across the serviceability program, the instruction crate, the Rust SDK, both CLIs and the daemon CLI: zero failures. `cargo clippy --tests -- -Dclippy::all -Dwarnings` clean, nightly rustfmt clean. Go serviceability SDK and `tools/stress` green. `go build -tags e2e` and `go vet -tags e2e` clean.
- **The e2e suite was not run locally** — no Docker daemon was available on the machine this was built on. The changed e2e lines are string edits to CLI invocations and every behavioural claim above is covered by tests that did run, but the four `knownIncompatibilities` entries can only be exercised against real older CLI images in CI.


bgm-malbeclabs:
Second pass over this branch. Four things, two of them worth acting on before merge.

**1. `cascadeKnownFail` on `user_delete` skips the whole teardown half of the matrix.** `e2e/compatibility_test.go:1271`. The phase runner is all or nothing: "if any non-noCascade step fails in a phase, all remaining phases are skipped" (`:1080`). `write/user_delete` is known-incompatible `before("0.39.0")` and 0.39.0 does not exist yet, so this fires for every version in the matrix and skips the 9 phases after `delete_start`: `delete_continue`, `delete_links`, `delete_users_done`, `delete_interfaces`, `wait_interfaces_removed`, `drain_devices`, `delete_devices`, `wait_devices_removed`, `delete_infrastructure`. That takes `link_delete`, `device_delete`, `exchange_delete`, `contributor_delete` and `location_delete` out of backward compatibility coverage, none of which has anything to do with access passes.

**2. Three of the four new `knownIncompatibilities` entries are unreachable.** `:169`. `write/user_delete_2`, `write/accesspass_close` and `write/accesspass_close_2` all live in phases the cascade skips, so they record `SKIP` and can never record `KNOWN_FAIL`. The two mechanisms disagree. Either the cascade goes and the real dependents (`user_wait_removed`, `user_wait_removed_2`) get listed, or these three entries go.

**3. The `--type` values are not the names anything prints.** `smartcontract/cli/src/accesspass/types.rs:7`. clap derives kebab-case, so the flag takes `solana-rpc`. `access-pass list` prints `solana_rpc` (`list.rs:171`) and `solana_validator` (`:166`). An operator reads the name off `list` and gets a parse error for a value nothing showed them. One `#[value(alias = "...")]` per variant closes it.

**4. Three spellings of one flag in one command family.** `access-pass set --accesspass-type` (`set.rs:22`), `access-pass close --type` (`close.rs:19`), `user delete --access-pass-type` (`user/delete.rs:20`). Learning it on one verb earns a parse error on the next.

Two more I have not verified, so treat them as pointers rather than findings: `tools/stress/device-orchestrator/pkg/exec/exec.go:118` hardcodes `Prepaid` behind an unchanged two argument `DeleteUser`, which works only because both stress scripts pass `--accesspass-type prepaid`; and the `get_accesspass` read added at `crates/doublezero-daemon-cli/src/disconnect.rs:177` looks redundant with the one `DeleteUserCommand::execute` already does, where a read error collapses to `Ok(None)` and would skip the delete before polling for 127 seconds.

bgm-malbeclabs:
Suggested fix for finding 1, and a correction to my own framing of it.

**Dropping the cascade alone would turn the matrix red, because the teardown really is blocked past the users.** `user create` increments `device.reference_count` (`processors/user/create_core.rs:394`) and only `user delete` decrements it (`user/delete.rs:244`), so `device delete` refuses with `ReferenceCountNotZero` (`device/delete.rs:132`) while user1 and user2 are still there, and `exchange`, `contributor` and `location` keep the device's reference in turn. `multicast group delete` refuses the same way while user1 subscribes (`multicastgroup/delete.rs:103`). A `noCascade` step that fails still calls `assert.NoError`, so those would be real failures rather than skips.

So the cascade is not wrong about the dependency. It is wrong about the extent: it skips the link and interface teardown too, and neither of those references a user.

**The small fix is to name the blocked steps instead of skipping everything after them.** Drop `cascadeKnownFail: true` from `user_delete` (`:1271`) and add the steps the lingering user actually blocks:

```go
"write/user_wait_removed":      {ranges: before("0.39.0")},
"write/user_wait_removed_2":    {ranges: before("0.39.0")},
"write/multicast_group_delete": {ranges: before("0.39.0")},
"write/device_delete":          {ranges: before("0.39.0")},
"write/device_delete_2":        {ranges: before("0.39.0")},
"write/device_wait_removed":    {ranges: before("0.39.0")},
"write/device_wait_removed_2":  {ranges: before("0.39.0")},
"write/exchange_delete":        {ranges: before("0.39.0")},
"write/contributor_delete":     {ranges: before("0.39.0")},
"write/location_delete":        {ranges: before("0.39.0")},
```

A `KNOWN_FAIL` without `cascadeKnownFail` records and moves on, so every version keeps coverage of `link_drain`, `link_drain_dzx`, `link_delete`, `link_delete_dzx`, the four `iface_wait_unlinked`, the four `device_interface_delete`, both `iface_wait_removed`, `exchange_clear_devices` and both `device_drain`. That is 17 steps back, and the results table says `KNOWN_FAIL` where the reason is known rather than `SKIP` where no reason is recorded. It also makes finding 2 go away: `user_delete_2`, `accesspass_close` and `accesspass_close_2` become reachable and record what they are listed for.

**What it costs.** Four polling steps now burn their full timeout instead of being skipped: `user_wait_removed`, `user_wait_removed_2`, `device_wait_removed`, `device_wait_removed_2`, at 30 seconds each, so about two minutes per old version in the matrix.

**If those minutes matter**, the alternative is to give the cascade a scope rather than a switch: `cascadeKnownFailTo []string` naming the steps to skip, checked in `execStep` instead of the phase loop. That keeps the fast skip and stops it spreading, at the price of touching the mechanism #2870 added.

martinsander00 (approved):

PR #4270, opened 2026-09-03
contributor-rewards: stop the keypair tests racing on a shared env var

bgm-malbeclabs:
`test_keypair_env_fallback` in `contributor-rewards` fails intermittently in CI. It failed on two unrelated pull requests today, and a re-run of the identical commit passed both times.

It is a race between tests, not a fault in the code under test.

## What happens

Three tests in `offchain/crates/contributor-rewards/tests/test_sanity.rs` set or clear the same process-wide variable:

- `test_keypair_cli_takes_precedence` sets `REWARDER_KEYPAIR_PATH`, then clears it
- `test_keypair_env_fallback` sets it, calls `load_keypair(&None)`, then clears it
- `test_keypair_not_provided_error` clears it, then expects `load_keypair(&None)` to fail

Cargo runs tests in parallel threads of one process, and `std::env` is per process, not per thread. When one of the other two clears the variable between `test_keypair_env_fallback`'s `set_var` and its `load_keypair(&None)`, the fallback finds nothing and `assert!(result.is_ok())` fails. The window is a few microseconds wide, which is why this shows up now and then rather than every run.

The other two keypair tests pass an explicit path, so the CLI argument wins and the variable is irrelevant to them. They need no change.

## Summary of Changes

- The three tests that touch `REWARDER_KEYPAIR_PATH` now carry `#[serial]`, so they run one at a time while the rest of the suite still runs in parallel.
- `serial_test` was already a workspace dependency, used by `doublezero-cli-core`, `config`, `smartcontract/cli` and `smartcontract/sdk/rs`. It is added to this crate's dev-dependencies only, so nothing new enters the build.

The production `load_keypair` is untouched. Reading the variable inside it is what makes the tests need real environment state, but the tests are verifying exactly that fallback, so serialising them is the fix rather than changing the signature its callers use.

## Diff Breakdown

| Category    | Files | Lines (+/-) | Net |
|-------------|-------|-------------|-----|
| Tests       |     1 | +13 / -0    | +13 |
| Config      |     2 | +4 / -0     |  +4 |
| Docs        |     2 | +3 / -0     |  +3 |
| **Total**   |     5 | +20 / -0    | +20 |

## Testing Verification

- Reproduced the failure before fixing it: 60 runs of `cargo test -p doublezero-contributor-rewards --test test_sanity` on the unchanged code produced 1 failure, on run 26, panicking at `test_sanity.rs:63` exactly as CI did.
- The same 60 runs after the change: 0 failures.
- Both loops ran the whole file, not the three tests alone, so the parallel scheduling that causes the race was in play.


bgm-malbeclabs:
Superseded by #4265, which landed on main as 6994a4f40 while this was in flight. Same fix: `use serial_test::serial` plus `#[serial]` on the same three tests — `test_keypair_cli_takes_precedence`, `test_keypair_env_fallback` and `test_keypair_not_provided_error` — with `serial_test` added as a dev-dependency and both changelogs updated. Closing this rather than resolving the conflicts, since there is nothing here the merged fix does not already do.

One thing worth recording, from before I saw #4265. I reproduced the race to confirm the diagnosis: 60 runs of `cargo test -p doublezero-contributor-rewards --test test_sanity` on the unfixed code failed once, on run 26, panicking at `test_sanity.rs:63` exactly as CI did. The same 60 runs with `#[serial]` applied failed none. So roughly a 1-in-60 flake rate, and the merged fix removes it.

PR #4271, opened 2026-09-03
geoprobe: reject forged, replayed and spoofed geolocation input

nikw9944:
Five defects in the RFC-16 measurement chain let unauthenticated input through: forged offsets were cached and published, `MeasurementSlot` was never compared despite being the RFC's named replay mitigation, unverified TWAMP probes mutated per-sender pair state, a completed-but-empty target scan was mistaken for a skipped one, and ICMP replies were matched without their source address.

Decisions a reviewer would otherwise have to reverse-engineer:

- **Invalid offsets are dropped, not recorded.** The public lake explorer aggregates `location_offsets` without filtering on `signature_valid`, and routing forgeries to a separate path would need a schema change. Abuse visibility is a `Warn` log.
- **Verification rejects a zero signature scalar rather than a list of bad keys.** `ed25519.Verify` does not screen small-order pubkeys: measured here, an unsigned datagram verifies on ~24% of messages with the all-zero key and on *every* message with the identity encoding. A zero S is what every forgery of that shape needs and no real signature has, so one check covers the class at all three sites where a verification key comes off the wire.
- **The slot window is asymmetric and its lead is derived, not chosen.** The 15 min of lag covers a DZD stamping from its own 5 min slot cache; the 10 min lead follows from how stale our own reference may be, since anything smaller rejects genuinely fresh offsets as replays.
- **`GetGeolocationUsers` errors when accounts came back but none decoded.** Callers act on emptiness — a geoprobe stops probing and empties its inbound allowlist — so a layout change or a lagging replica must not look like "no users". This is the one change here another component could notice.
- **Unverified probes still get a reply**, which RFC-16 allows, but off throwaway state and rate-limited with a floor: the reply is an order of magnitude larger than the probe, so that path is otherwise a reflection amplifier.

Out of scope: a probe carries no freshness field, so verifying it stops forged probes but not replayed ones, and closing that is a wire-format change. No wire format or Borsh layout is touched here. RFC: [rfc16-geolocation-verification.md](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc16-geolocation-verification.md) — step 6 of the inbound flow is amended.

Two unrelated commits ride along to get CI green, neither touching production code: a wheresitup ticker test deflake, and a twamp fuzz budget switched from wall time to execution count to avoid golang/go#75804.

## Testing Verification

- Each fix was reverted individually with its test in place, confirmed failing, then restored and confirmed passing.


tools/twamp/pkg/signed/reflector_linux.go
```diff
@@ -224,6 +228,29 @@ func (r *LinuxReflector) Run(ctx context.Context) error {
...
+			if !probe.Verify() {
+				if r.verifyInterval > 0 && !state.lastUnverifiedRx.IsZero() &&
+					now.Sub(state.lastUnverifiedRx) < r.verifyInterval {
+					continue
+				}
+				// Cap unverified replies at one per window per pubkey: the reply
+				// is an order of magnitude larger than the probe, so an unlimited
+				// reply path is a reflection amplifier.
+				state.lastUnverifiedRx = now
+				state = &senderState{}
```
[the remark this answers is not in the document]
nikw9944: Fixed in 42065853. The cap now uses `max(verifyInterval, minUnverifiedReplyInterval)` with a 1s floor, so it holds when `verifyInterval` is 0. Test `unverified replies stay capped with rate limiting disabled` builds a reflector with `verifyInterval: 0`, confirms the first spoofed probe is answered and the next four inside the window are not; it fails against the previous code.

controlplane/telemetry/internal/geoprobe/target_discovery.go
```diff
@@ -145,14 +148,14 @@ func (d *TargetDiscovery) discover(ctx context.Context) ([]ProbeAddress, []Probe
...
+			return false, nil, nil, nil, nil, nil, nil
 		}
 		d.lastSeenTargetUpdateCount = current
 	}
 
 	users, err := d.client.GetGeolocationUsers(ctx)
 	if err != nil {
-		return nil, nil, nil, nil, nil, fmt.Errorf("failed to fetch GeolocationUser accounts: %w", err)
+		return false, nil, nil, nil, nil, nil, fmt.Errorf("failed to fetch GeolocationUser accounts: %w", err)
 	}
```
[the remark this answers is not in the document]
nikw9944: Fixed the doc comment in 42065853 rather than the return value. Returning `scanned: true` from a scan that failed would claim the scan ran when it did not, and `discoverAndSend` checks `err` first and never reads `scanned` on that path. The comment now states that `scanned` is meaningful only when `err` is nil and that the other values are unset on error, which is the usual Go contract.

controlplane/telemetry/cmd/geoprobe-agent/main.go
```diff
@@ -718,6 +743,30 @@ func runOffsetListener(
...
+			log.Warn("Rejecting offset with measurement slot outside acceptance window",
+				"sender_pubkey", senderPK,
+				"addr", addr,
+				"measurement_slot", offset.MeasurementSlot,
+				"current_slot", currentSlot,
+				"max_lag_slots", maxOffsetSlotLag,
+				"max_lead_slots", maxOffsetSlotLead)
+			m.OffsetsRejected.WithLabelValues(geoprobe.RejectSlotOutOfWindow).Inc()
+			continue
+		}
```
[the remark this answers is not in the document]
nikw9944: Good catch — fixed in 42065853. This was a real hole in the fix: `getCurrentSlot` serves its cache indefinitely on RPC failure, so the window froze around the last fetched slot and fresh offsets would eventually be rejected as too far ahead.

Ingestion now goes through `getSlotReference`, which rejects a reference older than `maxSlotReferenceAge` (2 × `SlotCacheTTL` = 10m) rather than your proposed 5m: the cache refreshes lazily at 5m, so a 5m bound would reject on a single transient error at the refresh boundary. Two periods tolerates one missed refresh and still bounds the worst-case replay window at 10m + the 15m lag tolerance instead of unbounded. Composite offsets deliberately keep using `getCurrentSlot` — stamping a slightly stale slot beats the probe emitting nothing during a blip. Covered by `TestSlotReference`.

controlplane/telemetry/internal/geoprobe/signer_test.go
```diff
@@ -438,3 +438,22 @@ func TestNewOffsetSigner_ZeroSenderPubkey(t *testing.T) {
...
+		Version:         LocationOffsetVersion,
+		MeasurementSlot: 1,
+		Lat:             1.0,
+		Lng:             2.0,
+		MeasuredRttNs:   1000,
+		RttNs:           1000,
+	}
+
+	require.Error(t, VerifyOffset(offset))
+	require.Error(t, VerifyOffsetChain(offset))
```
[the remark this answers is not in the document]
nikw9944: Fixed in 42065853 — both calls now assert `authority pubkey is zero` via `require.ErrorContains`.

nikw9944:
Addressed all four review points in 42065853, and rebased onto main (v0.39.0) to replace the Update Branch merge commit with linear history — the rebased tree is byte-identical to that merge commit, and the CHANGELOG entry moved into the new `Unreleased` section since the release cut turned the old one into v0.39.0.

The substantive one was Bugbot's stale-slot finding: `getCurrentSlot` falls back to its cache indefinitely on RPC failure, so the replay window I added froze around the last fetched slot — replays near it stayed acceptable for the length of an outage, and fresh offsets would eventually be rejected as too far ahead. Ingestion now rejects a slot reference older than two refresh periods (10m); composite offsets keep the old behavior so the probe still emits during a blip.

Copilot's amplifier point was also correct: the unverified-reply cap keyed off `verifyInterval`, which is 0 when pair rate limiting is disabled, leaving the path it was meant to protect uncapped. It now has a 1s floor. Both fixes have tests that fail against the previous code.

Chose differently on one: `discover` still returns `scanned: false` on an RPC error. Returning `true` would claim a scan ran when it did not, and `discoverAndSend` checks `err` first, so the comment now documents that `scanned` is meaningful only when `err` is nil.

`go test ./controlplane/telemetry/... ./tools/twamp/...` passes except `tools/twamp/pkg/light`, which fails in my sandbox for lack of raw-socket permissions and is untouched by this branch (`git diff origin/main -- tools/twamp/pkg/light` is empty).

nikw9944:
Branch updated for main again (`c5815ae5`) — main had moved 8 commits and CHANGELOG.md was the only conflict, as before. Main added a CLI entry to the same `Unreleased` → `Changes` section; both blocks stay there, since this PR is unmerged and its entry does not belong in a released version. Verified mechanically: my Geolocation block is byte-identical to the previous head, the merged section is exactly that block plus main's verbatim, no entry from either side was dropped, and every released section matches main.

No code changed. For all 19 non-CHANGELOG files this PR touches, `git diff 2d863301 HEAD -- <file>` is empty.

No new review feedback since the last update — the four inline threads already carry replies, and nothing needing action arrived after them.

Local checks on the merge result: `go test -race ./tools/twamp/pkg/signed/` and the four geolocation packages pass; `golangci-lint run -c ./.golangci.yaml ./tools/twamp/pkg/signed/... ./controlplane/telemetry/...` reports 0 issues. e2e still needs a Docker host and has not been run here.

controlplane/telemetry/internal/geoprobe/target_discovery.go
```diff
@@ -91,14 +91,15 @@ func (d *TargetDiscovery) Tick(ctx context.Context, targetCh chan<- TargetUpdate
...
 	if err != nil {
 		d.log.Warn("Target discovery tick failed", "error", err)
 		return
 	}
 
-	// nil targets means the scan was skipped (target_update_count unchanged).
-	if targets == nil && inboundKeys == nil && icmpTargets == nil {
+	// A completed scan that matched nothing must still propagate: it is how a
+	// deregistered or delinquent user stops being probed.
+	if !scanned {
```
ben-dz, line 102: **Blocking** — A successful-but-empty user fetch now clears every target and the whole inbound allowlist, where the old all-nil guard kept the last good snapshot. `GetGeolocationUsers` returns `(empty, nil)` both for an empty `getProgramAccounts` reply and when every account fails to deserialize (`sdk/geolocation/go/client.go:203` skips and warns), so a lagging RPC replica or a `GeolocationUser` layout change silently takes the probe dark — `SetAuthorizedKeys(nil)` drops every inbound probe as unauthorized. The layout-skew case does not self-heal. Fix: treat "accounts returned but none parsed" as an error, not an empty scan.

controlplane/telemetry/cmd/geoprobe-agent/main.go
```diff
@@ -43,6 +43,28 @@ const (
...
+
+	// Acceptance window for an inbound offset's MeasurementSlot, RFC-16's replay
+	// mitigation. A DZD stamps offsets with a slot it caches for
+	// geoprobe.SlotCacheTTL (5m) and this agent compares against its own slot
+	// cache with the same TTL, so a legitimate offset can sit ~5m either side of
+	// our view before any RPC or finalization lag. maxOffsetSlotLag allows 15m of
+	// that skew in the past; maxOffsetSlotLead allows 5m in the future, for when
+	// our own cached slot is the stale one.
+	maxOffsetSlotLag  = uint64(15 * time.Minute / dzSlotDuration)
+	maxOffsetSlotLead = uint64(5 * time.Minute / dzSlotDuration)
```
ben-dz: **Blocking** — `maxSlotReferenceAge` lets the reference slot be up to 10 min old so one missed refresh does not stop ingestion, but `maxOffsetSlotLead` tolerates only 750 slots (5 min) of lead. Through most of that grace band a DZD stamping a current slot lands 750-1500 slots ahead and is rejected as a replay — the band the bound exists to preserve is a band where ingestion is dead, and it fails looking like an attack in the logs. Fix: derive the lead from `maxSlotReferenceAge` plus the DZD's own cache TTL, or shrink `maxSlotReferenceAge` to match the lead.

controlplane/telemetry/internal/geoprobe/signer.go
```diff
@@ -50,6 +50,16 @@ func (s *OffsetSigner) SignOffset(offset *LocationOffset) error {
 func VerifyOffset(offset *LocationOffset) error {
 	pubkey := solana.PublicKeyFromBytes(offset.AuthorityPubkey[:])
 
+	// ed25519.Verify accepts the all-zero (pubkey, signature) pair for a
+	// fraction of messages: the zero key decodes to a valid point of order 4
+	// rather than being rejected, and with S and R also zero the verification
+	// equation holds whenever the message hash lands on the right residue —
+	// about one message in four, which an attacker reaches by varying any field.
+	// No signer has the zero pubkey, so reject it before verifying.
+	if pubkey.IsZero() {
```
ben-dz: **Minor** — The guard rejects only the all-zero pubkey, one of eight small-order Ed25519 points. Measured on go1.25.6: identity encoding `0x01||00*31` with `R=0x01||00*31, S=0` verifies 100% of messages versus 24.7% for the case fixed here, and `IsZero()` rejects neither. The comment and CHANGELOG say the guard stops an unsigned datagram, which the next caller believes. Same guard at `tools/twamp/pkg/signed/packet.go:187` and `:346`. Fix: reject all small-order encodings.

tools/twamp/pkg/signed/reflector_linux.go
```diff
@@ -245,8 +278,8 @@ func (r *LinuxReflector) Run(ctx context.Context) error {
 			}
 
 			// Pair integrity: both probes must come from the same source IP.
-			// The pubkey allowlist (checked above) provides authentication;
-			// per-probe signature verification is left to the target.
+			// Only verified probes reach this state, so a spoofed source cannot
```
ben-dz: **Minor** — The comment and the RFC edit claim a spoofed source cannot claim or repoint the pair, but `ProbePacket.Verify` covers only `Seq`/`Sec`/`Frac`/`SenderPubkey` with no freshness check, so a captured probe replayed from another address verifies, takes the pair-0 slot and sets `pairSourceIP`; the real sender's probe is then dropped for source mismatch. It stops forged probes, not replayed ones.

controlplane/telemetry/cmd/geoprobe-agent/main.go
```diff
@@ -718,6 +776,30 @@ func runOffsetListener(
 
 		log.Debug("signature verification successful", "authority_pubkey", authorityPK)
 
+		// RFC-16 replay mitigation. A signature stays valid forever, and the
+		// only other freshness bound is receipt wall-clock, which a replay
+		// refreshes by definition — so MeasurementSlot has to be checked against
+		// the current ledger slot or a captured offset can be replayed
+		// indefinitely to pin this probe's attested reference point.
+		currentSlot, err := getCurrentSlot(ctx)
```
ben-dz: **Minor** — `getCurrentSlot` issues a blocking `GetSlot` with the shared retry defaults (10s x 4 attempts + backoff) from inside `runOffsetListener`'s receive loop, the only goroutine draining the socket. A slow or failing RPC stalls reads and overflows the socket buffer, and on failure `slotCachedAt` is not updated, so every packet re-runs the full retry. Fix: refresh from a background ticker.

ben-dz (changes requested):
An empty-but-successful user fetch now wipes all targets plus the inbound allowlist, and the new slot window rejects fresh offsets across the 5-10m band `maxSlotReferenceAge` exists to preserve. Plus a few smaller notes.

- `controlplane/telemetry/internal/geoprobe/target_discovery.go:107` — **Minor** — `d.cachedTargets` is assigned before the non-blocking send, so when the cap-1 channel is full the update hits `default`, is logged as skipped, and is never retried: the cache already compares equal, including on the forced full refresh. The empty update this PR adds is a one-shot, so losing it leaves a delinquent user probed until restart. Fix: assign inside the successful `case` arm.

controlplane/telemetry/internal/geoprobe/target_discovery.go
nikw9944: Fixed in dc7e0591. `GetGeolocationUsers` now returns an error when accounts came back but none decoded, so only a genuinely empty program produces an empty result. A single bad account among good ones is still skipped and warned, as before.

Two tests: `AllAccountsUndecodable` covers both causes (wrong owner, undecodable data) and `SkipsWrongOwner` now asserts the partial case still yields the good user. The old version of that test asserted `(empty, nil)` for a lone wrong-owner account, which is exactly the behavior you flagged, so it changed with the contract.

controlplane/telemetry/cmd/geoprobe-agent/main.go
nikw9944: Fixed in dc7e0591 by deriving the lead rather than shrinking the reference bound: `maxOffsetSlotLead = maxSlotReferenceAge / dzSlotDuration` (1500 slots). Shrinking `maxSlotReferenceAge` to 5m would restore the brittleness it was added for — a single transient error at the lazy 5m refresh boundary would stop ingestion.

Widening the lead costs nothing against replays: a captured offset carries an old slot, and a future one would have to be signed by the DZD itself. The DZD's own cache TTL pushes its slot *older*, so it belongs on the lag side, which already has 15m.

`TestOffsetSlotFresh` gains a "fresh offset against a maximally stale reference" case; it fails against the 5m literal.

controlplane/telemetry/internal/geoprobe/signer.go
nikw9944: Fixed in dc7e0591, and thank you — I reproduced your measurement before changing anything: on go1.25.6 the all-zero key verifies 24.0% of 400 messages and the identity encoding with `R=identity, S=0` verifies 100%. My guard missed the strictly more dangerous case.

Rather than blacklist encodings I reject a zero `S` scalar, which covers the whole class: `[S]B` lies in the prime-order subgroup, a small-order `R + [k]A` lies in the torsion subgroup, and those meet only at the identity, so any forgery of this shape needs `S ≡ 0 (mod L)` — and `Verify` already requires `S` canonical, making it exactly zero. A nonzero `S` is as hard to produce as an ordinary forgery. No table to get wrong and no new dependency.

Applied at all three sites (`signer.go`, `packet.go` probe and reply). Tests cover both forgeries at each site plus a real signature still verifying; the identity cases fail against the old `IsZero()` guard. The comments and the CHANGELOG now state what the guard actually covers.

tools/twamp/pkg/signed/reflector_linux.go
nikw9944: You are right, and both claims are corrected in dc7e0591 — the comment and the RFC step now say verification stops forged probes, not replayed ones, and name the missing freshness field as the reason.

I did not add freshness to the probe: `ProbePacket` has no spare field, so it would be a wire-format change, which is out of scope for this PR. Worth tracking separately if the replayed-pair denial matters on its own.

controlplane/telemetry/cmd/geoprobe-agent/main.go
nikw9944: Fixed in dc7e0591. A `slotRefresher` goroutine ticks at half the cache TTL and owns every `GetSlot` call; ingestion now reads the cached value and its age under the lock and never issues RPC, so a slow or failing replica cannot stall the socket drain or re-run the retry per packet. It composes with the age bound: when the refresher stops succeeding, ingestion fails closed rather than trusting a frozen slot. Composite offsets still call `getCurrentSlot` directly, off the receive path.

nikw9944:
All six findings addressed in dc7e0591, plus a merge of main (v0.40.0 and v0.41.0 were cut since the last update, so this branch's CHANGELOG entry moved down but stays under `Unreleased`). Individual replies are on each thread; the two blocking ones:

**Empty-vs-undecodable fetch.** `GetGeolocationUsers` now errors when accounts came back but none decoded, so only a genuinely empty program clears targets and the inbound allowlist. A lagging replica or a layout change no longer takes a probe dark.

**Slot window dead band.** `maxOffsetSlotLead` is derived from `maxSlotReferenceAge` rather than being a second literal, which removes the 5–10m band where a DZD stamping a current slot was rejected as a replay. I chose deriving the lead over shrinking the reference bound, since 5m would make a single transient RPC error at the refresh boundary stop ingestion — reasoning in the thread.

Also: the review-body finding about `d.cachedTargets` being assigned before the non-blocking send is fixed (caches update inside the successful `case` arm, so a dropped update is retried); the slot read moved off the socket-receive path onto a background refresher; and the reflector comment plus the RFC step no longer claim verification stops replayed probes.

The small-order finding deserves a callout: I reproduced the measurement first, and the identity encoding verifies **100%** of messages where the all-zero key verifies ~24%, so the guard I shipped missed the worse case. It now rejects a zero `S` scalar, which covers the whole class without a blacklist — argument in the thread. Each of the six fixes has a test that fails against the previous code, confirmed by reverting each one individually.

Not done: adding a freshness field to `ProbePacket` to close the replayed-pair denial, since that is a wire-format change and out of scope here.

Local verification: `go test` across `controlplane/telemetry/...`, `tools/twamp/...` and `sdk/geolocation/...` passes, `-race` clean on the two packages with concurrency, and `golangci-lint -c ./.golangci.yaml` reports 0 issues. `tools/twamp/pkg/light` fails in my sandbox for lack of raw-socket permission and is untouched by this branch. e2e needs a Docker host and has not been run here.

controlplane/telemetry/cmd/geoprobe-agent/main.go
```diff
@@ -43,6 +43,29 @@ const (
 	discoveryInterval                 = 60 * time.Second
 	defaultDeliveryDNSRefreshInterval = 5 * time.Minute
 	defaultDeliveryDNSTTL             = defaultDeliveryDNSRefreshInterval * 5 / 2
+
+	dzSlotDuration = 400 * time.Millisecond
```
ben-dz, line 47: **Optional** — `dzSlotDuration` is 400ms, but `internal/telemetry/epoch.go:18` documents the DZ ledger at ~367ms and uses 340ms as its safe fallback. The lead is now derived from `maxSlotReferenceAge` in slots, so at the real rate 1500 slots covers ~9.2m of the 10m the reference may be stale, leaving a narrow band where a fresh offset is still rejected. Same 9% optimism in the CHANGELOG figures.

ben-dz (approved):
All five posted findings verified fixed in dc7e0591/b84eff8c. The red `go-test` is a timing flake in `internet-latency-collector/wheresitup`, untouched here and passing locally.

controlplane/telemetry/cmd/geoprobe-agent/main.go
nikw9944: Fixed in 6f7508c6. `dzSlotDuration` is now 340ms — the same Solana-target-less-15% figure `internal/telemetry` projects epochs with — so at the real ~367ms rate the 10 minute lead covers ~10.8 min instead of ~9.2, and the 15 minute lag ~16.2 min. Under-estimating the slot duration is the safe direction here: the stated wall-time windows (and the CHANGELOG figures) become a floor, where over-estimating shrank them and rejected genuinely fresh offsets. Both bounds derive from the constant, so the boundary tests moved with it.

nikw9944:
Rebased onto main (9bc34caa) and addressed the last open thread.

- Slot window sizing (6f7508c6): `dzSlotDuration` 400ms → 340ms, matching the conservative figure `internal/telemetry` already uses. At the ledger's real ~367ms the 10 minute lead now covers ~10.8 min rather than ~9.2, so the narrow band where a fresh offset was rejected as a replay is gone and the CHANGELOG figures are a floor.
- The CHANGELOG conflict from the rebase was resolved by keeping both sides: the Geolocation block moved into main's current Unreleased section, nothing dropped.

Nothing else changed. Verified locally with `go test -race` over `controlplane/telemetry/{cmd/geoprobe-agent,cmd/geoprobe-target,internal/geoprobe}`, `tools/twamp/pkg/signed`, `sdk/geolocation/go` and `controlplane/internet-latency-collector/...`.

Still out of scope, as noted in the PR body: a signed TWAMP probe carries no freshness field, so verifying it stops forged probes but not replayed ones. Closing that needs a wire-format change and belongs in its own PR.

Issue #4101, opened 2026-07-28
sdk/rs: retry classification misses JSON-RPC response errors (Rust parity with #4100)


nikw9944:
## Problem

#4100 fixed retry classification for Go, but #4098's core finding — Go and Rust disagree on what's retryable — is only half resolved. The Rust SDK behind the `doublezero` CLI retries transport-kind errors only (`ClientErrorKind::Io` / `Reqwest` / `Middleware` in `is_retryable_rpc_error`, `smartcontract/sdk/rs/src/client.rs`) and never retries a decoded JSON-RPC error envelope.

That means the two shapes the 2026-07-28 incident and infra#2100 actually arrived in are still one-shot failures for every Rust CLI read:

- an HTTP status carried in the JSON-RPC error code (RPCPool's LB put `503` in the envelope — the #4098 shape)
- a transient message with no machine-readable code (`-32603 "Service unavailable, please try again later."` from Helius — the malbeclabs/infra#2100 shape)

Rust rode out 07-28 only because those 503s happened to surface at transport level (`Reqwest` errors). Had they arrived as decoded envelopes — as they did for Go — every CLI invocation would have failed on the first attempt.

## Concrete exposure

The QA payments testnet probe (malbeclabs/infra#2100) shells out to `doublezero access-pass get` twice per run for DZ-ledger reads. After the doublezero-shreds side adopts #4100, these CLI-mediated reads are the probe's last unretried transient-failure path. Any operator or automation using the CLI against a provider-fronted endpoint has the same exposure.

## Ask

Mirror #4100's classification in the Rust SDK's retry predicate:

- retry envelope codes that are really HTTP statuses (429/500/502/503/504) and the provider "busy / retry later" codes (`-32005`, `-32004`, `-32003`, `-32429`)
- retry transient message wording (`service unavailable`, `too many requests`, `bad gateway`, `gateway timeout`) when no recognizable code is set
- keep onchain/request-level rejections (`-32002` preflight failures, `-32602`, `-32601`) non-retryable
- keep `send_transaction` never retried on ambiguous failures (the same double-submit reasoning as #4100)

`crates/shred-oracle/src/retry.rs` already has a transient-error classifier and `backon` policies for the shreds repo's Rust side — worth aligning with (or sharing, per #4098's open design question about where common retry code should live) rather than growing a third opinion on retryability.

PR #4273, opened 2026-09-04
sdk/rs: retry JSON-RPC response errors that carry a transient status

ayushsingh82:
Follows #4100 (which fixed this for Go) into the Rust SDK behind the `doublezero` CLI. #4098's core finding — Go and Rust disagreed on what's retryable — was only half resolved: the Rust predicate `is_retryable_rpc_error` (`smartcontract/sdk/rs/src/client.rs`) matched only `ClientErrorKind::Io` / `Reqwest` / `Middleware` and never looked at a decoded JSON-RPC error envelope.

So the two shapes the 2026-07-28 incident and `malbeclabs/infra#2100` actually arrived in were still one-shot failures for every Rust CLI read:

- an HTTP status a provider load balancer put in the JSON-RPC `code` (RPCPool's `503` — the #4098 shape)
- transient wording with no machine-readable code (`-32603 "Service unavailable, please try again later."` from Helius — the infra#2100 shape)

Rust rode out 07-28 only because those particular 503s surfaced as `Reqwest` errors. Had they arrived as decoded envelopes, as they did for Go, every CLI invocation would have failed on the first attempt.

## What changed

`is_retryable_rpc_error` now also classifies `ClientErrorKind::RpcError`, mirroring the Go classifier in `tools/solana/pkg/jsonrpc`:

- retry an envelope `code` that is really an HTTP status (429/500/502/503/504)
- retry the "busy / retry later" node codes `-32005`, `-32004`, `-32429`
- retry transient message wording (`service unavailable`, `too many requests`, `bad gateway`, `gateway timeout`, `rate limited`) when no recognizable code is set
- keep onchain / request-level rejections non-retryable: `-32002` preflight, `-32602`, `-32601`, plus `-32003` (signature-verification failure) and `-32011` (no history) — deterministic, an identical retry gets the same answer

`-32003` is listed as retryable in the issue text, but #4100 deliberately excludes it as deterministic; I followed #4100.

No `send_transaction` path in this SDK is wrapped with this predicate — every retry site is a read (`get_account`, `get_program_accounts`, `get_signatures_for_address`, …) — so an accepted-but-unacknowledged send is never resent. The classifier keeps that property and its doc comment states it.

## Testing Verification

- `cargo test -p doublezero_sdk` — the new `client_tests` cases assert: an envelope `503`/`429` retries; `-32005`/`-32004`/`-32429` retry; a `-32603` whose message says "Service unavailable" retries while a `-32603` "Internal error" does not; `-32002`/`-32602`/`-32601`/`-32003`/`-32011` do not retry; transport (`Io`) still retries.


ayushsingh82:
@ben-dz @nikw9944 — addresses #4101, the Rust-SDK parity half that #4098 flagged as still open. Classification here mirrors your merged Go fix in #4100 (`tools/solana/pkg/jsonrpc`) — tagging you two since you authored/reviewed that one and filed #4101.

Issue #4162, opened 2026-08-07
build: publish plain semver tags for the root Go module


ben-dz:
## Problem

`github.com/malbeclabs/doublezero` has no usable Go module version. The newest unprefixed semver tag is **v0.8.0, from 2025-12-02**. Every release since is component-prefixed (`agent/v0.33.0`, `client/v0.33.0`, `controller/v0.33.0`, … cut 2026-07-31). Go's resolver ignores prefixed tags for the root module path — one only names a module version when a nested `go.mod` sits in that directory, and the only nested module in the repo is `controlplane/s3-uploader`.

So `go get github.com/malbeclabs/doublezero@v0.33.0` fails with unknown revision, and every Go consumer pins a pseudo-version instead. lake is on `v0.8.1-0.20260728210138-d904a1969083`, which decodes as "some untagged commit after v0.8.0."

Consequences: `go list -m -u` reports no available upgrade and dependabot can't propose one, and a bump PR has no tag range to cite — malbeclabs/lake#757 had to hand-enumerate its review surface package by package.

## Proposal

Push one unprefixed `vX.Y.Z` tag per release, at the same commit and through the same approved path as the nine component tags. The tag name is built in exactly one place, [`release.testnet.push.tags.yml:71`](https://github.com/malbeclabs/doublezero/blob/main/.github/workflows/release.testnet.push.tags.yml#L71), and both callers — the orchestrator's `push-tags` job and the manual dispatcher — funnel through it.

Add it as an explicit input (a `module_tag` boolean or a `prefix`), not by letting an empty `component` mean "bare tag." A typo must not be able to mint a root module version, because of the first guard below.

**Immutability.** Once any consumer resolves the tag, proxy.golang.org caches its tree hash permanently. Moving or deleting it afterwards doesn't fail cleanly — it gives every consumer `SECURITY ERROR: checksum mismatch` on their next build. The existing "tag already exists → error" check is right and must stay, but `skip_existing` currently downgrades a *mismatched* existing tag to a `::warning::` (lines 86-93), and the orchestrator passes that flag on re-runs. For the root tag it has to be a hard failure.

[...]

PR #4274, opened 2026-09-04
ci: publish a bare vX.Y.Z tag for the root Go module on testnet releases

ayushsingh82:
Addresses #4162, /cc @ben-dz per the issue.

`github.com/malbeclabs/doublezero` has had no usable Go module version since `v0.8.0` (2025-12-02). Every release since is component-prefixed (`agent/v0.33.0`, `client/v0.33.0`, …), and Go's resolver ignores prefixed tags for the root module path, so `go get github.com/malbeclabs/doublezero@vX.Y.Z` fails and every Go consumer pins a pseudo-version instead — `lake` is on `v0.8.1-0.20260728210138-d904a1969083`, and `lake#757` had to hand-enumerate its review surface package by package because there was no tag range to cite.

## What changed

`release.testnet.push.tags.yml` gets a new `module_tag` boolean input, mutually exclusive with `component`: when set, it pushes a bare `vX.Y.Z` tag instead of `component/vX.Y.Z`, through the same caller allowlist and approval gate as the nine component tags (nothing new can call this workflow without an approval story). The orchestrator's `push-tags` job gets a `push-module-tag` sibling — added to the three downstream `needs` lists so a module-tag failure fails the release rather than passing silently — and the manual dispatcher gets a matching `module-tag` job.

Two guards specific to the root tag, called out explicitly in the issue:

- **Immutability.** `skip_existing` no longer downgrades a *mismatched* existing tag to a warning. Once `proxy.golang.org` caches a tag's tree hash, moving it gives every consumer a checksum-mismatch error on their next build — so for the root tag a mismatch is now a hard failure, even on an orchestrator re-run. The existing "tag already exists, same commit → no-op" path is unchanged.
- **Ancestry.** The new tag's commit must descend from the previous bare tag (checkout now takes `fetch-depth: 0` so `merge-base --is-ancestor` has the history to check). A Go pseudo-version sorts by its base tag rather than its commit date, so cutting the tag off a non-descendant commit would let `go get -u` walk a consumer backwards onto a build missing commits its pseudo-version already includes.

## Testing Verification

Can't exercise the reusable workflow directly (needs Actions), so verified statically: both changed workflow files parse as valid YAML, and the inline shell block passes `bash -n`. I walked the new branches by hand against the existing test cases in the issue — module tag with no prior tag (skips ancestry check), module tag descending correctly, module tag off a non-ancestor commit (fails), mismatched existing module tag under `skip_existing: true` (now fails instead of warning), and the untouched component-tag path (`skip_existing` behavior unchanged).

### DIFF
--- a/.github/workflows/release.testnet.push.tags.components.yml
+++ b/.github/workflows/release.testnet.push.tags.components.yml
@@ -43,3 +43,16 @@ jobs:
     secrets:
       RELEASE_BOT_APP_ID: ${{ secrets.RELEASE_BOT_APP_ID }}
       RELEASE_BOT_PRIVATE_KEY: ${{ secrets.RELEASE_BOT_PRIVATE_KEY }}
+
+  module-tag:
+    name: "Tag root Go module with version ${{ github.event.inputs.version }}"
+    needs: approve
+    uses: ./.github/workflows/release.testnet.push.tags.yml
+    permissions:
+      contents: write
+    with:
+      version: ${{ github.event.inputs.version }}
+      module_tag: true
+    secrets:
+      RELEASE_BOT_APP_ID: ${{ secrets.RELEASE_BOT_APP_ID }}
+      RELEASE_BOT_PRIVATE_KEY: ${{ secrets.RELEASE_BOT_PRIVATE_KEY }}
--- a/.github/workflows/release.testnet.push.tags.yml
+++ b/.github/workflows/release.testnet.push.tags.yml
@@ -8,9 +8,15 @@ on:
         required: true
         type: string
       component:
-        description: 'The component to tag'
-        required: true
+        description: 'The component to tag. Required unless module_tag is true; must be empty when it is.'
+        required: false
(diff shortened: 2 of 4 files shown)


ayushsingh82:
@ben-dz this implements #4162 (your issue) — bare `vX.Y.Z` tag for the root Go module through the same approval-gated `release.testnet.push.tags.yml` path as the component tags, with the immutability + linear-ancestry guards described in the issue. Sibling to #4273 (#4101). Both are stuck on workflow approval — only Bugbot has run. Could you kick off CI when you get a chance?

ayushsingh82:
Rebased onto main (v0.40.0 cut) and re-pushed — switched the new module-tag secrets to `RELEASE_BOT_APP_ID`/`RELEASE_BOT_PRIVATE_KEY` to match the release-bot-app migration that landed on main since this was opened. Workflow YAML re-validated, changelog entry moved back under Unreleased.

PR #4281, opened 2026-09-05
cli: add access-pass plan and apply

juan-malbeclabs:
Related: [RFC-28](https://github.com/malbeclabs/doublezero/blob/jo/access-pass-plan-apply/rfcs/rfc28-declarative-access-passes.md)

## Summary of Changes

- New `doublezero access-pass plan` and `doublezero access-pass apply`, which reconcile access passes against a YAML document describing the access a fleet should have. `plan` reports the difference between the document and the ledger and writes nothing; `apply` sends the same difference after a confirmation. Granting six servers across four groups in both roles goes from 48 invocations to one reviewable file.
- The document covers both halves of a pass's access: the multicast publish/subscribe allowlists, and the IBRL (unicast) tenant. Each entry is keyed by `(client_ip, user_payer)` — the pass's PDA seeds.
- Every field is declarative. A group the document does not name is revoked; an entry with no `ibrl` has its tenant cleared. This is what makes the file a description of state rather than a list of additions, and it is why the revocations are shown before they happen.
- A declared subscribe that an EdgeSeat feed already grants is reported as satisfied and never written. Re-granting would spend a transaction, change nothing, and make every run report as changed — so this is what keeps a converged document a genuine no-op. Publisher is never feed-covered, so a publish gap on the same group is still reported.
- Two situations are refused rather than attempted, and both exit non-zero: an access pass that does not exist (`AddMulticastGroup*Allowlist` against an empty PDA silently mints a `Prepaid` pass with 1/1 seats and `last_access_epoch: 0`, so a typo'd address would produce a junk pass that looks real), and a group leaving both allowlists at once (the detach verbs send the role being *kept* as desired state, so revoking both entries first strands the roles on the User account with no legal write to remove them).
- `apply --json` writes exactly one object to stdout with `changed` first, so configuration management can drive it with `changed_when`. It requires `--auto-approve` or `--dry-run`, since there is no terminal to answer the prompt on.
- No new dependency and no program change: `serde_yaml` is already in the crate, and all writes go through existing instructions.

Two details worth a reviewer's attention:

- **The IBRL write targets the stored pass, not the declared IP.** A pass with `allow_multiple_ip` lives at the PDA seeded with `0.0.0.0` and serves any client IP, and `SetAccessPass` seeds its PDA from the `client_ip` argument — sending the declared address would write a *different account* than the plan described. `plan` also warns whenever a declared address resolves to a shared pass.
- **The IBRL write reads the pass first.** `SetAccessPass` overwrites `accesspass_type`, `last_access_epoch`, `allow_multiple_ip` and both seat caps from its arguments, so those are read back and re-sent unchanged; the transaction moves the tenant, pins the epoch, and touches nothing else. Declaring `ibrl` pins the epoch to unlimited, because the tenant and the epoch are one grant and a finite value fails a later `connect ibrl` at an unpredictable date.

## Diff Breakdown

| Core logic   |     3 | +1118 / -0  | +1118 |
| Tests        |     3 | +1125 / -0  | +1125 |
| Docs         |     3 | +400  / -0  |  +400 |
| Scaffolding  |     3 | +16   / -0  |   +16 |
| **Total**    |     9 | +2659 / -0  | +2659 |

Roughly half the diff is tests, and the tests live inline in the same three files as the core logic. Nothing is modified — every line is new.

<details>

- `smartcontract/cli/src/accesspass/plan.rs` (+567 core / +515 tests) — the diff engine and its renderer. Builds the multicast grants/revokes and the IBRL change, applies the feed-coverage and dual-revoke rules, batches the group/tenant/feed reads, and renders the plan.
- `smartcontract/cli/src/accesspass/apply.rs` (+371 core / +443 tests) — the write path. Prompts, sends each change, reports per item and continues past failures; contains the read-modify-write for `SetAccessPass` and the JSON contract.
- `smartcontract/cli/src/accesspass/desired.rs` (+180 core / +167 tests) — the document schema, strict parsing, and payer resolution (kept separate from parsing so a document can be validated with no keypair and no network).
- `rfcs/rfc28-declarative-access-passes.md` (+223) — design, alternatives considered, and six open questions.
- `docs/access-pass-plan-apply.md` (+171) — operator documentation: schema, reading a plan, what it refuses, automation.
- `smartcontract/cli/src/cli/{accesspass,command}.rs`, `accesspass/mod.rs` (+16) — subcommand enum, dispatch and module registration.

</details>

## Testing Verification

- 35 new unit tests against a mocked client (10 document, 17 plan, 8 apply); 472 pass in the crate.
- **Idempotency**: a document matching the ledger plans nothing, and a second `apply` reports `changed: false` and sends zero transactions.
- **Feed coverage**: an EdgeSeat pass with an empty `mgroup_sub_allowlist` but a covering feed reports the group as satisfied `via feed <code>` and issues no transaction, while still reporting the publisher gap on the same group.
- **Shared-pass targeting**: an entry declaring `203.0.113.10` that resolves to a pass stored at `0.0.0.0` sends `SetAccessPass` with `client_ip: 0.0.0.0`, preserving the EdgeSeat type, `allow_multiple_ip`, and 3/5 seat caps, with `last_access_epoch` pinned to `u64::MAX`.
- **Refusals**: a missing access pass and a group leaving both allowlists each plan zero writes and are reported as blocked.
- **Strict parsing**: `subscibe:` is rejected with the entry index, the valid alternatives and the line/column, before any RPC call — confirmed against the built binary, which exits 1 at parse time while a valid document proceeds to the network stage.
- **Lazy reads**: a document with no `ibrl` skips the tenant scan when no pass carries a tenant, but still reads them when one does — a pass with a tenant has to be cleared, so the scan cannot be skipped merely because the document is silent.
- Not yet exercised against a live ledger; no local devnet was running. An end-to-end run over `dev/local-devnet-multicast-setup.md` is the remaining verification.

### DIFF
--- a/smartcontract/cli/src/accesspass/apply.rs
+++ b/smartcontract/cli/src/accesspass/apply.rs
@@ -1,0 +1,814 @@
+//! Converges the ledger onto an access-pass definition document.
+//!
+//! Builds the same plan `plan` prints, shows it, asks for confirmation, then sends each change
+//! and reports the outcome per item. Each allowlist change is its own instruction and its own
+//! transaction — there is no multi-group allowlist instruction — so a run continues past a
+//! failure and names what did and did not land.
+
+use crate::{
+    accesspass::{
+        desired::AccessPassDocument,
+        plan::{build_plan, render_plan, AccessPassPlan, IbrlChange, Op, PlannedChange, Role},
+    },
+    doublezerocommand::CliCommand,
+    requirements::{CHECK_BALANCE, CHECK_ID_JSON},
+};
+use clap::Args;
+use doublezero_cli_core::CliContext;
+use doublezero_sdk::commands::{
+    accesspass::{get::GetAccessPassCommand, set::SetAccessPassCommand},
+    multicastgroup::allowlist::{
+        publisher::{
+            add::AddMulticastGroupPubAllowlistCommand,
+            remove::RemoveMulticastGroupPubAllowlistCommand,
+        },
+        subscriber::{
+            add::AddMulticastGroupSubAllowlistCommand,
+            remove::RemoveMulticastGroupSubAllowlistCommand,
+        },
+    },
+};
+use serde::Serialize;
+use std::{
+    io::{BufRead, Write},
+    path::PathBuf,
+};
+
+/// Reads an access-pass definition document and converges the ledger onto it.
+#[derive(Args, Debug)]
+pub struct ApplyAccessPassCliCommand {
+    /// Path to the access-pass definition document (YAML)
+    #[arg(long, short = 'f')]
+    pub file: PathBuf,
+    /// Show the plan and exit without writing anything
+    #[arg(long, default_value_t = false)]
+    pub dry_run: bool,
+    /// Skip the confirmation prompt
+    #[arg(long, alias = "force", default_value_t = false)]
+    pub auto_approve: bool,
+    /// Also list the grants that are already satisfied
+    #[arg(long, default_value_t = false)]
+    pub verbose: bool,
+    /// Output as pretty JSON
+    #[arg(long, default_value_t = false)]
+    pub json: bool,
+}
+
+#[derive(Debug, Serialize)]
+struct ApplyResult {
+    #[serde(flatten)]
+    change: PlannedChange,
+    /// `applied` or `failed`.
+    state: &'static str,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    signature: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    error: Option<String>,
+}
+
+#[derive(Debug, Serialize)]
+struct IbrlResult {
+    #[serde(flatten)]
+    change: IbrlChange,
+    /// `applied` or `failed`.
+    state: &'static str,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    signature: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    error: Option<String>,
+}
+
+#[derive(Debug, Serialize)]
+struct ApplyJson<'a> {
+    changed: bool,
+    counts: Counts,
+    results: &'a [ApplyResult],
+    ibrl_results: &'a [IbrlResult],
+    plan: &'a AccessPassPlan,
+}
+
+#[derive(Debug, Serialize)]
+struct Counts {
+    applied: usize,
+    failed: usize,
+    satisfied: usize,
+    blocked: usize,
+}
+
+impl ApplyAccessPassCliCommand {
+    pub async fn execute<C: CliCommand, W: Write, R: BufRead>(
+        self,
+        _ctx: &CliContext,
+        client: &C,
+        out: &mut W,
+        input: &mut R,
+    ) -> eyre::Result<()> {
+        client.check_requirements(CHECK_ID_JSON | CHECK_BALANCE)?;
+
+        // A JSON consumer has no terminal to answer the prompt on, so make the caller say up
+        // front that no one is watching rather than hanging on a read that never returns.
+        if self.json && !self.dry_run && !self.auto_approve {
+            eyre::bail!(
+                "--json requires --auto-approve (or --dry-run); there is no way to confirm"
+            );
+        }
+
+        let document = AccessPassDocument::from_path(&self.file)?;
+        let desired = document.resolve(client.get_payer())?;
+        let plan = build_plan(client, &desired)?;
+
+        if !self.json {
+            render_plan(out, &plan, self.verbose)?;
+        }
+
+        if self.dry_run {
+            if self.json {
+                emit_json(out, &plan, &[], &[])?;
+            } else {
+                writeln!(out, "\n[dry-run] nothing was sent.")?;
+            }
+            return blocked_result(&plan);
+        }
+
+        if plan.is_empty() {
+            if self.json {
+                emit_json(out, &plan, &[], &[])?;
+            }
+            return blocked_result(&plan);
+        }
+
+        if !self.auto_approve {
+            write!(out, "\nDo you want to perform these actions? [y/N]: ")?;
+            out.flush()?;
+            let mut answer = String::new();
+            input.read_line(&mut answer)?;
+            if !matches!(answer.trim().to_lowercase().as_str(), "y" | "yes") {
+                writeln!(out, "Aborted.")?;
+                return Ok(());
+            }
+            writeln!(out)?;
+        }
+
+        let mut results = Vec::with_capacity(plan.changes.len());
+        let mut ibrl_results: Vec<IbrlResult> = Vec::with_capacity(plan.ibrl_changes.len());
+        for change in &plan.changes {
+            let outcome = send(client, change);
+            match outcome {
+                Ok(signature) => {
+                    if !self.json {
+                        let sign = if change.op == Op::Grant { '+' } else { '-' };
+                        writeln!(
+                            out,
+                            "  {sign} {} {} {}  {signature}  ✓",
+                            change.role.label(),
+                            change.group,
+                            change.client_ip
+                        )?;
+                    }
+                    results.push(ApplyResult {
+                        change: change.clone(),
+                        state: "applied",
+                        signature: Some(signature),
+                        error: None,
+                    });
+                }
+                Err(err) => {
+                    if !self.json {
+                        let sign = if change.op == Op::Grant { '+' } else { '-' };
+                        writeln!(
+                            out,
+                            "  {sign} {} {} {}  {err}  ✗",
+                            change.role.label(),
+                            change.group,
+                            change.client_ip
+                        )?;
+                    }
+                    results.push(ApplyResult {
+                        change: change.clone(),
+                        state: "failed",
+                        signature: None,
+                        error: Some(err.to_string()),
+                    });
+                }
+            }
+        }
+
+        for change in &plan.ibrl_changes {
+            let label = match (&change.from, &change.to) {
+                (None, Some(to)) => format!("+ ibrl {to}"),
+                (Some(from), None) => format!("- ibrl {from}"),
+                (Some(from), Some(to)) => format!("~ ibrl {from} -> {to}"),
+                (None, None) => "~ ibrl (epoch only)".to_string(),
+            };
+            match send_ibrl(client, change) {
+                Ok(signature) => {
+                    if !self.json {
+                        writeln!(out, "  {label} {}  {signature}  ✓", change.client_ip)?;
+                    }
+                    ibrl_results.push(IbrlResult {
+                        change: change.clone(),
+                        state: "applied",
+                        signature: Some(signature),
+                        error: None,
+                    });
+                }
+                Err(err) => {
+                    if !self.json {
+                        writeln!(out, "  {label} {}  {err}  ✗", change.client_ip)?;
+                    }
+                    ibrl_results.push(IbrlResult {
+                        change: change.clone(),
+                        state: "failed",
+                        signature: None,
+                        error: Some(err.to_string()),
+                    });
+                }
+            }
+        }
+
+        let applied = results.iter().filter(|r| r.state == "applied").count()
+            + ibrl_results.iter().filter(|r| r.state == "applied").count();
+        let failed = results.len() + ibrl_results.len() - applied;
+
+        if self.json {
+            emit_json(out, &plan, &results, &ibrl_results)?;
+        } else {
+            writeln!(out, "\nApply complete! {applied} applied, {failed} failed.")?;
+        }
+
+        if failed > 0 {
+            eyre::bail!("{failed} of {} allowlist changes failed", results.len());
+        }
+        blocked_result(&plan)
+    }
+}
+
+/// The blocked items are the part of the document this run deliberately did not do, so the exit
+/// code has to say so even when everything else succeeded.
+fn blocked_result(plan: &AccessPassPlan) -> eyre::Result<()> {
+    if plan.blocked.is_empty() {
+        return Ok(());
+    }
+    eyre::bail!(
+        "{} declared access pass(es) could not be reconciled; see the blocked items above",
+        plan.blocked.len()
+    )
+}
+
+fn send<C: CliCommand>(client: &C, change: &PlannedChange) -> eyre::Result<String> {
+    let code = change.group.clone();
+    let client_ip = change.client_ip;
+    let user_payer = change.user_payer;
+
+    let signature = match (change.role, change.op) {
+        (Role::Publisher, Op::Grant) => {
+            client.add_multicastgroup_pub_allowlist(AddMulticastGroupPubAllowlistCommand {
+                pubkey_or_code: code,
+                client_ip,
+                user_payer,
+            })?
+        }
+        (Role::Publisher, Op::Revoke) => {
+            client.remove_multicastgroup_pub_allowlist(RemoveMulticastGroupPubAllowlistCommand {
+                pubkey_or_code: code,
+                client_ip,
+                user_payer,
+            })?
+        }
+        (Role::Subscriber, Op::Grant) => {
+            client.add_multicastgroup_sub_allowlist(AddMulticastGroupSubAllowlistCommand {
+                pubkey_or_code: code,
+                client_ip,
+                user_payer,
+            })?
+        }
+        (Role::Subscriber, Op::Revoke) => {
+            client.remove_multicastgroup_sub_allowlist(RemoveMulticastGroupSubAllowlistCommand {
+                pubkey_or_code: code,
+                client_ip,
+                user_payer,
+            })?
+        }
+        // IBRL is not an allowlist entry; it is written by `access-pass set` through
+        // `send_ibrl`, and the planner never puts it in `changes`.
+        (Role::Ibrl, _) => {
+            eyre::bail!("internal error: an IBRL change reached the allowlist writer")
+        }
+    };
+
+    Ok(signature.to_string())
+}
+
+/// Re-sends `access-pass set` with the tenant the document declares.
+///
+/// `set` overwrites `accesspass_type`, `last_access_epoch`, the `ALLOW_MULTIPLE_IP` flag and both
+/// seat caps from its arguments, so the pass is read back and those are sent unchanged — the write
+/// moves the tenant, pins the epoch, and touches nothing else. (`mgroup_*_allowlist` and
+/// `DZF_LOCKED` survive a `set` untouched; EdgeSeat feed seats are preserved by the program when
+/// both the stored and incoming types are EdgeSeat.)
+///
+/// The read is deliberately fresh rather than carried from the plan: the seat caps are live
+/// counters that the plan may already have seen go stale.
+fn send_ibrl<C: CliCommand>(client: &C, change: &IbrlChange) -> eyre::Result<String> {
+    let (_, pass) = client
+        .get_accesspass(GetAccessPassCommand {
+            client_ip: change.client_ip,
+            user_payer: change.user_payer,
+        })?
+        .ok_or_else(|| {
+            eyre::eyre!(
+                "access pass for {} / {} disappeared between plan and apply",
+                change.client_ip,
+                change.user_payer
+            )
+        })?;
+
+    // Address the pass that actually holds the grant. A shared pass is stored at 0.0.0.0, and
+    // `set` seeds the PDA from this value — sending the concrete IP would write a different
+    // account than the one the plan described.
+    let signature = client.set_accesspass(SetAccessPassCommand {
+        accesspass_type: pass.accesspass_type.clone(),
+        client_ip: pass.client_ip,
+        user_payer: pass.user_payer,
+        last_access_epoch: if change.to.is_some() {
+            u64::MAX
+        } else {
+            pass.last_access_epoch
+        },
+        allow_multiple_ip: pass.allow_multiple_ip(),
+        tenant: change.to_pk,
+        max_unicast_users: pass.max_unicast_users,
+        max_multicast_users: pass.max_multicast_users,
+    })?;
+
+    Ok(signature.to_string())
+}
+
+fn emit_json<W: Write>(
+    out: &mut W,
+    plan: &AccessPassPlan,
+    results: &[ApplyResult],
+    ibrl_results: &[IbrlResult],
+) -> eyre::Result<()> {
+    let applied = results.iter().filter(|r| r.state == "applied").count()
+        + ibrl_results.iter().filter(|r| r.state == "applied").count();
+    let total = results.len() + ibrl_results.len();
+    let json = serde_json::to_string_pretty(&ApplyJson {
+        changed: applied > 0,
+        counts: Counts {
+            applied,
+            failed: total - applied,
+            satisfied: plan.satisfied.len(),
+            blocked: plan.blocked.len(),
+        },
+        results,
+        ibrl_results,
+        plan,
+    })?;
+    writeln!(out, "{json}")?;
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::ApplyAccessPassCliCommand;
+    use crate::{
+        requirements::{CHECK_BALANCE, CHECK_ID_JSON},
+        tests::utils::create_test_client,
+    };
+    use doublezero_cli_core::testing::{block_on, cli_context_default_for_tests};
+    use doublezero_sdk::{AccountType, MulticastGroup, MulticastGroupStatus};
+    use doublezero_serviceability::state::accesspass::{
+        AccessPass, AccessPassStatus, AccessPassType,
+    };
+    use solana_sdk::{pubkey::Pubkey, signature::Signature};
+    use std::{collections::HashMap, io::Cursor, net::Ipv4Addr};
+
+    const IP: Ipv4Addr = Ipv4Addr::new(203, 0, 113, 10);
+
+    fn signature() -> Signature {
+        Signature::from([
+            120, 138, 162, 185, 59, 209, 241, 157, 71, 157, 74, 131, 4, 87, 54, 28, 38, 180, 222,
+            82, 64, 62, 61, 62, 22, 46, 17, 203, 187, 136, 62, 43, 11, 38, 235, 17, 239, 82, 240,
+            139, 130, 217, 227, 214, 9, 242, 141, 223, 94, 29, 184, 110, 62, 32, 87, 137, 63, 139,
+            100, 221, 20, 137, 4, 5,
+        ])
+    }
+
+    /// A client holding one group `g1` and a pass whose subscriber allowlist is `sub_allow`.
+    fn fixture(
+        sub_allow_has_g1: bool,
+    ) -> (
+        crate::doublezerocommand::MockCliCommand,
+        Pubkey,
+        tempfile::NamedTempFile,
+    ) {
+        let mut client = create_test_client();
+        let payer = Pubkey::new_unique();
+        let group_pk = Pubkey::new_unique();
+
+        client.expect_get_payer().returning(move || payer);
+        client
+            .expect_check_requirements()
+            .with(mockall::predicate::eq(CHECK_ID_JSON | CHECK_BALANCE))
+            .returning(|_| Ok(()));
+        client.expect_list_multicastgroup().returning(move |_| {
+            Ok(HashMap::from([(
+                group_pk,
+                MulticastGroup {
+                    account_type: AccountType::MulticastGroup,
+                    index: 1,
+                    bump_seed: 1,
+                    owner: Pubkey::new_unique(),
+                    tenant_pk: Pubkey::default(),
+                    multicast_ip: [239, 0, 0, 1].into(),
+                    max_bandwidth: 1_000_000_000,
+                    status: MulticastGroupStatus::Activated,
+                    code: "g1".to_string(),
+                    publisher_count: 0,
+                    subscriber_count: 0,
+                },
+            )]))
+        });
+
+        let sub_allow = if sub_allow_has_g1 {
+            vec![group_pk]
+        } else {
+            vec![]
+        };
+        let pass = AccessPass {
+            account_type: AccountType::AccessPass,
+            bump_seed: 255,
+            accesspass_type: AccessPassType::Prepaid,
+            client_ip: IP,
+            user_payer: payer,
+            last_access_epoch: u64::MAX,
+            connection_count: 0,
+            status: AccessPassStatus::Connected,
+            mgroup_pub_allowlist: vec![],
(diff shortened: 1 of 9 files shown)


smartcontract/cli/src/accesspass/apply.rs
```diff
@@ -0,0 +1,814 @@
...
+
+impl ApplyAccessPassCliCommand {
+    pub async fn execute<C: CliCommand, W: Write, R: BufRead>(
+        self,
+        _ctx: &CliContext,
+        client: &C,
+        out: &mut W,
+        input: &mut R,
+    ) -> eyre::Result<()> {
+        client.check_requirements(CHECK_ID_JSON | CHECK_BALANCE)?;
```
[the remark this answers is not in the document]
juan-malbeclabs: Fixed in 634411571. The requirement check now runs only when the command is going to write:

```rust
if !self.dry_run {
}
```

Right call — a preview sends nothing, and the operator most likely to run one is the one who holds neither a keypair nor credits. It also aligns the two verbs: `plan` already skips the check for exactly this reason.

What a preview *does* need a signer for still fails on its own terms rather than silently: a document saying `user_payer: me` is now refused when there is none (see the `desired.rs` thread), instead of resolving to the default pubkey and planning against passes nobody holds.

`dry_run_needs_no_keypair_or_balance` sets no `check_requirements` expectation at all, so mockall fails the test if the command asks.

smartcontract/cli/src/accesspass/apply.rs
```diff
@@ -0,0 +1,814 @@
...
+        let failed = results.len() + ibrl_results.len() - applied;
+
+        if self.json {
+            emit_json(out, &plan, &results, &ibrl_results)?;
+        } else {
+            writeln!(out, "\nApply complete! {applied} applied, {failed} failed.")?;
+        }
+
+        if failed > 0 {
+            eyre::bail!("{failed} of {} allowlist changes failed", results.len());
```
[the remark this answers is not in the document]
juan-malbeclabs: Fixed in 634411571 — `results.len() + ibrl_results.len()`, and the wording drops "allowlist" since the count no longer means only those:

```rust
eyre::bail!("{failed} of {} changes failed", results.len() + ibrl_results.len());
```

Confirmed the "1 of 0" case was reachable: `failed` is computed from both vectors a few lines above, so an IBRL-only run that failed reported a denominator of zero while the summary line right before it said "0 applied, 1 failed". `a_failed_ibrl_write_is_counted_in_the_total` covers it and fails against the old message.

smartcontract/cli/src/accesspass/apply.rs
```diff
@@ -0,0 +1,814 @@
...
+    plan: &AccessPassPlan,
+    results: &[ApplyResult],
+    ibrl_results: &[IbrlResult],
+) -> eyre::Result<()> {
+    let applied = results.iter().filter(|r| r.state == "applied").count()
+        + ibrl_results.iter().filter(|r| r.state == "applied").count();
+    let total = results.len() + ibrl_results.len();
+    let json = serde_json::to_string_pretty(&ApplyJson {
+        changed: applied > 0,
+        counts: Counts {
```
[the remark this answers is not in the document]
juan-malbeclabs: Fixed in 634411571, with the split you suggested:

```rust
let changed = if total == 0 { !plan.is_empty() } else { applied > 0 };
```

With no results the run wrote nothing — a dry run, or an empty plan — so `changed` answers the only question left, whether anything *would* move; after a real run it keeps reporting what landed. That also makes the field mean the same thing in `apply --dry-run --json` as in `plan --json`, which is what a driver comparing the two would expect.

Two tests: `dry_run_json_reports_changed_from_the_plan` (a pending grant, `changed: true`, `counts.applied: 0`) and `dry_run_json_reports_unchanged_for_a_converged_document`, so the fix cannot degenerate into "dry run always reports changed".

smartcontract/cli/src/accesspass/desired.rs
```diff
@@ -0,0 +1,347 @@
...
+            let user_payer = if raw_payer.eq_ignore_ascii_case("me") {
+                payer
+            } else {
+                Pubkey::from_str(raw_payer).map_err(|_| {
+                    eyre::eyre!(
+                        "access_passes[{index}] ({}) has an invalid user_payer: {raw_payer}",
+                        entry.client_ip
+                    )
+                })?
+            };
```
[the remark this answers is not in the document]
juan-malbeclabs: Fixed in 634411571 — `me` is refused when `get_payer()` answers with the default pubkey, naming both ways out (`--keypair`, or write the pubkey in the document).

The reason it has to be an error rather than a fallback: the default pubkey is a perfectly good PDA seed, so resolution succeeds and the run goes on to describe passes nobody holds. Every entry then reads as blocked or as a grant to make, and the document looks wrong rather than unresolvable — the failure mode is a misleading plan, not a missing one.

This is also what keeps the `--dry-run` fix on the neighbouring thread honest: dropping the requirement check means a run can now reach `resolve` with no signer configured, and this is the case where that matters.

Two tests, since `defaults.user_payer` resolves through the same path: `me_is_refused_when_there_is_no_signer` and `me_in_defaults_is_refused_when_there_is_no_signer`.

smartcontract/cli/src/accesspass/plan.rs
```diff
@@ -0,0 +1,1082 @@
...
+                changed: !plan.is_empty(),
+                plan: &plan,
+            })?;
+            writeln!(out, "{json}")?;
+        } else {
+            render_plan(out, &plan, self.verbose)?;
+        }
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
juan-malbeclabs: Fixed in 634411571. `blocked_result` moved to `plan.rs` and both verbs return it, so `plan` exits non-zero on exactly the documents `apply` refuses.

Worth noting the docs already promised this — `docs/access-pass-plan-apply.md`: "Blocked items are reported, and both verbs exit non-zero when there are any." So this was the code disagreeing with its own documentation, which is why no doc change goes with the fix.

Two tests: `plan_exits_non_zero_when_an_item_is_blocked` (which also asserts the blocked item is still printed — the exit code is added to the output, not substituted for it) and `plan_exits_zero_when_nothing_is_blocked`, because a plan that merely has work in it must stay a success or the gate is useless.

smartcontract/cli/src/accesspass/apply.rs
```diff
@@ -0,0 +1,814 @@
...
+
+        if self.json {
+            emit_json(out, &plan, &results, &ibrl_results)?;
+        } else {
+            writeln!(out, "\nApply complete! {applied} applied, {failed} failed.")?;
+        }
+
+        if failed > 0 {
+            eyre::bail!("{failed} of {} allowlist changes failed", results.len());
+        }
```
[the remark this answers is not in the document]
juan-malbeclabs: Fixed in 634411571, same change as the Copilot thread on line 240: the denominator is now `results.len() + ibrl_results.len()` and the text says "changes" rather than "allowlist changes".

juan-malbeclabs:
All six bot findings are fixed in `634411571` and answered on their own threads; the threads are resolved. Both duplicates of the IBRL count (Copilot 240, Bugbot 241) are the same one-line change.

| Finding | Fix |
|---|---|
| Requirement checks run under `--dry-run` | Checked only for a real apply; a preview needs neither keypair nor credits, like `plan` |
| Failure denominator omits IBRL | `results.len() + ibrl_results.len()`, and it no longer says "allowlist" |
| `changed: false` on `--dry-run --json` | With no results, `changed` reports the plan — the same meaning `plan --json` gives it |
| `user_payer: me` with no signer | Refused, naming `--keypair` and the explicit-pubkey way out |
| `plan` exits zero when blocked | Both verbs share `blocked_result` |

Two of these interact, which is worth flagging for review as one decision rather than two: dropping the requirement check is what lets a run reach `resolve` with no signer configured, and refusing `me` there is what stops that becoming a plan against the default pubkey. Taken separately either one is arguable; together they make `--dry-run` work for an operator with no keypair and fail loudly for the one document that cannot mean anything without one.

The `plan` exit-code change is the only one that alters a contract a caller may already depend on, and the documentation already claimed the new behaviour — `docs/access-pass-plan-apply.md`: "Blocked items are reported, and both verbs exit non-zero when there are any."

Seven new tests; `cargo test -p doublezero-serviceability-cli` is 480 green and clippy is clean. The CHANGELOG entry picks up the exit-code, dry-run and `me` behaviours.

**This branch is behind main and currently conflicts** (`CHANGELOG.md`, from releases cut since it was opened). I have not rebased it — say the word and I will, or I can leave it until review settles.

smartcontract/cli/src/accesspass/plan.rs
```diff
@@ -0,0 +1,1176 @@
...
+        // `true` against these allowlists — so once both entries are gone, `multicast unpublish`
+        // asks for subscriber=true and `multicast unsubscribe` asks for publisher=true, and
+        // neither is allowlisted any more. The roles are then stranded on the User with no legal
+        // write to remove them. Detach the host first, then revoke.
+        let dual_revoke: Vec<&str> = have_pub
+            .difference(&want_pub)
+            .filter(|code| have_sub.difference(&want_sub).any(|s| s == *code))
+            .copied()
+            .collect();
+        if !dual_revoke.is_empty() {
```
nikw9944, line 274: **Blocking** — A document that drops a group from both the publish and subscribe lists is refused, even though the write it guards against is legal. `check_mgroup_allowlists` in `subscribe.rs` is now called with `value.subscriber && !user.subscribers.contains(..)`, so a role the user already holds is not re-checked and the detach verbs succeed with both entries gone.

smartcontract/cli/src/accesspass/plan.rs
```diff
@@ -0,0 +1,1176 @@
...
+
+        for (role, want, have) in [
+            (Role::Publisher, &want_pub, &have_pub),
+            (Role::Subscriber, &want_sub, &have_sub),
+        ] {
+            for code in want.difference(have) {
+                // A feed already grants subscribe on its groups in its own metro, so granting it
+                // again spends a transaction and changes nothing. Publisher is never feed-covered.
+                if role == Role::Subscriber {
+                    if let Some(feed_code) = feed_granted.get(*code) {
```
nikw9944, line 332: **Blocking** — A declared subscribe that an EdgeSeat feed covers is reported satisfied and never written, but the feed only substitutes for the allowlist on the connect path. `UpdateMulticastGroupRoles` — what `doublezero multicast subscribe` sends — checks `mgroup_sub_allowlist` for any role the user gains, so the host is refused while the plan reports converged.

smartcontract/cli/src/accesspass/plan.rs
```diff
@@ -0,0 +1,1176 @@
...
+                reason: "no access pass at this PDA — create it with `doublezero access-pass set`"
+                    .to_string(),
+            });
+            continue;
+        };
+
+        // Access-pass resolution prefers a pass stored at the 0.0.0.0 PDA, which is valid for any
+        // client IP. So a lookup for a concrete address can legitimately land on a shared pass —
+        // and every group granted here is then granted to every host using that pass.
+        if pass.client_ip != entry.client_ip {
```
nikw9944, line 243: **Blocking** — Two entries whose passes resolve to the same account produce contradictory plans. With a shared `0.0.0.0` pass each entry reads the same allowlists and plans a revoke of the groups the others declared, so apply flips the state and the next run flips it back.

smartcontract/cli/src/accesspass/apply.rs
```diff
@@ -0,0 +1,1029 @@
...
+    // `set` seeds the PDA from this value — sending the concrete IP would write a different
+    // account than the one the plan described.
+    let signature = client.set_accesspass(SetAccessPassCommand {
+        accesspass_type: pass.accesspass_type.clone(),
+        client_ip: pass.client_ip,
+        user_payer: pass.user_payer,
+        last_access_epoch: if change.to.is_some() {
+            u64::MAX
+        } else {
+            pass.last_access_epoch
```
nikw9944, line 339: **Minor** — Clearing a tenant re-sends the pass's stored `last_access_epoch`; a finite elapsed epoch is rejected by both the SDK and the program, so IBRL cannot be revoked from an expired pass.

nikw9944 (commented):
Re-checked at b4194656 (main merged in, the PR's own files unchanged): all four findings remain open. Three of the plan's rules rest on program behavior that changed or was never checked — the dual-revoke block guards against a write `UpdateMulticastGroupRoles` now permits, feed coverage is treated as a full substitute for the subscriber allowlist, and two entries resolving to one shared pass plan contradictory revokes that never converge. Clearing a tenant on an expired pass also fails on the re-sent epoch.

smartcontract/cli/src/accesspass/plan.rs
```diff
@@ -0,0 +1,1176 @@
...
+) -> HashMap<&'a str, String> {
+    let mut granted = HashMap::new();
+    let Some(feeds) = feeds else {
+        return granted;
+    };
+    for seat in pass.feed_seats() {
+        let Some(feed) = feeds.get(&seat.feed_key) else {
+            continue;
+        };
+        for group in &feed.groups {
```
elitegreg: **Feed coverage ignores the metro gate, so a subscribe can be reported satisfied when nothing authorizes it**

The doc comment above says a feed grants subscribe on its groups "in its own metro", and `docs/access-pass-plan-apply.md` repeats it, but the loop has no metro condition: every group on every seated feed suppresses the grant.

The program does gate it. `check_feed_metro_coverage` (`smartcontract/programs/doublezero-serviceability/src/processors/feed/mod.rs:61`) rejects with `MetroMismatch` unless `feed.exchange == device_exchange`, and `load_feeds` runs that per seat on `SubscribeFeed`.

So for a pass holding seats on feeds in different metros, a declared subscribe that only the out-of-metro feed covers is reported as `satisfied via feed <code>`, no allowlist entry is written, and the host's `multicast subscribe` then fails on a device in the other exchange. That is the same shape of late, unpredictable failure the `last_access_epoch` pinning exists to prevent, and it is silent: the converged document keeps reporting a clean no-op.

The CLI cannot know which device the host will attach to, so a metro check is not the fix. Two options that keep the no-op property:

1. Name the metro in the `source` string, `feed mg-marketdata (xams)`, so the operator can see what the suppression is conditional on.
2. Push a warning when a pass holds seats across more than one exchange, alongside the existing shared-pass warning.

(1) alone would probably do it.

smartcontract/cli/src/accesspass/apply.rs
```diff
@@ -0,0 +1,1029 @@
...
+        })?;
+
+    // Address the pass that actually holds the grant. A shared pass is stored at 0.0.0.0, and
+    // `set` seeds the PDA from this value — sending the concrete IP would write a different
+    // account than the one the plan described.
+    let signature = client.set_accesspass(SetAccessPassCommand {
+        accesspass_type: pass.accesspass_type.clone(),
+        client_ip: pass.client_ip,
+        user_payer: pass.user_payer,
+        last_access_epoch: if change.to.is_some() {
```
elitegreg: **Clearing a tenant on a pass with an expired epoch will be rejected**

The `else` branch re-sends `pass.last_access_epoch` unchanged, and the SDK guards that value before it builds anything:

```rust
// smartcontract/sdk/rs/src/commands/accesspass/set.rs:25
if self.last_access_epoch > 0 && self.last_access_epoch != u64::MAX {
    let epoch = client.get_epoch()?;
    if self.last_access_epoch < epoch {
        return Err(eyre::eyre!("last_access_epoch {} cannot be in the past ...
```

So: a pass carrying a tenant and a finite epoch that has since passed, and a document that omits `ibrl`. `plan` shows `- ibrl <tenant>`, `apply` re-sends the stale epoch, and the SDK refuses. The item reports failed and the run exits non-zero with an error about an epoch the operator never mentioned, on a write whose whole purpose was to *remove* a grant.

Sending `0` when clearing would sidestep the guard, and it means "no epoch defined", which blocks every unicast type: the same thing removing the tenant intends. Whatever you pick, it is worth a mocked test, since none of the eight apply tests sets a finite past epoch.

smartcontract/cli/src/accesspass/apply.rs
```diff
@@ -0,0 +1,1029 @@
...
+
+        if failed > 0 {
+            // Both kinds of write are counted, not just the allowlist ones: an IBRL-only run
+            // that failed would otherwise report "1 of 0 changes failed".
+            eyre::bail!(
+                "{failed} of {} changes failed",
+                results.len() + ibrl_results.len()
+            );
+        }
+        blocked_result(&plan)
```
elitegreg: **`apply` writes the valid subset when the plan has blocked items**

`blocked_result` is returned early on the dry-run and empty-plan paths, so a preview refuses correctly, but on a real apply it is only reached here, after the prompt and after every send. A document with one blocked entry and nine good ones prompts, applies the nine, then exits non-zero.

Nothing half-applies within a pass, since the planner's `continue` drops the blocked entry's other changes too, so this is coherent as partial convergence. But it does not match the tool this is modeled on: `terraform apply` refuses a plan it cannot fully make rather than applying the valid subset. And the resulting exit code is ambiguous to the CI job the PR describes, which cannot tell a wholly-refused run from a successful partial one.

My preference would be to bail before the prompt when `plan.blocked` is non-empty, with `--allow-blocked` (or similar) for the operator who genuinely wants the subset. Either way it is worth being a decision rather than a fallthrough, and worth stating in `docs/access-pass-plan-apply.md`, which currently says only that blocked items make both verbs exit non-zero.

elitegreg (dismissed):
Design is sound and unusually well argued: the comments explain *why* at every non-obvious turn, the two refusals are real traps that I verified against the processors, and the tests are substantive rather than ceremonial. 496 pass in the crate on this head.

One blocking item, in the merge rather than in your own commits, plus three findings inline.

## The last merge of main clobbered `CHANGELOG.md`

`b419465` resolved `CHANGELOG.md` by taking the branch side wholesale. As it stands this PR deletes:

- the **v0.40.0, v0.41.0 and v0.42.0** release sections (the file jumps from Unreleased straight to v0.39.0), and
- every other **Unreleased** entry: the `connect --client-ip` CLI and Client entries, the five Geolocation entries from #4271, the `GetExactAccessPassCommand` SDK entry, and the Serviceability RFC-27 entry.

That accounts for essentially all 94 deletions in the diff. The fix is to redo that hunk against main's file and add only your five new CLI bullets under `### Changes`.

## Summary of the inline findings

| | Finding |
|---|---|
| High | Clearing a tenant on a pass with an expired epoch is rejected by the SDK's own guard |
| Medium | Feed coverage ignores the metro gate, so a subscribe can be reported satisfied when nothing authorizes it |
| Medium | `apply` writes the valid subset when the plan has blocked items, rather than refusing the run |

Two smaller things, neither worth blocking on:

- Aborting at the confirmation prompt returns `Ok(())` and skips `blocked_result`, so `apply` on a blocked document exits 0 if you answer `n` and non-zero if you answer `y`. It is the one path where the exit code disagrees with the others.
- `resolve` validates the tenant code and the payer carefully but not the group codes, so `publish: [""]` reaches the planner and surfaces as `unknown multicast group code(s):` with an empty name. Later and vaguer than the rest of that module's messages.

smartcontract/cli/src/accesspass/plan.rs
juan-malbeclabs: **Fixed in `929fcc4ba`** — the refusal is removed.

You are right, and the PR body was arguing from a version of the program that no longer exists. `31f3a3c33` (#4302, 9 Sep — four days after this PR was opened) changed the call to:

```rust
check_mgroup_allowlists(
    &accesspass,
    group_account.key,
    value.publisher && !user.publishers.contains(group_account.key),
    value.subscriber && !user.subscribers.contains(group_account.key),
)?;
```

and the function's own doc comment now says it outright: "Removals are always allowed, so a user can be cleaned up after a group leaves an allowlist." So the detach the block guarded against is legal, and the block was refusing a document that works.

`revoking_both_roles_of_one_group_is_blocked` is now `revoking_both_roles_of_one_group_is_planned` and asserts both revokes are planned with nothing blocked. The RFC keeps a short note on why the rule existed and why it went, so the next reader does not reintroduce it.

smartcontract/cli/src/accesspass/plan.rs
juan-malbeclabs: **Fixed in `929fcc4ba`** — this is now blocked, and it replaces the dual-revoke rule in the same spot.

Confirmed the non-convergence: with a shared `0.0.0.0` pass both entries read the same allowlists, so each plans a revoke of whatever the other declared, and apply flips the state every run. The existing warning was not enough — it told the operator the pass was shared, not that the document could never be satisfied.

Blocked only on a genuine disagreement. Two entries resolving to one pass and declaring the *same* access are redundant rather than contradictory, so they are planned once rather than twice (planning twice would send a duplicate write). Two tests: `two_entries_sharing_one_pass_with_different_access_are_blocked` and `two_entries_sharing_one_pass_with_the_same_access_plan_once`.

smartcontract/cli/src/accesspass/plan.rs
juan-malbeclabs: **Fixed in `929fcc4ba`** — the suppression is gone; the allowlist entry is always written.

Your reading is the correct one, and it is broader than the metro problem raised on the other feed thread. There is no feed path in `UpdateMulticastGroupRoles` at all — the comment above the loop states the rule plainly:

> Every pass type is authorized the same way here: each group must be on the pass's allowlist.

The feed substitution exists only in `create_subscribe.rs` (`value.subscriber && !feed_gated`), i.e. connect. So `doublezero multicast subscribe` → `cli/src/user/subscribe.rs:115` → `UpdateMulticastGroupRolesCommand` → that loop, and the host is refused while the plan reports converged. Both narrowings compound, so I did not try to model either: the CLI cannot know which device a host will attach to, and it cannot know whether the join arrives at connect or afterwards.

The coverage is still surfaced, as a warning rather than a skipped write, since an EdgeSeat operator seeing `+ subscriber` on a group their feed carries deserves to know why.

On the idempotency argument the rule was there to protect: it survives without it. The entry is written once and then matches, so a converged document is still a no-op — that property comes from the allowlist comparison, not from the feed rule. Cost is one transaction per feed-covered group, once.

`a_feed_granted_subscribe_is_satisfied_not_granted` is now `a_feed_granted_subscribe_is_still_written_to_the_allowlist`, asserting both grants are planned, that nothing is reported satisfied `via feed`, and that the warning names the feed.

smartcontract/cli/src/accesspass/apply.rs
juan-malbeclabs: **Fixed in `929fcc4ba`.** Confirmed against the guard in `smartcontract/sdk/rs/src/commands/accesspass/set.rs:25` — a finite `last_access_epoch` below the current epoch is rejected before the command builds anything, so IBRL could not be revoked from an expired pass.

Clearing now sends `0` when the stored epoch has already elapsed. Zero is "no epoch defined" rather than "expired", and blocks every unicast type, which is the end state removing the tenant intends anyway.

The clamp is deliberately narrow: an epoch still in the future is re-sent unchanged, because zeroing it would revoke unicast access the document never asked to remove. Two tests, since as you and @elitegreg both noted none of the apply tests set a finite past epoch: `clearing_a_tenant_on_an_expired_pass_sends_epoch_zero` and `clearing_a_tenant_preserves_an_epoch_still_in_the_future`.

(Worth noting for anyone writing a similar test: `create_test_client()` already mocks `get_epoch` at 10, and mockall honors the first expectation, so a second one is silently ignored — the fixtures pick their epochs relative to that.)

smartcontract/cli/src/accesspass/plan.rs
juan-malbeclabs: **Fixed in `929fcc4ba`**, though further than option (1) — the suppression itself is removed rather than annotated.

Your metro finding is correct and I verified `check_feed_metro_coverage` rejects with `MetroMismatch` unless `feed.exchange == device_exchange`. But @nikw9944's thread above found the larger half of the same problem: the feed substitution exists **only on the connect path**. `UpdateMulticastGroupRoles` — what `doublezero multicast subscribe` sends once the user exists — has no feed branch at all, and its own comment says every pass type is authorized against the allowlist.

So the coverage is narrowed twice over: connect only, and in-metro only. Naming the metro in the `source` string would have made one of those visible while leaving the other silent, and the plan would still have skipped a write the host needs. Rather than model either — the CLI cannot know which device a host attaches to, nor whether the join comes at connect or later — the allowlist entry is now always written.

The idempotency property the rule existed for survives: the entry is written once, then matches, so a converged document is still a no-op. Your option (1) lives on as a warning naming the feed, so the operator still sees the relationship; `docs/access-pass-plan-apply.md` gains a section spelling out both narrowings.

smartcontract/cli/src/accesspass/apply.rs
juan-malbeclabs: **Fixed in `929fcc4ba`** — this is the same defect @nikw9944 raised on `apply.rs:339`; your severity was the right one.

Took the `0` you suggested, with one narrowing: it applies only when the stored epoch has actually elapsed. A finite epoch still in the future is re-sent unchanged, since zeroing that would revoke unicast access the document never asked to remove. `get_epoch()` is on the trait already, so the check costs nothing extra.

Two mocked tests, as you asked — `clearing_a_tenant_on_an_expired_pass_sends_epoch_zero` and `clearing_a_tenant_preserves_an_epoch_still_in_the_future` — both asserting on the `SetAccessPassCommand` via `withf`.

smartcontract/cli/src/accesspass/apply.rs
juan-malbeclabs: **Fixed in `929fcc4ba`** — took your preference: `apply` now bails before the prompt when `plan.blocked` is non-empty, with `--allow-blocked` for the operator who wants the subset.

Agreed it should be a decision rather than a fallthrough, and the terraform comparison is the one that settles it: this tool is modeled on `plan`/`apply`, and the exit code has to carry one meaning for the CI job the PR describes. The refusal sends nothing, so non-zero-without-`--allow-blocked` now means "nothing ran" unambiguously.

Both smaller items are in too:

- **Aborting at the prompt** returns `blocked_result(&plan)` rather than `Ok(())`, so the exit code no longer depends on whether the operator answered `y` or `n`. With the change above this is only reachable under `--allow-blocked`, where exiting non-zero on `n` is the consistent answer.
- **Group codes** are validated in `resolve`, alongside the tenant code and the payer. `publish: [""]` is now rejected as `access_passes[0] (203.0.113.10) has an empty multicast publish code; remove the entry to declare no group` — before any RPC, with the entry index, rather than surfacing later as `unknown multicast group code(s):` with an empty name.

Tests: `a_blocked_plan_is_refused_before_anything_is_sent` (mockall panics on any unexpected write, so "nothing was sent" is asserted structurally), `allow_blocked_applies_the_valid_subset_and_still_exits_non_zero`, and `an_empty_multicast_group_code_is_rejected_with_its_entry`. `docs/access-pass-plan-apply.md` and the RFC both state the behavior now.

juan-malbeclabs:
All seven open findings are fixed in `929fcc4ba`, and the CHANGELOG blocker in `296b51c8c`. Each is answered on its own thread.

| Finding | Reviewer | Fix |
|---|---|---|
| CHANGELOG clobbered by the merge | @elitegreg | Redone against main; the diff is now **+5 / −0** |
| Feed coverage suppresses a needed write | @nikw9944 + @elitegreg | Suppression removed; coverage becomes a warning |
| Dual-revoke refusal guards a legal write | @nikw9944 | Refusal removed |
| Shared pass → contradictory plans | @nikw9944 | Blocked when the entries disagree; planned once when they agree |
| Tenant clear rejected on an expired epoch | @nikw9944 + @elitegreg | Sends `0` when the stored epoch has elapsed |
| `apply` writes the valid subset when blocked | @elitegreg | Refuses before the prompt; `--allow-blocked` opts in |
| Abort at prompt exits 0 · unvalidated group codes | @elitegreg | Both fixed |

Three things worth surfacing rather than leaving on threads.

**The two feed findings were one bug and a bigger one.** @elitegreg found that coverage ignores the metro gate; @nikw9944 found that coverage does not apply to `multicast subscribe` *at all*. Both are right and they compound — `UpdateMulticastGroupRoles` has no feed branch, and the connect path that does have one is metro-gated. So naming the metro in the `source` string would have left the larger half silent. The suppression is gone instead. The idempotency property it existed to protect survives on its own: the entry is written once, then matches.

**One PR-body claim was arguing from a program that no longer exists.** The dual-revoke refusal was justified by the detach verbs being unauthorized once a group left both allowlists. `31f3a3c33` (#4302) landed on 9 September, four days after this PR was opened, and changed `check_mgroup_allowlists` to authorize only roles being *gained*. Worth flagging because @elitegreg's review states both refusals were verified against the processors — that verification was correct for the other one. The RFC now records why the rule existed and why it went, so it does not get reintroduced.

**`--allow-blocked` is new surface**, added on @elitegreg's stated preference for terraform's refuse-the-whole-plan behavior. Reasonable to push back on if you would rather have no flag at all and an unconditional refusal.

503 tests pass in the crate, clippy clean. Still not exercised against a live ledger — that remains the open item.

nikw9944 (approved):
All four findings verified fixed at 296b51c8: the dual-revoke refusal is gone (`revoking_both_roles_of_one_group_is_planned` asserts both revokes); feed coverage no longer suppresses the subscriber write, matching `UpdateMulticastGroupRoles`'s allowlist check; two entries colliding on one pass PDA are blocked unless they agree; and `send_ibrl` sends epoch 0 for an elapsed stored epoch, clearing `SetAccessPassCommand`'s guard. Also read the new `--allow-blocked` refusal path and the docs/RFC rewrites.

nikw9944 (approved):
Head moved only by a merge of main (geoprobe Go telemetry), and the three access-pass source files are byte-identical to the approved version. Re-verified at 5834ee9b: no `dual_revoke` guard remains in `build_plan`, the feed branch pushes a warning instead of suppressing the grant, `by_pass` blocks entries that collide on one pass PDA, and `send_ibrl`'s clamp still sends 0 only for an elapsed stored epoch.

elitegreg (approved):
Verified at 5834ee9b. The three rule corrections in `929fcc4b` hold against the program: `UpdateMulticastGroupRoles` has no feed path, so the subscriber allowlist write is required and the coverage warning is the right treatment; `set.rs` preserves `DZF_LOCKED` (masking only `ALLOW_MULTIPLE_IP`) and EdgeSeat feed seats when both types are EdgeSeat, so `send_ibrl`'s read-modify-write is sound; and the elapsed-epoch clamp matches `SetAccessPassCommand`'s own rejection. The allowlist path needs no equivalent to `client_ip: pass.client_ip` because `resolve_accesspass_pda` prefers the 0.0.0.0 PDA and the processor accepts it via the `allow_multiple_ip` branch.

CHANGELOG is +5/-0 against main — my blocker is resolved. 503 pass in the crate on this head.

PR #4282, opened 2026-09-05
serviceability: make a feed a staked object (RFC-28 epic A)

bgm-malbeclabs:
Implements epic A of the Builder-Deployed Edge delivery plan, against RFC-28. Four commits, one per task, reviewable in order.

## Summary

- **`Feed` carries the RFC-28 stake terms.** `builder`, `stake_ref`, `spec_id`, `sla_hash`, `committed_rate_bits_per_sec`, and a lifecycle `status`. The rate is bits per second, not basis points: `bps` already means basis points elsewhere in DoubleZero and the two would be read wrong.
- **The feature ships dormant.** Setting the new fields needs `allow-staked-feeds`, which no cluster has. With the flag clear, a create carrying a builder is refused outright rather than silently dropping the stake terms, and `CreateFeed` otherwise behaves exactly as before. Same mechanism RFC-27 used for `require-ip-ownership-proof`.
- **New `StakeMirror` account**, one per builder, holding the stake a relayer saw on Solana. The DZ ledger cannot read a Solana account, so this is an assertion rather than a proof, and `relayer` records whose assertion it is.
- **`CreateFeed` refuses a rate the builder's stake tier does not cover**, reading the tier from the mirror. Two new errors, `StakeMirrorMissing` (119) and `StakeDoesNotCoverRate` (120), so a demo failure says which thing went wrong.
- **No version bump and no realloc instruction.** The fields append, short accounts decode with defaults per the `AccessPass` precedent, and `try_acc_write` already resizes on the next write.
- **Publish rights need no new instruction.** Covered by test rather than by code; see below.

**`status` does not use its own default.** A `Feed` written before this has no status byte, and `FeedStatus::default()` is `Pending`. Decoding a short account as `Pending` would pull every live catalog feed out of service on upgrade, so the decoder reads `Active` there. Same rule in the Rust program and in the Python and TypeScript SDKs.

**The stake mirror is located by PDA, not by position.** `CreateFeed` already ends in an optional `Permission` account. A second optional account at a fixed offset makes the two ambiguous whenever only one is sent, so the tail is searched for the expected mirror PDA instead.

**A4 is a test, not an instruction.** The plan asked for `CreateFeed` to add the builder to `mgroup_pub_allowlist`. It cannot: that allowlist lives on `AccessPass`, whose PDA is seeded on the client IP, and a feed does not carry one. It does not need to. Groups are created with `owner` set to the builder, and `AddMulticastGroupPubAllowlist` authorizes on `mgroup.owner == payer`, so the builder grants its own rights. The alternative would also duplicate the credit transfer and multiple-IP handling that instruction already does.

**`owner` and `relayer` are separate fields on `StakeMirror`.** They hold the same key today. They stop holding the same key the moment the trust-anchor question resolves to anything other than one signer.

## Left out

- The CLI does not expose the new `CreateFeed` arguments. Nothing can flip the feature flag yet, so there is nothing to drive.
- No instruction writes a `StakeMirror`. Who may sign one is the open trust-anchor decision, and writing the instruction before it is settled means writing it twice. This blocks the relayer work as well as the mirror's final shape.
- No SDK support for `StakeMirror`. The Python and TypeScript clients ignore account types they do not know, so nothing breaks, and nothing reads mirrors from an SDK yet.

## Open question for reviewers

`CreateFeed` is admin-only today (`FEED_AUTHORITY | FOUNDATION`, `owner = payer`). This PR makes `builder` an argument, so an admin deploys a feed on the builder's behalf. RFC-28 reads as though builders deploy their own. If that is the intent, it is a permission change, and it should land before more work builds on the current shape.

## Testing Verification

`make rust-lint` and `make sdk-test` clean. `cargo test -p doublezero-serviceability` green: 327 lib tests, 16 in `feed_test`.

Scenarios covered:

- A feed written before this decodes with the stake fields defaulted and `status` reading `Active`, asserted in Rust against a hand-built old-layout encoding and in the Python and TypeScript SDKs against a new `feed_legacy` fixture.
- With the flag clear, a create carrying a builder fails `NotAllowed`, and a create carrying none still succeeds and lands `Active`.
- Stake terms are refused unless they travel together: builder with no stake ref, spec, SLA or rate, and stake terms with no builder, each rejected.
- A rate one bit above the tier ceiling is refused, and a rate exactly at it is admitted, on both the 1 Gbps and 5 Gbps tiers.
- A builder with no mirror written and a create that omits the mirror account fail with distinct errors.
- The publish-rights walk runs with two separate signers: the admin creates the groups and the feed, and the builder alone adds itself to both groups' publisher allowlists. A change that quietly makes the second step need the admin's authority fails this test.

One pre-existing failure elsewhere in `make rust-test`, unrelated to this branch: `doublezero-daemon-cli`'s `probe_skips_a_local_address_that_cannot_reach_the_destination`, which depends on how the host routes 127.0.0.1 to 8.8.8.8. It fails identically on a clean tree.

## Size

About 560 lines of non-test code, over the ~500 guideline in AGENTS.md. The commits are one per task and split cleanly at any boundary; the natural cut is the two state commits from the two behavior commits. Happy to split if reviewers prefer.


smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
```diff
@@ -144,3 +206,102 @@ pub(crate) fn validate_feed_groups(groups: &[Pubkey]) -> Result<(), DoubleZeroEr
...
+    committed_rate_bits_per_sec: u64,
+) -> Result<(), DoubleZeroError> {
+    if mirror_account.data_is_empty() || mirror_account.owner != program_id {
+        msg!("No stake mirror written for builder {}", builder);
+        return Err(DoubleZeroError::StakeDoesNotCoverRate);
+    }
+
+    let mirror =
+        StakeMirror::try_from(mirror_account).map_err(|_| DoubleZeroError::InvalidAccountType)?;
+
```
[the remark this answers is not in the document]
bgm-malbeclabs: Done in 3727441. `verify_stake_covers_rate` now calls `mirror.validate()` right after the decode, before it reads the tier, so a mirror missing its stake, builder, or relayer is refused rather than trusted. The anchor is stale; the call is on the line above.

smartcontract/programs/doublezero-serviceability/tests/test_helpers.rs
```diff
@@ -85,6 +85,48 @@ pub async fn init_test() -> (BanksClient, Pubkey, Keypair, solana_program::hash:
...
+/// which cannot be derived until it is known.
+///
+/// Some accounts are written by an offchain service rather than by an instruction — `StakeMirror`
+/// is written by the cross-chain relayer. Seeding one directly is how a test exercises the read
+/// path before the writer exists.
+#[allow(dead_code)]
+pub async fn init_test_with_accounts(
+    program_id: Pubkey,
+    accounts: &[(Pubkey, Vec<u8>)],
+) -> (BanksClient, Keypair, solana_program::hash::Hash) {
```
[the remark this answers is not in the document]
bgm-malbeclabs: Leaving this one. `init_test_with_accounts` sits next to `init_test`, which it mirrors exactly apart from taking a program id and seeded accounts. Splitting the pair so one lives in the shared helpers and the other beside a single test is the churn, not the fix. Happy to move it if a reviewer disagrees.

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
```diff
@@ -110,6 +170,13 @@ pub fn process_create_feed(
         ],
     )?;
 
+    // Spend the stake on this feed. Ordering matters: the feed account is created first, so a
+    // failure anywhere above leaves the stake unclaimed and the builder can retry.
+    if let Some((mirror_account, mut mirror)) = claimed_mirror {
+        mirror.feed_key = *feed_account.key;
```
armcconnell, line 176: [Important] Nothing ever clears `feed_key`: `DeleteFeed` closes the feed without touching the mirror and no instruction writes a `StakeMirror`, so once a staked feed is deleted its stake returns `StakeAlreadyBacksFeed` forever and the builder must post a second deposit. Clear it in `DeleteFeed`, or refuse to delete a staked feed until the lifecycle work lands.

armcconnell (changes requested):
Checked against edge-feed-spec GLOSSARY.md v1.3.0 (blob 83f4e47): conforms.
See inline comments.

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
armcconnell, line 176: [Important] Nothing clears `feed_key`: `DeleteFeed` closes the feed without touching the mirror, and no instruction writes a `StakeMirror`, so a deleted staked feed leaves its stake returning `StakeAlreadyBacksFeed` forever and the builder must post a second deposit. Clear it in `DeleteFeed`, or refuse to delete a staked feed until the lifecycle work lands.

smartcontract/programs/doublezero-serviceability/src/state/feed.rs
```diff
@@ -30,6 +63,40 @@ pub struct Feed {
     pub name: String,              // 4 + len
     pub exchange: Pubkey,          // 32 (PDA seed, immutable) - the metro this feed serves
     pub groups: Vec<Pubkey>,       // 4 + 32*len - multicast groups joinable in this metro
+
+    // RFC-28 fields. Everything below is absent from feeds created before RFC-28, so every one of
+    // them decodes to a default on a short account and is written back on the next update.
+    /// The builder that deployed this feed and posted its stake, zero for a catalog feed with no
+    /// builder. This is the key the `StakeMirror` is written against.
```
armcconnell: [Minor] The mirror is written against `stake_ref`, not `builder`; `get_stake_mirror_pda` takes the stake.

armcconnell (changes requested):
Checked against edge-feed-spec GLOSSARY.md v1.3.0 (blob 83f4e47): conforms.
- [Minor] PR description — still describes the superseded builder-keyed mirror: one per builder, no `feed_key`, no `StakeAlreadyBacksFeed` (121). The CHANGELOG entry is current.

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
bgm-malbeclabs: Real bug, thanks. Fixed in 7aa601bbe.

I took the second option: `DeleteFeed` refuses a feed carrying a builder, new error `StakedFeedCannotBeDeleted` (122). Clearing `feed_key` instead would have kept delete working, but deleting a staked feed also bypasses RFC-28's thirty-day notice and strands its subscribers, so the delete is the thing that is wrong rather than the bookkeeping. Retirement is the path out and releases the stake with it, both in D2.

I also corrected the stale claim in the same function's comment that deleting a feed is always safe because feeds are not reference-counted. That holds for a catalog feed and not for a staked one.

Test `test_staked_feed_cannot_be_deleted`: create a staked feed, delete, assert 122, and confirm the feed and its claim survive.

smartcontract/programs/doublezero-serviceability/src/state/feed.rs
bgm-malbeclabs: Correct, that comment was left over from when the mirror was keyed on the builder. Rewritten in 7aa601bbe to say the mirror is keyed on `stake_ref` so that one deposit backs one feed.

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
bgm-malbeclabs: Same thread as the repost below. Fixed in 7aa601bbe: `DeleteFeed` refuses a feed carrying a builder (`StakedFeedCannotBeDeleted`, 122) rather than clearing the claim, because deleting a staked feed also bypasses the thirty-day notice and strands its subscribers.

bgm-malbeclabs:
Ready for another look. Everything raised is either fixed or answered, and I have resolved the threads that were fixed.

Since the last review:

- **`feed_key` was never cleared** (7aa601bbe). `DeleteFeed` now refuses a feed carrying a builder, `StakedFeedCannotBeDeleted` (122), rather than clearing the claim: deleting a staked feed also bypasses RFC-28's thirty-day notice and strands its subscribers, so the delete is the wrong operation rather than the bookkeeping being wrong. Also corrected the stale comment in that function claiming deletion is always safe because feeds are not reference-counted, which holds for a catalog feed and not a staked one.
- **`Feed::builder` doc** said the mirror is written against it; the mirror is keyed on `stake_ref`. Rewritten.
- Earlier: the authorization iterator filters by key rather than pointer identity, the decoded mirror is validated before its tier is trusted, and two unit tests assert specific errors.

Two threads left open on purpose, both mine to argue rather than fixes I skipped:

- `init_test_with_accounts` in `test_helpers.rs` — it sits beside `init_test` and mirrors it exactly. Splitting the pair is churn. Happy to move it if you disagree.
- `StakeMirror.bump_seed` — kept because every account in this program carries one whether or not it signs a CPI, and diverging costs more than the byte saves.

Rebased onto main after #4216 landed. Full serviceability suite passes, `make rust-lint` clean, and the PR is mergeable.

smartcontract/programs/doublezero-serviceability/src/processors/feed/create.rs
```diff
@@ -61,14 +101,22 @@ pub fn process_create_feed(
     let globalstate = GlobalState::try_from(globalstate_account)?;
     authorize(
         program_id,
-        accounts_iter,
+        &mut authorize_iter,
         payer_account.key,
         &globalstate,
         permission_flags::FEED_AUTHORITY | permission_flags::FOUNDATION,
```
martinsander00, line 107: `CreateFeed` still requires `FEED_AUTHORITY` or `FOUNDATION`, and `owner` is the admin who pays. RFC-28 says the builder deploys the feed. We should pen an edge-builder issue and add it to the delivery plan. I think the end state should be that a builder should have the stake and when deploying the feed it should immediately become available.

martinsander00 (approved):

armcconnell (approved):
Checked against edge-feed-spec GLOSSARY.md v1.3.0 (blob 83f4e47): conforms.
- [Minor] PR description — stale: `StakeMirror` is one per stake, not one per builder, and there are four new errors (119-122), not two. The CHANGELOG entry is current.

PR #4283, opened 2026-09-06
feed: admit subscribers only to a publishing feed

bgm-malbeclabs:
Stacked on #4282, which adds `Feed.status`. Review that one first; this diff is the three files above the base.

## Summary

- A feed admits a subscriber only while its status is `Active`. Checked at both places a seat is spent: `SubscribeFeed` and `CreateSubscribeUser`.
- Retiring, halting, or slashing a feed now takes effect on the subscriber side by itself. No walk over affected access passes, no oracle sweep.
- New error `FeedNotActive` (122).

## Why this is small

The delivery plan lists seat termination as out of scope because it "needs a walk over affected passes or an oracle sweep, and it is a whole epic." It does not. Both admission paths already decode the `Feed` to check metro coverage, so the status check is a few lines in a function that has the feed in hand. Nothing has to find the passes, because the passes ask on their way in.

**The check is not in `check_feed_metro_coverage`, even though that is the one shared function.** `UnsubscribeFeed` runs through it too. Gating there would leave a user holding a seat on a retired feed with no way to release it, which is worse than the problem being fixed. So the rule lives in `require_feed_admits` and is called from the two places that spend a seat. Two call sites for one rule is the cost of not trapping users; the doc comment says so, because the obvious refactor is to move it down.

**Pending feeds can still be sold, just not connected to.** A staked feed is `Pending` until the attestor clears it. Selling a seat ahead of that is legitimate, so `SetAccessPassFeeds` is unchanged. Refusing to sell a seat on a `Retired` feed is a separate policy call and belongs to the oracle.

**Live subscribers are not evicted.** The check runs on the way in, so anyone already connected keeps the connection until they disconnect. That is a real limit, not an oversight. Evicting live users is a much narrower problem than walking every pass and can wait for the lifecycle work.

## What this cannot test yet

There is no instruction that halts or retires a feed, so the only non-`Active` status reachable today is `Pending`, via a staked feed. The test uses that. `Halted` and `Retired` coverage belongs with D1 and D2 when they add the transitions.

The other untestable invariant: unsubscribing from a non-`Active` feed must keep working. Reaching that state needs a feed that goes non-`Active` after a user joins, which no instruction can do yet. The doc comment carries the reasoning until D1 lands.

## Testing Verification

- `test_non_active_feed_admits_no_subscriber`: creates a staked feed (lands `Pending`), provisions a paid seat on it with the correct metro and group, and confirms `CreateSubscribeUser` is refused with 122. Asserts `current_users` stayed at 0, so a refused join does not burn the seat.
- The first version of this change gated only `SubscribeFeed` and the test caught it: `CreateSubscribeUser` reaches the seat through `enforce_feed_metro_gate`, a second path. That is why the rule is shared rather than inline.
- The six existing metro-gate tests still pass unchanged, so an `Active` feed is unaffected on both paths.
- Full `doublezero-serviceability` suite: 53 test binaries, no failures. `make rust-lint` clean.
- Pre-existing and unrelated: `doublezero-daemon-cli`'s `ip_proof::tests::probe_skips_a_local_address_that_cannot_reach_the_destination` fails on this machine, verified identical on a clean tree.

## Size

About 30 non-test lines.


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
...
+  - A feed admits a subscriber only while its status is `Active`. `SubscribeFeed` and `CreateSubscribeUser` both check it where the seat is spent, so a feed that is pending conformance, halted, or retired stops taking subscribers the moment its status changes, and retiring one needs no sweep over the access passes already holding a seat for it. The check is deliberately not in the shared coverage path, which `UnsubscribeFeed` also runs: gating there would leave a user holding a seat on a retired feed with no way to release it. New error `FeedNotActive` (122). Live subscribers are unaffected until they disconnect; evicting them is separate work.
```
armcconnell: [Minor][docs] `sweep` is banned; name the operation — "walk over the access passes".

smartcontract/programs/doublezero-serviceability/src/processors/feed/mod.rs
```diff
@@ -91,6 +94,29 @@ pub fn enforce_feed_metro_gate(
     )?;
     // feed_account is guaranteed Some here (check returns Err otherwise).
     let feed_account = feed_account.ok_or(DoubleZeroError::FeedAccountRequired)?;
+    require_feed_admits(feed_account.key, &Feed::try_from(feed_account)?)?;
```
armcconnell, line 97: [Minor] Decodes the feed a second time; `check_feed_metro_coverage` just deserialized the same account, so return the `Feed` from it and pass that in.

smartcontract/programs/doublezero-serviceability/src/processors/feed/mod.rs
```diff
@@ -91,6 +94,29 @@ pub fn enforce_feed_metro_gate(
...
     accesspass.try_add_feed_user(feed_account.key)?;
     Ok(*feed_account.key)
 }
+
+/// Reject a feed that is not publishing, before a seat is spent on it.
+///
+/// This is the whole enforcement of feed lifecycle on the subscriber side. A feed that is pending
+/// conformance, halted by its builder, or retired stops admitting subscribers the moment its
+/// status changes, so retirement and slashing need no sweep over the access passes that already
```
armcconnell, line 106: [Minor][docs] `sweep` -> `walk`.

armcconnell (approved):
Checked against edge-feed-spec GLOSSARY.md v1.3.0 (blob 83f4e47): does not conform; see findings.
See inline comments.

PR #4285, opened 2026-09-07
builder-stake: size the deposit against the committed rate

bgm-malbeclabs:
Implements B2. Stacked on #4284; review that first, this diff is the six files above the base.

## Summary

- `ProgramConfig` gains a tier table: three 2Z amounts, one per RFC-28 rate tier, set by the admin through `ConfigureProgram`.
- `InitializeBuilderStake` looks up the amount for the rate the builder commits to and pins it on the stake as `required_2z_amount`.
- `BuilderStake::is_funded()` answers whether the stake holds what it owes.
- A malformed table is refused: a zero amount, or a higher tier that costs less than a lower one.

**A table, not a price feed.** RFC-28 quotes about \$100k / \$200k / \$500k but says the deposit is "denominated in 2Z, fixed at the price prevailing when the tier is set". So an admin writes three numbers and no oracle is involved.

**The rate ceilings are compiled in, not configured.** 1 Gbps and 5 Gbps have to agree with `StakeTier::max_rate_bits_per_sec` in the serviceability program on the DZ ledger. If those two drift, a builder funds one tier here and its feed is checked against a different one there. Two configurable copies is exactly how that drift happens.

**The requirement is pinned per stake, not read live.** Repricing a tier does not move what an existing stake owes, so a builder who fully funded a stake cannot wake up under-funded because the admin changed a number. This is the stronger reading of "fixed at the price prevailing when the tier is set", and it is one field. There is a test for it.

**`Deposit` still accepts any positive amount, which is a departure from how issue #9 was written.** Requiring the full amount in one transfer would make "a stake exists implies it is funded" true, which is tempting. Against it: a builder may reasonably fund from more than one source or split a large transfer, and nothing bad happens with a short stake, because only a funded stake gets mirrored to the DZ ledger and so a short one backs no feed. The guard that actually matters is `Withdraw` in B3, which must not drop a stake below its requirement regardless. Over-funding is allowed too, since RFC-28 lets a builder withdraw the excess after the hold.

**An unset table sizes nothing rather than sizing everything at zero.** `required_2z_amount` returns `None` for a zero entry, so a program unpaused before it was configured takes no stake instead of handing out free ones.

## Struct change

`BuilderStake` grew `required_2z_amount`, so its size assertion moves from 136 to 144. Nothing is deployed, so there is no migration. I also dropped the `ProgramConfig` size assertion: it is allocated at 10kb, so a new setting grows into slack, which is the same reasoning `revenue-distribution` gives.

## Testing Verification

`make test-sbf`, 16 tests (3 unit, 12 integration, plus the id check):

- Every tier boundary, at and either side of each ceiling: 1 Gbps and one unit over, 5 Gbps and one unit over, and `u64::MAX`.
- A stake starts unfunded, two part-deposits reach the requirement with the last unit tipping it over, and over-funding is accepted.
- An unpaused program with no tier table refuses to take a stake at all.
- Three malformed tables are refused (a hole at the first tier, 5 Gbps cheaper than 1 Gbps, unmetered cheaper than 5 Gbps), and a well-formed one lands and reads back.
- Raising every tier tenfold leaves an existing funded stake funded and its requirement unchanged, while the next stake posted pays the new price.
- The seven B1 tests still pass unchanged.

## Size

About 90 non-test lines.


solana/programs/builder-stake/src/processor.rs
```diff
@@ -270,14 +314,16 @@ fn try_initialize_builder_stake(
     builder_stake.builder = *builder_info.key;
     builder_stake.stake_index = stake_index;
     builder_stake.committed_rate_bits_per_sec = committed_rate_bits_per_sec;
+    builder_stake.required_2z_amount = required_2z_amount;
```
armcconnell, line 319: [Important] The requirement is pinned at creation, before any deposit, and `InitializeBuilderStake` is permissionless and costs only rent, so a builder can pre-create stakes in bulk to lock in today's tier amount and fund them years later at a price a repricing was meant to replace. Pin at the deposit that first funds the stake, or re-read the table while the stake is short.

solana/programs/builder-stake/src/processor.rs
```diff
@@ -21,14 +21,14 @@ use spl_token_interface::instruction as token_instruction;
...
 };
 
-// A change to either size means every deployed account of that type has to be migrated, so make
-// the change deliberate rather than incidental.
-const _: () = assert!(size_of::<BuilderStake>() == 136);
-const _: () = assert!(size_of::<ProgramConfig>() == 176);
+// A change to this size means every deployed BuilderStake has to be migrated, so make the change
+// deliberate rather than incidental. The program config needs no such check: it is allocated at
+// 10kb, so a new setting grows into slack rather than needing a realloc.
+const _: () = assert!(size_of::<BuilderStake>() == 144);
```
armcconnell, line 32: [Minor] `ProgramConfig` is still 176 bytes — `TierParameters` came out of the storage gap, `StorageGap<4>` to `<3>` — so the dropped assertion still holds and was the tripwire for a setting that overruns the gap.

solana/programs/builder-stake/tests/tier_test.rs
```diff
@@ -0,0 +1,211 @@
...
+
+/// A stake is funded once it holds its requirement, and deposits accumulate to get there.
+#[tokio::test]
+async fn test_stake_is_funded_once_it_holds_the_requirement() {
+    let mut t = common::start_test().await;
+    t.initialize_and_unpause().await;
+
+    let builder = t.builder.insecure_clone();
+    let builder_key = builder.pubkey();
+    let source = t.builder_2z_key;
```
armcconnell: [Minor][docs] Bare `source` is banned; use `source_token_account`.

armcconnell (changes requested):
Checked against edge-feed-spec GLOSSARY.md v1.3.0 (blob 83f4e47): does not conform; see findings.
See inline comments.

solana/programs/builder-stake/src/processor.rs
bgm-malbeclabs: Good catch, and you are right that rent is not a barrier. Fixed in 5393c815c.

The requirement now follows the tier table while a stake is short and freezes the moment it is funded. That keeps what the pin was for, a builder who paid in full cannot be made short by a later repricing, without letting anyone bank a price by opening empty stakes. `try_deposit` re-reads the table when `!is_funded()`, before the transfer.

I took "re-read while short" over "pin at first funding deposit" because the latter still lets a builder pin today's price with a one-unit deposit.

New test `test_an_unfunded_stake_pays_the_current_price`: open a stake, raise every tier tenfold, and the old amount no longer funds it while the new one does. `test_repricing_a_tier_leaves_existing_stakes_alone` still covers the funded side.

solana/programs/builder-stake/src/processor.rs
bgm-malbeclabs: You are right, and I had the reasoning wrong. I dropped it because the account is allocated at 10kb so it never needs a realloc, but that is not what the assertion was catching. It is the tripwire for a setting that overruns the storage gap instead of coming out of it, which is exactly what `TierParameters` did correctly here (`StorageGap<4>` to `<3>`, still 176).

Restored in 5393c815c, with a comment saying what it is actually for.

solana/programs/builder-stake/tests/tier_test.rs
bgm-malbeclabs: Renamed to `source_token_account` in 5393c815c, in both the test and the `deposit` builder's parameter.

solana/programs/builder-stake/tests/tier_test.rs
```diff
@@ -0,0 +1,211 @@
...
+        .await
+        .unwrap();
+
+    let builder = t.builder.insecure_clone();
+    t.send(
+        common::initialize_builder_stake(&builder.pubkey(), 0, ONE_GBPS),
+        &[&builder],
+    )
+    .await
+    .expect_err("no tier table means no deposit can be sized");
```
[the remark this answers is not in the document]
bgm-malbeclabs: Taken. The three tier failure tests now assert the exact `InstructionError` at index 0 through a new `assert_instruction_error` helper: `InvalidInstructionData` for a malformed table, `InvalidAccountData` for an unset one. Fixed in 5393c815c.

armcconnell (approved):
Checked against edge-feed-spec GLOSSARY.md v1.3.0 (blob 83f4e47): conforms.
No findings.

PR #4297, opened 2026-09-09
sentinel: read v1 access-request transactions

martinsander00:
Resolves: https://github.com/malbeclabs/infra/issues/2567

## Summary

- Sentinel asks the Solana node for version 1 when it fetches a landed access-request transaction.
- The lookup reads the JsonParsed instruction list so a v1 body still yields the access-request account.

## Testing

- One unit test feeds a version 1 JsonParsed body whose request-access instruction names a known account at index 2, and sentinel returns that account.
- Another unit test feeds a version 1 body with a grant-access instruction and a foreign program instruction, and sentinel returns no accounts.


bgm-malbeclabs:
Reviewed this alongside #4298 and malbeclabs/doublezero-shreds#719. The rewrite itself is good, but I think it landed on a function that no longer runs.

### `get_access_requests_from_signature` has no caller

Three references in the repo, none of them a call:

- `offchain/crates/sentinel/src/client/solana.rs:66` — trait declaration
- `:111` — trait impl, forwards to the inherent method
- `:217` — the inherent method this PR rewrites

The running sentinel takes requests from `get_access_requests()` (`solana.rs:279`), which is `getProgramAccounts` with a discriminator memcmp filter. The poller calls that one (`sentinel/poller.rs:80`). An account fetch carries no transaction version, so it cannot return -32015. `get_transaction_with_config` appears exactly once in the crate, inside the dead function, and `git log -S get_access_requests_from_signature` shows the signature path died in be0e5460f, "sentinel: entirely remove websocket support".

So the CHANGELOG line "so a v1 request no longer fails with error -32015" does not describe a path that runs today. Could you share the log lines behind infra#2567? If that error was really seen, it came from somewhere else.

Suggestion: delete `get_access_requests_from_signature`, its trait method, and `Error::MissingAccountKeys`, then move the JsonParsed rewrite to the place that will actually break.

### Where the same root cause does bite

`offchain/crates/solana-cli/src/command/shreds/payments.rs:169` reads other people's escrow-funding transactions off Solana L1 with `Base64` and `max_supported_transaction_version: Some(0)`. A v1 transaction from any payer makes the RPC return -32015, and `?` aborts the whole `shreds payments` listing.

Raising the cap alone will not fix it. Line 175 calls `.decode()`, which goes through `VersionedTransaction`, and `VersionedMessage` in solana-message 3.1.0 has only `Legacy` and `V0` arms (`src/versions/mod.rs:56-59`). A v1 body returns `None`, and the `continue` then drops every payment event in that transaction without a word. That site needs exactly the rewrite in this PR.

Four more call sites still ask for `Some(0)`:

- `smartcontract/sdk/rs/src/client.rs:260`
- `offchain/crates/solana-client-tools/src/payer.rs:246`
- `offchain/crates/solana-cli/src/command/shreds/publisher_rewards/status.rs:564`
- `doublezero-shreds` `crates/onchain-admin/src/command/show/escrow_history.rs:225`

I checked that the node puts no ceiling on the requested value: `maxSupportedTransactionVersion: 9` is accepted without complaint. So one shared constant in a common crate would settle this once, instead of hand-editing an integer per PR.

### Two notes on the new code, if it stays

1. A short or unreadable Passport instruction returns `Err` instead of `continue` (`solana.rs:414` and `:418`). A different discriminator continues, so a transaction carrying one good `request_access` plus any junk Passport instruction yields no PDAs at all. `should_retry` (`error.rs:125`) matches only `Error::RpcClient`, so `InstructionInvalid` is terminal for that signature. `continue` on both is consistent with the discriminator mismatch and cannot lose a valid request.

2. The tests do not exercise v1. `parsed_transaction` sets `version` and then discards it by returning `encoded.transaction`, and `access_request_pdas` never reads a version. `transactionConfig` is not a field of `UiParsedMessage` (`solana-transaction-status-client-types`, `src/lib.rs:136-142`), and there is no `deny_unknown_fields`, so serde drops it. The tests pass identically with `version: 0`. They do prove the parser reads a JsonParsed instruction list, which is worth having. They prove nothing about a real v1 response. That matters because `EncodedTransaction` is untagged: a v1 shape change fails inside `get_transaction_with_config` as "data did not match any variant", before any of this code runs. A fixture captured from a real v1 `getTransaction` would close it.

### Credit where it is due

The Base64 to JsonParsed switch fixes a second bug the description does not claim. The old path read `static_account_keys()` only, so a v0 transaction that named the AccessRequest PDA through an address-lookup table resolved to the wrong key, or fell out as `MissingAccountKeys`. JsonParsed makes the node resolve lookups first. Worth saying in the CHANGELOG. The old `ix.data[..8]` also panicked on a short Passport instruction, and the guarded read fixes that.

40 tests pass on the branch, and clippy is clean with `-Dwarnings`.

martinsander00:
Sentinel does not fetch access-request transactions. The poller calls `get_access_requests()`, which is `getProgramAccounts`. A v1 request still writes the AccessRequest account, and sentinel still sees it. `get_access_requests_from_signature` has no caller after websocket support was removed.

Closing this PR. The live v1 read is the subscription API collect check: malbeclabs/doublezero-shreds#719.

PR #4298, opened 2026-09-09
validator-debt: read v1 blocks

martinsander00:
Resolves: https://github.com/malbeclabs/infra/issues/2565

## Summary

- The debt job asks the Solana node for version 1 when it loads a leader slot.
- Signature and fee math stay the same.


bgm-malbeclabs:
I do not think this one-line change does anything.

The config it edits sets `transaction_details: Some(TransactionDetails::Signatures)`. On that path the node never reads `max_supported_transaction_version`. In `solana-transaction-status` 3.1.14, `ConfirmedBlock::encode_with_options` passes the field to `validate_version` only in the `Full` and `Accounts` arms (`src/lib.rs:338-374`). The `Signatures` arm just maps each transaction to its signature string.

I confirmed it against mainnet. `getBlock` on slot 445684240 with `transactionDetails: "signatures"` and **no** `maxSupportedTransactionVersion` at all returns a normal result with rewards. That block holds v0 transactions, so the parameter is plainly ignored.

So the debt job cannot have been failing with -32015 here, and version 1 changes nothing. The change is harmless, but the CHANGELOG claims a fix that did not happen:

> The debt job asks the Solana node for version 1 when it loads a leader slot, so one v1 transaction no longer refuses the whole block.

Could you share the log lines behind infra#2565? Either the failure came from a different call, or this PR should close.

Same note as on #4297: the place that will really break is `offchain/crates/solana-cli/src/command/shreds/payments.rs:169`. It reads third-party transactions off Solana L1 with `Base64` and `Some(0)`, and its `.decode()` would still drop v1 bodies even after a cap bump, because `VersionedMessage` in solana-message 3.1.0 has only `Legacy` and `V0` arms.

martinsander00:
The debt job asks for `transactionDetails: "signatures"`. The node ignores `maxSupportedTransactionVersion` on that path, so raising it from 0 to 1 changes nothing.

Closing this PR. The live v1 read is the subscription API collect check: malbeclabs/doublezero-shreds#719.

PR #4302, opened 2026-09-09
serviceability: authorize only new multicast roles

martinsander00:
## Summary

- Multicast updates check allowlists only when a user gains a publisher or subscriber role.
- Existing EdgeSeat feed subscriptions remain intact when users add allowed publishing.

## Testing
- New tests recreate a publisher addition while the direct subscriber allowlist omits an existing feed role.

This change addresses [infra#2596](https://github.com/malbeclabs/infra/issues/2596).


nikw9944 (approved):
Logic looks right: skipping the allowlist for an already-held role is state-neutral, and a genuinely new unallowlisted role is still rejected. One stale comment on the authorization model.

- `smartcontract/programs/doublezero-serviceability/src/processors/multicastgroup/subscribe.rs:333` — The comments describing the multicast authorization model now contradict the code beneath them. This loop comment says each group "must be on the pass's allowlist", and the `check_mgroup_allowlists` doc (line 63-68) says a caller runs the allowlist check or the EdgeSeat feed gate, "never neither".

PR #4305, opened 2026-09-09
serviceability/sdk: track the access pass used by each user

martinsander00:
Resolves: [infra#2594](https://github.com/malbeclabs/infra/issues/2594)

## Summary
- New user accounts record the access pass used during creation.
- Feed changes, user deletion, and disconnect use the recorded access pass.
- Operators must delete legacy users individually before deleting their tenant.

## Testing Verification
- SDK tests verify that legacy feed changes list the available access passes.
- Feed program tests verify subscription and removal behavior.


smartcontract/sdk/rs/src/commands/accesspass/get.rs
```diff
@@ -40,10 +41,92 @@ impl GetAccessPassCommand {
...
+                "Recorded access pass {} for user {user_pk} has the wrong account type",
+                user.accesspass_pk
+            ),
+            Err(err) => Err(err).wrap_err_with(|| {
+                format!(
+                    "Failed to load recorded access pass {} for user {user_pk}",
+                    user.accesspass_pk
+                )
+            }),
+        };
```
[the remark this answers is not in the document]
martinsander00: this fix is retarded

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file.
...
 - Serviceability
   - `UpdateMulticastGroupRoles` authorizes only roles a user gains. An existing feed subscription no longer needs a direct subscriber allowlist entry when the user adds publishing. (malbeclabs/infra#2596)
+  - New user accounts record the selected access-pass address, and deletion rejects another pass while preserving legacy address validation.
```
nikw9944, line 17: Update smartcontract/programs/doublezero-serviceability/README.md too?

smartcontract/programs/doublezero-serviceability/src/processors/user/delete.rs
```diff
@@ -141,15 +141,21 @@ pub fn process_delete_user(
...
+        let (accesspass_pda, _) = get_accesspass_pda(program_id, &user.client_ip, &user.owner);
+        let (accesspass_dynamic_pda, _) =
+            get_accesspass_pda(program_id, &Ipv4Addr::UNSPECIFIED, &user.owner);
+        assert!(
+            accesspass_account.key == &accesspass_pda
+                || accesspass_account.key == &accesspass_dynamic_pda,
+            "Invalid AccessPass PDA",
+        );
+    } else {
+        assert_eq!(
```
nikw9944, line 154: **Blocking** — Delete now demands the exact access pass recorded on the user, but the feed-seat paths still accept either the exact-IP or the 0.0.0.0 pass, so a seat taken on one pass account is released against the other and stays consumed until the feed reports `FeedSeatFull`. `multicastgroup/feed.rs:111` validates the pass against `(user.client_ip, user.owner)` only, with no reference to `user.accesspass_pk`.

smartcontract/sdk/rs/src/commands/accesspass/get.rs
```diff
@@ -40,10 +41,92 @@ impl GetAccessPassCommand {
...
+                        "  {pk}: {:?}, client IP {}, {} connections",
+                        accesspass.accesspass_type,
+                        accesspass.client_ip,
+                        accesspass.connection_count
+                    )
+                })
+                .collect::<Vec<_>>()
+                .join("\n");
+            eyre::bail!(
+                "Legacy user {user_pk} matches multiple access passes:\n{choices}\nRetry with --access-pass <ADDRESS>."
```
nikw9944, line 92: **Blocking** — A legacy user whose payer holds both a 0.0.0.0 pass and an exact-IP pass now fails to delete instead of resolving to the 0.0.0.0 pass, which is what `CreateUserCommand` used at connect. `DeleteUserCommand::new` callers — `tenant delete` in both the CLI and the SDK — cannot supply `--access-pass`, so that sweep aborts with no way to complete it.

crates/doublezero-daemon-cli/src/disconnect.rs
```diff
@@ -167,7 +170,7 @@ impl Disconnect {
 
             spinner.inc(1);
             writeln!(out, "⚡  Removing account: {pubkey}")?;
-            match ledger.delete_user(*pubkey) {
+            match ledger.delete_user(*pubkey, self.access_pass) {
```
nikw9944: **Minor** — `delete_users` forwards the single `--access-pass` value to every user in the sweep, so a payer with more than one user cannot clear them in one run once any user needs the flag.

nikw9944 (changes requested):
Re-checked at c46b4ff; all three findings remain open. The recorded-pass rule is enforced only in DeleteUser, so a feed seat can be taken against one access pass and released against another. And a legacy user matching two passes now fails to delete rather than resolving as before, with no `--access-pass` escape hatch on the `tenant delete` path.

CHANGELOG.md
martinsander00: https://github.com/malbeclabs/doublezero/commit/6ce98dcb2972b7bc04fb5aee474b2bdc68d227a0

smartcontract/programs/doublezero-serviceability/src/processors/user/delete.rs
martinsander00: https://github.com/malbeclabs/doublezero/commit/6ce98dcb2972b7bc04fb5aee474b2bdc68d227a0

smartcontract/sdk/rs/src/commands/accesspass/get.rs
martinsander00: https://github.com/malbeclabs/doublezero/commit/6ce98dcb2972b7bc04fb5aee474b2bdc68d227a0

crates/doublezero-daemon-cli/src/disconnect.rs
martinsander00: https://github.com/malbeclabs/doublezero/commit/6ce98dcb2972b7bc04fb5aee474b2bdc68d227a0

smartcontract/sdk/rs/src/commands/user/delete.rs
```diff
@@ -35,6 +39,16 @@ impl DeleteUserCommand {
             .get_user()
             .map_err(|e| eyre::eyre!(e))?;
 
+        if user.accesspass_pk == Pubkey::default() && self.accesspass_pk.is_none() {
```
nikw9944: **Blocking** — Deleting any user created before this change now fails unless the caller passes `--access-pass`, even when exactly one access pass matches, so `doublezero disconnect` stops working for the whole existing fleet until each operator looks up an address. The new guard in `DeleteUserCommand::execute` returns `legacy_user_accesspass_selection_error` for every legacy user with no flag, bypassing the single-candidate branch in `resolve_user_accesspass`; `SubscribeFeedCommand` and `UnsubscribeFeedCommand` carry the same guard.

nikw9944 (changes requested):
Re-checked at 6ce98db: the feed-seat binding and the disconnect flag scoping are both fixed and dropped, and the README now documents the field. The remaining blocker moved in the wrong direction — `--access-pass` is now mandatory for every legacy user, not just an ambiguous one, so delete and the feed commands fail for the entire existing fleet where a single matching pass used to resolve on its own.

smartcontract/sdk/rs/src/commands/user/delete.rs
martinsander00: https://github.com/malbeclabs/doublezero/commit/f9ceed71ddbbc33af74a238374737e1e413423d7

nikw9944 (approved):
Re-checked at f9ceed7, CI fully green. The blanket `--access-pass` requirement is gone — `legacy_user_accesspass_selection_error` is deleted and `resolve_user_accesspass` auto-resolves a single legacy candidate again, with a test pinning it. The feed-seat binding, disconnect flag scoping and README rows are all still in place. No findings remain.

PR #4308, opened 2026-09-10
geoprobe-target: bound offset replay with a per-key slot floor

### LINKED ISSUE #4286: geoprobe-target: replayed offsets still reach location_offsets (no MeasurementSlot check)
Follow-up from malbeclabs/doublezero#4271, which closed the forgery hole in the geoprobe measurement chain. That PR added RFC-16's `MeasurementSlot` replay check to **geoprobe-agent** only. **geoprobe-target** — the process that writes the `location_offsets` table lake's public geolocation explorer aggregates — has no slot comparison at all.

### 1. No replay bound on the target's ingest path

`controlplane/telemetry/cmd/geoprobe-target/main.go` `handleOffset` now drops offsets whose signature chain does not verify, which is the important half. But a signature stays valid forever, so a **validly signed offset captured off the wire and replayed** passes the gate unchanged and is written to ClickHouse. `git grep MeasurementSlot` in that file returns only display code.

This is the same attack RFC-16's `MeasurementSlot` field exists to stop, against the surface that feeds the public page. It is narrower than the forgery hole malbeclabs/doublezero#4271 closed — it needs an on-path capture rather than just a UDP socket — which is why it was deliberately left out of that PR rather than growing it.

The agent's implementation is the reference to follow: `offsetSlotFresh`, `slotReference` and the `maxOffsetSlotLag` / `maxOffsetSlotLead` / `maxSlotReferenceAge` constants in `controlplane/telemetry/cmd/geoprobe-agent/main.go`. Note that geoprobe-target does not currently hold a ledger RPC client, so wiring a slot source into it is the non-trivial part of this change and should be scoped before implementing. Fail closed on an unusable slot, as the agent does.

One sizing note carried over from review of that PR: the agent sets `maxSlotReferenceAge = 2 * geoprobe.SlotCacheTTL` (10m) while `maxOffsetSlotLead` is 5m. Because a legitimate offset's slot is at most the true slot, a reference staler than the lead allowance already rejects fresh offsets while still admitting replays near the frozen slot — so the two bounds interact, and a target-side implementation should not copy the 2x multiplier without deciding it is what it wants.

### 2. `MinCache.Update` still resets `best`'s clock on an equal RTT

`controlplane/telemetry/internal/geoprobe/mincache.go:120` reads `if rttNs <= c.best.rttNs` and, on that branch, replaces `best` — which also resets its `receivedAt`. This is the identical defect malbeclabs/doublezero#4271 fixed in the agent's own `offsetCache.Put`, where the comment now reads: "An equal-RTT offset must not replace best, because replacing also resets best's receivedAt clock — a replayed offset would otherwise hold best forever."

`MinCache` was not touched by that PR. In geoprobe-target the cache drives console and JSON output rather than anything persisted, so the impact is lower than it was in the agent — but the code is wrong in the same way and should match its sibling.

### 3. `signature_valid` is now a constant `true`

Small cleanup in the same function. Since `handleOffset` returns early on a failed chain, `verifyError` is always nil by the time the ClickHouse row is built, so the `sigErrStr` computation is dead and `signature_valid` is written `true` unconditionally.

Two things worth deciding rather than just deleting:

- With `-verify-signatures=false` (the flag defaults to `true`, and infra's ansible sets `verify_signatures: true`, so this is not the deployed configuration), nothing is verified but rows are still written with `signature_valid = true`. That is a mislabel, and it matters more now that the column is supposed to carry a real assertion. Recording `false` or a distinct "unverified" state would be more honest than asserting a check that did not run.
- Historical rows written before malbeclabs/doublezero#4271 include `signature_valid = false`, and lake's geolocation handlers do not filter on the column — `grep signature_valid` across lake's `api/` finds it only in test fixtures. So forged offsets already in `mainnet-beta.location_offsets` keep rendering on the public explorer even though the intake is now closed. Closing that is lake-side work, tracked separately; it is noted here so whoever picks this up knows the column's consumer does not currently read it.

## Context

All of the above was found while reviewing malbeclabs/doublezero#4271 (implemented by instance 413). None of it is a regression from that PR — items 1 and 2 predate it, and item 3 is a consequence of its fix. Read that PR's diff for the shape the fixes should follow.

nikw9944:
## Summary of Changes

* geoprobe-target bounds offset replay with a per-signing-key `MeasurementSlot` floor: an offset more than 2 minutes below the highest slot that key has proven is dropped, as are repeats once the floor has stood still for 30 minutes.
* A valid signature never expires, so a captured offset could previously be replayed into `location_offsets` — the table the public geolocation explorer aggregates — forever. malbeclabs/doublezero#4271 closed the forgery half of RFC-16's mitigation here; this closes the replay half.
* `signature_valid` no longer asserts a check that did not run, and `MinCache.Update` no longer restarts `best`'s expiry clock on an equal RTT (matching the agent's own cache).
* Fixes #4286

**The floor comes from the offset stream, not a ledger clock.** geoprobe-target holds no RPC connection by design — RFC-16 makes it example target software whose premise is verifying with no chain access. `MeasurementSlot` is signed, so the highest slot a key has proven is a lower bound on real time a replay can repeat but not advance. `CLAUDE.md` records the invariant.

**The floor also has a ceiling** (added in review): it will not advance further than elapsed wall time allows. Without one it is a ratchet, and a single anomalous slot — a probe pointed at the wrong ledger RPC needs no attacker — would wedge that key out of the table until restart.

**30m reads looser than the agent's 15m lag, but the steady-state window is ~7 minutes** — a healthy sender's floor advances every ~5. `maxFloorStall` only covers a probe riding out its own RPC outage, where a dropped measurement is lost permanently rather than deferred.

**This does not make the ingest path trustworthy.** `VerifyOffsetChain` checks the signature against the `AuthorityPubkey` in the same datagram, and this daemon has no allowlist of known probe keys, so any freshly generated keypair can still write attacker-chosen coordinates. Closing that needs the onchain geoprobe registry, which the lake indexer already ingests and this daemon deliberately cannot reach — target enforces freshness, consumer enforces identity. An issue covering both halves is being filed separately.

**Rejections are logged, not counted.** geoprobe-target has no prometheus metrics, so these `Warn` lines are the only signal that a sender stopped being ingested. Each carries a stable message plus separate `reason` (`slot_regressed` / `floor_stalled` / `slot_jumped`), `authority_pubkey`, `sender_pubkey`, `offset_slot`, `floor_slot` and `floor_age_seconds` fields. A counter would be a reasonable follow-up.

## Testing Verification

* Every behavior change has a test confirmed to fail with the production change reverted, then restored — not inferred.
* Adversarial cases: a forged offset stamped with slot 2^40 claiming a real geoprobe's `SenderPubkey` does not lock that geoprobe out; a sustained replay against a stalled floor stays rejected across repeated sweeps rather than outliving its entry and reseeding from itself.
* The ceiling is checked against honest traffic too: catch-up after a 25-minute sender RPC outage, a cluster sustaining 200ms slots, and a slot-cache refresh step arriving with no elapsed time are all accepted.
* `make e2e-test RUN=TestE2E_Geoprobe` was **not** run — it needs Docker and cEOS, unavailable here. Worth a reviewer's time before merge: this changes ingest acceptance, so a false rejection loses a measurement permanently rather than deferring it.


controlplane/telemetry/cmd/geoprobe-target/main.go
```diff
@@ -214,7 +262,92 @@ func (rl *rateLimiter) cleanup(ctx context.Context) {
...
+	entry, exists := f.entries[authority]
+	if !exists {
+		f.entries[authority] = &floorEntry{slot: slot, advancedAt: now, lastSeen: now}
+		return true, "", slot, 0
+	}
+	entry.lastSeen = now
+
+	age := now.Sub(entry.advancedAt)
+	switch {
+	case slot > entry.slot:
```
ben-dz, line 322: **Blocking** — A single anomalous high `MeasurementSlot` from a real sender removes it from `location_offsets` permanently: the floor advances with no ceiling, never moves down, and rejections refresh `lastSeen` so the entry is never swept. Every later genuine offer then hits `slot_regressed` (326) and returns before `floor_stalled` (328), so the branch meant as the escape hatch can never fire — only a process restart clears it. Reachable without an attacker: one cycle against the wrong `-ledger-rpc-url` stamps a Solana-L1-height slot. `advancedAt` is already tracked, so bounding the advance by elapsed real time rejects the bad datagram instead of absorbing it — the shape of `maxOffsetSlotLead` in geoprobe-agent, at no cost to replay defense. Make sure this is tolerant of solana block times that may drop to 200ms. This should be a sanity check against huge steps forward, rather than making sure the results are exact.

controlplane/telemetry/internal/geoprobe/mincache.go
```diff
@@ -117,15 +117,18 @@ func (c *MinCache[T]) Update(value T) UpdateInfo {
 		info.Result = UpdateBest
 		return info
 	}
-	if rttNs <= c.best.rttNs {
-		// New record low: reset best's clock, clear backup.
+	if rttNs < c.best.rttNs {
+		// New record low: reset best's clock, clear backup. An equal-RTT value
```
ben-dz, line 121: **Minor** — The comment says an equal-RTT value must not replace `best` because a repeat would push best's expiry out. It still does: an equal sample falls to the backup path, where `rttNs <= c.backup.rttNs` (136) holds, and is then promoted into `best` with its fresh `receivedAt` every TTL — one datagram replayed for 10h against a 1h `maxAge` is still `Best()`. Trim the claim; the slot floor is what bounds replay. Same wording in CHANGELOG.md:14.

ben-dz (commented):
A single anomalous high `MeasurementSlot` wedges a sender out of `location_offsets` permanently — the floor has no advance ceiling, and `slot_regressed` short-circuits before the `floor_stalled` escape hatch can fire, so only a process restart recovers. Reachable by one measurement cycle against the wrong ledger RPC, on hosts DoubleZero does not operate and which expose no metrics. Separately, the `mincache` comment claims a replay bound the change does not deliver — equal-RTT samples still pin `best` through the backup-promote cycle.

- `CHANGELOG.md:12` — The bullet announces a "per-sender `MeasurementSlot` floor", then three sentences later explains the floor is keyed by `AuthorityPubkey` and deliberately not by `SenderPubkey`. That keying is the entry's one security-relevant detail and the headline clause states it backwards; the PR title already says "per-key".

controlplane/telemetry/cmd/geoprobe-target/main.go
nikw9944: Fixed in 6955b6eb — the floor now refuses to advance further than elapsed wall time allows, so the anomalous datagram is rejected instead of ratcheted in and the sender keeps being ingested.

The allowance is `maxFloorAdvanceSlack + elapsed/minSlotDuration`, with `minSlotDuration = 200ms` as you asked — half the nominal rate, so a cluster running at full speed has headroom rather than tripping the check. `maxFloorAdvanceSlack` is 2 × `SlotCacheTTL` worth of slots, covering a jump with near-zero elapsed time when consecutive offers straddle the sender's slot-cache refresh. Deliberately loose: it is a sanity check against a huge step, not a clock.

Four tests: the wedge regression (non-vacuous — it fails if I remove the ceiling), honest catch-up after a 25-minute sender RPC outage, a sustained 200ms cluster, and the no-elapsed-time cache-refresh step.

One case the ceiling does not cover, now called out in the `accept` doc comment: seeding a never-seen key has no reference to check against, so an anomalous slot in a key's very first offer still sticks until restart. That needs the probe's *first* datagram after target startup to be the bad one; a persistent wrong-RPC misconfiguration stamps every offer consistently and never wedges anything.

controlplane/telemetry/internal/geoprobe/mincache.go
nikw9944: Right — trimmed in 6955b6eb. The comment now says only what the change does (an equal sample stops restarting `best`'s `receivedAt` on the fast path, matching the geoprobe agent's cache) and states outright that this is not a replay bound, since an equal repeat still reaches `best` through the backup promote. Same correction applied to the CHANGELOG bullet.

nikw9944:
All three points addressed in 6955b6eb, and the branch is rebased onto the current `nikw9944/adhoc-413` (25 commits; the one conflict was the shared Geolocation CHANGELOG list, resolved by keeping the base's bullets verbatim and prepending this PR's three).

**Blocking — the wedge.** The floor now has a ceiling: it refuses to advance further than elapsed wall time allows, so an anomalous `MeasurementSlot` is rejected (`slot_jumped`) rather than absorbed, and the sender keeps being ingested. Computed against a 200ms `minSlotDuration` per your note, plus a fixed slack for a cache-refresh step with no elapsed time. Four tests, including a non-vacuous regression for the wedge itself.

**Minor — the mincache comment.** Trimmed. It no longer claims a replay bound; the slot floor is named as what actually provides one. Same fix in the CHANGELOG bullet.

**Docs.** CHANGELOG headline now reads "per-key", matching the PR title and the `AuthorityPubkey` keying explained later in the entry. The entry also now mentions the ceiling.

Not changed, and worth a look before you approve: seeding a never-seen key has no reference to compare against, so an anomalous slot in a key's *first* offer after process start still sticks until restart. I documented it on `accept` rather than fixing it — closing it needs an absolute slot bound, which means a ledger clock that geoprobe-target deliberately does not have. Reaching it requires the very first datagram to be the bad one; a persistent wrong-RPC misconfiguration stamps every offer consistently and wedges nothing.

Verified locally: `go build ./controlplane/telemetry/...`, `go vet`, `gofmt`, and the full `geoprobe-target`, `geoprobe-agent` and `internal/geoprobe` test packages.

controlplane/telemetry/cmd/geoprobe-target/main.go
```diff
@@ -214,7 +281,99 @@ func (rl *rateLimiter) cleanup(ctx context.Context) {
...
+// nothing to check against, so it is also the one place an anomalous high slot
+// still sticks until restart.
+func (f *slotFloor) accept(authority [32]byte, slot uint64) (ok bool, reason string, floorSlot uint64, floorAge time.Duration) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	now := f.nowFunc()
+	entry, exists := f.entries[authority]
+	if !exists {
+		f.entries[authority] = &floorEntry{slot: slot, advancedAt: now, lastSeen: now}
```
ben-dz, line 377: **Blocking** — A probe misconfigured to the wrong ledger RPC is ingested fine, then disappears from `location_offsets` permanently the moment someone fixes it. Its first offer seeds the floor at the wrong cluster's height (seeding has no ceiling), so it is accepted; once repointed, every genuine offer is `slot_regressed` (352) — that branch still returns before `floor_stalled` (354) and rejections refresh `lastSeen`, so nothing recovers it short of a restart. Simulated: 2160/2160 offers rejected over the 18h after remediation. The two mitigations in the doc comment cancel out — a persistent misconfiguration is exactly what guarantees the bad-first-datagram precondition it calls unlikely.

ben-dz (changes requested):
Both round-1 findings verified fixed: the advance ceiling rejects the anomalous slot instead of ratcheting it in (1 rejection, not 2880, in simulation), and it holds at a sustained 200ms and even 100ms block time with no false rejections; the `mincache` comment, test doc and changelog now all state it is not a replay bound. One new blocking issue: the absorbing state survives through seeding, so a probe misconfigured to the wrong cluster is ingested until someone fixes it and then disappears from the table permanently.

controlplane/telemetry/cmd/geoprobe-target/main.go
nikw9944: Fixed in 13675b62 — a seed is now checked against the highest slot other live keys have proven, so the wrong-cluster first offer is rejected instead of becoming an unreachable floor. The probe is not ingested while misconfigured, and it seeds and keeps being ingested the moment someone repoints it.

Every sender stamps `MeasurementSlot` from the same ledger, so what the other keys have proven is the one reading of the current height available without a ledger connection — which geoprobe-target does not have, by design. The allowance is `maxFloorAdvance(referenceAge)`, the same ceiling the advance path uses, so a reference that is itself stale comes out more permissive rather than less. A key whose own floor has stalled past `maxFloorStall` stops counting, or one parked high slot would go on authorizing the next wrong-cluster seed.

Three tests, all non-vacuous (each fails with the check removed): your scenario end to end, including a seed plus 12 accepted offers over the hour after remediation; no false rejection for a genuine new sender against a reference 0, 4m and 30m stale; and a stalled key refusing to authorize the seed it would otherwise have vouched for.

One case this does not close, now stated in the `accept` doc comment: a target whose only sender is the misconfigured one has no reference, so that seed still sticks until restart. I do not think local state can close it — once the sender is repointed its genuine offers are indistinguishable from a replay, and letting them pull the floor back down is the replay the bound exists to stop. Happy to trade that if you prefer: a stall-based downward reseed would recover it, at roughly one accepted datagram per 30 minutes to a replayer holding two captures.

controlplane/telemetry/cmd/geoprobe-target/main.go
nikw9944: Same absorbing-state family as the seed case you raised in round 2, now closed in 13675b62 — the seed path has a ceiling too, so neither an anomalous advance nor an anomalous first offer can ratchet a floor out of reach.

The advance ceiling itself is unchanged; note it is a94b9eea after the rebase onto main, not the 6955b6eb I cited above.

nikw9944:
Rebased onto main (which now carries #4271) and closed the round-2 blocking finding.

**The seeding wedge** — 13675b62. A seed is now checked against the highest slot other live keys have proven, so a probe pointed at the wrong ledger RPC has its first offer rejected rather than absorbed into an unreachable floor. It is not ingested while misconfigured, and it seeds and keeps being ingested as soon as it is repointed. The reference comes from the offset stream, not a slot lookup, since geoprobe-target holds no ledger connection; the allowance reuses `maxFloorAdvance`, and a key whose own floor has stalled past `maxFloorStall` stops counting so one parked high slot cannot keep authorizing the next bad seed. Three non-vacuous tests.

**Not closed, deliberately:** a target whose only sender is the misconfigured one has no reference, so that seed still sticks until restart — now stated in the `accept` doc comment. After remediation such a sender's genuine offers are indistinguishable from a replay, and letting them pull the floor down is the replay this bound exists to stop. Details and the alternative in the thread.

**Not changed:** the `mincache` comment and changelog wording from round 1 already read as you verified them; nothing this round touched them.

The previous e2e backward-compatibility failure was branch drift, not a defect — the branch predated the commit that made `--accesspass-type` required on `user delete`, and release 0.42.0 auto-entered the dynamically discovered compat matrix. The rebase resolves it.

controlplane/telemetry/cmd/geoprobe-target/main.go
```diff
@@ -214,7 +310,109 @@ func (rl *rateLimiter) cleanup(ctx context.Context) {
...
+		}
+		f.entries[authority] = &floorEntry{slot: slot, advancedAt: now, lastSeen: now}
+		return true, "", slot, 0
+	}
+	entry.lastSeen = now
+
+	age := now.Sub(entry.advancedAt)
+	switch {
+	case slot > entry.slot:
+		if slot-entry.slot > maxFloorAdvance(age) {
```
ben-dz: **Blocking** — One replayed old datagram silently suppresses a real geoprobe for as long as the capture is old. Replayed before the genuine sender after a restart — or after any gap past `floorEntryTTL`, once the entry is swept — it seeds the floor low, and the advance ceiling here then rejects the sender's real slots as `slot_jumped` until elapsed time closes the gap: a 1h-old capture costs 50m, a 24h-old one 23h50m, a 7-day-old one ~7 days. Bounding the seed below the live reference is what closes it; line 374 fires only on `slot > ref`. The doc comment at 352-354 still says the live stream recovers "within minutes", true before the ceiling.

controlplane/telemetry/cmd/geoprobe-target/main.go
```diff
@@ -214,7 +310,109 @@ func (rl *rateLimiter) cleanup(ctx context.Context) {
...
+// genuine offers are indistinguishable from a replay, and letting them pull the
+// floor back down is the replay this bound exists to stop.
+func (f *slotFloor) accept(authority [32]byte, slot uint64) (ok bool, reason string, floorSlot uint64, floorAge time.Duration) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	now := f.nowFunc()
+	entry, exists := f.entries[authority]
+	if !exists {
+		if ref, refAge, hasRef := f.referenceSlot(authority, now); hasRef &&
```
ben-dz: **Blocking** — A single minted keypair parked at a low slot stops every geoprobe that is not already seeded from being ingested at all. `referenceSlot` takes the maximum over live keys, so when the attacker's key is the only live one — the state after any target restart, which it wins by sending continuously against a 30s probe cadence — every genuine seed is more than `maxFloorAdvance` above it and is rejected as `slot_jumped`. Rejected seeds create no entry, so genuine senders never become references themselves and the lockout is stable: 0 accepted, 5760 rejected over 48h in simulation, from about one attacker datagram every 20 minutes.

ben-dz (changes requested):
Separate from copilot's findings, two below.

PR #4309, opened 2026-09-10
solana-cli: add a Squads vault path to shreds validator-client-rewards claim

### LINKED ISSUE #4185: solana-cli: add a Squads vault path to shreds validator-client-rewards claim
Depends on #4183 and #4184

Target submodule: [malbeclabs/doublezero/offchain](https://github.com/malbeclabs/doublezero/tree/main/offchain)

## Summary

`ClaimValidatorClientRewards` takes the `ValidatorClientRewards` manager as its only signer, and `shreds validator-client-rewards claim` refuses to build anything unless the wallet is that manager. A manager held by a Squads vault cannot claim at all, because a vault is a PDA and no local keypair can sign for it.

Add a vault path built on `OptionalSquadsArgs` from the `squads` module in `doublezero-solana-client-tools`. With `--multisig`, the vault stands in for the wallet as manager, and the command prints base58 encoded transactions for import into Squads instead of sending them.

- Keep `validate_manager` on both paths, against the wallet key on the direct path and the vault key on the vault path. The account it reads is already fetched alongside `ProgramConfig`, so the vault path costs no extra round trip, and refusing there is what stops a payload that the multisig can never execute from consuming an approval round. Its message names the wallet today and has to name whichever actor was checked.
- The vault path builds a `SolanaConnection` and no `Wallet`, which lets `--multisig` run on a machine with no keypair. Every signing and sending option in `WriteVerbOptions` is inert there (`--keypair`, `--fee-payer`, `--dry-run`, `--with-compute-unit-price`, and `--verbose`), and the output should state that nothing was signed or sent.
- The payload carries `CheckCliVersion` and the claim instruction, and no compute budget instructions. Squads sets the budget on its own execute transaction, and a compute budget instruction reached through a CPI is a no-op that only burns compute units.
- The default destination becomes the associated token account of the vault, which will usually not exist. Prepend `create_idempotent` when that account is missing, and put it in every payload rather than only the first, so payloads stay independently executable. It costs roughly 106 bytes and three account keys per payload, which has to come out of each payload's budget, and it drops out of a later run once the account exists.
- Creating the destination that way makes the vault pay the rent, so the vault has to hold SOL.
- Retire `MAX_CLAIM_EPOCHS_PER_TX`. The claim instruction carries a `Vec<ClaimHoldingId>`, so its size is measurable: grow the bundled instruction while the trial transaction still fits. `try_batch_instructions_with_common_signers` cannot be reused, because it packs independent instructions rather than growing one, it takes `&[&Keypair]` that the vault path does not have, and splitting claim into one instruction per holding costs about a third of the packing density (19 holdings per direct transaction against 13).
- Count `CheckCliVersion`, the compute unit limit, and the compute unit price in the direct path's trial whether or not a price is configured, so packing does not shift with the flag.
- Count the fee payer in the direct path's trial as well. `Wallet::new_transaction` signs with both the fee payer and the signer, so a `--fee-payer` distinct from the wallet costs another signature and another writable signer key, 96 bytes, and drops the measured maximum from 19 holdings to 16. A trial built with one signer while the send uses two produces oversized transactions.
- The direct path measures a v0 transaction and the vault path measures the legacy message the base58 string actually carries. Same 1232-byte ceiling, different encodings, so the two paths cannot share one measured length.
- A new `--max-transactions N` caps how many transactions one invocation produces: the direct path executes the first N, the vault path prints the first N payloads. Holdings are discovered in epoch order, so "first" is deterministic. The default is all of them, which is today's behavior, and `--max-transactions 1` is the one-proposal-at-a-time flow that keeps a terminal readable.
- When transactions are withheld, report how many, and how to reach them: re-run once the printed payloads have executed, or name the epochs explicitly.
- A human copies each payload into the Squads UI by hand, so the output has to make losing track impossible.
  - Number every payload (`Vault transaction 2 of 3`)
  - State the holdings, the epochs, and the pre-claim total each payload covers
  - Keep the base58 string alone on its own line, so a copy takes exactly the payload
  - Separate the blocks
  - Close with a recap naming every payload by number and epoch range
- Print the destination token account's authority alongside the destination itself. Neither the processor nor the command constrains who owns the destination, so `--destination-token-account` can point a payload at a third party, and the approvers are reading a base58 blob. The authority is the one field that tells them where the 2Z lands.
- The transactions are independent, so any import order works and a partial import is safe. State that in the output, together with the fact that re-running the command discovers only the holdings that are still outstanding.
- `print_vault_transaction` writes with `println!`, but the verb contract passes `out: &mut impl Write`. Add a writer-taking variant to the `squads` module and have `print_vault_transaction` delegate to it, so the payloads follow the same writer as the rest of the command's output.

elitegreg:
Resolves: #4185

## Summary of Changes
- `shreds validator-client-rewards claim` gains a Squads vault path. With `--multisig` (and optional `--vault-index`), the vault stands in for the wallet as manager, no keypair is loaded, and the command prints base58 payloads for import into Squads instead of signing and sending. Each payload carries `CheckCliVersion` and the claim, and no compute budget instructions, since Squads sets the budget on its execute transaction.
- The vault path keeps the manager check, against the vault key, so a payload the multisig can never execute is refused before it consumes an approval round. The mismatch message now names whichever actor was checked.
- The default destination on the vault path is the vault's associated token account. When it is missing, every payload prepends `create_idempotent`, so each payload stays independently executable, and the output says the vault pays the rent. An explicit `--destination-token-account` that does not exist is refused. The direct path still requires the destination to exist.
- `MAX_CLAIM_EPOCHS_PER_TX` is retired. Each transaction grows one holding at a time while a trial still fits. The direct path measures the v0 transaction the wallet sends, always counting the fee payer and both compute budget instructions so packing does not shift with the flags. The vault path measures the legacy message the base58 string carries against the Squads payload budget. Measured capacities: 19 holdings per direct transaction, 16 with a distinct `--fee-payer`, 12 per vault payload into an existing destination, 10 when every payload creates it.
- New `--max-transactions N` caps how many transactions one invocation produces. The output reports how many were withheld and how to reach them.
- Vault output is written for a human copying payloads by hand: every payload is numbered, states its holdings, epochs and pre-claim total, keeps the base58 alone on its own line, and the command closes with a recap and the note that payloads are independent and a re-run discovers only what is still outstanding. The destination token account's authority is printed on both paths.
- `solana-client-tools`: add `squads::try_write_vault_transaction`, a writer-taking form that `try_print_vault_transaction` now delegates to, and `Wallet::try_transaction_size`, which measures the transaction `new_transaction` would build with the same signer set.

## Diff Breakdown
| Core logic   |     1 | +538 / -192   |  +346 |
| Scaffolding  |     2 | +47  / -23    |   +24 |
| Tests        |     3 | +299 / -11    |  +288 |
| Docs         |     3 | +8   / -0     |    +8 |
| **Total**    |     6 | +892 / -226   |  +666 |

Test and core lines are split by test module within each Rust file, so the per-category numbers are approximate. Roughly half the change is the claim command itself and a third is new tests.

<details>

- [`offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs`](https://github.com/malbeclabs/doublezero/pull/4309/files#diff-16db6ff15c9c64afd49a8d8581ec8293b10e12589d90e6c7c3e6210a4a266c4c) — actor selection (wallet or vault), destination resolution and ATA creation, measured packing, `--max-transactions`, and the vault output
- [`offchain/crates/solana-client-tools/src/payer.rs`](https://github.com/malbeclabs/doublezero/pull/4309/files#diff-29650db255707f841bfb88b839b3e7ffee34d3f9e1b231b03133e127cd3b910f) — `Wallet::try_transaction_size` and the shared signer list it measures with
- [`offchain/crates/solana-client-tools/src/squads.rs`](https://github.com/malbeclabs/doublezero/pull/4309/files#diff-f18cf5de60d5bc6914e0ec4f2d1dd7ca6f2c513524ea70db9469be176b76821f) — `try_write_vault_transaction`, with the printing form delegating to it

</details>

## Testing Verification
- Unit tests pin the measured packing capacities (19, 16, 12, 10 holdings) and check that every packed batch fits as sent, with and without a compute unit price, while one more holding does not. The vault batches are also run through the checked Squads encoder, which accepts each batch and refuses each batch plus one holding.
- Tests cover the growth loop itself (grows, splits, refuses a holding that fits nowhere), the `--max-transactions` withheld note, the create-ATA instruction appearing in every payload, `--vault-index` requiring `--multisig`, `--max-transactions 0` being rejected, and the manager mismatch message naming the wallet or the vault.
- A `Wallet` with a distinct fee payer measures exactly 96 bytes larger than one without, and the writer-taking Squads printer puts the base58 alone on its own line.
- Manually on Solana devnet, with no keypair available to the command: `--multisig` pointed at a non-Squads account fails with the "not the Squads v4 program" message before anything is built, and a real devnet multisig gets through the multisig check to the validator client rewards lookup. No devnet client is managed by a Squads vault, so no end-to-end payload was produced against a live multisig.


offchain/crates/solana-client-tools/src/payer.rs
```diff
@@ -205,6 +192,27 @@ impl Wallet {
...
+    // The fee payer signs first, since `try_new_transaction` takes the first signer as the
+    // payer. `try_new` refuses a fee payer equal to the signer, so the two are distinct.
+    fn signers(&self) -> Vec<&Keypair> {
+        let mut signers = Vec::with_capacity(2);
+        if let Some(ref fee_payer) = self.fee_payer {
+            signers.push(fee_payer);
+        }
+        signers.push(&self.signer);
+        signers
+    }
```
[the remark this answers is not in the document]
elitegreg: Fixed in b19f1a75c: `Wallet::signers` now skips the signer when the fee payer is the same key, with a test that a wallet naming one key twice measures the same as one naming it once.

nikw9944 (commented):
Two things to change before merge: vault payloads are packed to within one holding of the size ceiling, so a memo typed at import is refused — `squads.rs` names that hazard exactly; and an existing `--destination-token-account` is accepted on the vault path without checking its authority, while the missing-destination branch already enforces the vault-only rule. Four smaller items on the epoch range, the pre-claim total under `--max-transactions`, `CheckCliVersion` in a deferred payload, and the unbounded execute CU.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -218,175 +287,449 @@ impl ClaimCommand {
...
+                Ok(wallet.try_transaction_size(&instructions)? <= MAX_TRANSACTION_SIZE)
+            })?,
+            Actor::Vault { vault_key, .. } => pack_holdings(&holdings, |batch| {
+                let instructions = vault_instructions(
+                    &check_cli_version_ix,
+                    create_destination_ix.as_ref(),
+                    build_claim_ix(batch)?,
+                );
+                let payload = Message::new(&instructions, Some(vault_key)).serialize();
+                Ok(payload.len() <= vault_transaction_payload_budget(instructions.len()))
```
nikw9944: **Blocking** — Every vault payload is packed to within one holding of the size ceiling, so a memo typed in the Squads import dialog pushes it over and the import is refused. `squads.rs:95-101` names this hazard exactly — "a payload sized to the last byte is a payload that a memo breaks" — and nothing here reserves for it or lets the operator shrink a payload.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -218,175 +287,449 @@ impl ClaimCommand {
...
+                            format!("unpacking destination token account {destination_key}")
+                        })?;
+                if destination_token.mint != self.rewards_token_mint {
+                    bail!(
+                        "destination {destination_key} mint mismatch: expected {}, found {}",
+                        self.rewards_token_mint,
+                        destination_token.mint
+                    );
+                }
+                (destination_token.owner, None)
```
nikw9944: **Blocking** — An existing `--destination-token-account` is accepted after checking only its owner program and mint, never its authority, so a vault payload can route every outstanding reward into a token account the proposer controls while approvers see only base58. The same command already refuses a non-vault destination in the missing-destination branch at line 333.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -218,175 +287,449 @@ impl ClaimCommand {
...
+        while end < holdings.len() && fits(&holdings[start..=end])? {
+            end += 1;
+        }
+        batches.push(&holdings[start..end]);
+        start = end;
+    }
+    Ok(batches)
+}
+
+fn epoch_range(batch: &[HoldingToClaim]) -> String {
```
nikw9944: **Minor** — `epoch_range` collapses a batch to `first..=last`, so a payload covering epochs 700, 705 and 712 prints as `700..=712`. The pre-PR code printed the exact list, and the vault recap is what a human uses to pick which blob to approve; `write_withheld_note` then suggests feeding that range back to `--subscription-epoch`.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -218,175 +287,449 @@ impl ClaimCommand {
...
             }
+            wallet.write_verbose_output(out, &[tx_sig]).await?;
         }
+    }
 
-        Ok(())
+    if last_executed {
+        writeln!(
+            out,
+            "\nPre-claim total: {total_pre_balance} ({executed_holdings}/{total_holdings} holding(s) claimed across {batch_count} transaction(s))."
```
nikw9944: **Minor** — `total_pre_balance` includes the batches `--max-transactions` withheld but prints beside the executed count: `Pre-claim total: 910 (19/91 holding(s) claimed ...)` where 910 covers all 91. Before this PR every batch was always attempted, so the two agreed.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -218,175 +287,449 @@ impl ClaimCommand {
...
+/// The payload a vault imports: the version check, the destination create when
+/// the account is missing, and the claim. No compute budget instructions, since
+/// Squads sets the budget on its own execute transaction and a budget
+/// instruction reached through a CPI is a no-op that only burns compute units.
+fn vault_instructions(
+    check_cli_version_ix: &Instruction,
+    create_destination_ix: Option<&Instruction>,
+    claim_ix: Instruction,
+) -> Vec<Instruction> {
+    let mut instructions = vec![check_cli_version_ix.clone()];
```
nikw9944: **Minor** — `CheckCliVersion` enforces the onchain minimum CLI version, but in a vault payload it is evaluated days later at execute time rather than milliseconds after it is stamped. A floor raised while approvals are being collected reverts every outstanding payload and burns the round.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -218,175 +287,449 @@ impl ClaimCommand {
...
+    ];
+    instructions.extend(compute_unit_price_ix.cloned());
+    instructions
+}
+
+/// The payload a vault imports: the version check, the destination create when
+/// the account is missing, and the claim. No compute budget instructions, since
+/// Squads sets the budget on its own execute transaction and a budget
+/// instruction reached through a CPI is a no-op that only burns compute units.
+fn vault_instructions(
```
nikw9944, line 715: **Minor** — Vault payloads carry no compute budget, so the limit is whatever the Squads app puts on its execute transaction, against this command's own estimate of ~390k CU for a 12-holding payload. If it falls short the failure lands after the approvals are spent; I could not settle the app's behavior from the repo, and the README's devnet confirmation covers single-instruction payloads.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -218,175 +287,449 @@ impl ClaimCommand {
...
+            })?,
+        };
+
+        let emitted_count = self
+            .max_transactions
+            .map_or(batches.len(), |max| max.get().min(batches.len()));
+        let (emitted, withheld) = batches.split_at(emitted_count);
 
         let total_holdings = holdings.len();
         let total_pre_balance = holdings.iter().fold(0u64, |total, holding| {
```
bgm-malbeclabs: **Blocking.** `total_pre_balance` sums every discovered holding, and `execute_direct` prints it as `Pre-claim total`. With `--max-transactions` the withheld holdings are in that sum, so the line reports money this invocation did not claim. An operator reconciling a claim reads that number as the amount that moved. The vault path gets this right per payload with `batch_pre_balance`. Sum over `emitted`, or rename the line to say it covers every outstanding holding.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -218,175 +287,449 @@ impl ClaimCommand {
...
+                );
+                Ok(wallet.try_transaction_size(&instructions)? <= MAX_TRANSACTION_SIZE)
+            })?,
+            Actor::Vault { vault_key, .. } => pack_holdings(&holdings, |batch| {
+                let instructions = vault_instructions(
+                    &check_cli_version_ix,
+                    create_destination_ix.as_ref(),
+                    build_claim_ix(batch)?,
+                );
+                let payload = Message::new(&instructions, Some(vault_key)).serialize();
```
bgm-malbeclabs, line 427: **Blocking.** The packer aims at the budget to the last byte. Budget is 1232 - 384 - 2 = 846, and a full 12-holding payload measures 810. One holding costs 41 bytes (32 for the account key, 9 for the data), so a greedy packer always stops with less than 41 bytes left. 36 here.

`squads.rs` lists three costs the 384-byte reserve does not cover, and the first is a memo typed at import: 4 bytes and the text on the create instruction, and again on the approval. A 20-character memo is 48 bytes and breaks the import of every payload, not one of them. The same 36 bytes also have to cover a rent payer that is not the creator (+96) and each extra wrapper account key (+33).

Subtract a fixed memo allowance from the budget the packer aims at, so the payloads this command emits keep the room that comment says they need.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
bgm-malbeclabs: Non-blocking. `epoch_range` prints the first and last epoch, but the holding set is usually sparse. `discover_holdings` probes 91 consecutive epochs and keeps only the ones that exist for this mint, and a re-run after a partial claim is exactly the sparse case this command promises to handle. A batch covering 700, 705 and 712 prints `epochs 700..=712`.

On the vault path that line and the recap are the only place the epochs appear, so a human copying payload 3 of 5 cannot tell what is in it. The direct path used to print the list.

It costs most in `write_withheld_note`, which tells the operator to name those epochs with `--subscription-epoch`. Expanding a printed range names epochs that hold nothing, and `validate_explicit_holdings` then warns and skips them.

Print the list, or collapse to `a..=b` only when the batch really is consecutive. `test_epoch_range_collapses_a_single_epoch` uses three consecutive epochs, so it does not cover this.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -218,175 +287,449 @@ impl ClaimCommand {
...
             total.saturating_add(holding.pre_balance)
         });
-        let batches = holdings.chunks(MAX_CLAIM_EPOCHS_PER_TX).collect::<Vec<_>>();
-        let batch_count = batches.len();
 
         writeln!(
             out,
             "Shred subscription - Claim Validator Client Rewards \
-             (client_id={}, mint={}, holdings={total_holdings}, transactions={batch_count})",
+             (client_id={}, mint={}, holdings={total_holdings}, transactions={emitted_count})",
```
bgm-malbeclabs: Non-blocking. The header pairs the full holding count with the emitted transaction count. With `--max-transactions 1` over 91 holdings it reads `holdings=91, transactions=1`, and only the next line corrects it. Count what `emitted` covers.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -218,175 +287,449 @@ impl ClaimCommand {
...
+                (destination_token.owner, None)
+            }
+            (None, Actor::Wallet(_)) => bail!(
+                "destination token account {destination_key} does not exist. \
+                     Run: `spl-token create-account --owner {actor_key} {} --fee-payer {actor_key}`",
+                self.rewards_token_mint
+            ),
+            (None, Actor::Vault { vault_key, .. }) => {
+                ensure!(
+                    self.destination_token_account.is_none(),
```
bgm-malbeclabs: Non-blocking. This refuses `--destination-token-account` set to exactly the vault's own associated token account when that account does not exist yet, and the message then tells the operator to drop the flag to reach the same address. Compare the supplied key against the derived address rather than test whether the flag was given.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -218,175 +287,449 @@ impl ClaimCommand {
...
-            bail!(
-                "destination {destination} mint mismatch: expected {}, found {}",
-                self.rewards_token_mint,
-                destination_token.mint
-            );
-        }
+            }
+        };
+
+        let check_cli_version_ix = super::super::build_check_cli_version_instruction()?;
```
bgm-malbeclabs, line 396: Non-blocking. `CheckCliVersion` rides inside each vault payload, so it runs at execute, days after the payload was written. Raise the onchain minimum CLI version while a proposal sits in approval and every approved payload fails at execute. That is the wasted approval round the manager check at line 205 exists to prevent. The direct path checks immediately, so this exposure comes with the vault path.

offchain/crates/solana-client-tools/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- add `squads::try_write_vault_transaction`, which writes the import payload to an `impl Write` the way a verb's `out` requires. `try_print_vault_transaction` delegates to it ([malbeclabs/doublezero#4185](https://github.com/malbeclabs/doublezero/issues/4185))
+- `Wallet` signs once when a directly constructed wallet names the same key as fee payer and signer. `try_new` already refuses that pairing, so only callers building the struct themselves are affected
```
bgm-malbeclabs: Non-blocking. "`Wallet` signs once when a directly constructed wallet names the same key as fee payer and signer" describes a fix this diff does not make. The old body already had `if self.signer.pubkey() != fee_payer.pubkey() { signers.push(&self.signer); }`, so the extracted `signers()` behaves the same, and `test_transaction_size_signs_once_when_the_fee_payer_is_the_signer` passes against main. Reword it as the refactor it is, or drop it. Commit `b19f1a75c` carries the same title.

bgm-malbeclabs (dismissed):
Good shape overall. Measured packing is the right call over the fixed cap, and the trial that always counts the fee payer and both compute budget instructions means the sent transaction is never larger than what was measured. Checking the manager against the vault before anything gets built saves a wasted approval round. The tests that pin the capacities and then re-run every packed batch through the checked encoder are the right shape, not just the number.

Two blockers inline: the `Pre-claim total` line reports money this run did not claim, and the vault payloads pack to within 36 bytes of the Squads budget, which a memo typed at import spends. Five non-blocking notes after those.

Minor, no comment needed: `pack_holdings` treats any `fits` error as fatal, so a build failure aborts the claim instead of ending the batch; `last_executed` is really `any_executed`; and the header's "N more transaction(s) withheld" duplicates `write_withheld_note`.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
elitegreg: Fixed in 029f19deb. `vault_transaction_packing_budget` is the payload budget less `SQUADS_IMPORT_MEMO_RESERVE_BYTES`, 72 bytes for a 32-byte memo on the create and again on the approval, and the packer now aims at that instead of at the budget the encoder enforces. Dropping CheckCliVersion from the payload paid for most of it, so a payload goes from 12 holdings to 11, and from 10 to 9 where it also creates the destination. A test asserts every packed batch still has the memo's room left against the hard budget.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
elitegreg: Fixed in 029f19deb. The existing-destination arm now runs through `validate_existing_destination`, which takes the vault key on the vault path and refuses any authority that is not the vault, naming both keys. The direct path passes `None` and is unchanged.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
elitegreg: Fixed in 029f19deb. `epoch_list` collapses only genuinely consecutive runs, so a sparse batch prints `700, 705..=707, 712`. The withheld note now prints its epochs space separated so the line pastes straight after `--subscription-epoch`.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
elitegreg: Fixed in 029f19deb. The header count and the pre-claim total both fold over the emitted batches now, via `emitted_totals`, so they cover only what the invocation claims. The withheld holdings are accounted for by `write_withheld_note` alone.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
elitegreg: Fixed in 029f19deb by dropping CheckCliVersion from vault payloads. In a deferred payload it says nothing about the version that built it, only that the floor did not move, so it bought the failure mode and nothing else. The direct path keeps it.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
elitegreg: Addressed in 029f19deb, though not settled. Each payload now prints the compute units it needs at execute, and the preamble says Squads sets that budget on its own execute transaction and points at the figure to check it against. The app behavior is still unconfirmed from here, so this makes the number visible rather than assuming a limit.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
elitegreg: Fixed in 029f19deb. `emitted_totals` folds over the emitted batches, and both the header count and the `Pre-claim total` line use it, so neither reports a withheld holding. A test pins that over a split batch list.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
elitegreg: Fixed in 029f19deb. Added `SQUADS_IMPORT_MEMO_RESERVE_BYTES`, 72 bytes derived as 4 plus 32 on the create and again on the approval, and `vault_transaction_packing_budget` which is the payload budget less that. The packer aims at the packing budget and the encoder keeps enforcing the hard one. Payload capacity is 11 holdings into an existing destination and 9 where it creates one, the drop softened by CheckCliVersion coming out of the payload. Not covering the rent payer or the extra wrapper keys: those are app behaviors the reserve already assumes away, and pricing them would cost more capacity than the hazard warrants.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
elitegreg: Fixed in 029f19deb. `epoch_list` collapses a run only where it is really consecutive, so `700, 705, 706, 707` prints as `700, 705..=707`. The withheld note names its epochs exactly, space separated, ready to paste after `--subscription-epoch`. The test now uses a sparse set rather than three consecutive epochs.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
elitegreg: Fixed in 029f19deb. The header counts what `emitted` covers. The discovery line above it still prints the full outstanding count, and the duplicate withheld line in the header is gone, since `write_withheld_note` says it with more detail.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
elitegreg: Fixed in 029f19deb. The branch compares `destination_key` against `get_associated_token_address(vault_key, mint)`, so naming the vault's own associated token account explicitly is the same request as leaving the flag off.

offchain/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
elitegreg: Fixed in 029f19deb by taking CheckCliVersion out of vault payloads entirely. It guards the moment a transaction is built, which is not the moment this payload runs, so in a deferred payload it only supplies the revert. The direct path, which sends milliseconds after building, keeps it.

offchain/crates/solana-client-tools/CHANGELOG.md
elitegreg: Fixed in 029f19deb. You are right that main already deduped at `payer.rs:146`. Reworded as the refactor it is: extracting `Wallet::signers` so `try_new_transaction` and `try_transaction_size` measure and sign with the same set, behavior unchanged.

elitegreg:
Pushed 029f19deb, replies on each thread above.

On the three from the review body: `pack_holdings` still treats a `fits` error as fatal on purpose, since a build failure there is a bug rather than a size signal and ending the batch on it would quietly emit short transactions. `last_executed` is now `any_executed`. The header no longer duplicates `write_withheld_note`.

Two things worth a second look. Vault payload capacity moved to 11 holdings into an existing destination and 9 where every payload creates it, down one each: the 72-byte memo reserve costs more than that, and dropping CheckCliVersion from the payload gives most of it back. And the execute compute budget is surfaced rather than solved, since each payload now prints the compute units it needs but the Squads app's own limit is still unconfirmed from here.

nikw9944 (approved):
All six findings from the last round are fixed in `029f19de` and verified in the code, with tests covering the memo-reserve packing budget, the vault-authority check on an existing destination, the sparse epoch list, and the emitted-only totals. The execute compute is now reported per payload rather than bounded — the Squads app's own limit stays unconfirmed, which the vault output says plainly. lgtm.

PR #4310, opened 2026-09-10
serviceability: retire a feed, with the notice its seat holders are owed

bgm-malbeclabs:
Implements D2. Stacked on #4307, which gave a feed a way to stop and start; this gives it a way to end.

## The lifecycle it completes

| From | `RetireFeed` | `FinalizeFeedRetirement` |
|---|---|---|
| `Pending` | to `Retiring`, no wait | — |
| `Active` | to `Retiring`, thirty days | — |
| `Halted` | to `Retiring`, thirty days | — |
| `Retiring` | refused, already started | to `Retired`, once the notice elapses |
| `Retired` | refused, terminal | refused, nothing to finish |

A retiring feed neither halts nor resumes. That is the whole difference between retirement and halt, and it has a test.

## Four decisions worth arguing with

**A `Pending` feed retires with no wait.** The notice exists for seat holders, and a feed that was never `Active` admitted none: `require_feed_admits` refuses subscribers unless the feed is publishing. So the wait is zero rather than absent, which keeps one path through retirement instead of two branches. It has to be retirable at all because `DeleteFeed` refuses a staked feed, so refusing here as well would leave a feed that never went live with no way out — the same shape of hole as the one found in halt on #4307.

**Finalizing is permissionless**, unlike every other feed instruction. The clock decided when retirement started, and this instruction can only agree with it. Requiring an authority adds no safety and one failure mode: a feed sitting in `Retiring` forever because whoever held the key stopped caring. A seat holder given a date deserves the date rather than someone's attention.

**`FeedRetired` is a `msg!` line, not a new event mechanism.** This program has none: no `sol_log_data`, no return data. Nothing consumes the event yet either, because releasing the stake on retirement waits for the slashing work. That is the same reasoning the plan already applied to `StakeLocked` and `StakeReleased` under `solana/`: inventing a mechanism for a single absent consumer buys nothing. The line's shape is fixed so a future consumer can parse it without it changing.

**`Retiring` takes discriminant 4**, not a value between `Halted` and `Retired`. Those are written into live accounts, and renumbering them would reinterpret every stored feed rather than adding a state.

## Testing Verification

Fifteen tests in `feed_lifecycle_test.rs`, six of them new:

- Retiring an active feed starts a notice at least thirty days out, measured against the validator's clock rather than assumed.
- A `Pending` feed's notice ends at or before now.
- A retiring feed refuses halt, resume **and** a second retire, each with its own error.
- Finalizing before the notice elapses is refused and the feed stays `Retiring`.
- Once elapsed, a signer holding nothing finalizes it, which is what permissionless means.
- Finalizing a feed that is not retiring is refused.

The elapsed and not-yet cases seed `retires_at` in the past and the future rather than moving the validator's clock. The comparison against `Clock` is what those two tests are about; how the timestamp gets there is `retire`'s business and has its own test.

The full serviceability suite passes, plus the SDK blast radius a `Feed` field change now reaches: the fixture generator, `feed.bin`, and the Python and TypeScript decoders, which also gain `FEED_STATUS_RETIRING`. 126 Python tests and 149 TypeScript pass against the regenerated fixture, and `make check-fixtures` is clean.

TypeScript reads `retiresAt` as `BigInt.asIntN(64, r.readU64())`, matching how the seat timestamps in the same file handle a signed 64-bit value; there is no `readI64`.


ben-malbeclabs (commented):
Reviewed the program, the SDK decoders and the tests. The instruction wiring is complete in every hand-written match — `unpack`, `instruction_name`, the args formatter, `entrypoint`, `From<DoubleZeroError>`, `From<u32>`, and the range comment in the error round-trip test — which is the trap called out on #4307. Discriminants are appended rather than inserted, `FeedStatus::Retiring = 4` does not renumber stored bytes, `retires_at` appends at the struct tail so pre-RFC-28 accounts still decode, and `finalize_retirement.rs` pins identity with both the owner check and the `AccountType` discriminant despite being permissionless. I found no substitution, rent or griefing vector in the permissionless path.

Two Critical and eleven Important findings inline. Two of them are product calls rather than defects: whether `Retiring` is meant to stop reconnections immediately, and whether a builder should be able to retire its own feed.

Evidence for the two test findings, since CI is green and would stay green: three mutations to `retires_at` survive the whole suite (the `Pending` branch writing `0` instead of `now`, `RETIREMENT_NOTICE_SECONDS * 1000`, and `finalize` zeroing `retires_at` while overwriting `halted_by`), and on the SDK side replacing the decoder read with a constant leaves all 126 Python and 149 TypeScript tests passing. The field this PR adds is unpinned onchain and in both decoders.

Two findings could not be anchored inline because their files are not in this PR:

**Important** — `smartcontract/programs/doublezero-serviceability/src/processors/feed/delete.rs:59`. The comment says retirement is a staked feed's path out and that it lands in D2, but this is D2 and `delete.rs:60` still refuses on a non-default `builder` with no status exemption, so a `Retired` staked feed stays undeletable with its bond locked. That also makes `retire.rs:62-64`'s justification for the `Pending` path self-defeating: retiring a staked `Pending` feed leaves it with no way out either.

**Important** — `crates/doublezero-serviceability-instruction/src/feed.rs:160`. Neither instruction has a builder, so nothing outside the program's own tests can invoke either one; a repo-wide grep for `RetireFeed` and `FinalizeFeedRetirement` hits only `entrypoint.rs`, `instructions.rs`, the CHANGELOG and `feed_lifecycle_test.rs`. Note `finalize_retirement`'s layout is `[feed, payer, system]`, one account shorter than every sibling, so it cannot reuse `common::build_with_permission` unchanged.

Glossary pass against `edge-feed-spec/GLOSSARY.md` v1.3.0 is clean: no banned word in any added line, and the one nearby `source` use is the qualified `upstream source`.

smartcontract/programs/doublezero-serviceability/src/processors/feed/retire.rs
```diff
@@ -0,0 +1,96 @@
...
+    // The notice exists for seat holders, and a feed that was never `Active` has none: it admits
+    // no subscriber until it publishes. So the wait is zero rather than absent, which keeps one
+    // path through retirement instead of two.
+    feed.retires_at = if feed.status == FeedStatus::Pending {
+        now
+    } else {
+        now.checked_add(RETIREMENT_NOTICE_SECONDS)
+            .ok_or(DoubleZeroError::InvalidArgument)?
+    };
+    feed.status = FeedStatus::Retiring;
```
ben-malbeclabs, line 85: **Critical** — this transition enters a state nothing honors: `require_feed_admits` (`processors/feed/mod.rs:121`) gates on `status != Active`, so `Retiring` is refused exactly as `Retired` is and a seat holder who reconnects during the thirty days is locked out on day one. Admit `Retiring` on the reconnect path and keep refusing it where a new seat is provisioned.

`require_feed_admits` is the only subscriber-side lifecycle gate, reached from `enforce_feed_metro_gate` (`mod.rs:105`) via `user/create_subscribe.rs:211` and `multicastgroup/subscribe_feed.rs`. `try_add_feed_user` (`state/accesspass.rs:374`) increments `current_users`, so reconnect re-runs the gate and reverts `FeedNotActive`. Nothing evicts live users, so the only population the notice could serve is the one that cannot come back, and `retires_at` is read in exactly one place, `finalize_retirement.rs:54`.

If the intent is that retirement stops all new connections immediately and the thirty days only defers stake release, then this code is right and the PR title and the `retires_at` doc are wrong instead.

smartcontract/programs/doublezero-serviceability/tests/feed_lifecycle_test.rs
```diff
@@ -531,3 +538,247 @@ async fn test_a_staked_feed_cannot_resume_without_its_mirror() {
...
+        program_id,
+        DoubleZeroInstruction::RetireFeed(FeedRetireArgs {}),
+        feed_accounts(feed, globalstate),
+        &payer,
+    )
+    .await;
+
+    let f = feed_of(&mut banks_client, feed).await;
+    assert_eq!(f.status, FeedStatus::Retiring);
+    assert!(
```
ben-malbeclabs, line 619: **Critical** — both notice assertions are one-sided bounds, so `retires_at` is never pinned to a value the processor could get wrong. Assert the computed timestamp against the observed clock rather than `>=` here and `<=` at line 642.

Three mutations survive 15/15: the `Pending` branch writing `0` instead of `now` (the `<=` at line 642 is satisfied by the `i64` default, so the test cannot tell a zero wait from no notice written), `RETIREMENT_NOTICE_SECONDS * 1000` for an 82-year notice, and `finalize` zeroing `retires_at` while overwriting `halted_by`.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
...
 - Serviceability
+  - `RetireFeed` (variant 122) and `FinalizeFeedRetirement` (variant 123), with a new `Retiring` status and a `retires_at` timestamp on `Feed`. Retiring starts a thirty-day notice to seat holders and is terminal: a retiring feed neither halts nor resumes, and the only state after it is `Retired`. A `Pending` feed retires with no wait, because the notice exists for seat holders and a feed that never went `Active` admitted none; it still has to be retirable, since `DeleteFeed` refuses a staked feed and refusing here too would leave it with no way out. Finalizing is permissionless: the clock has already decided, so requiring an authority would only let a feed sit in `Retiring` because whoever held the key stopped caring. `FeedRetired` is a `msg!` line rather than a new event mechanism, since this program has none and nothing consumes it yet. New errors `FeedNotRetirable` (126), `FeedNotRetiring` (127) and `RetirementNoticeNotElapsed` (128).
```
ben-malbeclabs: **Important** — 148 words of design rationale in one bullet. Keep the two instructions, the new status and field, the thirty-day notice and the three error codes; the reasoning belongs in the PR body.

sdk/serviceability/testdata/fixtures/generate-fixtures/src/main.rs
```diff
@@ -1520,6 +1520,7 @@ fn generate_feed(dir: &Path) {
         committed_rate_bits_per_sec: 1_000_000_000,
         status: FeedStatus::Pending,
         halted_by: Pubkey::default(),
+        retires_at: 0,
```
ben-malbeclabs: **Important** — `retires_at` is generated as its own default with no `RetiresAt` entry in the fixture's `FixtureMeta` field list, so no Python or TypeScript test reads the new field. Give it a distinct nonzero timestamp, add the meta entry, and assert it in both feed fixture tests.

Replacing the decoder read with a constant left all 126 Python and 149 TypeScript tests passing, so nothing pins the field's offset, width or sign. `status` is `Pending` here too, so `FEED_STATUS_RETIRING = 4` is exercised by nothing, and the TypeScript `BigInt.asIntN(64, ...)` at `state.ts:1306` is meaningful only for a negative value that no fixture produces.

smartcontract/programs/doublezero-serviceability/src/processors/feed/retire.rs
```diff
@@ -0,0 +1,96 @@
...
+        "Invalid GlobalState Account Owner"
+    );
+    assert!(feed_account.is_writable, "PDA Account is not writable");
+
+    let mut feed = Feed::try_from(feed_account)?;
+    let globalstate = GlobalState::try_from(globalstate_account)?;
+    require_feed_writer(
+        program_id,
+        accounts_iter,
+        payer_account.key,
```
ben-malbeclabs, line 57: **Important** — retirement is terminal but authorizes through `require_feed_writer`, which was built for reversible halts and admits the feed's own builder, so a builder can unilaterally end a feed with sold seats and lock its own bond. `delete.rs:40`, the other terminal feed instruction, requires `FEED_AUTHORITY`/`FOUNDATION`.

`mod.rs:151` returns `Ok` as soon as `feed.builder == payer`, and its doc justifies that by halt being the builder's lever for upstream rotation, which does not carry over to an irreversible transition. RFC-28 is not in `rfcs/`, so I could not check whether builder-initiated retirement is intended.

smartcontract/programs/doublezero-serviceability/src/state/feed.rs
```diff
@@ -111,6 +118,13 @@ pub struct Feed {
     /// moment an operator halts, and with `Retired` unreachable and `DeleteFeed` refusing a staked
     /// feed, nothing else stops one.
     pub halted_by: Pubkey, // 32
+    /// When this feed's retirement notice elapses, as a unix timestamp, zero when it is not
+    /// retiring.
+    ///
+    /// Set when retirement starts and never moved, so the date a seat holder was given is the
+    /// date that arrives. A feed that never sold a seat gets `now`, because the notice exists for
+    /// seat holders and a feed that was never `Active` has none.
+    pub retires_at: i64, // 8
```
ben-malbeclabs, line 127: **Important** — no borsh round-trip covers `FeedStatus::Retiring` or a nonzero `retires_at`; the unit test module at the foot of this file is unchanged. The comment at lines 28-33 names discriminant 4 as the load-bearing compatibility decision and no test pins that byte.

smartcontract/programs/doublezero-serviceability/tests/feed_lifecycle_test.rs
```diff
@@ -531,3 +538,247 @@ async fn test_a_staked_feed_cannot_resume_without_its_mirror() {
     .await;
     assert_custom_at_ix0(&result, custom_code(DoubleZeroError::StakeMirrorMissing));
 }
+
+/// A feed in `Retiring`, seeded with the notice ending at `retires_at`.
```
ben-malbeclabs, line 543: **Important** — no test drives `retire` and then `finalize`, so the timestamp `retire` computes is never the timestamp `finalize` compares. The `Pending` path makes the end-to-end case free, since its wait is zero and needs no clock warp, and `catalog_feed` then retire then finalize covers the not-yet-elapsed case against a real thirty-day notice rather than `i64::MAX`.

The justification for seeding carries over from the `halted_feed` helper at line 328, where a staked feed genuinely cannot reach `Halted` until G1. `retire` reaches `Retiring` for real at lines 585 and 617, so that constraint does not apply here.

smartcontract/programs/doublezero-serviceability/tests/feed_lifecycle_test.rs
```diff
@@ -531,3 +538,247 @@ async fn test_a_staked_feed_cannot_resume_without_its_mirror() {
...
+    get_account_data(banks_client, feed)
+        .await
+        .expect("the feed should exist")
+        .get_feed()
+        .expect("it should be a feed")
+}
+
+/// Retiring an active feed starts the notice its seat holders are owed.
+#[tokio::test]
+async fn test_retiring_an_active_feed_starts_the_notice() {
```
ben-malbeclabs, line 586: **Important** — `require_feed_writer` has no negative test on `RetireFeed`, so nothing proves the repo's most irreversible feed transition is gated at all, and its builder shortcut is never exercised for retire. Every retire in this file signs with the globalstate authority.

`test_the_builder_is_authorized_on_its_own_feed` at line 297 proves a stranger gets `NotAllowed` on halt, but a wiring mistake in `retire.rs` alone would pass the suite.

smartcontract/programs/doublezero-serviceability/tests/feed_lifecycle_test.rs
```diff
@@ -531,3 +538,247 @@ async fn test_a_staked_feed_cannot_resume_without_its_mirror() {
...
+    let (mut banks_client, program_id, payer, globalstate, feed) = catalog_feed("leaving").await;
```
ben-malbeclabs, line 587: **Important** — `Halted` to `Retiring` is untested, the only supported transition in the state machine with no test at all; every retire test starts from a catalog or active feed. `cluster_with_halted_feed` at line 385 already builds the fixture for the resume tests.

smartcontract/programs/doublezero-serviceability/tests/feed_lifecycle_test.rs
```diff
@@ -531,3 +538,247 @@ async fn test_a_staked_feed_cannot_resume_without_its_mirror() {
...
+        &mut banks_client,
+        recent_blockhash,
+        program_id,
+        DoubleZeroInstruction::RetireFeed(FeedRetireArgs {}),
+        feed_accounts(feed, globalstate),
+        &payer,
+    )
+    .await;
+
+    let f = feed_of(&mut banks_client, feed).await;
```
ben-malbeclabs, line 612: **Important** — no test asserts the whole `Feed` after either new transition, though it derives `PartialEq`, and `halted_by` is asserted nowhere on the retire or finalize paths. Retiring a feed halted by an operator could silently drop or keep that key with no test either way.

smartcontract/programs/doublezero-serviceability/tests/feed_lifecycle_test.rs
```diff
@@ -531,3 +538,247 @@ async fn test_a_staked_feed_cannot_resume_without_its_mirror() {
...
+    );
+}
+
+/// Once the notice has elapsed anyone may finish the retirement.
+///
+/// Permissionless on purpose. The clock already decided, so this instruction can only agree with
+/// it, and requiring an authority would let a feed sit in `Retiring` forever because whoever held
+/// the key stopped caring. The signer here holds nothing.
+#[tokio::test]
+async fn test_anyone_finalizes_once_the_notice_has_elapsed() {
```
ben-malbeclabs, line 770: **Important** — nothing tests that `Retired` is terminal: neither instruction is attempted on a retired feed, so the `Retired` arm of the `matches!` guard at `retire.rs:65` is unreached. This test already leaves a retired feed in hand after line 767.

smartcontract/programs/doublezero-serviceability/src/processors/feed/retire.rs
bgm-malbeclabs: Fixed, both halves. `require_feed_admits` now takes `Active | Retiring`, so a holder who reconnects on day one keeps the service the notice promises. `SetAccessPassFeeds` refuses a feed that is `Retiring` or `Retired` unless the pass already carries a seat on it, so re-provisioning an existing holder still works and nobody buys in during the notice.

Neither half had a test, and I checked that by reverting each: the suite stayed green. Now `test_a_retiring_feed_still_admits_its_seat_holders` (`feed_metro_gate_test.rs`) and `test_a_retiring_feed_keeps_its_seats_and_sells_no_new_one` (`set_access_pass_feeds_test.rs`) each fail on their own mutation.

smartcontract/programs/doublezero-serviceability/tests/feed_lifecycle_test.rs
bgm-malbeclabs: Fixed in `3e7dad10`. Both assertions now read the clock either side of the transaction and require `retires_at` to land inside `before + NOTICE ..= after + NOTICE`, so a zero, a default, and a multiplied constant all fail.

CHANGELOG.md
bgm-malbeclabs: Trimmed to 56 words in `3e7dad10`: the two instructions, `Retiring`, `retires_at`, the thirty days and the three error codes. The reasoning is in the PR body.

sdk/serviceability/testdata/fixtures/generate-fixtures/src/main.rs
bgm-malbeclabs: Fixed in `3e7dad10`. The fixture is now `Retiring` with `retires_at: -1_764_547_200`, and `HaltedBy` and `RetiresAt` are in `FixtureMeta`, asserted in both the Python and TypeScript feed tests.

Negative on purpose, so a signed read is distinguishable from an unsigned one. I checked both are pinned by mutating them: replacing the Python decoder value with a constant fails one test, and changing the TypeScript `asIntN` to `asUintN` fails one test.

smartcontract/programs/doublezero-serviceability/src/state/feed.rs
bgm-malbeclabs: Fixed in `3e7dad10`. Two tests in the module at the foot of the file: one round-trips a `Retiring` feed with a nonzero `retires_at`, and one walks every status and asserts its discriminant byte in the encoding. Both index the status byte from the end of the buffer (`len - 32 - 8 - 1`), so appending a field later moves the index with it rather than silently reading the wrong byte.

smartcontract/programs/doublezero-serviceability/tests/feed_lifecycle_test.rs
bgm-malbeclabs: Added `test_a_pending_feed_retires_and_finalizes_end_to_end` and `test_a_real_notice_cannot_be_finalized_early` in `3e7dad10`. The first runs retire then finalize with no clock warp; the second catalogs a feed, retires it against the real thirty days, and gets `FeedRetirementNotElapsed` from the timestamp retire actually wrote.

smartcontract/programs/doublezero-serviceability/tests/feed_lifecycle_test.rs
bgm-malbeclabs: Added `test_a_stranger_cannot_retire_a_feed` in `3e7dad10`. A funded stranger gets `NotAllowed`, then the feed's own builder retires it in the same bank, so the refusal is about the signer and not about the instruction being misrouted.

smartcontract/programs/doublezero-serviceability/tests/feed_lifecycle_test.rs
bgm-malbeclabs: Added `test_a_halted_feed_retires_and_keeps_its_halter` in `3e7dad10`, built on `cluster_with_halted_feed`. It asserts `halted_by` still names the operator after the transition, which is the part that could have been dropped silently.

smartcontract/programs/doublezero-serviceability/tests/feed_lifecycle_test.rs
bgm-malbeclabs: Covered in `3e7dad10`. `test_a_halted_feed_retires_and_keeps_its_halter` asserts `halted_by` across retire, and `test_a_pending_feed_retires_and_finalizes_end_to_end` asserts the fields finalize writes and leaves alone.

smartcontract/programs/doublezero-serviceability/tests/feed_lifecycle_test.rs
bgm-malbeclabs: Added `test_retired_is_terminal` in `3e7dad10`. It retires and finalizes, then tries retire, halt, resume and finalize on the retired feed and asserts each returns its own error code rather than one shared failure.

smartcontract/programs/doublezero-serviceability/src/processors/feed/retire.rs
bgm-malbeclabs: Keeping the builder, and the Critical above is what makes that defensible.

RFC-28 gives the builder retirement as its own lever: a builder that wants out of running a feed should be able to stop, and needing a foundation signature to stop serving turns an exit into a support request. I cannot paste the text here, but builder-initiated retirement is intended rather than an oversight.

Your objection was sound against the code as it stood, though, and for a reason neither of us framed as authorization: retirement was terminal *and* immediate, so a builder ended a paid feed the instant it signed. With `3e7dad10` the notice does what it is named for. Existing holders keep service for thirty days, no new seat is sold in that window, and the builder cannot shorten it. The builder chooses when the thirty days start, not whether holders get them.

The bond is the builder's own stake, so locking it is a cost to the signer rather than to anyone else.

`delete.rs` is a different thing and I would not loosen it. Retire is a state change with notice; delete destroys the account and refunds rent, with no window in which anyone can notice.

smartcontract/programs/doublezero-serviceability/src/instructions.rs
```diff
@@ -265,8 +266,10 @@ pub enum DoubleZeroInstruction {
 
     WriteStakeMirror(StakeMirrorWriteArgs), // variant 119
 
-    HaltFeed(FeedHaltArgs),     // variant 120
-    ResumeFeed(FeedResumeArgs), // variant 121
+    HaltFeed(FeedHaltArgs),                             // variant 120
+    ResumeFeed(FeedResumeArgs),                         // variant 121
+    RetireFeed(FeedRetireArgs),                         // variant 122
+    FinalizeFeedRetirement(FeedFinalizeRetirementArgs), // variant 123
```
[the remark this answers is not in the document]
bgm-malbeclabs: Added in `0ed11387`. `retire_feed` takes `[feed, globalstate]` on the permission-appending path with the other gated feed verbs. `finalize_feed_retirement` takes `[feed]` and goes through `common::build` instead: it is permissionless, so there is no authority to check and no globalstate to send. The builder test now asserts both tags and both account lists.

smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set_feeds.rs
```diff
@@ -146,7 +146,22 @@ pub fn process_set_access_pass_feeds(
...
+        // the pass is untouched, so this refuses adding one rather than keeping one.
+        if matches!(feed.status, FeedStatus::Retiring | FeedStatus::Retired)
+            && !prior_seats.iter().any(|s| s.feed_key == feed_key)
+        {
+            msg!(
+                "Feed {} is {}, so it takes no new seat",
+                feed_key,
+                feed.status
+            );
+            return Err(DoubleZeroError::FeedNotActive.into());
```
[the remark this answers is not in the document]
bgm-malbeclabs: Added in `ec84f65b1`, which landed after this review: `test_a_retiring_feed_keeps_its_seats_and_sells_no_new_one` in `set_access_pass_feeds_test.rs`. It provisions two feeds, retires one, re-provisions the holder's pass and asserts both seats survive, then refuses a second pass a seat on the retiring feed and gives it one on the live feed to show the refusal is about the feed. Dropping the guard fails that test alone.

smartcontract/programs/doublezero-serviceability/src/processors/feed/retire.rs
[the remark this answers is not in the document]
bgm-malbeclabs: You are right, and fixed in `0ed11387`. `DeleteFeed` refuses a `Retiring` feed with a new `RetiringFeedCannotBeDeleted` (129).

Worth naming what this is and is not. A `FEED_AUTHORITY` key could already delete an active catalog feed outright, with no notice at all, so this does not stop an authority ending a feed abruptly. What it stops is the contradiction: once retirement has started, thirty days is what `Retiring` means, and an instruction that closes the account mid-notice empties the state of that meaning.

`test_a_retiring_feed_cannot_be_deleted_before_its_notice_ends` covers it, and `test_a_retired_feed_deletes` covers the other side, since a guard that outlived the notice would strand a finished retirement with no way to clean it up.

smartcontract/programs/doublezero-serviceability/tests/feed_lifecycle_test.rs
```diff
@@ -531,3 +538,500 @@ async fn test_a_staked_feed_cannot_resume_without_its_mirror() {
...
+/// Seeded rather than driven so the notice can be placed in the past or the future without moving
+/// the validator's clock. The comparison against `Clock` is what these tests are about; how the
+/// timestamp got there is `retire`'s business and has its own tests above.
+fn retiring_feed(
+    program_id: Pubkey,
+    code: &str,
+    exchange: Pubkey,
+    builder: Pubkey,
+    retires_at: i64,
+) -> (Pubkey, Vec<u8>) {
```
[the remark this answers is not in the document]
bgm-malbeclabs: Partly, and the real paths are there now. `ec84f65b1` added `test_a_pending_feed_retires_and_finalizes_end_to_end`, which drives `RetireFeed` then `FinalizeFeedRetirement` with no seeding, and `test_a_real_notice_cannot_be_finalized_early`, which retires a catalog feed against the full thirty days and gets `RetirementNoticeNotElapsed` from the timestamp `retire` actually wrote. `test_retiring_an_active_feed_starts_the_notice` pins that timestamp to a clock window either side of the transaction.

The seeded helper stays for the three tests that need `retires_at` in the past or the future without moving the validator clock. Reaching those through `RetireFeed` would mean warping the bank clock thirty days, which is slower and tests less: what those cases are about is the comparison against `Clock`, not how the timestamp got written.

ben-malbeclabs (approved):

PR #4325, opened 2026-09-11
release: push tags with the app token, not a personal token

bgm-malbeclabs:
## Summary of Changes

- The testnet tag job checks out the repo with a token minted from the release bot app instead of `DOUBLEZERO_PAT`. The personal token was set on 2025-09-10, a fine-grained token lives at most 366 days, and its expiry failed the v0.40.0 release at the checkout step in `push-tags`: `could not read Username for 'https://github.com': terminal prompts disabled`, exit 128.
- An app token has no calendar expiry, so this failure mode is gone rather than reset for another year. Every other job in the release already uses this app.
- `GITHUB_TOKEN` is not an option here: a tag pushed with it starts no workflow, so the package builds that `verify-cloudsmith` waits for would never run. A tag pushed with an app token does start them.
- The token is scoped to `doublezero` alone, narrower than the `doublezero,infra` scope the other jobs ask for.
- Both callers of the reusable workflow pass the two app secrets in place of the PAT: the release orchestrator and the manual components dispatcher.

After this merges, `DOUBLEZERO_PAT` can be deleted from the repository secrets. Nothing in `.github/` reads it.

The release bot app needs **Contents: write** on this repo, which it already has; it pushes the version-bump branches in `open-prs` with the same token.

## Diff Breakdown

| Config/build      |     3 | +22    / -5     |    +17 |
| Docs              |     1 | +1     / -0     |     +1 |
| **Total**         |     4 | +23    / -5     |    +18 |

All workflow wiring, no product code.

<details>

- `.github/workflows/release.testnet.push.tags.yml` — mints the app token and checks out with it; the `secrets:` contract takes the app id and private key instead of the PAT
- `.github/workflows/release.testnet.yml` — the orchestrator passes the two app secrets
- `.github/workflows/release.testnet.push.tags.components.yml` — the manual dispatcher passes the two app secrets

</details>

## Testing Verification

- `actionlint` on all three workflows, clean. This matters more than usual because a reusable workflow's `secrets:` contract is only checked at call time: a caller left on the old name fails the whole release at the point it can least afford to.
- Confirmed both callers were updated; `grep -r DOUBLEZERO_PAT .github/` returns nothing.
- Checked the repo has no tag ruleset, so no actor allowlist blocks the app from pushing `<component>/vX.Y.Z`.
- Not exercised end to end. The honest test is a dry run of `release.testnet` with `dry_run: true`, which runs the tag job through checkout and stops before the push. Worth doing before the next real release.


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
...
   - The Agave toolchain install retries, and a failed one now fails the job. Eight workflow steps across five workflows ran `sh -c "$(curl -sSfL .../install)"` once with no retry, so a transient reset from `release.anza.xyz` killed a job before it ran anything. That form also swallowed a failed fetch: the command substitution comes back empty, `sh -c ""` exits 0, and the step passed having installed nothing. The eight are now one composite action at `.github/actions/solana-toolchain` that fetches and runs as separate steps, checks `solana --version` actually runs, clears partial state between attempts, bounds every wait so a stalled handshake or hung transfer reaches the backoff instead of sitting until the job times out, and backs off. Each caller keeps the version it used before; `solana.yml` and `offchain.local-validator.yml` are on v3.0.12 and the other three on v3.0.4, which is drift worth settling separately.
```
martinsander00, line 15: Could we instead clarify that the app credentials have no PAT-style calendar expiry? Each minted installation token still expires after about one hour.

.github/workflows/release.testnet.push.tags.yml
```diff
@@ -55,11 +58,23 @@ jobs:
...
+      # starts no workflow, so the package builds this release waits for would
+      # never run. Not a personal token either, because those expire after a
+      # year and take the release down with them.
+      - uses: actions/create-github-app-token@v2
+        id: app
+        with:
+          app-id: ${{ secrets.RELEASE_BOT_APP_ID }}
+          private-key: ${{ secrets.RELEASE_BOT_PRIVATE_KEY }}
+          owner: malbeclabs
+          repositories: doublezero
```
martinsander00, line 71: Could we instead add `permission-contents: write` to this `create-github-app-token` step? Without it, the token inherits every permission granted to the app.

CHANGELOG.md
bgm-malbeclabs: Reworded in 8fe09f4. The entry now says the app credentials carry no PAT-style calendar expiry, and that each run mints its own installation token, which lives about an hour and is scoped to `contents: write`.

.github/workflows/release.testnet.push.tags.yml
bgm-malbeclabs: Added in 8fe09f4, with the same comment `release.docker.client.yml` carries. The app needs `contents: write` here anyway, for both the checkout and the tag push, so nothing else was being used.

martinsander00 (approved):

PR #4327, opened 2026-09-13
serviceability: prefer the accesspass that clears the epoch check

### LINKED ISSUE #4244: sdk: GetAccessPassCommand should prefer a usable pass, not just the dynamic one
`GetAccessPassCommand::execute` ([`get.rs:26-38`](https://github.com/malbeclabs/doublezero/blob/main/smartcontract/sdk/rs/src/commands/accesspass/get.rs#L26-L38)) returns the `0.0.0.0` pass whenever that account exists, and falls back to the exact-IP PDA only when it is absent. It never considers whether the pass it returns is *usable*.

So a payer holding an epoch-stale `0.0.0.0` pass **and** a valid prepaid pass at their exact IP is turned away: `doublezero connect ibrl` gets the stale dynamic pass, the `last_access_epoch >= epoch` check fails, and it prints `Unable to find a valid AccessPass` — even though `processors/user/create_core.rs` accepts either PDA and would have taken the exact-IP one.

An `EdgeSeat` pass is a realistic source of the stale half: multicast is not epoch-gated (`epoch_allows_connection` / `UserType::is_epoch_gated`), so nothing keeps its `last_access_epoch` current.

The fix belongs in `GetAccessPassCommand`, not in a caller's pre-flight. `CreateUserCommand` ([`user/create.rs:36-38`](https://github.com/malbeclabs/doublezero/blob/main/smartcontract/sdk/rs/src/commands/user/create.rs#L36-L38)) resolves the accesspass account through the same command, so whatever resolution a pre-flight trusts has to be the resolution that picks the PDA the transaction sends — otherwise the check and the transaction disagree. That means the command needs an epoch read so it can evaluate both candidates and prefer one that passes the epoch check.

Reported by @juan-malbeclabs in review of #4230, which does not change this behavior.

ayushsingh82:
## Problem

`GetAccessPassCommand::execute` ([`get.rs:26-38`](https://github.com/malbeclabs/doublezero/blob/main/smartcontract/sdk/rs/src/commands/accesspass/get.rs#L26-L38)) returns the dynamic (`0.0.0.0`) access pass whenever that account exists, and falls back to the exact-IP PDA only when it is absent — it never checks whether the pass it returns is *usable*.

A payer holding an epoch-stale dynamic pass **and** a valid prepaid pass at their exact IP gets turned away: `doublezero connect ibrl` resolves to the stale dynamic pass, the `last_access_epoch >= epoch` check fails, and it prints `Unable to find a valid AccessPass` — even though [`processors/user/create_core.rs`](https://github.com/malbeclabs/doublezero/blob/main/smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs#L308) would have accepted either PDA and taken the exact-IP one. An EdgeSeat pass is a realistic source of the stale dynamic half, since multicast is never epoch-gated (`UserType::is_epoch_gated`), so nothing keeps its `last_access_epoch` current.

Closes #4244.

## Fix

New `GetAccessPassCommand::execute_usable(client, user_type)` evaluates both the dynamic and exact-IP candidates against `epoch_allows_connection` — the same check the on-chain program enforces — and prefers whichever one clears it, falling back to the dynamic pass (matching `execute`'s default) only when neither does. It only reads the current epoch when *both* candidates exist, so it's a no-op cost change for the common single-candidate case. `execute` itself, and its existing test coverage, are untouched — every other caller of `GetAccessPassCommand` is unaffected.

`CreateUserCommand` and `CreateSubscribeUserCommand` (the SDK commands that actually build the `create_user`/`create_subscribe_user` transaction) switch to `execute_usable`. So does `doublezero connect`'s own preflight: per the issue, "whatever resolution a pre-flight trusts has to be the resolution that picks the PDA the transaction sends" — `LedgerClient::get_accesspass` gained a `user_type` parameter, threaded through `check_accesspass`/`require_accesspass` in `crates/doublezero-daemon-cli/src/connect.rs`, so the bare `doublezero connect`'s own epoch re-check (which reads the same `AccessPass` `require_accesspass` returns) sees the same pass the IBRL leg's transaction will use.

## Testing Verification

- `smartcontract/sdk/rs/src/commands/accesspass/get.rs`: 4 new tests for `execute_usable` — prefers the exact-IP pass when the dynamic one is epoch-stale (the reported bug), prefers dynamic when both are usable, falls back to dynamic when neither clears the epoch check, and confirms no epoch read happens when only one candidate exists (would panic on the unmocked `get_epoch()` otherwise). `cargo test -p doublezero_sdk` — 216/216 passing, including all pre-existing `execute()` coverage unmodified.
- `cargo test -p doublezero-daemon-cli` — 223/223 passing (covers `connect.rs`'s dispatch, bare-connect, and preflight paths through the updated `LedgerClient` trait).
- `cargo test -p doublezero-serviceability-cli` — 437/437 passing (covers the `CliCommand::get_accesspass_usable` addition).
- `cargo clippy --all-targets -- -Dclippy::all -Dwarnings` clean on every touched crate (`doublezero_sdk`, `doublezero-daemon-cli`, `doublezero-serviceability-cli`, `doublezero`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 ### Changes
 
+- Client / SDK
+  - `connect` and `GetAccessPassCommand`'s callers (`CreateUserCommand`, `CreateSubscribeUserCommand`) now prefer whichever of the dynamic (`0.0.0.0`) and exact-IP access passes actually clears the epoch check for the user type being created, instead of always taking the dynamic pass when one exists. A dynamic pass left over from a never-epoch-gated EdgeSeat multicast subscription could go stale while a valid exact-IP prepaid pass sat unused at the same address, so `doublezero connect ibrl` picked the stale pass and failed with "Unable to find a valid AccessPass" even though a usable one existed. New `GetAccessPassCommand::execute_usable` (only reads the epoch when both candidates exist); `execute` is unchanged for every other caller. (#4244)
+
 ## [v0.40.0](https://github.com/malbeclabs/doublezero/compare/client/v0.39.0...client/v0.40.0) - 2026-09-11
 
 ### Breaking
--- a/client/doublezero/src/main.rs
+++ b/client/doublezero/src/main.rs
@@ -100,14 +100,16 @@ impl<C: CliCommand + Sync> doublezero_daemon_cli::LedgerClient for LedgerAdapter
         &self,
         client_ip: std::net::Ipv4Addr,
         user_payer: solana_sdk::pubkey::Pubkey,
+        user_type: doublezero_sdk::UserType,
     ) -> eyre::Result<Option<doublezero_serviceability::state::accesspass::AccessPass>> {
         Ok(self
             .client
-            .get_accesspass(
+            .get_accesspass_usable(
                 doublezero_sdk::commands::accesspass::get::GetAccessPassCommand {
                     client_ip,
                     user_payer,
                 },
+                user_type,
             )?
             .map(|(_, accesspass)| accesspass))
     }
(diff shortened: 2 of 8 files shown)


ayushsingh82:
@elitegreg @juan-malbeclabs could you take a look when you have a chance?

PR #4328, opened 2026-09-13
serviceability: prune User test literals in multicastgroup/tenant/user commands

ayushsingh82:
## Problem

PR #4080 added a manual `impl Default for User` and switched test literals across `smartcontract/cli`, `smartcontract/sdk/rs`, and `crates/doublezero-daemon-cli` to end with `..Default::default()` — but as merged, the struct-update tail only absorbed `feed_pks`; every literal still named the other ~20 fields explicitly. #4084 asks to lean into the pattern: prune each literal down to only the fields a test actually asserts on or that drive the command's behavior.

#4235 did the first slice (`smartcontract/cli/src/user/*`). This is the second slice: `smartcontract/sdk/rs`.

## Changes

Pruned `User` literals in:
- `commands/tenant/delete.rs`
- `commands/multicastgroup/{subscribe,subscribe_feed,unsubscribe_feed}.rs`
- `commands/user/{delete,requestban,update}.rs`

For each literal I traced which fields the command under test actually reads (or which the test's own mock setup independently depends on) before dropping the rest — e.g. `DeleteUserCommand` reads `owner`/`device_pk`/`tenant_pk`/`publishers`/`subscribers` but never `user_type`/`cyoa_type`/`dz_ip`/`tunnel_net`/etc., so only the former survive per-test. `commands/accesspass/get.rs` was already at this shape (from the earlier #4244 work), so it's untouched here.

`crates/doublezero-daemon-cli` is **not** included — `connect.rs` alone has ~70 more `User` literals, well past this repo's ~500-line PR guidance, so that's left for a follow-up PR.

## Testing Verification

- `cargo test -p doublezero_sdk --lib`: 212/212 passing (no literal's pruning changed a test's outcome).
- `cargo clippy -p doublezero_sdk --all-targets -- -Dclippy::all -Dwarnings`: clean.
- `cargo +nightly fmt -p doublezero_sdk -- --check`: clean.
- Removed now-unused imports (`NetworkV4`, `UserCYOA`, `UserStatus`, `UserType`, `Ipv4Addr`) file-by-file as literals no longer referenced them.


ayushsingh82:
@nikw9944 could you take a look when you have a chance?

PR #4329, opened 2026-09-13
release: publish aarch64/arm64 client packages alongside x86_64/amd64

### LINKED ISSUE #4176: packaging: publish aarch64 rpm/deb artifacts for the client
We publish apt and yum repositories but no ARM64 artifacts, so the client cannot be installed from a package on aarch64 hosts.

The Cloudsmith setup script succeeds and correctly detects the platform (`distro=amzn version=2 arch=aarch64`), the repo installs, and metadata fetches. The install then fails because the index is empty:

```
$ sudo yum install doublezero
No package doublezero available.
```

Every published artifact is `x86_64` (rpm) or `amd64` (deb), across all packages, not just `doublezero`. Package metadata identifiers are `binary, x86_64, rpm, any-distro/any-version`. Since packages publish as `any-distro/any-version`, distro is not the constraint — architecture is.

### This is packaging only, not a port

Verified on `client/v0.33.0`, AWS Graviton (`t4g.small`), aarch64, with Rust 1.97.1 (per `rust-toolchain.toml`) and Go 1.26:

- `doublezero` builds clean, reports `client version: 0.33.0`
- `doublezerod` builds clean, statically linked `ELF 64-bit ARM aarch64`
- Daemon starts, discovers its public IP, and `doublezero latency` returns successful probes against testnet devices (nyc, lax, lon, ams, fra), all `reachable: true`

No architecture-specific changes were needed.

### Ask

Add `aarch64`/`arm64` build targets to the release pipeline and publish alongside the existing x86_64 artifacts. Graviton and Ampere hosts are common enough that this comes up in evaluations.

Note `INSTALL.md` currently documents x86_64 only; the support matrix should be updated if this ships.

ayushsingh82:
## Problem

Closes #4176.

We publish apt and yum repositories but no ARM64 artifacts, so the client cannot be installed from a package on aarch64 hosts. Every published artifact is `x86_64` (rpm) / `amd64` (deb) — since packages publish as `any-distro/any-version`, architecture is the only real constraint. Verified on the issue itself: the client and daemon both build clean and run correctly on Graviton with no source changes needed — this is packaging only, not a port.

## Fix

- `doublezerod` (`CGO_ENABLED=0`, pure Go): add `arm64` to its `goarch` matrix — cross-compiles natively, no extra toolchain.
- Rust `doublezero` CLI: new `doublezero-cli-arm64` / `doublezero-cli-mainnet-beta-arm64` build entries targeting `aarch64-unknown-linux-musl`, using `cargo-zigbuild` instead of a musl-gcc cross toolchain. Reason: `aws-lc-sys`'s hand-written arm assembly build script feeds raw `--target=<rust-triple>` flags straight to the C compiler, which a plain `aarch64-linux-musl-gcc` (or a bare `zig cc` CC-shim) does not accept as-is — `cargo-zigbuild` rewrites that invocation for zig's own target syntax. The existing x86_64 build entries and their musl-gcc setup are untouched (separate build ids).
- Wired the new arm64 build ids into the existing `archives`/`nfpms` `ids:` lists so both the tarball and deb/rpm packages carry the arm64 binaries.
- CI (`release.client.yml` and the shared `release.daily.yml`) installs zig (pinned tarball from ziglang.org) + `cargo-zigbuild`, and adds the `aarch64-unknown-linux-musl` rustup target. In `release.daily.yml` — shared by several pure-Go daily-release components — this is gated on `inputs.component == 'client'` so unrelated components don't pay the cost.
- `client/INSTALL.md`: one-line note that packages now support both architectures.

## Testing Verification

- `cargo zigbuild --release --package=doublezero --bin=doublezero --target aarch64-unknown-linux-musl` (run locally): produces a statically-linked ARM64 ELF binary (`file` confirms `ELF 64-bit LSB executable, ARM aarch64 ... statically linked`).
- `CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build ./client/doublezerod/cmd/doublezerod/main.go`: produces a statically-linked ARM64 ELF binary.
- All touched YAML files re-validated for syntax.
- **Not verified**: the actual GoReleaser-Pro pipeline end-to-end (build-id grouping into archives/nfpm packages, the CI zig-install step, Cloudsmith publish) — I don't have goreleaser-pro or a Cloudsmith token to run this outside CI. Recommend a `--snapshot` dry run (or a real tag push to a test channel) to confirm the packaging/publish steps before this goes out on a real release.


ayushsingh82:
@armcconnell could you take a look when you have a chance?

PR #4330, opened 2026-09-13
docs: wait for the dpkg lock before the scripted deb install

### LINKED ISSUE #3540: client/install: scripted install hangs silently on fresh Ubuntu LTS hosts
## Summary

The scripted install path documented in `client/INSTALL.md` can hang indefinitely at `RUN: Attempting to install 'apt-transport-https' ...` on fresh Ubuntu LTS hosts. Restarting the script clears it up, so the impact is cosmetic — but the user sees zero feedback for several minutes and reasonably concludes the installer is broken.

Reported by a user during client install on Ubuntu LTS.

## Root cause

`client/INSTALL.md` directs users to pipe `https://dl.cloudsmith.io/public/malbeclabs/doublezero/setup.deb.sh` to bash. That script (a Cloudsmith template, not ours) does:

```bash
apt-get install -y "$tool" &>$tmp_log
```

with all output redirected into a throwaway log. On a freshly-booted Ubuntu LTS, `unattended-upgrades.service` holds `/var/lib/dpkg/lock-frontend`, so `apt-get install` blocks waiting for the lock. apt's default `DPkg::Lock::Timeout=-1` means wait forever, and apt's normal `Waiting for cache lock: held by process N (unattended-upgr)...` message is swallowed by the redirect. The user sees no output and assumes a hang.

It would eventually unblock on its own (typically within ~20 min, once the background upgrade finishes). On retry it usually skips the install entirely because `dpkg -l | grep apt-transport-https` now matches.

## Options

We don't own the Cloudsmith template, but we have real options:

1. **Ship our own `setup.deb.sh`** hosted from our release pipeline. Gives us full control — wait for the lock with a visible message, use a real timeout, don't swallow output, optionally skip `apt-transport-https` entirely on modern distros where it's a transitional empty package (Ubuntu 18.04+, Debian 10+).
2. **Cloudsmith template customization** if our plan supports it, or file an upstream issue/PR against Cloudsmith's template.
3. **Doc-only mitigation** in `client/INSTALL.md`:
   - Promote the manual install path above the scripted one for fresh hosts (the manual steps don't swallow apt's lock-wait output).
   - Add a preflight one-liner before the `curl | bash`:
     ```bash
     while sudo fuser /var/lib/dpkg/lock-frontend >/dev/null 2>&1; do
       echo "waiting for apt lock..."; sleep 2
     done
     ```

Cheapest fix with real impact is (3). (1) is the durable fix if we want a polished install UX.

## Repro

1. Boot a fresh Ubuntu 22.04 or 24.04 host.
2. Immediately (while `unattended-upgrades` is active) run:
   ```
   curl -1sLf 'https://dl.cloudsmith.io/public/malbeclabs/doublezero/setup.deb.sh' | sudo -E bash
   ```
3. Observe indefinite hang at `RUN: Attempting to install 'apt-transport-https' ...`.

ayushsingh82:
## Problem

Closes #3540.

The scripted install path documented in `client/INSTALL.md` can hang indefinitely at `RUN: Attempting to install 'apt-transport-https' ...` on fresh Ubuntu LTS hosts. Restarting the script clears it up, so the impact is cosmetic — but the user sees zero feedback for several minutes and reasonably concludes the installer is broken.

Root cause: `client/INSTALL.md` directs users to pipe `setup.deb.sh` (a Cloudsmith template, not ours) to bash. That script runs `apt-get install -y "$tool" &>$tmp_log`, redirecting all output to a throwaway log. On a freshly-booted Ubuntu LTS host, `unattended-upgrades.service` holds `/var/lib/dpkg/lock-frontend`, so `apt-get install` blocks waiting for the lock. apt's default `DPkg::Lock::Timeout=-1` means wait forever, and apt's normal `Waiting for cache lock: held by process N (unattended-upgr)...` message is swallowed by the redirect. It typically unblocks within ~20 min once the background upgrade finishes.

## Fix

Of the three options in the issue (ship our own `setup.deb.sh`, customize/PR the Cloudsmith template, or a doc-only mitigation), this takes the cheapest one with real impact: a preflight one-liner in `client/INSTALL.md` before the `curl | bash`, waiting for the dpkg lock and printing why, so the wait is visible instead of silent. The durable fix (our own install script with a real timeout and non-swallowed output) is a separate, bigger change — happy to track that separately if wanted.

## Testing Verification

Docs-only change; verified the added shell snippet's syntax and that it matches the exact one-liner proposed in the issue.


ayushsingh82:
@ben-dz could you take a look when you have a chance?

PR #4331, opened 2026-09-14
internet-latency-collector: rank lossy ripe atlas targets last and split the probe blacklist by role

elitegreg:
## Problem

The collector kept one "unresponsive" list for two different failures: a probe that stops running
measurements as a source, and a probe that stops answering pings as a target. A target
failure therefore also barred the probe from sourcing.

Separately, a target that answered only a fraction of the pings aimed at it never tripped
the staleness check, which asks only whether a measurement exported anything in the last
hour.

And when every candidate for a metro was marked, excluding them all emptied the candidate
set, so no measurement was wanted for that metro and reconciliation deleted the one it had.

## Why it matters

A metro with no measurement is a coverage gap that pages. A metro whose probe is needlessly
barred from sourcing cascades measurement rebuilds into every other measurement it feeds.

## Fix

- Split the blacklist by role: a target failure bars targeting only, a source failure bars both.
- Measure per-target loss in a rolling window and mark a target that stays above the threshold.
- Rank marked targets last instead of excluding them, so a metro always keeps a target.
- Persist target marks across restarts.


controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -601,12 +626,22 @@ func (c *Collector) exportSingleMeasurementResults(ctx context.Context, measurem
...
 	// Process results - use slice to preserve all samples
 	var records []exporter.Record
 	for _, result := range results {
 		// Parse latency from result (now also returns probe ID)
 		latency, timestamp, probeID := c.parseLatencyFromResult(result)
 
+		targetAttempts++
+		if latency > 0 {
+			targetSuccesses++
+		}
```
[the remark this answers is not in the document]
elitegreg: Good catch, confirmed and fixed in deb4f86df.

The export cursor advances only past results carrying a latency, so every timeout newer than the last success came back from each incremental query and was tallied again — six times an hour at the 10 minute export interval. That inflates the ratio for a target that is merely losing its most recent pings, which is exactly the target we would not want to rotate away from.

Fixed as suggested with a separate cursor rather than by touching the export cursor: `MeasurementMeta.TargetLossCursor` records the newest result already counted, and `exportSingleMeasurementResults` only tallies results newer than it. Stale detection keeps the export cursor it depends on. Results are compared at one second granularity, so a result sharing the cursor's second is skipped — undercounting biases away from blacklisting a usable target, which is the safe direction here.

The other findings in the review are addressed in the same commit:

- Counters now update after `WriteRecords` succeeds, so a failed export no longer recounts the batch.
- `EvaluateTargetLoss` resets only once a window has enough attempts to be judged. Previously a measurement with fewer than five sources could never reach 30 attempts in an hour and so was never judged at all.
- Target and source exclusions are now separate lists. A probe that fails as a target is barred from target selection only and stays available to source measurements, since NAT breaks inbound rather than outbound. This changes pre-existing behavior, so `ConfigureMeasurements_UnresponsiveTargetDoesNotBlacklistSources` was updated to assert the new split.
- Added `ExportSingleMeasurementResults_RecordsTargetLoss` (tallies, replayed timeouts not recounted, failed export counting nothing) and `ConfigureMeasurements_LossyTargetIsRotated` for the full path.

nikw9944 (approved):

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -1361,12 +1444,17 @@ func (c *Collector) fetchFallbackProbesForUnresponsiveLocations(ctx context.Cont
 		if len(match.NearbyProbes) == 0 {
 			continue
 		}
-		if len(filterResponsiveProbes(match.NearbyProbes, measurementState)) > 0 {
-			continue // at least one probe is still responsive — no fallback needed
+		hasUnmarked := false
+		for _, probe := range filterSelectableTargets(match.NearbyProbes) {
+			if !measurementState.IsTargetUnresponsive(probe.ID) {
```
juan-malbeclabs, line 1540: The role split does not survive this path, so the cascade the PR targets is still reachable.

This condition now fires on a target-only mark, and line 1474 then *replaces* the location's probe list: `result[i].NearbyProbes = filterValidProbes(GetProbesInRadius(..., anchorsOnly=false))`. `NearbyProbes` is also exactly what `generateWantedMeasurements` hands to `filterResponsiveProbes` for source selection (collector.go:1548).

Scenario, the xsin shape from `..._UnresponsiveTargetDoesNotBlacklistSources` but with xsin's single anchor P and no second candidate:

1. P stops answering pings, so the staleness check calls `AddUnresponsiveTarget(P)`.
2. Next cycle every xsin candidate is marked, so this branch runs the wider fetch. `anchorsOnly=false` returns every Connected probe in radius, a superset of the anchors whose nearest member is very likely a non-anchor Q rather than P.
3. `generateWantedMeasurements` then picks Q as xsin's **source** probe for every measurement whose target sorts before xsin. Those compare as "outdated source probes", and Step 7 stops and recreates them.

So a target-only failure still changes the metro's source probe and tears down unrelated measurements, one indirection later than before. Carrying the fallback probes in a separate field consumed only by `rankTargets` would close it.

Not introduced here (pre-PR the same trigger fired through `filterResponsiveProbes`), but the split is what makes it fixable, and it is untested: the new `..._TriggeredByTargetMarks` test returns an empty probe set, so the replacement branch never runs.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -20,8 +35,18 @@ type MeasurementState struct {
 }
 
 type MetadataTracker struct {
-	Metadata           map[int]MeasurementMeta  `json:"metadata"`
+	Metadata map[int]MeasurementMeta `json:"metadata"`
+
+	// UnresponsiveProbes holds probes that failed as a measurement source, meaning
+	// they stopped running measurements at all. Such a probe is excluded from both
+	// source and target selection.
```
juan-malbeclabs: `Such a probe is excluded from both source and target selection` no longer describes the code. `rankTargets` (collector.go:220) consults `IsTargetUnresponsive`, which is the union of both lists, and only *orders* the union last; it never drops anything. A probe in `UnresponsiveProbes` is therefore still selected as a target when it is a metro's only candidate, which the `a probe unresponsive as a source is ranked last too` subtest pins exactly. Suggest "ranked last in target selection", so the next reader does not treat a source failure as a hard target bar.

Related, on the same split: `GetUnresponsiveTargets` returns only `UnresponsiveTargets` while `IsTargetUnresponsive` is the union, so the `unresponsive_target_count` field logged at collector.go:1492 under-reports how many probes are actually demoted.

controlplane/internet-latency-collector/internal/ripeatlas/collector_test.go
```diff
@@ -1578,8 +1622,12 @@ func TestInternetLatency_RIPEAtlas_ConfigureMeasurements_UnresponsiveTargetDoesN
 	err := c.configureMeasurements(t.Context(), locationMatches, false, 1, stateDir, 1*time.Minute)
 	require.NoError(t, err)
 
-	// Probe 6726 should be marked unresponsive (target staleness via Phase 1)
-	require.True(t, c.measurementState.IsProbeUnresponsive(6726), "Target probe 6726 should be marked unresponsive")
+	// Probe 6726 should be barred from target selection (target staleness via Phase 1)
+	require.True(t, c.measurementState.IsTargetUnresponsive(6726), "Target probe 6726 should be barred as a target")
```
juan-malbeclabs, line 1981: One assertion short of pinning the PR's headline claim. Probe 6726 is xsin's source probe in measurement 1001 (target xams), and the whole point of the split is that marking 6726 as a failed *target* must leave 1001 alone. The test asserts `Contains(stoppedMeasurements, 1002)` but never asserts 1001 survived, so a regression that re-blacklisted the probe for sourcing (and so swapped xsin's source) would still pass here. `require.NotContains(t, stoppedMeasurements, 1001)` would lock it down.

CHANGELOG.md
```diff
@@ -43,6 +43,11 @@ All notable changes to this project will be documented in this file.
...
+  - Target selection ranks probes carrying an unresponsive-target mark last rather than dropping them, so a metro whose every candidate is marked keeps its measurement pointed at the nearest one. Dropping them emptied the metro's candidate set, and reconciliation then read the absent target as unwanted and deleted the measurement, trading partial data for none until the marks expired up to 24h later. (malbeclabs/doublezero#4182)
+  - Rotate a RIPE Atlas target probe that replies but drops most pings. The staleness check only asked whether a measurement exported anything in the last hour, so a target answering a small fraction of the time looked healthy indefinitely while its circuits took turns falling out of the freshness window. Ping outcomes now accumulate in a rolling hourly window per measurement, and a target above the loss threshold is marked, which ranks it behind any healthier candidate.
+  - Target-failure marks survive a collector restart. `Load` never decoded them, so the list came back empty and the next save wrote the empty list back over it.
```
juan-malbeclabs: This bullet describes a defect that only ever existed inside this branch. `unresponsive_targets` does not exist on main, and the missing decode was both introduced and fixed here (547dd5a), so no released collector ever lost target marks across a restart. As written a reader will go looking for the version that had the bug. Folding the persistence guarantee into the first bullet, or dropping this one, would read truer.

juan-malbeclabs (commented):
Reviewed the role split, the rolling loss window, the ranking change and the state format against the full files rather than the hunks.

Verified:
- `rankTargets` can never empty a metro's candidate set. `GetNearestSourcesSorted` returns an empty non-nil slice for `maxCount <= 0`, so the single-candidate, every-candidate-marked and empty-input cases all fall through correctly.
- Every read site is on the right role: `filterResponsiveProbes`/`IsProbeUnresponsive` for sources, `filterSelectableTargets`/`rankTargets`/`IsTargetUnresponsive` for targets, and Step 4b's skip correctly widened to the union.
- Loss window: the threshold direction is right (`> MaxTargetLossRatio`), the ratio is only computed once `attempts >= MinTargetAttemptsForLossCheck` so there is no divide-by-zero, and a short window accumulates instead of being discarded. With `Packets: 1` one result really is one ping, so the five-source arithmetic in the comment holds.
- State: a legacy `unresponsive_probes: [N, N]` file still loads, and `unresponsive_targets` plus the per-measurement tallies round-trip through `Save`/`Load`.
- No new shared state is read outside `ms.mu`; the added fields are scalars copied by value.

`go build ./controlplane/internet-latency-collector/...`, `go test`, `go test -race` and `go vet` on `internal/ripeatlas` all pass. The non-test production diff is about 300 added lines, within the repo's 500-line guidance.

Four comments below. The first is the one I would want resolved or explicitly deferred before merge: the source-side cascade this PR sets out to remove is still reachable through the fallback fetch.

CHANGELOG.md
```diff
@@ -43,6 +43,11 @@ All notable changes to this project will be documented in this file.
...
   - The Agave toolchain install retries, and a failed one now fails the job. Eight workflow steps across five workflows ran `sh -c "$(curl -sSfL .../install)"` once with no retry, so a transient reset from `release.anza.xyz` killed a job before it ran anything. That form also swallowed a failed fetch: the command substitution comes back empty, `sh -c ""` exits 0, and the step passed having installed nothing. The eight are now one composite action at `.github/actions/solana-toolchain` that fetches and runs as separate steps, checks `solana --version` actually runs, clears partial state between attempts, bounds every wait so a stalled handshake or hung transfer reaches the backoff instead of sitting until the job times out, and backs off. Each caller keeps the version it used before; `solana.yml` and `offchain.local-validator.yml` are on v3.0.12 and the other three on v3.0.4, which is drift worth settling separately.
+- Internet latency collector
```
juan-malbeclabs: These four bullets are inside the already-released `v0.40.0` section, so they will never reach a release.

Section offsets in this file: `## Unreleased` is line 5, `## [v0.41.0] - 2026-09-16` line 14, `## [v0.40.0] - 2026-09-11` line 34, `## [v0.39.0]` line 78. This new heading and its bullets sit at 46-50, inside the v0.40.0 block.

`scripts/release/bump-version.sh` promotes only the Unreleased body — line 35 requires `^## Unreleased$`, line 36 awks that section, line 40 inserts the new dated heading beneath it. So these notes are dropped from every future release and are retroactively attributed to a release dated 2026-09-11, before the code existed. The convention is to add entries under the existing `## Unreleased` / `### Changes` section rather than creating a new dated section.

Also: `## Unreleased` already carries a `- Collector` bullet at line 11 for this same RIPE Atlas work, and `- Collector` is the heading this file uses for the component throughout (lines 11, 224, 350, 721) — so these belong under that bullet rather than under a new `- Internet latency collector` one.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -885,7 +947,27 @@ func (c *Collector) configureMeasurements(ctx context.Context, locationMatches [
 					slog.String("reason", reason),
 					slog.Time("created_at", time.Unix(meta.CreatedAt, 0)),
 					slog.Time("last_export_at", time.Unix(meta.LastExportAt, 0)))
-				measurementState.AddUnresponsiveProbe(meta.TargetProbeID)
+				measurementState.AddUnresponsiveTarget(meta.TargetProbeID)
```
juan-malbeclabs: This swap removes the only mechanism that ever took a fully dead probe out of the **source** pool.

After this diff `AddUnresponsiveProbe` has exactly one production call site, collector.go:1011 in Step 4b, and it sits behind `if source.LastResponseAt == 0 { continue }` at line 998. `UpdateSourceProbeResponse` advances `LastResponseAt` for any uploaded result including a timeout, so a probe that uploads *nothing at all* never satisfies that gate and can never enter `UnresponsiveProbes`.

Scenario — the case `fetchFallbackProbesForUnresponsiveLocations` exists for: metro cmh's nearest probe P goes offline while RIPE still reports it `Connected`. P is cmh's target and, being nearest, cmh's source in every measurement whose target sorts before cmh.
1. cmh's measurement never exports, so this line marks P in `UnresponsiveTargets` only (pre-PR it went to `UnresponsiveProbes`).
2. `filterResponsiveProbes` (collector.go:1548) consults `IsProbeUnresponsive`, which does not read `UnresponsiveTargets`.
3. P is re-enlisted as cmh's source every cycle, forever, with `LastResponseAt` frozen at 0 — so Step 4b can never mark it either.

Every cmh-sourced circuit stays dark, with only the explicitly observation-only `ripeatlas_sources_without_samples` gauge as a signal. The rationale in the PR body, "failing to answer pings says nothing about sending them", is true for a NAT'd probe but false for an offline one, and the staleness signal at this line cannot tell the two apart. The loss check is what distinguishes them: a probe answering 15% is NAT-like and belongs in `UnresponsiveTargets`; a measurement that has *never* exported is the offline case and arguably still belongs in both lists.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -609,12 +638,36 @@ func (c *Collector) exportSingleMeasurementResults(ctx context.Context, measurem
...
+		// Results are counted at one second granularity, so a result sharing the
+		// cursor's second is skipped. Undercounting biases away from blacklisting a
+		// usable target, which is the safe direction to err in.
+		if resultAt := timestamp.Unix(); resultAt > lossCursor {
+			targetAttempts++
+			if latency > 0 {
+				targetSuccesses++
+			}
+			if resultAt > newestResult {
+				newestResult = resultAt
```
juan-malbeclabs, line 681: `newestResult` has no upper clamp, and it flows straight into `TargetLossCursor` in `RecordTargetResults`. One future-dated result therefore parks the cursor ahead of wall clock and permanently disables loss counting for that measurement.

`resultAt > lossCursor` at line 658 is the only gate, and `timestamp` comes from the probe-reported `timestamp` field in `parseLatencyFromResult`. A single **timeout** from a clock-skewed probe is enough, because only `latency > 0` advances `maxTimestamp`, so a timeout never touches the export cursor and nothing else notices. After that every subsequent real result fails `resultAt > lossCursor`, `attempts` stays where it was, and `EvaluateTargetLoss` returns `false` for the life of the measurement.

Clamping to the `now` already passed into `RecordTargetResults` closes it: `if resultAt > newestResult && resultAt <= now`.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -898,7 +980,7 @@ func (c *Collector) configureMeasurements(ctx context.Context, locationMatches [
 			// Skip measurements whose target is already marked unresponsive —
 			// source probes in these measurements will have stale LastResponseAt
 			// because the target isn't replying, not because the sources are broken
-			if measurementState.IsProbeUnresponsive(meta.TargetProbeID) {
+			if measurementState.IsTargetUnresponsive(meta.TargetProbeID) {
```
juan-malbeclabs: This guard used to be self-limiting; now that a marked target is deliberately kept, the skip is permanent and that measurement's source failures are never detected.

Before this PR a marked target dropped the location from `wantedMeasurements` and the measurement was removed, so the exemption ended within a cycle. With rank-last the measurement survives (that is what `..._LastMarkedCandidateIsKept` pins), and `IsTargetUnresponsive` is re-satisfied every cycle, so this `continue` fires forever.

That matters because sources are drawn only from locations sorting after the target, so a given probe is a source in a limited set of measurements — for the alphabetically **first** metro, the second location's probe is a source in exactly that one measurement and nowhere else. When it stops running measurements its `LastResponseAt` ages past `probeTimeout`, but `AddUnresponsiveProbe` at line 1011 is never reached, the probe is never rotated, and the circuit stays dark.

The intent here is "the target is not answering, so do not blame the sources", which is right — but it should key off *this measurement's* observed target health rather than a standing probe-level mark, so a permanently kept marked target does not permanently exempt its sources.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -11,6 +11,21 @@ import (
 const (
 	// UnresponsiveProbeExpiry is how long a probe stays blacklisted before being retried.
 	UnresponsiveProbeExpiry = 24 * time.Hour
+
+	// TargetLossWindow is how long loss against a target probe accumulates before the
+	// ratio is judged. It matches the staleness timeout so both checks reason over the
+	// same horizon.
+	TargetLossWindow = time.Hour
```
juan-malbeclabs, line 18: `TargetLossWindow` (1h) is exactly the default management interval (`defaultRipeAtlasMeasurementInterval = 1 * time.Hour`, cmd/collector/main.go:34), and the duration gate compares against a `currentTime` that drifts, so roughly half the cycles miss the window by seconds and slip the verdict a whole hour.

The ticker fires a fixed `measurementInterval` apart, but `currentTime` (collector.go:923) is cycle start **plus that cycle's pre-work**: `GetLocations`, `GetProbesForLocations` with its per-location `CallDelay` sleeps, the Step 2 fallback fetches with theirs, and `GetAllMeasurements`. That is seconds to tens of seconds, and different every cycle. `EvaluateTargetLoss` then resets `TargetWindowStart` to that same drifting value, so the next cycle judges only when the new pre-work is at least as long as the previous one; whenever a cycle happens to be faster, `now - start` lands at 3599 and the measurement is skipped for another full hour. Detection latency is therefore 1h or 2h at random, and the post-expiry re-mark in the `AddUnresponsiveTarget` comment can take 2h of serving bad data. Making the window comfortably shorter than the management interval, or comparing against the tick rather than mid-cycle wall clock, removes the coin flip.

Both new window tests seed `windowStart` at `-2 * TargetLossWindow`, so neither can observe this.

On `MinTargetAttemptsForLossCheck` just below: the comment derives 30 from "at the 10 minute sampling interval a source contributes 6 attempts an hour", but `samplingInterval` and `len(meta.Sources)` are both in scope at the call site. Raise `samplingInterval` to 1h and 30 attempts needs 30 sources; drop it to 1m and the ratio is judged on 30 seconds of data. Deriving the minimum from those two values would keep the constant honest across configurations.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -35,6 +60,20 @@ type MeasurementMeta struct {
 	Sources        []SourceProbeMeta `json:"sources"`
 	CreatedAt      int64             `json:"created_at"`
 	LastExportAt   int64             `json:"last_export_at,omitempty"`
+
+	// Rolling ping tallies against the target probe, reset each time the window is
+	// judged. Attempts counts every result a source uploaded; successes counts those
+	// that carried a latency back.
+	TargetWindowStart int64 `json:"target_window_start,omitempty"`
```
juan-malbeclabs, line 108: These tallies are keyed by measurement ID and are not carried across a recreation, so any unrelated churn silently discards the window.

Step 8 builds the new metadata at collector.go:1356 as `MeasurementMeta{TargetLocation, TargetProbeID, Sources, CreatedAt}` — none of `TargetWindowStart`, `TargetAttempts`, `TargetSuccesses` or `TargetLossCursor` carries over — and Step 7 calls `RemoveMetadata` on the old ID. Crucially Step 5 recreates on **any** source-set change (`sourcesMatch` at 1119 and 1168), not only on a target change.

So a measurement 55 minutes into a >50%-loss window loses that window because some other metro's nearest Connected probe flapped. With 3 sources a measurement needs roughly 2h to reach `MinTargetAttemptsForLossCheck`, so in a deployment where any metro's source probe flips more often than that, the loss check silently never fires — and nothing logs it, because `EvaluateTargetLoss` returns `false` indistinguishably from a healthy target. The PR body itself notes that one un-blacklisting "changed that metro's source probe in three unrelated measurements", and every such churn resets windows fleet-wide.

Either carry the `Target*` fields over when `spec.TargetProbe.ID` matches the old `meta.TargetProbeID`, or key the window by probe ID rather than measurement ID.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -221,35 +264,168 @@ func (ms *MeasurementState) UpdateSourceProbeResponse(measurementID int, probeID
...
+	}
+
+	attempts, successes = meta.TargetAttempts, meta.TargetSuccesses
+
+	// A window that closes short stays open and keeps accumulating rather than being
+	// reset. Resetting here would discard the evidence every hour, and a measurement
+	// with few enough sources never reaches the minimum within one window: at the
+	// 10 minute sampling interval a source contributes 6 attempts an hour, so fewer
+	// than 5 sources could never be judged at all.
+	if attempts < MinTargetAttemptsForLossCheck {
```
juan-malbeclabs, line 443: This early return never advances `TargetWindowStart`, so once a window closes short the duration gate at line 309 is permanently satisfied and the tallies accumulate without bound. The verdict is then rendered over an arbitrarily long span that mixes an old outage with current health.

Concretely, 2 sources at the 10-minute sampling interval give 12 attempts/hour:
- hour 1: 12 attempts, 1 success — below the minimum, not judged, window start untouched
- hour 2: 24 attempts, 2 successes — still below, not judged
- hour 3: the target fully recovers and answers all 12 — judged at 36 attempts / 14 successes, ratio 0.61 > 0.5 → `lossy = true`

A probe that has answered every single ping for the last hour is barred as a target for 24h and its measurement torn down. The comment above correctly argues against discarding the evidence every hour, but the fix wants an upper bound too: reset once `TargetWindowStart` is older than some multiple of `TargetLossWindow`, or age the tallies, rather than carrying them indefinitely.

The same missing reset also lets outage evidence survive a recovery through the stale path: while a measurement is stale the `continue` at collector.go:952 skips `EvaluateTargetLoss` entirely, yet RIPE keeps uploading timeouts and `RecordTargetResults` keeps adding to this same never-reset window.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -221,35 +264,168 @@ func (ms *MeasurementState) UpdateSourceProbeResponse(measurementID int, probeID
...
+		return false, attempts, successes
+	}
+
+	meta.TargetWindowStart = now
+	meta.TargetAttempts = 0
+	meta.TargetSuccesses = 0
+	ms.tracker.Metadata[measurementID] = meta
+
+	lossRatio := 1 - float64(successes)/float64(attempts)
```
juan-malbeclabs, line 452: The ratio pools every source's outcomes but charges the verdict to the target probe, so loss on one *source's* path marks a target that answered every ping that reached it.

There is no per-source attribution here: `attempts`/`successes` are summed across all of a measurement's sources by `RecordTargetResults`. This bites hardest exactly where the fan-in is thinnest. `generateWantedMeasurements` enlists only locations sorting *after* the target (collector.go:1530-1541), so the alphabetically penultimate metro's measurement has exactly **one** source — and the ratio then *is* that single circuit's reachability, with nothing to average it against. At 10-minute sampling that source contributes 6 attempts/hour, so after roughly 5h the window closes and a healthy target is marked and rotated onto a worse probe sitting behind the same broken path.

This also reverses the decision deliberately recorded a few lines away at collector.go:1024-1028 for #4153 — "Rotating the probe is not the answer either way ... and it is expensive" — for a signal that cannot distinguish target loss from source-path loss. Worth either requiring a minimum source count before a verdict, or tallying per source and demanding that a majority of sources see the loss.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -221,35 +264,168 @@ func (ms *MeasurementState) UpdateSourceProbeResponse(measurementID int, probeID
...
+}
+
+// AddUnresponsiveTarget records that a probe failed as a measurement target. Unlike
+// AddUnresponsiveProbe this does not bar the probe from sourcing measurements, since
+// failing to answer pings says nothing about sending them.
+func (ms *MeasurementState) AddUnresponsiveTarget(probeID int) {
+	ms.mu.Lock()
+	defer ms.mu.Unlock()
+
+	for _, entry := range ms.tracker.UnresponsiveTargets {
```
juan-malbeclabs: This dedupe returns early without refreshing `MarkedAt`, so continuous fresh failure evidence cannot extend a mark — the probe unblocks exactly 24h after its *first* mark.

Probe A is marked lossy at t=0. Every window thereafter `EvaluateTargetLoss` judges A lossy again, but this loop no-ops and `MarkedAt` stays t=0. At t=24h `PruneExpiredUnresponsiveTargets` drops the mark although the collector judged A lossy minutes earlier. `rankTargets` then ranks A first again on pure distance, so a metro that had rotated to a healthier farther candidate is torn down and retargeted at the known-bad nearest probe, and needs another full window of bad data before it can be re-marked — then a second teardown to move back off it. Two recreations and an hour or more of bad data every 24h, indefinitely.

Separately, this loop ignores expiry, so its correctness quietly depends on `PruneExpiredUnresponsiveTargets` having run earlier in the same `configureMeasurements` pass. Refreshing `MarkedAt` on a repeat mark (or checking expiry here) would remove both the flap and the ordering dependency. `AddUnresponsiveProbe` has the same shape, but it mattered less before, because a marked target used to lose its measurement rather than keep it.

juan-malbeclabs (changes requested):
Second pass, deeper on the loss window and on what the role split changed at each mark site. The four items in my earlier review still stand; these are additional and I verified each against the code rather than the hunks.

Two of them I would treat as blocking:
- **collector.go:950** removes the only mechanism that ever took a fully dead probe out of the *source* pool, so such a probe is re-enlisted as a source forever and its circuits go permanently dark.
- **CHANGELOG.md:46** puts the four new bullets inside the already-released `v0.40.0` section, so `scripts/release/bump-version.sh` will never promote them.

The rest are loss-window correctness issues: the window is never reset when it closes short, it pools every source's outcomes but charges the verdict to the target, it is keyed by measurement ID and so is discarded by any unrelated recreation, and its cursor has no upper clamp. Each of those can either mark a healthy target or silently prevent the check from ever firing, so I think they want addressing before this carries a rotation decision.

Re-verified unchanged from the first pass: build, `go test`, `go test -race`, `go vet` and `gofmt` are clean, `rankTargets` cannot empty a metro's candidate set, the `> MaxTargetLossRatio` direction is right, there is no divide-by-zero, and the legacy `[]int` state format still loads.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -852,6 +910,10 @@ func (c *Collector) configureMeasurements(ctx context.Context, locationMatches [
 	}
 
 	// Step 4: Prune expired unresponsive probes so they get retried
+	if pruned := measurementState.PruneExpiredUnresponsiveTargets(); pruned > 0 {
```
juan-malbeclabs, line 927: `--dry-run` mutates and persists shared state, and this diff adds a new *destructive* read on that path.

In this file `dryRun` is consulted only at collector.go:1212 (Step 7 removal) and 1287 (Step 8 creation). Step 4 is entirely unguarded: `PruneExpiredUnresponsiveTargets`, `PruneExpiredUnresponsiveProbes`, `AddUnresponsiveTarget` (950, 970) and the new `EvaluateTargetLoss` (960) all run, and the `Save()` at 1047 fires whenever anything got marked. `EvaluateTargetLoss` is not a query — on a judged window it zeroes `TargetAttempts`/`TargetSuccesses` and re-anchors `TargetWindowStart`.

`--dry-run` is a persistent flag documented at cmd/collector/main.go:427 as "Log what would be created without actually creating measurements or jobs", and `create-measurements` (main.go:298) constructs the collector with a nil state, so `configureMeasurements` falls into collector.go:858-862 and loads the daemon's own state file from `--state-dir` (default at main.go:424).

So an operator running `create-measurements --dry-run` to inspect intent discards up to a full window of loss evidence per measurement, can blacklist probes, and writes the zeroed windows back over the file the running daemon shares. Gating Step 4's mutations (or at least the `Save`) on `!dryRun` would make the flag mean what it says.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -78,6 +120,7 @@ func (ms *MeasurementState) Load() error {
 
 	var tracker MetadataTracker
 	tracker.Metadata = intermediate.Metadata
+	tracker.UnresponsiveTargets = intermediate.UnresponsiveTargets
```
juan-malbeclabs, line 165: No migration for the target-failure marks that are already sitting in `unresponsive_probes`, so the first cycle after deploy reproduces the teardown this PR is meant to remove.

On the merge base the staleness path called `measurementState.AddUnresponsiveProbe(meta.TargetProbeID)` — that is the very line this PR changes to `AddUnresponsiveTarget` at collector.go:950. So today's production state files hold target-failure entries in `unresponsive_probes`. This line adds decoding for the new list but migrates nothing, and `filterResponsiveProbes` (collector.go:238) still reads `IsProbeUnresponsive`, which only sees the old list.

Net on deploy day: those probes stay barred from *sourcing* for up to 24h, contradicting the PR's premise; then `PruneExpiredUnresponsiveProbes` drops the entry, the metro's source flips back to its nearest probe, and every measurement sourcing from it is stopped and recreated — the exact cascade bullet 1 describes, one last time. This is live state, not hypothetical: the Unreleased entry at CHANGELOG.md:12 documents cmh probe 1009793 going around the mark/expire/re-select loop on mainnet-beta.

`Load` already holds `tracker.Metadata`, so a one-time migration is mostly determinable here — a marked probe id that equals some entry's `TargetProbeID` was marked by the staleness path. It is not perfectly unambiguous, since Step 4b could have marked the same probe as a source, but either that heuristic or simply dropping the pre-upgrade marks on first load avoids paying the cascade one more time on rollout.

controlplane/internet-latency-collector/internal/ripeatlas/state_test.go
```diff
@@ -436,3 +436,123 @@ func TestInternetLatency_RIPEAtlas_State_TimestampTracker_Structure(t *testing.T
...
+		ms := newState()
+
+		ms.RecordTargetResults(999, 100, 0, time.Now().Unix(), time.Now().Unix())
+		lossy, attempts, _ := ms.EvaluateTargetLoss(999, time.Now().Unix()+windowSecs)
+		require.False(t, lossy)
+		require.Zero(t, attempts)
+	})
+}
+
+func TestInternetLatency_RIPEAtlas_State_UnresponsiveTargetsRoundTrip(t *testing.T) {
```
juan-malbeclabs, line 631: Nothing covers the 24h expiry of `UnresponsiveTargets`, and it is the part of this change with the least cover.

`TestInternetLatency_RIPEAtlas_State_UnresponsiveProbeExpiry` (state_test.go:347) exists for the source list, but no test references `PruneExpiredUnresponsiveTargets` or an expired target mark — this round-trip test only exercises a fresh one. I checked by mutation: patching the expiry that `IsTargetUnresponsive` passes to `hasLiveEntry`, and making `PruneExpiredUnresponsiveTargets` keep every entry, both leave the package green. state.go:397-399 (the `pruned++` branch) and state.go:341-343 (the dedupe return) get zero hits.

That matters more than it did for the source list: under rank-last a mark that never expires pins a metro to a dead probe rather than retrying it, which is the opposite of the intended behaviour and would pass CI today.

controlplane/internet-latency-collector/internal/ripeatlas/collector_test.go
```diff
@@ -2004,3 +2052,453 @@ func TestInternetLatency_RIPEAtlas_ExportSingleMeasurementResults_LossCountsAsRe
...
+			mu.Unlock()
+			return &MeasurementResponse{Measurements: []int{id}}, nil
+		},
+		StopMeasurementFunc: func(_ context.Context, measurementID int) error {
+			mu.Lock()
+			stoppedMeasurements = append(stoppedMeasurements, measurementID)
+			mu.Unlock()
+			return nil
+		},
+		GetMeasurementResultsIncrementalFunc: func(_ context.Context, _ int, _ int64) ([]any, error) {
```
juan-malbeclabs, line 2728: This mock is never called, so neither new `configureMeasurements` test covers the path it looks like it covers.

The collector at line 2327 (and at 2443 in `..._LastMarkedCandidateIsKept`) leaves `exporter` unset, so the `if c.exporter != nil` guard at collector.go:1218 skips the only `exportSingleMeasurementResults` call in `configureMeasurements`. I confirmed by replacing both mock bodies with `panic(...)`: the package still passes.

Consequently the tallies these tests turn on (`TargetAttempts: 100`, `TargetSuccesses: 15`) are hand-seeded through `SetMetadata` and never produced by the counting block at collector.go:660-670. Nothing covers export-counts -> window-closes -> mark -> rotate end to end, so a defect in how attempts reach the window — a wrong cursor comparison, or `newestResult` and `now` swapped at the `RecordTargetResults` call site, which compiles because both are `int64` — is caught only by the unit-level `..._RecordsTargetLoss` test. Wiring a real exporter and letting the results flow through `exportSingleMeasurementResults` would close that, and would also exercise the cursor gating these tests currently bypass.

Minor related prose point: the changelog calls this a "rolling hourly window", but it is tumbling (reset only when judged) and, per my comment on state.go:18, in production it is frequently not an hour.

juan-malbeclabs (commented):
Third and last pass — four more items, all verified against the code. Changes are already requested; this adds to that rather than changing the verdict.

The first two are the ones I would prioritise: a one-shot deploy-day regression that reproduces the exact teardown bullet 1 claims to fix, and `--dry-run` now destructively mutating shared state through a new code path. The other two are test gaps where the new behaviour has no cover at all.

CHANGELOG.md
nikw9944: Moved under `## Unreleased` / `### Changes` / the existing `- Collector` bullet in f0b22f44. You are right that `bump-version.sh` only promotes the Unreleased body, so these would have been dropped from every release and dated before the code existed. Also changed "rolling hourly window" to "hourly window" per your note on the tumbling/rolling distinction. (instance 613)

CHANGELOG.md
nikw9944: Dropped in f0b22f44. Agreed: `unresponsive_targets` does not exist on main and the missing decode was both introduced and fixed on this branch, so no released collector ever lost target marks. Nothing for a reader to go looking for. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
nikw9944: Fixed in b87d8275. The wider fetch now lands in a new `LocationProbeMatch.FallbackTargetProbes` that only target selection reads; `NearbyProbes` is untouched, so source selection cannot see it.

`..._TriggeredByTargetMarks` now has two subtests: one where the fetch finds nothing (the marked candidate stays for `rankTargets`), and one where it returns a nearer non-anchor probe — your xsin shape. That one asserts `NearbyProbes` still holds only the original probe, that the fallback becomes the target, and that `filterResponsiveProbes` over the location's own probes still yields the original as its source. `..._TriggeredWhenAllUnresponsive` asserted the old replacement behaviour and was updated to the new field. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
nikw9944: Confirmed and fixed in aa00d3ae. `UpdateSourceProbeResponse` is called at collector.go:674, outside the `latency > 0` guard, so any uploaded result including a timeout advances `LastResponseAt` — a probe that uploads nothing leaves it at zero and could never pass the gate.

Step 4b now treats a zero past the measurement's first hour as the offline case rather than an unpopulated field, with reason `no_responses_since_created`. Covered by `..._SourceThatNeverResponded`.

One thing worth your eye: combined with the fix for your collector.go:983 comment, this only reaches measurements whose target *is* answering, since a measurement judged stale in the cycle skips Step 4b entirely. That keeps it well clear of most of Step 4c's population, but it does now rotate some sources that Step 4c's comment deliberately declines to rotate for #4153. If you would rather that population stay observation-only, the alternative is marking stale targets in both lists instead. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
nikw9944: Fixed in 47475e34. `configureMeasurements` now tracks the measurements whose target it judged stale or lossy in this cycle and Step 4b keys the skip off that, not off `IsTargetUnresponsive`. Your reading was right: under rank-last the standing mark is re-satisfied every cycle, so the exemption never lifted.

`..._KeptMarkedTargetStillChecksSources` pins it — a target marked in an earlier cycle and kept, with a stale source that must still be marked. Reverting the guard to the probe-level check fails it. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
nikw9944: Fixed in 3a1ea4a4. A future-dated result is now skipped outright rather than counted: gating on `resultAt <= now` only for the cursor advance would still have recounted the same result every cycle, since `resultAt > lossCursor` stays true. Skipping it neither moves the cursor nor inflates attempts.

New subtest in `..._RecordsTargetLoss`: one timeout dated 48h ahead, then a real result — attempts and cursor stay at zero, and the later result is still counted. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/state.go
nikw9944: Fixed in 97229ed3, with an ordering correction in ea4e9e3e. Past `MaxTargetLossWindowAge` the tallies are now dropped unjudged.

One correction on the constant: `4 * TargetLossWindow` does not fix your scenario. At 12 attempts/hour the tallies cross `MinTargetAttemptsForLossCheck` at hour 3, which is inside a four-window cap, so the verdict is still rendered at 36/14 and the target still marked. The cap has to fire before the attempt count crosses, so it is `2 * TargetLossWindow`.

The cap is also checked ahead of the attempt-count gate, because a cycle that drifts past it keeps accumulating and the next one would judge the window anyway. So no verdict is ever rendered over more than two windows of evidence.

The cost, recorded in the constant's comment: a measurement with too few sources to reach the minimum inside two windows is never judged at all. Given your state.go:329 point about thin fan-in that seems the right direction, and the two-source floor added for it makes the same trade explicit. Your exact 3-hour scenario is now a test asserting no mark. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/state.go
nikw9944: Fixed in 9332c3f5. Both Add functions now restamp an existing entry through a shared `refreshOrAppendMark`, so fresh failure evidence extends the mark. That also removes the ordering dependency you noted — an expired entry is restamped rather than left reading as expired.

Tested for both lists by seeding a mark 23h old, re-marking, and asserting the persisted `marked_at` moved forward. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/state.go
nikw9944: Fixed in c1f30493. `configureMeasurements` snapshots each removed measurement's metadata by target location before Step 7, and Step 8 carries `TargetWindowStart`, `TargetAttempts`, `TargetSuccesses` and `TargetLossCursor` over when `spec.TargetProbe.ID` matches the old `TargetProbeID`. A new target still starts clean.

`..._WindowSurvivesSourceChurn` recreates a measurement 40 minutes into a window because of a source change and asserts all four fields survive. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/state.go
nikw9944: Fixed in ea4e9e3e. A window is now judged from `TargetLossWindow - TargetLossWindowGrace`, with a 5 minute grace, and the comment on the constant records why: the window equals the default management interval and the clock is cycle start plus that cycle's pre-work, so any faster cycle lands short and slips the verdict a full hour.

Test seeds the start at `-(TargetLossWindow - 30s)` and expects a verdict; it fails without the grace. The existing "not judged before it closes" subtest moved to the grace boundary.

On `MinTargetAttemptsForLossCheck`: left a constant, since the sampling interval is fixed in production, with a sentence in the comment now saying it assumes the 10 minute interval. Deriving it from `samplingInterval` and `len(meta.Sources)` is a fair follow-on. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/state.go
nikw9944: Fixed in 12d39f50, along the lines of your first suggestion. `EvaluateTargetLoss` is not called unless `len(meta.Sources) >= 2`; with two sources contributing equally one dead path caps at exactly 0.5, which does not clear the threshold. Skips log at Debug, so they are no longer indistinguishable from a healthy target.

Per-source tallying is the better answer and is left to its own change — noted in the comment. `..._SingleSourceTargetNotMarked` covers the penultimate-metro shape: a closed window at 95% loss carried by one source produces no mark. Two existing tests seeded a single source and were given a second one. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
nikw9944: Fixed in 157f5d31. The Step 4 `Save()` is gated on `!dryRun` and the log line says the marks were not persisted.

`..._DryRunDoesNotPersistMarks` asserts the state file is not created at all by a dry-run cycle that marks a target. The in-memory mutation from `EvaluateTargetLoss` still happens under the flag; gating the save keeps it off the file the daemon shares, which is the part that reaches the operator. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/state.go
nikw9944: Fixed in 2ce0c74b. `Load` reclassifies any `unresponsive_probes` entry whose probe ID matches some metadata entry's `TargetProbeID` into `unresponsive_targets`, keeping `MarkedAt` so the mark expires when it always would have. The count is logged at startup.

On the ambiguity you flagged: it runs exactly once, recorded by a new `unresponsive_marks_migrated` flag in the file, so a source failure marked later that happens to be another measurement's target probe is not moved on the next restart. That still leaves the one-shot misclassification possible on rollout for a probe Step 4b had marked as a source while it was also a target — accepted, and the flag stops it recurring. Tested against a main-era file with one entry of each kind. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/state.go
nikw9944: Fixed in 560774ea. The comment now says a probe in `UnresponsiveProbes` is excluded from source selection and ranked last in target selection, not excluded from it.

On the count: the log line now reports the union of both lists as `demoted_target_count` alongside both lengths, so `unresponsive_target_count` is no longer readable as the number of probes actually demoted. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/collector_test.go
nikw9944: Added in b5a70d28: `require.NotContains(t, stoppedMeasurements, 1001)`, with a comment saying why it is the assertion that pins the split. You are right that a regression re-barring 6726 for sourcing would have passed as written. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/state_test.go
nikw9944: Added in b5a70d28. `..._State_UnresponsiveTargetExpiry` backdates one of two target marks past 24h and asserts `IsTargetUnresponsive` reports false for it, `GetUnresponsiveTargets` omits it, `PruneExpiredUnresponsiveTargets` returns 1 then 0, and the live mark survives. Making the prune keep every entry fails it. The dedupe return at state.go:341-343 is now covered by the repeat-mark test added for your state.go:340 comment. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/collector_test.go
nikw9944: Fixed in b5a70d28. `..._LossyTargetIsRotated` now runs a real CSV exporter and fills the window by calling `exportSingleMeasurementResults` with 36 results across two sources, 6 answered, instead of hand-seeding via `SetMetadata`. It asserts the counted tallies (36/6) and that the results mock is called again during reconciliation, so the pre-removal export path is really exercised. Breaking the counting block fails it.

Only `TargetWindowStart` is still moved back, since nothing in the test can wait an hour for the window to close — noted in a comment. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/state.go
nikw9944: Right, and fixed in 34adbec5: the cap now compares against `MaxTargetLossWindowAge + TargetLossWindowGrace`, so a 3- or 4-source window arriving at an age of about two windows is judged deterministically rather than on cycle drift, while a 2-source window is still dropped unjudged by three. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -910,15 +1030,19 @@ func (c *Collector) configureMeasurements(ctx context.Context, locationMatches [
...
+				// uploaded, a timeout included, so a source that has uploaded
+				// nothing in the measurement's whole first hour is not running the
+				// measurement at all. Since the staleness path above now marks only
+				// UnresponsiveTargets, this is the sole route by which a probe that
+				// is offline while RIPE still reports it Connected leaves the source
+				// pool; treating zero as "not populated yet" left it enlisted forever.
 				if source.LastResponseAt < probeTimeout {
 					reason := "no_recent_responses"
+					if source.LastResponseAt == 0 {
+						reason = "no_responses_since_created"
```
juan-malbeclabs: This closes the gap I flagged, but it acts on the one signal this file already decided not to act on, and the decision record twenty lines below now argues against the code above it.

Step 4c at 1060-1070 watches exactly this population — `sourcesWithoutSamples` counts nothing but `source.LastResponseAt == 0` on measurements older than `sourceSampleGracePeriod` (2h) — and states why it only observes: *"Rotating the probe is not the answer either way — the drops observed so far recovered unaided — and it is expensive, because a metro's probe is a source for every measurement whose target sorts before it. See #4153."*

With the zero guard gone, `probeTimeout` (1h) reaches that population first. A source that #4153 says recovers unaided is now barred from sourcing for 24h, and being its metro's nearest probe it is the source in every measurement whose target sorts before it, all of which Step 5 then stops and recreates. That is the cascade bullet 1 of this PR exists to remove, re-entered through a different door. The recreation also zeroes `LastResponseAt` for every *other* source in those measurements, restarting their clocks; and if the cause really is RIPE not dispatching the enlistment, the replacement is a fresh enlistment that can fail the same way.

Step 4c also stops meaning what it says. Both gates land on the same cycle (4b needs `CreatedAt < now-1h`, 4c `CreatedAt < now-2h`, and at T+2h+δ both are satisfied), so the gauge reports once in the very cycle that marks the probe and the measurement is gone by the next — instead of standing while the source is dark, which is what an observation-only signal is for.

The finding asked for *a* reachable route out of the source pool, not this one. A narrower one leaves #4153 alone: at 955 `never_exported` **is** the offline case — the measurement has produced nothing at all — while `no_recent_exports` at 959 is the NAT-like one. Marking `meta.TargetProbeID` in both lists on `never_exported` only takes an offline-but-Connected probe out of the source pool without touching a source that is merely silent on one circuit. It misses a probe that is nobody's target, which is just the alphabetically last metro, and that one is covered by the existing `no_recent_responses` path once it has uploaded anything at all.

If the rotation is the call you want, that is defensible — but then #4153's block needs to record the reversal rather than keep arguing the opposite.

juan-malbeclabs:
Re-reviewed at 34adbec. The fifteen findings from the previous pass are each addressed by a commit that does the thing rather than working around it, and I checked them against the code rather than the messages:

- fallback probes now land in `FallbackTargetProbes` and only target selection reads them, so the wider fetch can no longer swap a metro's source probe
- Step 4b keys its skip to `targetFailedThisCycle` instead of a standing mark, so a kept marked target no longer exempts its sources forever
- `resultAt <= countedUpTo` clamps a future-dated result, so one clock-skewed timeout cannot park `TargetLossCursor` past wall clock
- `MaxTargetLossWindowAge` bounds a short window, `TargetLossWindowGrace` takes the drift coin-flip out of the hourly gate, and the two interact as the comment claims (2 sources -> 24 attempts at two windows, under the minimum, dropped by three)
- the window carries across a recreation only when the target probe is unchanged
- `refreshOrAppendMark` restamps a repeat mark, which also drops the ordering dependency on the prune
- `migrateTargetMarks` runs once per file behind a persisted flag, with the heuristic's limits written down
- Step 4's marks are no longer persisted under `--dry-run`
- `len(meta.Sources) < 2` keeps a single circuit's reachability from convicting a target
- changelog entries are under `## Unreleased` / `- Collector`, and the branch-only persistence bullet is gone

`go test ./internal/ripeatlas/...` passes locally and all 42 checks are green.

One thing left, inline on collector.go: dropping the `LastResponseAt == 0` guard in Step 4b closes the gap I raised, but it does so on the population Step 4c deliberately only observes, and the #4153 rationale sitting twenty lines below now argues against the code above it. A narrower route through `never_exported` gets the same offline-but-Connected probe out of the source pool without reopening the cascade this PR is about. Happy to approve once that is either narrowed or the #4153 note is updated to record the reversal.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
nikw9944: Agreed, and reverted in b9f5261f — the Step 4b change was reaching #4153's population and the decision record twenty lines below was arguing against the code above it, which is not a state to merge.

Taken your narrower route: on the staleness branch `never_exported` marks `meta.TargetProbeID` in both lists, since a measurement that has produced nothing at all in its first hour has an offline target and an offline probe cannot source either, while `no_recent_exports` stays target-only as the NAT-like case. Step 4b's `LastResponseAt == 0` skip is back, with a comment pointing at #4153 and at the `never_exported` branch so the next reader does not remove it again. The miss you noted — a probe that is nobody's target — is the alphabetically last metro only, and `no_recent_responses` covers it once it has uploaded anything; that is recorded in the comment too.

Covered by `..._StaleTargetListsBySeverity`, two cases: `LastExportAt == 0` marks both lists and the probe drops out of `filterResponsiveProbes`; a non-zero `LastExportAt` older than an hour marks the target list only and the probe still sources. Flipping the branch either way fails the matching case. (instance 613)

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
nikw9944: Correction to my earlier reply here: the fix I described in aa00d3ae was reverted in b9f5261f. Removing the `LastResponseAt == 0` skip took Step 4b into the population Step 4c observes and deliberately does not rotate per #4153, so the offline probe now leaves the source pool through the `never_exported` staleness reason instead — see the thread on collector.go:1044. (instance 613)

juan-malbeclabs (approved):
b9f5261 takes the narrow route and the last blocker is gone.

The `LastResponseAt == 0` skip is back in Step 4b, so Step 4c keeps the population it was built to watch and the #4153 block no longer argues against the code above it. The offline probe leaves the source pool through the staleness reason instead: `never_exported` marks both lists, `no_recent_exports` stays target-only. That split is the right one — a measurement that has produced nothing at all in its first hour is a probe that is not running it, while a probe that answered and stopped is the inbound-only case the role split exists for — and it restores exactly the pre-PR treatment of `never_exported` rather than adding a new rotation trigger. The acknowledged gap, a probe that is nobody's target and has never uploaded anything, is only the alphabetically last metro and is named in the comment.

Checked the rest of the change once more at this head:

- `migrateTargetMarks` moving a legacy mark to the target list alone is recoverable as the new note claims: a still-offline probe is re-marked in both lists on the next `never_exported` cycle, and one that has merely stopped answering is correctly target-only from then on.
- `AddUnresponsiveProbe` on the target does not collide with the rest of Step 4 — `targetFailedThisCycle` skips that measurement, and the `IsProbeUnresponsive` guard skips it everywhere it is a source.
- `filterResponsiveProbes` still gates source selection on the source list alone, and `rankTargets` still demotes the union, so a metro whose only candidate is offline keeps a target.
- `StaleTargetListsBySeverity` pins both branches and, usefully, asserts the effect on source selection rather than just the flag.

`go vet` and `go test ./internal/ripeatlas/...` clean locally, 42/42 checks green, no conflicts with main.

PR #4336, opened 2026-09-15
ci: install the platform-tools that build-sbf is asked for

elitegreg:
## Summary of Changes
- `cargo build-sbf --tools-version v1.54` has been silently ignored since anza published platform-tools **v1.51.1**. The flag is only honored when cargo-build-sbf can prove the version exists: it looks under `~/.cache/solana`, and finding nothing there, resolves `https://github.com/anza-xyz/platform-tools/releases/latest`. v1.51.1 was published after v1.54, so that redirect now names v1.51.1, the requested v1.54 reads as "newer than latest", and the tool logs at a level nobody sees and falls back to its built-in default v1.51, whose Cargo is 1.84.
- Cargo 1.84 cannot parse the edition2024 crates the solana 3.0 tree pulls, so the failure lands nowhere near the flag. `rust-build`, `rust-test` and `rust-validator-test` all die on ``feature `edition2024` is required`` in `offchain/crates/contributor-rewards`, a workspace member no program depends on, which is why the same tree builds and tests fine locally (where v1.54 is already in `~/.cache/solana` and therefore accepted).
- `scripts/install-sbf-tools.sh` installs the requested release into `~/.cache/solana/<version>/platform-tools` before the build, which takes the network lookup out of the decision. Every site that passes `--tools-version` now calls it first: `smartcontract/Makefile` (`build-programs`, `test-sbf`), `smartcontract/test/run_record_test.sh`, `start-test.sh`, `start-test-permissions.sh`, `e2e/docker/base.dockerfile`, `release.testnet.yml` and `release.devnet.smartcontract.daily.yml`. `solana/Makefile` is left alone: those programs pass no `--tools-version`, keep their own workspace and toolchain, and their compiled bytes are pinned by `solana/programs/sha256sums_*.txt`.
- The script is idempotent, retries the download, and stages into a directory holding no `platform-tools` child before renaming it into place — cargo-build-sbf reads every directory under `~/.cache/solana` as an installed version, so a half-extracted tree would be taken at face value. A tree carrying the binaries but no `rust/lib`, which is what an interrupted extraction leaves, reads as absent and is replaced; the e2e base image checked for exactly that state and deleted it, so that check gives way to the install. `builder-base` gains `bzip2`, which the tarball needs and ubuntu:24.04 does not carry.

## The same build comes out two ways, and that part is unexplained
Pinning the tools version only settles the build when the strict `cargo-build-sbf` runs, and which one runs is not determined:

- **Lenient** — `cargo build-sbf` sets up platform-tools and then drives the **host** Cargo (1.97.1), which parses edition2024 fine, so the build goes green whatever platform-tools it settled on. Runs through 2026-09-14 and both green runs on this branch.
- **Strict** — it drives the **platform-tools** Cargo instead. Without the pin above that is 1.51's Cargo 1.84.0, and the build dies on `edition2024`. This is #4332.

Both are Agave v3.0.4's `cargo-build-sbf` from the same install path, with the same flags. The failing runs correlate with a `Swatinem/rust-cache` miss and the passing ones with a hit, but the mechanism is not established, so treat that correlation as a lead rather than a cause.

`.github/actions/solana-toolchain` now logs the resolved `cargo build-sbf` path and version, so the next occurrence says which binary ran instead of leaving it to be inferred. It also removes any `cargo-build-sbf` in `$CARGO_HOME/bin`, which Cargo prefers over the one just installed — a guard against one way the variance could arise, though its own output shows no run has actually carried such a copy.

The pin makes the build survive both paths, which is what makes CI green regardless. It does not make the two paths compile the programs identically: the lenient path ignores `--tools-version` and builds with v1.51's rustc.

## Testing Verification
- Ran `scripts/install-sbf-tools.sh v1.54` against an empty `HOME`: downloads, unpacks to `rust/`, `llvm/`, `version.md`, and reports `cargo 1.89.0`. A second run detects the install and exits without touching the network. Deleting `rust/lib` from that cache makes the third run reinstall rather than accept it.
- `make -C smartcontract build-programs env=localnet` builds all four programs on the branch, exercising the already-installed path.
- `rust-build`, `rust-test`, `rust-validator-test` and `rust-cli-static` pass on this branch, twice. **Both runs were rust-cache hits and both took the lenient path, so neither exercised the strict path this PR is meant to fix.** The second run's log confirms it: `cargo build-sbf` resolved to the Agave install at 3.0.4, the installer put v1.54 in place, and `cargo build-sbf` then fetched and linked v1.51 anyway and built green through the host Cargo. Forcing a cache miss is what would test the failing condition.
- The e2e shards exercise the new base image, including the added `bzip2` and the install step that replaced the `rust/lib` corruption sweep.


scripts/install-sbf-tools.sh
```diff
@@ -0,0 +1,93 @@
...
+readonly ATTEMPTS=3
+readonly CONNECT_TIMEOUT=15
+readonly FETCH_TIMEOUT=900
+
+readonly CACHE_DIR="${HOME}/.cache/solana"
+readonly DEST="${CACHE_DIR}/${VERSION}/platform-tools"
+
+# cargo-build-sbf reads the tools straight out of this directory, so a complete
+# tree here is all it needs; it makes the rustup link itself.
+if [[ -x "${DEST}/rust/bin/cargo" && -x "${DEST}/rust/bin/rustc" ]]; then
```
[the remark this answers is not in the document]
elitegreg: Fixed in b1ace9e5a. The check is now a shared `installed()` helper requiring `rust/bin/cargo`, `rust/bin/rustc` and a `rust/lib` directory, so a tree the extraction left half-written reads as absent and gets replaced. The same helper guards the staged download before it is moved into place, so a bad download leaves nothing behind for the next run to trust.

The e2e base image was checking for exactly this state and deleting it; that block gives way to the install, which repairs it instead.

smartcontract/Makefile
```diff
@@ -34,6 +38,13 @@ test-programs: test-sbf
 		-p doublezero-serviceability \
 		-p doublezero-geolocation
 
+# cargo build-sbf and cargo test-sbf fall back to their built-in platform-tools,
+# whose Cargo is too old for this tree, unless the version they are asked for is
+# already installed. Install it first rather than let them choose.
+.PHONY: sbf-tools
+sbf-tools:
+	@bash $(ROOT_DIR)/scripts/install-sbf-tools.sh $(SBF_TOOLS_VERSION)
```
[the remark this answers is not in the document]
elitegreg: Fixed in b1ace9e5a. `start-test.sh`, `start-test-permissions.sh` and the three `cargo build-sbf` steps in `e2e/docker/base.dockerfile` now install the tools first, so every site that passes `--tools-version` is covered. The dockerfile needed `bzip2` in `builder-base` as well, since the platform-tools tarball is bz2 and ubuntu:24.04 does not carry it.

`solana/Makefile` is deliberately left alone: those programs pass no `--tools-version`, keep their own workspace and toolchain, and their compiled bytes are pinned by `solana/programs/sha256sums_*.txt`.

scripts/install-sbf-tools.sh
```diff
@@ -0,0 +1,105 @@
...
+# cargo-build-sbf reads the tools straight out of this directory, so a complete
+# tree here is all it needs; it makes the rustup link itself.
+#
+# rust/lib is checked alongside the binaries because an interrupted extraction
+# leaves the tree in exactly that state: the binaries are there and the standard
+# library is not. cargo-build-sbf takes the version directory's existence as
+# proof the version is installed and would build against the wreckage, so an
+# incomplete tree has to read as absent here and be replaced below.
+installed() {
+  [[ -x "$1/rust/bin/cargo" && -x "$1/rust/bin/rustc" && -d "$1/rust/lib" ]]
```
bgm-malbeclabs: `installed()` only looks at `rust/`. `cargo-build-sbf` also uses `platform-tools/llvm/bin/{clang,llvm-ar,llvm-objdump,llvm-objcopy}` and exports them as `CC`, `AR`, `OBJDUMP` and `OBJCOPY` (`main.rs:164-173`), and this tree compiles C (blake3, per the Makefile's own note).

So take the case the comment above already describes, just the other subtree: an extraction interrupted after `rust/` and before `llvm/`. This test passes, the script prints "already installed" and exits 0, `install_if_missing` skips the re-download because the directory is not empty, and the old `corrupted_toolchain` sweep only looked at `rust/` too. Every later build on that cache fails on a missing `clang`, and nothing repairs it.

Adding `-x "$1/llvm/bin/clang"` to the same test closes it.

scripts/install-sbf-tools.sh
```diff
@@ -0,0 +1,105 @@
...
+# name is what marks the version installed. This is the same test the cached
+# tree had to pass above; a download that fails it leaves nothing behind for the
+# next run to trust.
+if ! installed "${staging}/tree"; then
+  echo "platform-tools ${VERSION} unpacked incomplete; refusing to install it" >&2
+  exit 1
+fi
+
+mkdir -p "${CACHE_DIR}/${VERSION}"
+rm -rf "${DEST}"
```
bgm-malbeclabs, line 143: `rm -rf "${DEST}"` followed by a plain `mv` is destructive on a cache that two builds share. The e2e `setup` job runs on `self-hosted` and the BuildKit mount `sbf-solana-3.0.4` is `sharing=shared`, so two PR image builds on one host write the same `~/.cache/solana`.

Run B can delete the tree run A is reading. Worse: if B recreates `DEST` between A's `rm -rf` and A's `mv`, `mv src dst` onto an existing directory nests instead of replacing, and the cache ends up holding `platform-tools/tree/rust/...`. That directory is not empty, so `cargo-build-sbf` reads the version as installed. The old sweep only deleted a tree it had already judged corrupt, so this is new exposure.

`mv -T`, or a lock file around the whole install, is enough.

.github/actions/solana-toolchain/action.yml
```diff
@@ -85,3 +85,30 @@ runs:
...
+          if [ -e "$stale" ]; then
+            echo "Removing $stale, which would shadow the Agave ${{ inputs.version }} $tool"
+            rm -f "$stale"
+          else
+            echo "No $stale to remove"
+          fi
+        done
+
+        echo "cargo build-sbf now resolves to: $(command -v cargo-build-sbf)"
+        cargo build-sbf --version
```
bgm-malbeclabs: `cargo build-sbf --version` cannot answer the question this step exists to answer. Agave builds the banner before it parses arguments: `main.rs:371-377` formats `crate_version!()` with the compile-time constant `DEFAULT_PLATFORM_TOOLS_VERSION`, which is `v1.51`. The line reads `platform-tools v1.51` on a fixed run and a broken one alike.

This PR makes it worse than useless: with v1.54 the only installed version, `get_base_rust_version` finds no `~/.cache/solana/v1.51/platform-tools/rust/bin/rustc` and returns an empty string, so the rustc line prints blank. The next person debugging this reads "v1.51" in the log and draws exactly the wrong conclusion.

`~/.cache/solana/v1.54/platform-tools/rust/bin/rustc --version` reports the real thing. The `command -v` line above is the half that works.

.github/actions/solana-toolchain/action.yml
```diff
@@ -85,3 +85,30 @@ runs:
...
+
+    # Cargo looks for a subcommand in $CARGO_HOME/bin before it looks at PATH, and
+    # Swatinem/rust-cache carries that directory from run to run, so a cargo-build-sbf
+    # left there would quietly shadow the one this action installs. No run has been
+    # observed carrying one, so this is a guard rather than a fix for anything seen;
+    # what earns its place is the last two lines, which put the resolved path and
+    # version in the log. The same build has come out both ways on this repo -- one
+    # run driving the host Cargo, the next driving the platform-tools Cargo, which
+    # fails on edition2024 -- and the logs could not say which binary ran.
+    - name: Pin and report the SBF build tools
```
bgm-malbeclabs, line 97: `solana.yml` sets `defaults.run.working-directory: solana`, but workflow defaults do not reach steps inside a composite action, so this step runs at the repository root. `solana-test-sbf` otherwise installs only what `solana/rust-toolchain.toml` pins (1.91); the root pins 1.97.1. This is the first `cargo` call at the root in that job, so rustup downloads 1.97.1 there — a toolchain that job never needs, and one more network step that can fail it.

A `working-directory` on the step, or a guard so it runs only where an SBF build follows.

e2e/docker/base.dockerfile
```diff
@@ -175,17 +176,14 @@ RUN --mount=type=cache,id=sbf-cargo-${SOLANA_VERSION}-${CARGO_LOCK_HASH},target=
...
+# here: a cache left half-extracted by an interrupted build reads as absent and
+# is replaced, rather than being deleted and re-downloaded by the build itself.
 RUN --mount=type=cache,id=sbf-solana-${SOLANA_VERSION},target=/root/.cache/solana \
-    for pt_dir in /root/.cache/solana/*/platform-tools; do \
-        if [ -e "$pt_dir" ] && [ ! -d "$pt_dir/rust/lib" ]; then \
-            echo "Removing corrupted platform-tools cache: $pt_dir"; \
-            rm -rf "$pt_dir"; \
-        fi; \
-    done
+    scripts/install-sbf-tools.sh v1.54
```
bgm-malbeclabs: `v1.54` now appears four times in this file (186, 194, 201, 208), and `release.testnet.yml` (389 against 390-392) and `release.devnet.smartcontract.daily.yml` (49 against 52) split the same way. Bump the build lines and miss the install line and the fallback comes back, failing with the same `edition2024` error many layers from its cause.

The file already has `ARG SOLANA_VERSION` as the precedent. An `ARG SBF_TOOLS_VERSION=v1.54` read by all four lines removes the trap.

bgm-malbeclabs (changes requested):
The diagnosis is right and the writeup is the best part of this PR. I checked `validate_platform_tools_version` in agave v3.0.4: it does return the requested version as soon as an installed copy parses `>=` the request, before it ever resolves `releases/latest`. The install paths, the OS and arch names and the asset name all match what agave builds, and `solana/Makefile` passes no `--tools-version`, so it takes the early return and its pinned checksums stay put.

Two things I want changed before this lands: the completeness check misses `llvm/`, which is half the toolchain the build uses, and the `--version` line cannot report what it claims to report. Three smaller notes inline.

On verification: the PR says both green runs were rust-cache hits that took the lenient path, so the strict path this fixes is still unexercised. That is an honest disclosure rather than a defect, but it does leave the fix untested against the failure it targets. One run with a forced cache miss would settle it.

nikw9944 (approved):
lgtm

scripts/install-sbf-tools.sh
bgm-malbeclabs: Fixed in `46b4f89db`. `installed()` now also requires `llvm/bin/clang`, so the extraction that stopped between `rust/` and `llvm/` reads as absent and is replaced instead of being reported as installed. Verified both ways against a staged cache: a tree with `rust/bin/{cargo,rustc}` and `rust/lib` but no `llvm/` reads as absent, and adding `llvm/bin/clang` makes it read as installed. The real v1.54 cache on this machine still passes, so the stricter test does not force a reinstall of a good tree.

scripts/install-sbf-tools.sh
bgm-malbeclabs: Fixed in `46b4f89db`, with the lock rather than `mv -T`, since `mv -T` is GNU-only and this script runs on macOS too. The whole install sits inside a `mkdir` mutex, which closes the `rm -rf` to `mv` gap as well as the read-while-deleted case, and it re-checks `installed()` after waiting so a build that queued behind the writer exits rather than reinstalling. A stale lock is taken after 900s.

Verified: the lock directory exists while an install runs, is gone after it fails, and the staging directory is cleaned with it.

.github/actions/solana-toolchain/action.yml
bgm-malbeclabs: You are right, and it is worse than useless in exactly the way you describe. Fixed in `46b4f89db`: the step now walks `~/.cache/solana/*/platform-tools` and reports each `rust/bin/rustc --version`, so the log says what a build would actually pick up, and says "incomplete" or "none installed" instead of a blank line.

.github/actions/solana-toolchain/action.yml
bgm-malbeclabs: Fixed by the same change, which is the part I liked about it. Reporting the cached `rustc` by path removed the only `cargo` invocation in that step, so nothing there resolves a toolchain and rustup has no reason to fetch the root 1.97.1 for `solana-test-sbf`. No `working-directory` needed, because no step in the action runs `cargo` any more.

e2e/docker/base.dockerfile
bgm-malbeclabs: Fixed in `46b4f89db`. `ARG SBF_TOOLS_VERSION=v1.54` in the base image feeds the install line and all three build lines; `release.testnet.yml` gets a step-level `env` and `release.devnet.smartcontract.daily.yml` a workflow-level one. `grep -rn "v1\.54"` across `.github/` and the dockerfile now returns only the three declarations and one prose mention.

scripts/install-sbf-tools.sh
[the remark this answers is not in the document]
bgm-malbeclabs: Already addressed on the branch in `b1ace9e5a`, before this review: `installed()` requires `rust/lib` alongside the binaries for exactly the reason given. `46b4f89db` extends the same test to `llvm/bin/clang`, which is the same failure one subtree over and the one this comment does not reach.

smartcontract/Makefile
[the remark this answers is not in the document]
bgm-malbeclabs: Stale by the time it was written. `smartcontract/test/start-test.sh`, `smartcontract/test/start-test-permissions.sh` and `e2e/docker/base.dockerfile` all call `scripts/install-sbf-tools.sh` before their build on this branch, as does `run_record_test.sh`. `grep -rn "cargo build-sbf" ` finds no call site without an install ahead of it.

bgm-malbeclabs (approved):

scripts/install-sbf-tools.sh
elitegreg: Right, and the comment above described the case without covering it. `llvm/bin/clang` is in the test as of 46b4f89db (yours), so a tree with `rust/` and no `llvm/` reads as absent and is replaced. Verified against a scratch cache: moving `llvm/` aside makes the next run reinstall rather than report "already installed".

scripts/install-sbf-tools.sh
elitegreg: Locked in 46b4f89db, with a re-check once the lock is held. I have put one thing back on top: the lock serializes this script against itself, but `cargo-build-sbf` never takes it and `install_if_missing` creates the version directory before it downloads. In the gap the `rm -rf` opens, it can be the writer that recreates `DEST`, and the plain `mv` then nests exactly as you described. `mv -T` where it exists closes that; BSD `mv` has no `-T` and no shared cache to protect.

Also raised `LOCK_TIMEOUT` from 900s: the download alone is allowed `ATTEMPTS * FETCH_TIMEOUT`, so at 900s a slow-but-live install has its lock declared stale and a second writer put beside it. It is now derived from those two, so the deadline cannot drift from the retry policy.

.github/actions/solana-toolchain/action.yml
elitegreg: Agreed, and the second half is the worse half. Gone in 46b4f89db, replaced by the rustc in each cache entry read by path.

.github/actions/solana-toolchain/action.yml
elitegreg: Fixed by the same change: reading the binaries by path leaves no `cargo` call in the step, so nothing pulls the root toolchain into `solana-test-sbf`.

e2e/docker/base.dockerfile
elitegreg: Done in 46b4f89db: `ARG SBF_TOOLS_VERSION` in the dockerfile feeding all four lines, and one `env` in each release workflow feeding both the install and the build.

PR #4339, opened 2026-09-15
internet-latency-collector: skip reconciliation when the measurement picture is incomplete

### LINKED ISSUE #4169: internet-latency-collector: a RIPE Atlas API error deletes all measurement metadata
## Summary

When the RIPE Atlas API returns an error, `GetAllMeasurements` failure is handled by substituting an **empty list**, which the orphaned-metadata cleanup then reads as "none of our measurements exist any more" and deletes every metadata entry. One expired API credential cost the mainnet-beta collector its entire measurement state.

The credential expiry was unavoidable. Losing the state was not, and it is what turned a fixable outage into a multi-day one.

## What happened on mainnet-beta

```
2026-08-08 00:00:00   RIPE Atlas API key expires
2026-08-08 00:05:37   first  "API request failed with status: 401"
2026-08-08 00:55:57   "Cleaning up orphaned metadata entries"  count: 29
                      state file -> {"metadata": {}}   (21 bytes)
2026-08-10 15:52      "No Samples Widespread" alert finally fires
2026-08-10 17:42      recovered by hand-rebuilding the state file
```

Fifty minutes after the key expired, the collector deleted all 29 metadata entries and then sat idle for 65 hours.

## Mechanism

`collector.go:819-820` — an API error is downgraded to an empty result set:

```go
c.log.Warn("Failed to get existing measurements", slog.String("error", err.Error()))
existingMeasurements = []Measurement{}
```

`collector.go:1165-1175` — everything in state is then orphaned, because `activeIDs` is built from that empty list:

```go
orphanedIDs := []int{}
for id := range measurementState.GetAllMetadata() {
    if !activeIDs[id] { orphanedIDs = append(orphanedIDs, id) }
}
```

There is no way for the cleanup to tell "the API says we have no measurements" from "we could not reach the API".

## Consequences observed

1. **Exports stop entirely.** The export path iterates metadata; with none, nothing is fetched or exported even for measurements that are alive and producing results.
2. **Measurements are orphaned at the provider, still billing.** `Removing measurement` never fired, because the API list was empty. All 29 measurements kept running at RIPE for the full 65 hours, at roughly 62,640 credits/day, with nothing collecting the results.
3. **Export cursors are lost**, so there is no automatic backfill once the API recovers. RIPE retains results ~10 days, so the data is still fetchable, but only out of band.
4. **A restart rebuilds the whole mesh.** With no metadata, every measurement hits "No metadata, can't verify sources, recreate" and also lands in `toRemove`, which costs a full recreation and the ~23h dispatch warm-up documented in #4153.
5. **One epoch lost entirely.** Epoch 199 (2026-08-08 06:00 → 2026-08-10 01:53) has **0 ripeatlas rows and 0 circuits**, against 435 circuits for `wheresitup`. That is the first epoch with zero ripeatlas coverage in the whole retained history (epochs 150-200).

## Suggested fix

Do not let a failed fetch stand in for an empty one. Minimally, skip the orphan cleanup when the fetch errored:

- return the error from `configureMeasurements` rather than continuing with `existingMeasurements = []`, or
- carry a `fetchOK bool` and gate the Step 7b cleanup on it.

The same substitution also drives the create/remove reconciliation, so on an API error the cycle currently believes every wanted measurement is missing. Gating the whole reconciliation, not just the cleanup, is the safer shape.

Worth considering alongside: state deletion is irreversible and silent at INFO level. A `.bak` on destructive state rewrites, or a WARN when the cleanup removes more than some fraction of entries, would have made this visible on day one.

## Related

- #4153 — the dispatch/warm-up behaviour that makes a forced mesh rebuild expensive.
- Companion issue on the detection gap: this ran for 63 hours before any alert fired.

nikw9944:
## Summary

- A failed `GetAllMeasurements` now fails the management cycle instead of reading as an empty fleet. Reconciling against an empty set made every wanted measurement `toCreate` and every tracked one unaccounted for, so one API error rebuilt all 30.
- An unreadable state file holds off **both** cycles, retried each cycle so a file repaired out of band resumes without a restart. Export is gated because it ends by saving the state file.
- `MeasurementState.Save` writes a temp file and renames over the target, and refuses to overwrite a file `Load` never read. That invariant belongs on the type owning the file.
- Only a path with nothing at it counts as a first deploy: a dangling symlink and an unmounted state directory also surface as ENOENT, and reading either as an empty fleet reaches the same wipe. The management cycle now creates the state directory, as export already did.
- The `missing metadata` → `toRemove` branch is unchanged by design: deleting a measurement the collector cannot account for is correct cleanup. The bug was reconciling before the picture was complete.

Fixes malbeclabs/doublezero#4169, and addresses the mechanism half of malbeclabs/doublezero#4131.

## Why

Full fleet wipes were observed in production , each resetting `CreatedAt`/`LastExportAt` and every source's `LastResponseAt`, buying a new fleet with RIPE credits and leaving a data gap:

Any `GetAllMeasurements` failure triggered it (timeout, 5xx, reset, or a request cancelled by shutdown), as does a restart reaching the same empty-tracker state through the non-atomic `Save` and warn-only `Load`. 

## Testing Verification

- `configureMeasurements` with an erroring `GetAllMeasurements`: no `CreateMeasurement`, no `StopMeasurement`, metadata survives in memory and on disk.
- With a corrupt state file, `RunRipeAtlasMeasurementCreation` returns an error and makes zero RIPE Atlas calls; after repair, the next call reaches the API. `ExportMeasurementResults` likewise makes no API call and leaves the file untouched — the ungated version overwrote it with `{"metadata":{}}`.
- `Save` refuses an existing file `Load` never read, replaces by rename rather than truncating in place (via inode identity), leaves no temp file, and leaves the target intact when the rename cannot land.
- A dangling symlink and a missing parent directory fail the load rather than reading as empty; `Save` renames onto a symlink's target instead of replacing the link; a new state file lands `0600`. Each guard was mutation-checked: reverting it fails its own test.


controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -822,10 +857,15 @@ func (c *Collector) configureMeasurements(ctx context.Context, locationMatches [
 	wantedMeasurements := c.generateWantedMeasurements(locationMatches, probesPerLocation, measurementState)
 
 	// Step 4: Get all existing measurements
+	// A failed fetch is not an empty fleet: substituting an empty list made every wanted
+	// measurement look missing and every tracked one orphaned, so one API blip deleted and
+	// recreated the entire fleet (#4169). Losing an interval of reconciliation is the
+	// cheaper failure.
 	existingMeasurements, err := c.client.GetAllMeasurements(ctx, c.env)
```
juan-malbeclabs, line 938: **A successful fetch that returns an empty page still takes the "fleet is empty" path.**

The fix covers the error return, but `GetAllMeasurements` returns `(nil, nil)` for a 200 with an empty first page (`client.go:359`) — a typo'd or wrong `--env` tag, RIPE-side filter/index lag, a paging glitch. `ExportMeasurementResults` guards this at line 415 (`if len(measurements) == 0 { return nil }`); `configureMeasurements` does not.

The outcome is worse than the error case this PR fixed:
- Step 7b builds `activeIDs` from an empty `doubleZeroMeasurements`, so **every** tracker entry is orphaned, removed, and the empty tracker is **persisted** by the `Save()` at line 1238.
- `existingByTarget` is empty, so every wanted measurement lands in `toCreate` — duplicates for the same target/probe pair.
- `toRemove` iterates `doubleZeroMeasurements`, so it is empty: the real measurements are never stopped. Their metadata is gone, so the collector can never reclaim them — they run forever burning credits.

The #4169 error path at least stopped-then-recreated. This one leaks the fleet. Suggest treating `len(existingMeasurements) == 0` as "picture incomplete" too, or deriving the precondition from the same place as the state-load gate.

Worth noting that `TestInternetLatency_RIPEAtlas_MeasurementCreation_GatedOnStateLoad` already drives this path and asserts it as the success case: after the state file is repaired to `{"metadata": {}}`, its `GetAllMeasurements` returns `[]Measurement{}` while two locations are wanted, so `require.Positive(t, apiCalls)` is satisfied by creating measurements against an empty tracker and an empty fleet. That fixture would need a real fleet for the post-repair run to exercise reconciliation, and would need updating if the empty-page case becomes a refusal.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -58,7 +63,8 @@ func (ms *MeasurementState) Load() error {
 
 	file, err := os.Open(ms.filename)
 	if os.IsNotExist(err) {
-		// File doesn't exist yet, keep empty tracker
+		// A clean empty state is a correct read of a first deploy, so it counts as loaded.
+		ms.loaded = true
```
juan-malbeclabs: **A missing state file satisfies the new gate, and `rm` is the only remediation an operator has for the refusal it produces.**

`Load` sets `loaded = true` on `ENOENT`, so `ensureMeasurementStateLoaded` succeeds and both cycles proceed against an empty tracker. Verified with a probe test: empty tracker + a healthy `GetAllMeasurements` returning a live fleet, and `configureMeasurements` stops every measurement (`stopped=[1001 1002]`) via the "missing metadata" branch at line 1137 and recreates them.

The gate blocks *corrupt* but not *absent*, and the two are one `rm` apart. An operator who sees `Refusing to manage measurements: measurement state could not be loaded` every cycle has no in-process recovery path (`Save` refuses too), so deleting the file is the obvious repair — and it triggers exactly the wipe this PR exists to prevent. Same shape for an unmounted or wrong `--state-dir`, a `/var/lib` volume that mounts after the unit starts, or a host rebuild.

The collector already has the information to tell these apart: the tracker is empty **and** the API reports N live DoubleZero measurements. Since `StopMeasurement` is irreversible on RIPE Atlas while the state file is recoverable, the guard belongs on the destructive action — refuse bulk removal when the tracker is empty and the fleet is not, or cap removals per cycle, with a metric on the refusal.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -101,24 +107,63 @@ func (ms *MeasurementState) Load() error {
...
+	// os.CreateTemp creates with 0600, so carry the existing mode (or os.Create's 0644)
+	// over rather than tightening it on every rewrite.
+	mode := os.FileMode(0644)
+	if info, err := os.Stat(ms.filename); err == nil {
+		mode = info.Mode().Perm()
 	}
-	defer file.Close()
 
-	encoder := json.NewEncoder(file)
+	tmp, err := os.CreateTemp(filepath.Dir(ms.filename), filepath.Base(ms.filename)+".tmp")
```
juan-malbeclabs: **`Save` now needs write permission on the *directory*, where `os.Create` needed it only on the file.**

`os.CreateTemp` + `os.Rename` both require `+w` on `filepath.Dir(ms.filename)`; a truncate-in-place write did not. Verified: file `0644` inside a `0555` dir — `os.Create(file)` succeeds (old code worked), new `Save()` returns `failed to create temp timestamp file: ... permission denied`.

Same class: `ENOSPC` (atomic save needs room for a second full copy, where truncate freed the old blocks first), and a Docker `-v file:file` bind mount, where rename-over-target fails but truncate worked.

This matters because of the amplification in the next comment — every `Save` in `configureMeasurements` is warn-only, so a persistent failure is silent.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -101,24 +107,63 @@ func (ms *MeasurementState) Load() error {
...
+// a file we cannot see with an empty tracker and reach that same outcome.
 func (ms *MeasurementState) Save() error {
 	ms.mu.Lock()
 	defer ms.mu.Unlock()
 
-	file, err := os.Create(ms.filename)
-	if err != nil {
-		return fmt.Errorf("failed to create timestamp file: %w", err)
+	if !ms.loaded {
+		if _, err := os.Stat(ms.filename); err == nil {
```
juan-malbeclabs: **The refusal guard only fires when `os.Stat` returns a nil error, so it is bypassed exactly when the file may exist and be unreadable.**

Any non-`ENOENT` stat failure — `EACCES` on a path component, `ELOOP`, `ENOTDIR`, `EIO`, or a dangling symlink — is read as "no file here" and the write proceeds, replacing a file we cannot see with the empty in-memory tracker. That is the one thing the guard was added to prevent.

```go
if !ms.loaded {
    if _, err := os.Stat(ms.filename); err == nil || !errors.Is(err, fs.ErrNotExist) {
        return fmt.Errorf("refusing to overwrite a measurement state file that was never loaded: %s", ms.filename)
    }
}
```

i.e. refuse unless absence is positively established. `Load` has the mirror-image asymmetry at line 65.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -101,24 +107,63 @@ func (ms *MeasurementState) Load() error {
...
+	if err := tmp.Sync(); err != nil {
+		return fmt.Errorf("failed to sync temp timestamp file: %w", err)
+	}
+	if err := tmp.Close(); err != nil {
+		return fmt.Errorf("failed to close temp timestamp file: %w", err)
+	}
+	if err := os.Chmod(tmpName, mode); err != nil {
+		return fmt.Errorf("failed to set mode on temp timestamp file: %w", err)
+	}
+	if err := os.Rename(tmpName, ms.filename); err != nil {
```
juan-malbeclabs: **`os.Rename` replaces a symlink at the state path; `os.Create` wrote *through* it.**

Verified: with `ripe_atlas_timestamps.json` a symlink into another directory, `Load` reads the target through the link (so `loaded = true`), and after one `Save` the path is no longer a symlink while the real target still holds its original content, frozen and orphaned. No error, and the mode is even preserved because `os.Stat` follows the link, so nothing in the logs shows the indirection was destroyed.

If the state file is ever managed as a symlink into a persistent volume — a standard pattern when `/var/lib` is ephemeral — every write after the first lands on the ephemeral path. The next redeploy re-converges the link, `Load` succeeds on the stale copy, and everything created since is unaccounted for: the line 1137 wipe again.

The dangling-link variant is worse: `os.Stat` and `os.Open` both report `ENOENT`, so `Load` declares a clean empty state, the refusal guard does not fire, and the link is destroyed on the first write. `os.Lstat` to detect it, or `filepath.EvalSymlinks` before deriving the rename target.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -101,24 +107,63 @@ func (ms *MeasurementState) Load() error {
...
+		tmp.Close()
+		os.Remove(tmpName)
+	}()
+
+	encoder := json.NewEncoder(tmp)
 	encoder.SetIndent("", "  ")
 	if err := encoder.Encode(ms.tracker); err != nil {
 		return fmt.Errorf("failed to encode timestamp file: %w", err)
 	}
+	if err := tmp.Sync(); err != nil {
```
juan-malbeclabs, line 365: **The atomic write is not crash-durable, and it leaks temp files.**

`tmp.Sync()` fsyncs the file but the parent **directory** is never fsynced after `os.Rename`, so the doc comment's atomicity claim holds for a process kill but not for the power loss / hard reset the incident notes describe as "during a restart" — the rename itself can be lost, leaving the previous version. Bounded (stale, not torn), but it is the durability the `Sync()` was presumably added to buy. An `os.Open(filepath.Dir(...))` + `Sync()` after the rename closes it.

Separately, the deferred `os.Remove(tmpName)` cannot run on `SIGKILL`/OOM-kill between `CreateTemp` and `Rename`, and nothing ever prunes the orphans. The name is `ripe_atlas_timestamps.json.tmp<digits>` (the pattern has no `*`, so the random string is appended). `Save` is called once per created measurement (line 1322), so a crash-looping rebuild can strand several full copies in the state dir. The two new tests assert `require.Len(t, entries, 1)` on the clean paths only. A `<base>.tmp*` sweep in `ensureMeasurementStateLoaded` would cover it.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -101,24 +107,63 @@ func (ms *MeasurementState) Load() error {
...
-		return fmt.Errorf("failed to create timestamp file: %w", err)
+	if !ms.loaded {
+		if _, err := os.Stat(ms.filename); err == nil {
+			return fmt.Errorf("refusing to overwrite a measurement state file that was never loaded: %s", ms.filename)
+		}
+	}
+
+	// os.CreateTemp creates with 0600, so carry the existing mode (or os.Create's 0644)
+	// over rather than tightening it on every rewrite.
+	mode := os.FileMode(0644)
```
juan-malbeclabs: **Minor, but the comment is factually wrong and the rename changes ownership.**

`os.Create` uses `0666 &^ umask`, not a fixed `0644`. Under a hardened umask (`UMask=0077`) the state file used to be created `0600` and is now hard-coded `0644`. Production is a no-op today since the shipped unit sets no `UMask=`, but the comment should not assert `0644` is what `os.Create` does.

Two related points: `os.Rename` installs a **new inode owned by the running process**, whereas truncate-in-place preserved the existing uid/gid — so a file deliberately chowned for an out-of-band reader changes owner on the first save. And `info.Mode().Perm()` copies whatever is at the path: `TestInternetLatency_RIPEAtlas_State_SaveFailureLeavesTargetIntact` shows the path can legitimately be a directory, in which case a succeeding rename would have left the state file `0755`.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -730,9 +731,41 @@ func sourcesWithoutSamples(measurements []Measurement, state *MeasurementState,
...
+		return nil, err
+	}
+	c.measurementStateLoaded = true
+	return c.measurementState, nil
+}
+
 func (c *Collector) RunRipeAtlasMeasurementCreation(ctx context.Context, dryRun bool, probesPerLocation int, stateDir string, samplingInterval time.Duration) error {
 	c.log.Info("Running RIPE Atlas measurement creation")
 
+	if _, err := c.ensureMeasurementStateLoaded(stateDir); err != nil {
```
juan-malbeclabs, line 837: **The management path never creates the state dir, so a missing one reads as "clean empty state" rather than a load failure.**

`ExportMeasurementResults` does `os.MkdirAll(stateDir, 0755)` at line 399; `RunRipeAtlasMeasurementCreation` and `Run`'s pre-goroutine `ensureMeasurementStateLoaded` do not. With a nonexistent state dir, `ensureMeasurementStateLoaded` returns **no error** (`Load` -> `ENOENT` -> `loaded = true`), so the new gate lets the cycle through against an empty tracker, and the subsequent `Save` fails with `failed to create temp timestamp file: ... no such file or directory`.

The damaging route is the `create-measurements` one-shot (`cmd/collector/main.go:298`): a typo'd or not-yet-created `--state-dir` stops every live measurement through the missing-metadata branch at 1137, recreates them, cannot persist the new metadata (warn-only `Save`), and exits 0. Same shape for the initial synchronous management run at 1547 if `/var/lib` mounts a moment after the unit starts, or after a package purge / host rebuild.

Correcting my earlier wording on this thread: in the long-running process the blast radius is one cycle, not every cycle forever. The creation loop's `SetMetadata` at 1321 keeps the new metadata in memory, and the export goroutine's `MkdirAll` creates the directory on its first tick, so later `Save`s land. At minimum `MkdirAll` here too, so a missing directory is distinguishable from a missing file.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -730,9 +731,41 @@ func sourcesWithoutSamples(measurements []Measurement, state *MeasurementState,
...
+//
+// Both the management and the export cycle must refuse to run while this returns an error.
+// Management would reconcile against an empty tracker and delete the fleet; export would
+// save that empty tracker over the file and hand the next management cycle the same
+// outcome one interval later.
+func (c *Collector) ensureMeasurementStateLoaded(stateDir string) (*MeasurementState, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.measurementState == nil {
```
juan-malbeclabs, line 821: **`stateDir` is used only on the first call and silently ignored afterwards, and the export path now mutates collector state where it previously used a throwaway.**

The deleted `ExportMeasurementResults` fallback rebuilt the timestamp path from its own `stateDir` argument every call. Verified: `ExportMeasurementResults(ctx, dirA)` then `RunRipeAtlasMeasurementCreation(ctx, dirB)` leaves `c.measurementState.filename` pointing at dirA — management reads and writes dirA's file while `os.MkdirAll(dirB)` creates a directory that is never touched. Silent wrong-file operation, no error.

Not reachable from today's single `Run(c.cfg.StateDir)` caller, but it is a live footgun for a second entry point or a per-env collector. Either take the path at construction and drop the parameter, or refuse when the incoming `stateDir` disagrees with the bound filename.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -796,13 +829,15 @@ func (c *Collector) RunRipeAtlasMeasurementCreation(ctx context.Context, dryRun
 }
 
 func (c *Collector) configureMeasurements(ctx context.Context, locationMatches []LocationProbeMatch, dryRun bool, probesPerLocation int, stateDir string, samplingInterval time.Duration) error {
-	// Step 1: Get measurement state (shared instance from Run(), or fallback for tests)
+	// Step 1: Get measurement state (shared instance from RunRipeAtlasMeasurementCreation,
+	// or fallback for standalone/test usage).
 	measurementState := c.measurementState
```
juan-malbeclabs, line 908: **The gate lives in the two entry points, not at the point of damage, and this read bypasses it.**

`configureMeasurements` — the function that actually stops measurements — reads `c.measurementState` lock-free and never consults `measurementStateLoaded`. Two consequences:

1. The struct comment added at line 83 now claims `mu` protects `measurementState`, which this line violates. It is benign today only by accident of ordering: `Run` pre-seeds at line 1537 before the goroutines start, so the sole writer at line 748 never fires again and `-race` stays quiet. That pre-seed now looks redundant (both cycles call `ensure` themselves) and is the natural thing for a follow-up to delete — at which point the two goroutines race on 748 vs 834 and can build two `MeasurementState` instances for the same file whose `Save`s overwrite each other.
2. `ensureMeasurementStateLoaded` assigns `c.measurementState` **before** attempting `Load` (line 748), so after a failed load the field is non-nil-but-unloaded. The `== nil` fallback at 835-841 can therefore never catch the unloaded case, and is dead in every production path. Any caller reaching `configureMeasurements` directly gets the shared empty tracker and reconciles against it.

Worth either moving the gate into `configureMeasurements`, or passing the state in as a parameter so there is no way to reach it ungated. (Relatedly, the new test sets `measurementStateLoaded: true` on a field `configureMeasurements` never reads, so that half of the fixture is inert.)

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -396,14 +400,11 @@ func (c *Collector) ExportMeasurementResults(ctx context.Context, stateDir strin
...
 
-	measurementState := c.measurementState
-	if measurementState == nil {
-		// Fallback for standalone/test usage without Run()
-		timestampFile := filepath.Join(stateDir, TimestampFileName)
-		measurementState = NewMeasurementState(timestampFile)
-		if err := measurementState.Load(); err != nil {
-			return err
-		}
+	measurementState, err := c.ensureMeasurementStateLoaded(stateDir)
```
juan-malbeclabs, line 435: **The export gate returns upstream of every sample-accounting metric, so an unloadable state file is an indefinite outage with no distinguishing signal.**

This return is above `LatencySamplesPerCollectionIntervalExpected` (line 476) and `...Missing` (line 535), so those counters simply stop moving — a dashboard keyed on the missing-samples metric reads flat/green while nothing at all is being exported. The only signals are `RipeatlasMeasurementManagementFailuresTotal`, `CollectionFailuresTotal` and one log line per interval; there is no gauge distinguishing "held off on state load" from "healthy but idle" (confirmed against `internal/metrics/metrics.go`).

This matters most for the case the PR is defending against: a **zero-byte** file is exactly what the old truncate-in-place `Save` left behind on a kill, `json.Decoder.Decode` returns `io.EOF`, so `Load` errors forever and `Save` also refuses — nothing can repair it in-process. The fleet keeps burning credits server-side while the collector exports zero samples until a human intervenes, and the intervention most likely to occur is the `rm` that opens the wipe path. The PR's "costs an interval without reconciliation" framing holds for the API-error path, but not for this one.

A `ripeatlas_state_file_loaded` 0/1 gauge (or an alert on the failure counters) would make the held-off state visible.

juan-malbeclabs (commented):
## Review: `internet-latency-collector: skip reconciliation when the measurement picture is incomplete`

The three changes here are the right ones and the reasoning in the description is sound — a failed fetch should not read as an empty fleet, and `Save` owning its own invariant is the right place for it. The findings below are about the parts of the same failure class the PR does not reach, and the new failure modes the atomic rewrite introduces. Several were verified with throwaway probe tests against this branch; those are marked.

Inline comments cover the findings that land on changed lines. Four more fall outside the diff:

---

### 1. The same fail-open substitution survives on the probe side, with a worse outcome (`internal/ripeatlas/client.go:161`, `:178`)

`fetchProbesWithErrorHandling` returns `[]Probe{}, nil` on error in **both** branches, so a failing `/probes/?radius=` request — same API, same key, same rate limit as the measurements call this PR just hardened — makes a location look probe-less rather than erroring. `GetProbesForLocations` even has an `if err != nil { return nil, err }` at `client.go:271` that can never fire.

Downstream: the location is dropped by `generateWantedMeasurements` (`collector.go:1457` and `1462`), so it is absent from `wantedTargets`, and `collector.go:1149` sends its live measurement to `toRemove`. It also shrinks the source set of the other measurements, so the `sourcesMatch` comparison at `collector.go:1120` fails and they land in `measurementsToRecreate` -> `toRemove`.

**Verified** with a probe test on this branch: healthy state file fully describing a 2-measurement fleet, healthy `GetAllMeasurements`, probes endpoint returning empty lists for every location →

```
stopped=[1001 1002] created=[] remaining_metadata=0
RunRipeAtlasMeasurementCreation err = <nil>
```

Every measurement stopped, **nothing recreated**, state file emptied, and the cycle returns `nil` — so it is counted in `RipeatlasMeasurementManagementRunsTotal` as a **success**. `StopMeasurement` is irreversible on RIPE Atlas, so this is strictly worse than #4169, which at least rebuilt what it destroyed. A probes-endpoint outage and a measurements-endpoint outage are the same kind of event; only one of them is now guarded.

This is the altitude point on the PR as a whole: the invariant worth stating is "never take a destructive action unless the inputs are complete", and it is currently enforced as a per-call-site error return on one of the two inputs. A single precondition next to the state-load gate — state loaded, locations non-empty, fleet fetched, wanted set non-empty — would cover both, and would make the "reconcile nothing" guarantee structural rather than incidental.

### 2. `Save` is now fail-closed but all four of its callers in `configureMeasurements` are still fail-open (`collector.go:1006`, `1202`, `1238`, `1322`)

Every one only `c.log.Warn`s and continues, and the rewrite *added* failure modes the old buffered `os.Create` + `defer file.Close()` never had (`CreateTemp`, `Sync`, the checked `Close`, `Chmod`, `Rename`, the new refusal, and the need for transient space for a second copy). The probability of this path went up, not down.

The damage is already done by the time the error is seen: `StopMeasurement` + `RemoveMetadata` have run before the `Save` at 1202, and the creation loop keeps creating after a failed `Save` at 1322. A failure at 1322 leaves a just-created measurement with no metadata, which line 1137 reads next cycle as "remove due to missing metadata" — an unbounded create/stop churn burning credits, visible only as a repeated `Warn`. The in-memory tracker papers over it for the life of the process, so the bill arrives at the next restart.

Given the PR's own reasoning, a `Save` failure in the removal and creation loops seems like it should abort the cycle the way the fetch failure now does.

### 3. Data race on `MeasurementMeta.Sources` (`state.go:261`, read at `collector.go:719`, `956`, `960`, `968`)

`GetMetadata` and `GetAllMetadata` return shallow value copies whose `Sources` slice header still points at the tracker's backing array, and `UpdateSourceProbeResponse` writes `meta.Sources[i].LastResponseAt` in place. So the export goroutine writes under `ms.mu` while the management goroutine reads the same words with no lock at all.

**Verified** — `go test -race` on this branch:

```
WARNING: DATA RACE
Write at 0x00c00010aad8 by goroutine 71:
  ...ripeatlas.(*MeasurementState).UpdateSourceProbeResponse()
      .../internal/ripeatlas/state.go:261
Previous read at 0x00c00010aad8 by goroutine 72:
  ...
```

Both sides are reachable from `Run`'s two goroutines: export calls `UpdateSourceProbeResponse` per result (`collector.go:622`), management reads the field in the unresponsive-probe detection. Existing tests never run the two goroutines together, which is why `-race` passes today.

Beyond the UB, the read is logically torn in a way that feeds the destructive path: a stale zero at line 956 skips a genuinely dead source probe, and a stale old value at line 960 marks a **healthy** probe unresponsive → `AddUnresponsiveProbe` → `wantedMeasurements` regenerated with a smaller source set → `sourcesMatch` fails → remove-and-recreate.

Pre-existing rather than introduced here, but this PR rewrites the locking comment on `mu` and is the hardening pass on this type. Fix is a deep copy of `Sources` in the two getters.

### 4. `InitializeMeasurementMetrics` is a third, independent load site with a dead branch (`collector.go:126-140`)

It builds its own `MeasurementState` and `Load`s the same file at startup, bypassing the new accessor. Its `if !os.IsNotExist(err)` at line 131 can never be false: `Load` returns `nil` for a missing file (a fact this PR formalises in the new comment at `state.go:66`) and wraps every other error, and `os.IsNotExist` does not unwrap. So the explicit zeroing and its "initializing metrics to zero" log are unreachable.

Combined with the new hold-off-and-retry recovery, `RipeatlasTotalMeasurements` is permanently wrong after a recovery: it is never `Set`, only `Inc()`'d at 1326 and `Dec()`'d at 1196. Start with a corrupt file, repair it out of band, management resumes exactly as advertised — and the gauge stays at 0 and thereafter tracks only deltas, under-reporting the fleet by ~30 for the process lifetime.

---

### Verified negatives, for the record

- No interface change needed: `RipeAtlasCollectorInterface` (`internal/collector/collector.go:15-19`) exposes only `Run` / `InitializeCreditBalance` / `InitializeMeasurementMetrics`, none of which changed signature.
- **No crash-restart loop.** `Run`'s goroutines only log and increment counters, and `internal/collector/run.go:70` pushes to `errChan`/`cancel()` only on a `Run()` error, which now happens solely for non-positive intervals. The per-cycle retry claim holds for the daemon.
- `MockClient`'s nil-func default for `GetAllMeasurements` returns `([]Measurement{}, nil)`, so no existing `configureMeasurements` test regresses on the new fetch error.
- `go test -race ./internal/ripeatlas/` and `golangci-lint` are both clean on this branch.

### Follow-up worth filing separately

`internal/wheresitup/state.go` carries the identical hazard, unhardened: `Load` (line 73) silently resets `jt.Jobs = []JobEntry{}` and returns `nil` on a decode failure, so a torn file is indistinguishable from an empty one, and `Save` (line 102) is still a truncate-in-place `os.Create`. `AddJobIDs`/`AddJobIDsWithCircuits` do Load-then-Save, so a corrupt read is immediately made permanent and every outstanding WIU job is orphaned — paid for, never collected. Both collectors share `StateDir` and run under the same process, so one `SIGKILL` can catch either mid-write. Same class as #4131. There is no atomic-write helper anywhere in the repo, so a small shared one would cover both writers.

juan-malbeclabs (commented):
Second pass, focused on the `loaded` flag, the atomic `Save`, error propagation, and the overlap with #4331. Summary of what held up and what did not.

**Verified sound**

- The `loaded` flag is set on every path where `Load` legitimately succeeds, including the missing-file path (`state.go:67`), so a clean first deploy does not leave `Save` refusing forever. `Save` also sets it, and the two flags (`MeasurementState.loaded` and `Collector.measurementStateLoaded`) cannot disagree in the direction that would re-`Load` over accumulated in-memory state: while `measurementStateLoaded` is false both cycles refuse, so nothing accumulates.
- `measurementStateLoaded` is written and read only inside `ensureMeasurementStateLoaded`, under `c.mu`, so the flag itself is synchronized across the two goroutines. The unlocked read of `c.measurementState` at `collector.go:834` is benign today only because `Run` seeds the pointer before either goroutine starts — see the inline note there.
- No partial mutation is left behind by the new `GetAllMeasurements` error return: `fetchFallbackProbesForUnresponsiveLocations` and `generateWantedMeasurements` are read-only, and `PruneExpiredUnresponsiveProbes` / the staleness loops / every `Save` are all downstream of the fetch.
- The new error returns do not abort a loop that previously continued. Both tickers in `Run` log and keep going, `run.go:35` only warns, and no caller treats `*CollectorError` differently from a plain error, so there is no crash-restart loop and the per-cycle retry claim holds.
- `Save`'s temp file is created in `filepath.Dir(ms.filename)`, so the rename is same-filesystem; the deferred `Close`/`Remove` cover every error return; a failed rename leaves the target byte-for-byte intact (the new test asserts this).
- `go build ./controlplane/internet-latency-collector/...` and `go test -race ./controlplane/internet-latency-collector/...` are clean on this branch.

**Did not hold up**: the `missing metadata` -> `toRemove` branch can still be reached with an incomplete picture. Three routes survive and are in the inline comments: a missing state file (`state.go:67`), a missing state dir (`collector.go:763`), and a 200 with an empty measurement page (`collector.go:864`). The probe-side fail-open at `client.go:161` is a fourth, and the worst, since it destroys without recreating.

**Merge order with #4331** (`gm/ripeatlas-lossy-probe-rotation`, which rewrites the same two files)

- `collector.go` and `state.go` auto-merge cleanly between the two branches; only `collector_test.go` and `state_test.go` conflict, and only because both append at the end of the file. The merged `state.go` keeps both `ms.loaded = true` sites and #4331's `tracker.UnresponsiveTargets = intermediate.UnresponsiveTargets`, and the merged `collector.go` keeps both the load gate and `rankTargets`.
- Semantically the two are complementary rather than competing: this PR closes the "reconcile against an incomplete picture" route, #4331 closes the "candidate set went empty" route. #4331's new persisted fields (`unresponsive_targets`, `target_*` on each meta) are decoded before `loaded` is set, so this PR's `Load`/`Save` invariant carries over unchanged, and #4331's `RecordTargetResults` mutates persisted state from the export goroutine, which is exactly why the export gate added here has to stay in front of it.
- One real dependency, which is my main reason to land #4331 first or together with this one: the hold-off this PR adds ends in a full fleet rebuild on the first cycle after recovery, because #4331's `rankTargets` is what stops a blacklisted target probe from emptying a location's candidate set. Detail in the inline comment on `ensureMeasurementStateLoaded`.

Not approving yet given the surviving fail-open routes, but the direction is right and the atomic `Save` plus the load gate are both clear improvements on their own.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
```diff
@@ -730,9 +731,41 @@ func sourcesWithoutSamples(measurements []Measurement, state *MeasurementState,
 	return byLocation, total, sample
 }
 
+// ensureMeasurementStateLoaded creates the shared measurement state if needed and reads it
+// from disk, retrying on every call until a read succeeds, so a state file repaired out of
+// band resumes the collector without a restart. A missing file is a clean empty state,
```
juan-malbeclabs: **A hold-off longer than an hour ends in the fleet rebuild this PR exists to prevent.**

Because export is gated too, `LastExportAt` stays frozen at whatever the file last recorded for the whole duration of the hold-off. The first management cycle after the file finally loads therefore sees every measurement's `LastExportAt` older than `probeTimeout` (`now - 3600`, line 902), takes the `no_recent_exports` branch at 914, and calls `AddUnresponsiveProbe(meta.TargetProbeID)` for all 30. `newUnresponsiveProbes == 30` then regenerates `wantedMeasurements` at 1015 with every target probe blacklisted, `filterResponsiveProbes` drops each single-anchor location ("No responsive probes found for location", 1462), `wantedTargets` comes back empty, and the loop at 1133 sends all 30 live measurements to `toRemove` through `!wantedTargets[targetLocation]`. Stopped, nothing created, and the probes stay blacklisted for `UnresponsiveProbeExpiry` (24h). A metro with a second anchor fares no better: the alternate probe wins selection, `targetProbeChanged` fires, and the measurement is torn down and recreated anyway.

Concretely: torn state file at T, operator repairs it at T+2h -> `Run`'s synchronous startup management run at 1547 (or the next hourly tick, which lands on the same instant as an export tick at every hour boundary, so ordering is a coin flip) rebuilds the entire fleet with fresh credits and reset `CreatedAt`/`LastExportAt`.

The staleness cascade itself is pre-existing — any export outage longer than an hour triggers it — but gating export is a new systematic trigger, and it fires on precisely the cycle this doc comment promises "resumes the collector without a restart". #4331 removes it: `rankTargets` ranks marked probes last instead of dropping them, so no location loses its target and `wantedTargets` never empties. Worth either landing #4331 first, or skipping the staleness pass on the first cycle after a hold-off.

juan-malbeclabs (approved):
Approving. The core change is right and the mechanism is sound: failing the management cycle on a `GetAllMeasurements` error, and holding off both cycles on an unreadable state file, close the route that wiped the fleet three times in five weeks. I verified the specifics rather than taking the description on trust:

- The `loaded` flag is set on every legitimate success path, including the missing-file path, so a clean first deploy will not have `Save` refusing forever. It is written and read only under `c.mu`.
- No partial mutation survives the new error return: the prune, both staleness passes, every `Save`, `StopMeasurement` and `CreateMeasurement` are all downstream of the fetch.
- No loop that previously continued now aborts. Both tickers log and continue, and no caller branches on the error type, so the "retried each cycle without a restart" claim holds.
- Atomic `Save` is correct in the essentials: the temp file is in `filepath.Dir(target)`, the deferred close and remove cover every error return, and a failed rename leaves the target intact.
- `go test -race -count=1` passes on the package, and on the whole `internet-latency-collector` tree.

The comments are all non-blocking and fall into two groups. Several are fail-open routes that survive this fix and are worth separate follow-ups rather than scope creep here, the sharpest being that a 200 with an empty measurement page still takes the "fleet is empty" path, and that `fetchProbesWithErrorHandling` returns an empty slice with a nil error so a probes-endpoint outage empties every location. The rest are narrow consequences of the rename-based `Save`: directory write permission is now required, a symlink at the state path is replaced rather than written through, there is no parent-directory fsync, and a temp file leaks on SIGKILL.

One merge-order note. The `ensureMeasurementStateLoaded` comment thread describes the one place where this change adds a systematic trigger: a hold-off longer than an hour freezes `LastExportAt`, so the first management cycle after the file loads sees every measurement stale, marks every target probe, and empties `wantedTargets` for single-anchor locations, ending in the rebuild this PR exists to prevent. #4331's `rankTargets` removes that cascade by ranking marked targets rather than dropping them. I would land #4331 first or alongside this one. The two auto-merge cleanly in `collector.go` and `state.go`; only the two test files conflict, and only because both append at end of file.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -58,7 +63,8 @@ func (ms *MeasurementState) Load() error {
 
 	file, err := os.Open(ms.filename)
 	if os.IsNotExist(err) {
-		// File doesn't exist yet, keep empty tracker
+		// A clean empty state is a correct read of a first deploy, so it counts as loaded.
+		ms.loaded = true
 		return nil
```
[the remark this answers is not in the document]
nikw9944: Fixed in 3e6a06c7. `Load` now clears `loaded` before it opens the file, so a reload that fails after an earlier success no longer leaves `Save` treating the stale in-memory tracker as a trusted description of what is now on disk.

`TestInternetLatency_RIPEAtlas_State_LoadClearsLoadedOnFailure` covers it, and fails against the previous code.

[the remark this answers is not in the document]
nikw9944: Fixed in 3e6a06c7, from both ends.

`Load` now accepts the missing-file path only when nothing at all sits at the state path *and* its parent is a real directory (`pathIsAbsent`). A missing or unmounted parent is no longer a clean empty state; it fails the load and keeps both cycles gated.

The management cycle also creates the state directory now, as export already did, so a not-yet-created `--state-dir` is a first deploy rather than an unreadable path. That was the more likely route in practice — see the thread on `collector.go:763`.

Covered by `TestInternetLatency_RIPEAtlas_State_LoadRejectsUnreadablePaths`.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
```diff
@@ -101,24 +107,63 @@ func (ms *MeasurementState) Load() error {
...
 	defer ms.mu.Unlock()
 
-	file, err := os.Create(ms.filename)
-	if err != nil {
-		return fmt.Errorf("failed to create timestamp file: %w", err)
+	if !ms.loaded {
+		if _, err := os.Stat(ms.filename); err == nil {
+			return fmt.Errorf("refusing to overwrite a measurement state file that was never loaded: %s", ms.filename)
+		}
+	}
```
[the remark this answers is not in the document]
nikw9944: Fixed in 3e6a06c7. The guard now refuses unless absence is positively established:

```go
if _, err := os.Lstat(ms.filename); err == nil || !errors.Is(err, fs.ErrNotExist) {
    return fmt.Errorf("refusing to overwrite a measurement state file that was never loaded: %s", ms.filename)
}
```

`Lstat` rather than `Stat` for the dangling-symlink case, and the inverted condition means EACCES, ELOOP, ENOTDIR and EIO all refuse instead of reading as "no file here" — which was the guard failing exactly when it mattered.

`Load` had the mirror-image asymmetry and is fixed the same way.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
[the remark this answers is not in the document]
nikw9944: Fixed in 3e6a06c7. A new state file now keeps the temp file's `0600`; the mode is copied only from an existing **regular** file, so measurement metadata cannot become world-readable because the file was recreated. The comment claiming `os.Create` uses `0644` is gone — it was wrong, as the thread on line 133 also notes.

`TestInternetLatency_RIPEAtlas_State_SavePermissions` asserts both halves.

controlplane/internet-latency-collector/internal/ripeatlas/collector_test.go
```diff
@@ -1947,3 +1947,209 @@ func TestInternetLatency_RIPEAtlas_ExportSingleMeasurementResults_LossCountsAsRe
...
+		{
+			LocationMatch: collector.LocationMatch{LocationCode: "lon", Latitude: 51.5074, Longitude: -0.1278},
+			NearbyProbes:  []Probe{{ID: 200, Address: "2.2.2.2", Latitude: 51.5074, Longitude: -0.1278}},
+			ProbeCount:    1,
+		},
+	}
+
+	err := c.configureMeasurements(t.Context(), locationMatches, false, 1, stateDir, 1*time.Minute)
+	require.Error(t, err, "a failed measurement fetch should fail the cycle")
+	require.Contains(t, err.Error(), "failed to get existing measurements")
```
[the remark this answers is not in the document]
nikw9944: Tightened in 3e6a06c7. The fetch-failure test now pins the whole message and the cause, so a regression that fails at a different stage fails the test:

```go
require.EqualError(t, err, "failed to get existing measurements: "+apiErr.Error())
require.ErrorIs(t, err, apiErr)
```

The two state-load tests (lines 2099 and 2145 in your numbering) now assert the stage as well as the wrapper — `"failed to load measurement state: failed to decode timestamp file:"` — rather than exact equality, because the tail is `encoding/json`'s wording and pinning it would make the test brittle against a Go upgrade for no added coverage.

I did not add log-record assertions. All three tests already pin the behavior that matters at the stage in question (`require.Zero` on API calls and on location lookups, metadata surviving, the file left byte-for-byte intact), which is stronger evidence than a log line at a fixed index and does not break when a log call is added.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
nikw9944: Fixed the "absent is not the same as unreadable" half in 3e6a06c7, and I agree with your framing on the rest.

A missing state file now only satisfies the gate when nothing sits at the path *and* its parent is a real directory, and the management cycle creates the state directory itself. That removes the unmounted / wrong `--state-dir` / host-rebuild routes, and with them most of the pressure toward `rm` as the remediation: the refusal an operator now sees comes from a genuinely unreadable file, not a missing directory.

What I have **not** done is guard the destructive action itself — refusing bulk removal when the tracker is empty and the fleet is not, or capping removals per cycle. You are right that this is where the guard belongs, given `StopMeasurement` is irreversible while the state file is recoverable, and it subsumes the empty-page route on the `collector.go:864` thread. It is a behavior change to reconciliation with its own fixture work (as you noted, `..._GatedOnStateLoad` currently asserts the empty-fleet path as the success case), so I have kept it out of this PR rather than grow it further. Filed as a follow-up.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
nikw9944: Fixed in 3e6a06c7, using your formulation — refuse unless absence is positively established, with `Lstat` so a dangling symlink does not read as absent:

```go
if _, err := os.Lstat(ms.filename); err == nil || !errors.Is(err, fs.ErrNotExist) {
    return fmt.Errorf("refusing to overwrite a measurement state file that was never loaded: %s", ms.filename)
}
```

`Load`'s mirror-image asymmetry is fixed too: the missing-file path now requires nothing at the path *and* a real parent directory.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
nikw9944: Fixed in 3e6a06c7, both variants.

`Save` derives the rename target through `filepath.EvalSymlinks`, so the temp file is created next to the real target and renamed onto it. A state file symlinked into a persistent volume keeps its indirection. `TestInternetLatency_RIPEAtlas_State_SaveWritesThroughSymlink` asserts the path is still a symlink afterwards and that the write landed on the target; it fails against the previous code.

The dangling variant is closed at the `Load` end, which is where it had to be: `os.Open` returns ENOENT, so nothing downstream could tell it from a first deploy. `Load` now checks `Lstat` on the path before accepting the missing-file branch, so a broken link fails the load, `loaded` stays false, and `Save`'s refusal fires and preserves the link for repair.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
nikw9944: Both fixed in 3e6a06c7.

Parent-directory fsync after the rename, which is the durability the `Sync()` was meant to buy:

```go
if err := syncDir(filepath.Dir(target)); err != nil {
    return fmt.Errorf("failed to sync timestamp file directory: %w", err)
}
```

Temp-file sweep in `Load`, pruning `<base>.tmp*` older than an hour. The age floor is deliberate: an unconditional sweep could delete a temp file another process has in flight (the `create-measurements` one-shot alongside the daemon), and since every `Save` in `configureMeasurements` is warn-only that would be a silent state loss — trading a bounded leak for a worse failure. `TestInternetLatency_RIPEAtlas_State_LoadPrunesStaleTempFiles` asserts the stale file goes and a recent one stays.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
nikw9944: Fixed in 3e6a06c7. The comment is gone — it was wrong, as you say — and the default is now `0600` rather than a hard-coded `0644`, so a hardened umask is no longer widened by a rewrite.

The mode is copied only from an existing **regular** file, which also covers the directory case you flagged in `..._SaveFailureLeavesTargetIntact`: a succeeding rename can no longer pick up `0755` from a directory at the path.

The ownership change is inherent to rename-based atomic writes and I have left it. Worth recording that it is a real difference from truncate-in-place for anyone who chowns the file for an out-of-band reader.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
nikw9944: Fixed in 3e6a06c7. `ensureMeasurementStateLoaded` now does the `MkdirAll` itself, so both entry points get it and a missing state dir is distinguishable from a missing file:

```go
if err := os.MkdirAll(stateDir, 0755); err != nil {
    return nil, fmt.Errorf("failed to create state directory: %w", err)
}
```

The now-redundant `MkdirAll` at the top of `ExportMeasurementResults` is removed, since `ensure` runs on the next line.

That closes the `create-measurements` one-shot route you identified as the damaging one, and combined with `Load` no longer accepting a missing parent directory, a genuinely unavailable path now fails the load and holds both cycles off rather than reading as an empty fleet. Thanks for the correction on blast radius — it matched what I saw.

Three existing tests called `configureMeasurements` directly against a `state/` dir they never created, and were relying on exactly this fail-open behavior; they create it now.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
nikw9944: Not fixed here — deferring deliberately, and flagging that it needs a narrower condition than `len(existingMeasurements) == 0`.

You are right that the outcome is worse than the error path: the fleet leaks rather than being stopped-and-recreated. But refusing on an empty fetch alone would also refuse the genuine bootstrap case, where an empty tracker and an empty fleet are both correct and the cycle must create. The condition that separates them is disagreement between the two — tracker non-empty while the fleet is empty (metadata describes measurements the API does not report), which is the mirror of the "tracker empty, fleet not" case on the `state.go:67` thread.

So the right shape is the single precondition you describe in the review body, or a guard on the destructive action, rather than a per-call-site length check. Both routes plus the `client.go:161` probe-side fail-open want the same fix, and it needs the `..._GatedOnStateLoad` fixture rebuilt around a real fleet, as you noted. Filed as a follow-up rather than grown into this PR.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
nikw9944: Not moved in this PR. Agreed on the analysis, including that the `== nil` fallback is dead in every production path and that the pre-seed at 1537 is now redundant and is the natural thing for a follow-up to delete.

Two things I did rather than restructure: the `MkdirAll` and the `Load` hardening mean the ungated path has fewer ways to produce an empty tracker in the first place, and the tests that reach `configureMeasurements` directly now create their state dir instead of depending on the fail-open.

Moving the gate into `configureMeasurements` or passing the state in as a parameter is the correct fix and I would rather do it as its own change than bundle a signature change with the guards here — particularly with #4331 rewriting the same function. Filed as a follow-up, called out explicitly so the pre-seed is not deleted on its own.

You are right that the new test's `measurementStateLoaded: true` is inert for `configureMeasurements`; I have left it, since the field is what the gated entry points read and removing it would make the fixture misleading in the other direction once the gate moves.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
nikw9944: Not addressed here — no metric added. Agreed that "held off on state load" and "healthy but idle" are indistinguishable on a dashboard today, and that a `ripeatlas_state_file_loaded` 0/1 gauge is the right signal.

Narrowing the case somewhat: the zero-byte file you describe as unrepairable in-process is the one the atomic `Save` stops being produced going forward, and the `Load` and `MkdirAll` changes in 3e6a06c7 remove the missing-directory and dangling-symlink routes into an indefinite hold-off. What remains is a genuinely corrupt file, which does need a human and should be visible as such.

Adding a gauge means touching `internal/metrics` and the alerting that reads it, which I would rather not fold into this change. Filed as a follow-up.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
nikw9944: Not changed, deliberately. The footgun is real and your trace is right, but it is unreachable from today's single `Run(c.cfg.StateDir)` caller, and the guards in this PR are all for failures the code can actually reach now.

Taking the path at construction and dropping the parameter is the fix I would pick — it makes the wrong-file case unrepresentable rather than merely detected. That is a signature change across the entry points, which is worth its own review, especially with #4331 in flight on the same file. Filed as a follow-up.

One thing this PR does change: `ensureMeasurementStateLoaded` now calls `MkdirAll(stateDir)` on every call, so a second entry point passing a different `stateDir` creates that directory while still operating on the bound file. That makes the divergence more visible on disk, not less — an argument for doing the constructor fix rather than leaving it.

controlplane/internet-latency-collector/internal/ripeatlas/collector.go
nikw9944: Not addressed in code — this is a merge-order decision, and I agree with your recommendation to land #4331 first or alongside.

The cascade is as you describe and I am not going to paper over it with a first-cycle skip: that would add a special case to the staleness pass to compensate for a candidate-set problem that #4331 fixes properly via `rankTargets`. Flagging it for the operator as a sequencing constraint on this PR rather than a code change, and it is in the summary comment so it is not buried in a thread.

Worth noting the hold-off is now shorter in the cases this PR touches: a missing state directory and a dangling symlink no longer produce one at all. A corrupt file still can, so the constraint stands.

controlplane/internet-latency-collector/internal/ripeatlas/state.go
nikw9944: Acknowledged, not changed — this is inherent to rename-based atomic writes and I do not think it should be traded away.

You are right about all three cases: `+w` on the directory, `ENOSPC` needing room for a second copy, and a `-v file:file` bind mount where rename-over-target fails. A truncate-in-place write survives each of them. But the failure mode it survives them with is a torn file, which is what wiped the fleet three times in five weeks, so failing the save is the better outcome than half-writing it.

What I have done in 3e6a06c7 is remove the amplification you point at in the next comment for the worst case: an interrupted or refused save can no longer leave the process with a state file it will silently treat as authoritative, since `Load` clears `loaded` and the refusal now fires on any stat error rather than only a clean success.

The warn-only `Save` callers are the part that makes a persistent failure silent, and I agree they should abort the cycle the way the fetch failure now does. That is your review item 2 and it changes reconciliation's control flow in four places, so I have filed it as a follow-up rather than fold it in here. The deployment cases above are exactly why it matters, and I have said so in the follow-up.

nikw9944:
Thanks — the verified probe tests made this much easier to act on, particularly the dangling-symlink and missing-state-dir traces.

Pushed 3e6a06c7, which fixes the defects in this PR's own new code and leaves the surviving fail-open routes to follow-ups, per your split.

**Fixed**

- `Load` clears `loaded` before opening, so a failed reload drops the trust an earlier success established.
- A missing file counts as a first deploy only when nothing sits at the path *and* its parent is a real directory. A dangling symlink and an unmounted state directory no longer read as an empty fleet.
- `Save`'s refusal now requires absence to be positively established, via `Lstat` — EACCES, ELOOP, ENOTDIR, EIO and a broken link all refuse instead of being read as "no file here".
- `Save` renames onto a symlink's target rather than replacing the link.
- A new state file defaults to `0600` instead of inheriting the umask; the mode is copied only from an existing regular file, which also fixes the directory case.
- Parent-directory fsync after the rename.
- `Load` prunes temp files an interrupted `Save` could not clean up, with an age floor so it cannot delete an in-flight temp from another process.
- The management cycle creates the state directory, as export already did.

Five new tests, each verified to fail against the previous code. Three existing tests called `configureMeasurements` directly against a `state/` dir they never created and were relying on the fail-open; they create it now. The error assertions you and Copilot flagged now pin the failing stage.

**Deferred, filed as follow-ups**

The probe-side fail-open in `client.go`, the empty-measurement-page route, the four warn-only `Save` callers, the `Sources` data race, `InitializeMeasurementMetrics`' dead branch and the gauge that stays wrong after a recovery, moving the gate into `configureMeasurements`, binding `stateDir` at construction, a `ripeatlas_state_file_loaded` gauge, and the same hazard in `internal/wheresitup/state.go`.

On the empty-page case specifically: refusing on `len(existingMeasurements) == 0` alone would also refuse a genuine bootstrap, where an empty tracker and an empty fleet are both correct. The condition that separates them is disagreement between tracker and fleet, which is the same guard the missing-file thread asks for — so both want the single precondition from your review body rather than a length check per call site.

**Merge order:** agreed, #4331 should land first or alongside. I have not added a first-cycle skip to work around the staleness cascade, since `rankTargets` fixes the cause. Treating that as a sequencing constraint for whoever merges.

nikw9944:
Follow-ups from this review are filed, so the deferrals in the threads above have somewhere to land:

- malbeclabs/doublezero#4355 — destructive reconciliation on incomplete inputs: the `client.go` probe-side fail-open, the empty-measurement-page route, the four warn-only `Save` callers, and the gate's placement. Notes that the pre-seed at `Run` must not be deleted before the gate moves.
- malbeclabs/doublezero#4356 — the `Sources` data race, `InitializeMeasurementMetrics`' dead branch and the gauge that stays wrong after a recovery, the missing held-off signal, and binding `stateDir` at construction.
- malbeclabs/doublezero#4357 — the identical torn-read hazard in `internal/wheresitup/state.go`, with the suggestion to extract a shared atomic-write helper.

nikw9944:
Rebased onto main at c1bb994d. No new code changes — the review fixes are unchanged, but the commits were rewritten by the rebase, so the SHAs cited in the threads above have moved: 3e6a06c7 is now 0593b78f, and the comment tidy 63b0c138 is now c1bb994d.

**malbeclabs/doublezero#4331 has landed**, which settles the merge-order note on the `ensureMeasurementStateLoaded` thread. `rankTargets` is now in the same tree as the hold-off, so the first management cycle after a recovery no longer empties `wantedTargets` for single-anchor locations. That was my one open sequencing risk and it is gone.

Conflicts were confined to `state.go` and the two test files, and were additive in every case: #4331's `migrateTargetMarks` / `MigratedTargetMarks` and its `migratedTargetMarks` field sit alongside this PR's `loaded` flag and atomic `Save`; the test files collided only where both branches append at the end.

Verified on the merged tree rather than assumed:

- `go build`, `go vet`, `golangci-lint` (0 issues), and `go test -race -count=1` all clean across `internet-latency-collector`.
- This PR's five state-file guard tests and #4331's `TestInternetLatency_RIPEAtlas_State_EvaluateTargetLoss` all pass together.
- The export gate is still the first statement in `ExportMeasurementResults`, ahead of #4331's `RecordTargetResults`, which is the ordering called out in the second-pass review.
- Diffed this branch's added lines before and after the rebase: identical apart from one `if err != nil {` that #4331 now supplies as context.

Nothing else changed. The follow-ups remain malbeclabs/doublezero#4355, malbeclabs/doublezero#4356 and malbeclabs/doublezero#4357.

Issue #4186, opened 2026-08-13
solana-cli: add a Squads vault path to shreds validator-client-rewards set-proportion


karl-dz:
Depends on #4185

Target repo: [doublezerofoundation/doublezero-offchain](https://github.com/doublezerofoundation/doublezero-offchain)

## Summary

`SetValidatorClientRewardsProportion` takes the `ValidatorClientRewards` manager as its only signer, the same authority `ClaimValidatorClientRewards` requires, and `shreds validator-client-rewards set-proportion` has no vault path. A manager held by a Squads vault cannot set a proportion.

- Add the `OptionalSquadsArgs` vault path the same way `claim` gets it, including the writer-taking variant of `print_vault_transaction`. One instruction and no batching, so one payload, encoded through the checked encoder rather than the budget accessor.
- The vault path builds a `SolanaConnection` and no `Wallet`. `build_wallet` fails outright when no keypair is loadable, and the vault path has nothing to sign with.
- Leave `set_compute_unit_limit` out of the payload. Squads sets the budget on its own execute transaction, and a compute budget instruction reached through a CPI is a no-op that only burns compute units.
- Call `validate_manager` before building the instruction, against the wallet key on the direct path and the vault key on the vault path. This command reads no accounts at all today, so it gains a fetch of the `ValidatorClientRewards` account. It sends whatever the wallet signs and lets the program reject it, so a wallet that is not the manager gets an `invalid account data` failure rather than a message naming `--multisig`, and a vault that is not the manager would get a payload nothing can execute.
- `validate_manager` lives in `claim` and now has a second caller, so hoist it into the `validator_client_rewards` module rather than copying it.

nikw9944:
<details>
<summary>System Prompt</summary>

# Code Review

You are a Claude Code instance managed by Roadbike, performing a thorough code review.
You must NOT make any code changes — review only. Do NOT post the review to GitHub —
the operator will read your findings in the dashboard and decide what to do.

## Assignment

- **PR**: [malbeclabs/doublezero#4332](https://github.com/malbeclabs/doublezero/pull/4332)
- **Title**: solana-cli: add a Squads vault path to shreds validator-client-rewards set-proportion
- **Repository**: malbeclabs/doublezero
- **Branch**: `gm/set-proportion-squads-vault` → `main`
- **Linked Issue**: #4186 — solana-cli: add a Squads vault path to shreds validator-client-rewards set-proportion

## PR Description

<pr-description>
[...]

PR #1608, opened 2025-09-10
Enable AccessPass with client_ip=0.0.0.0 to dynamically learn the user’s IP on first connection

juan-malbeclabs:
This pull request introduces several enhancements and fixes across the CLI and onchain programs, mainly focused on AccessPass management, multi-IP support, and improved CLI usability. The most notable changes include allowing AccessPass creation without a specified client IP, supporting connections from multiple IPs, updating CLI argument handling, and refining test coverage for these scenarios.

**AccessPass multi-IP support and dynamic IP handling:**

* Added the ability to create AccessPasses without specifying a `client_ip`, enabling dynamic IP learning on first connection and compatibility with allowlists. Also introduced the `allow_multiple_ip` argument for supporting connections from multiple IPs. [[1]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR11-R17) [[2]](diffhunk://#diff-dd0b8fce2acca223e0b8169ce27bb2986acabb1d568a7ccc5fb4c94464c0d595L24-R24) [[3]](diffhunk://#diff-dd0b8fce2acca223e0b8169ce27bb2986acabb1d568a7ccc5fb4c94464c0d595R34-R36) [[4]](diffhunk://#diff-dd0b8fce2acca223e0b8169ce27bb2986acabb1d568a7ccc5fb4c94464c0d595L68-R74)
* Updated onchain logic and CLI to handle AccessPasses with `client_ip=0.0.0.0` and to display flags indicating dynamic/multi-IP support. [[1]](diffhunk://#diff-49622a8ec624eea6e893b5e01e9b19c0868b2807a1b9f901d251feca47bf32abR41) [[2]](diffhunk://#diff-49622a8ec624eea6e893b5e01e9b19c0868b2807a1b9f901d251feca47bf32abR96) [[3]](diffhunk://#diff-49622a8ec624eea6e893b5e01e9b19c0868b2807a1b9f901d251feca47bf32abR146-R153) [[4]](diffhunk://#diff-49622a8ec624eea6e893b5e01e9b19c0868b2807a1b9f901d251feca47bf32abR162) [[5]](diffhunk://#diff-49622a8ec624eea6e893b5e01e9b19c0868b2807a1b9f901d251feca47bf32abL180-R184) [[6]](diffhunk://#diff-49622a8ec624eea6e893b5e01e9b19c0868b2807a1b9f901d251feca47bf32abL193-R197)

**CLI usability improvements:**

[...]

PR #4341, opened 2026-09-15
client: honor connect --client-ip for ip-bound access passes

### LINKED ISSUE #4333: client: restore `--client-ip` on `connect` for IP-bound access passes
## Summary

`doublezero connect ibrl --client-ip <ip>` is accepted but silently discarded. On a host
whose auto-discovered IP differs from the IP its AccessPass was issued for, the connect
fails and the only remedy is editing the packaged `doublezerod` systemd unit and
restarting the daemon as root.

Proposal, in two coupled parts:

1. Honor `--client-ip` again, but only when an AccessPass already exists at the exact
   `(client_ip, user_payer)` PDA — a pass whose IP the issuing authority pinned. Passes
   that are not IP-bound (wildcard `0.0.0.0` PDA, or `ALLOW_MULTIPLE_IP`) must keep
   rejecting the flag: for those the program accepts any globally-routable address, and a
   client-chosen IP is exactly the squatting hole RFC-27 exists to close.
2. Waive the RFC-27 proof *requirement* for such a pass, so a host that cannot originate
   from the pinned IP can still connect once `require-ip-ownership-proof` is set. This is
   an onchain change and an RFC-27 amendment — see the proposal section for the
   trust-boundary trade-off it carries.

## Current behavior

The flag parses (it is `global = true` on `Connect`, and the `ibrl` subcommand declares no
`--client-ip` of its own), prints a deprecation warning, and is then never read again:

```
⚠️  WARNING: --client-ip on the CLI is deprecated and will be ignored.
    Set --client-ip on the daemon (doublezerod) instead.
```

The only reference to `Connect::client_ip` in the file is the `is_some()` check that emits
that warning (`crates/doublezero-daemon-cli/src/connect.rs:502-508`). The field is typed
`Option<String>` and is never parsed into an `Ipv4Addr`, so a malformed value is accepted
just as silently. The address actually used comes from `resolve_client_ip(daemon)`
(`connect.rs:512` → `helpers.rs:32`), which reads the daemon's `/v2/status`.

## Reproduction

Host with an AccessPass issued for `B`, whose daemon discovers `A` (multi-homed, or an
egress path where `ifconfig.me` reports a different address than the one the pass names):

```
$ doublezero connect ibrl --client-ip B
⚠️  WARNING: --client-ip on the CLI is deprecated and will be ignored. ...
❌  Unable to find a valid AccessPass for the IP: A UserPayer: <payer>
Error: A valid AccessPass is required to connect. Please contact support to obtain one.
```

Today's only workaround is `doublezerod -client-ip B`. The shipped unit
(`client/doublezerod/cmd/doublezerod/doublezerod.service`) carries no such flag, so this
means a systemd drop-in plus a daemon restart — root, and disruptive to any other tunnel
on the host.

## Why it was removed

Not an oversight. #3034 moved provisioning out of the CLI into the daemon's onchain
reconciler; the CLI became "enable the reconciler and wait". Once the daemon owned
provisioning, a CLI-supplied IP had no way to reach the code that uses it, so the flag was
reduced to a warning. #4010 moved the verb into `doublezero-daemon-cli` unchanged.

## What a fix has to cover

The AccessPass gate alone is not sufficient — three layers each hold the IP independently:

1. **Authorization (the gate itself).** Query the exact-IP PDA directly, *not*
   `GetAccessPassCommand`: that helper prefers the wildcard `0.0.0.0` pass and only falls
   back to the exact-IP one (`smartcontract/sdk/rs/src/commands/accesspass/get.rs:28-34`),
   so a payer holding both would be resolved to the dynamic pass and wrongly admitted.
   Onchain the corresponding check is `create_core.rs:229-231`.

2. **Data plane — this is the blocker.** The reconciler filters onchain users by
   `userIP.Equal(n.clientIP)` (`client/doublezerod/internal/manager/manager.go:508-509`)
   and, for plain IBRL, uses `n.clientIP` verbatim as the tunnel source
   (`manager.go:677`). A CLI-only override would create an onchain User at `B` that the
   daemon — still on `A` — never matches, never provisions, and never tears down: an
   orphaned account and a connect that hangs rather than fails. The flag therefore has to
   reach the daemon (new field on the enable/provision request, or a `/v2` endpoint that
   sets the reconciler's client IP for this session), not just the ledger call.

3. **RFC-27 proof.** The verifier signs only the source address it observes
   (`crates/doublezero-ip-verifier/src/server.rs:280`); `connect` binds the request to the
   daemon's client IP where it can (`ip_proof.rs:174-175,213`) and hard-fails on a
   mismatch (`connect.rs:417-441`). Two sub-cases:

   - Host *can* originate from `B`: make `B` the binding source and the proof matches.
     Nothing else needed.
   - Host *cannot* originate from `B` (asymmetric routing, verifier unreachable from that
     source, NAT): no proof for `B` is obtainable, and once `require-ip-ownership-proof`
     is set the creation is rejected onchain. This is what the proposal below addresses.

## Proposal: waive the proof requirement for an authority-pinned IP

An IP-bound AccessPass is already an attestation. `SetAccessPass` is permissioned
(`processors/accesspass/set.rs:155-163`): the caller must pass the `ACCESS_PASS_ADMIN`
check (foundation allowlist, sentinel authority, feed authority, or a Permission account)
or be an administrator of the tenant being added. A registrant cannot self-issue a pass, so
a pass stored at `(B, payer)` means a privileged party asserted that `payer` may use `B`.
RFC-27 scopes its threat to wildcard passes on exactly this reasoning (RFC lines 21-22, 48),
and `payer_is_sentinel` is existing precedent for a privileged issuer vouching in place of a
wire proof.

The change is **onchain**, in `validate_ip_ownership_proof`
(`smartcontract/programs/doublezero-serviceability/src/ip_proof.rs:97-112`) — the branch
that returns `IpOwnershipProofRequired` for a missing proof. A client-side-only relaxation
is not sufficient: with the feature flag set, the program rejects the proofless creation
regardless of what the CLI decides. A supplied proof must still be validated in full either
way. Because this narrows RFC-27's enforcement surface, it needs an RFC amendment, not just
a code change.

### Trust-boundary caveat: the tenant-admin issuance path

`is_tenant_admin` lets a tenant administrator issue passes. Under a blanket exemption, a
tenant admin could issue an IP-bound pass naming a third party's IP to their own payer, then
create a User at that IP with no proof — squatting the `(client_ip, user_type)` User PDA and
pointing device tunnel provisioning at that third party. That is the RFC-27 attack reached
through a different door. The exemption moves the trust boundary from "the verifier observed
it" to "every ACCESS_PASS_ADMIN and every tenant administrator names IPs correctly."

### Options

1. **Exempt any IP-bound pass.** Simplest; widest trust boundary; admits the tenant-admin
   path above.
2. **Exempt only passes issued under `ACCESS_PASS_ADMIN`.** Closes the tenant-admin path,
   but `AccessPass` does not record its issuer today — needs a new field and a migration.
3. **Explicit per-pass waiver flag** (e.g. `WAIVES_IP_PROOF`, alongside `ALLOW_MULTIPLE_IP`
   in `state/accesspass.rs:171`). The waiver becomes a deliberate, auditable act by the
   issuer rather than an implicit consequence of pass shape, and strict-by-default is
   preserved. Recommended.

## Open questions

- Which of the three waiver options above? If 3, who is allowed to set the flag — the same
  `ACCESS_PASS_ADMIN` set, or a narrower one?
- Should the waiver also suppress the client-side *mismatch* abort (`connect.rs:417-441`),
  or only the missing-proof case? A mismatch still means the verification request and the
  tunnel leave by different paths, which is worth surfacing even when a proof is not
  required.
- Should `connect` still attempt to obtain a proof when a waiver applies, attaching it
  opportunistically? Costs one request, and keeps the stronger evidence on-chain where the
  host can produce it.
- Should the override be per-invocation, or persisted by the daemon across restarts like
  the enable/disable state?
- Interim, independent of the above: when `--client-ip` disagrees with the daemon's
  address, is a hard error better than the current warn-and-ignore? Today the operator is
  told the value was ignored and then sees an unrelated AccessPass error naming a
  different IP.

## Affected code

- `crates/doublezero-daemon-cli/src/connect.rs:114-116, 502-512` — flag declaration and warning
- `crates/doublezero-daemon-cli/src/helpers.rs:32` — `resolve_client_ip`
- `client/doublezerod/internal/manager/manager.go:508-509, 677` — reconciler match and tunnel src
- `client/doublezerod/internal/runtime/clientip.go:82` — daemon-side discovery
- `smartcontract/sdk/rs/src/commands/accesspass/get.rs:28-34` — wildcard-first resolution
- `smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs:229-231` — onchain IP check
- `smartcontract/programs/doublezero-serviceability/src/ip_proof.rs:97-112` — missing-proof enforcement branch
- `smartcontract/programs/doublezero-serviceability/src/processors/accesspass/set.rs:155-163` — who may issue a pass
- `smartcontract/programs/doublezero-serviceability/src/state/accesspass.rs:171` — pass flags
- `rfcs/rfc27-ip-verification.md:21-22,48` — RFC scoping the threat to wildcard passes

juan-malbeclabs:
Resolves: #4333

Two halves, both of #4333. The client half honors `connect --client-ip`; the onchain half waives the RFC-27 proof *requirement* for a pass whose address an authority pinned. They ship together because either alone leaves the flag unusable for the host that asked for it: without the waiver a pinned host still fails when a verifier is configured, and without the client half there is no way to ask for the pinned address.

## Summary of Changes

- `connect --client-ip` is honored again instead of being parsed, warned about and discarded. It is accepted only when the payer holds an AccessPass at the exact `(client_ip, user_payer)` PDA and that pass is not flagged `allow_multiple_ip`: a pass that authorizes any address does not authorize a caller-chosen one, which is the squatting hole RFC-27 exists to close. This is the same predicate `create_user` applies, so a refusal arrives up front rather than as a late `IpOwnershipProofRequired`, and an absent pass and a pass flagged `allow_multiple_ip` are refused separately, since what the operator has to do about each differs.
- **Onchain:** with `RequireIpOwnershipProof` set, a creation carrying no proof is now accepted when the AccessPass is bound to the address being claimed — stored at that `client_ip` and not flagged `allow_multiple_ip`. Such a pass is itself an attestation: `SetAccessPass` is permissioned, so a registrant cannot self-issue one. The waiver covers *absence* only; a supplied proof is still validated in full. `validate_ip_ownership_proof` moves below the access-pass checks in `create_user_core`, which is what establishes that the pass names this address and this owner before the waiver can rely on it. RFC-27 is amended to match.
- New `GetExactAccessPassCommand` in the Rust SDK does the CLI's lookup. `GetAccessPassCommand` resolves the wildcard pass first and falls back to the exact one, so a payer holding both would have been wrongly admitted.
- The pinned address reaches the reconciler, which is what made the flag a no-op before: it matches onchain users on the client IP and, for plain IBRL, uses it verbatim as the GRE tunnel source. `POST /enable` now takes an optional `client_ip` and reconciles even when already enabled, so the previous address's services are torn down and the new one's provisioned.
- The CLI verifies the address is globally routable and assigned to an interface that is up and running on this host, and the daemon repeats both checks rather than trusting the caller. Possession is not ownership — RFC-27 proofs remain the answer to that — but it stops a caller naming a third party's address, and an address the kernel does not hold cannot source a tunnel. `create_user` is not the backstop for either: it rejects the onchain *user*, while what `/enable` writes is the daemon's *pin*, and a pin no user can exist at matches nothing, tears the host's services down and survives every restart. The CLI check is best-effort (an enumeration failure defers to the daemon); the daemon's is fatal, since nothing downstream is left to catch it.
- The pin is persisted with the reconciler-enabled state, so a restart does not silently revert the host to its discovered address and tear the tunnel down. Every state write persists the pin *in effect* rather than the request being served, so the running daemon and the state file cannot disagree: an enable without the flag leaves the pin alone, and so does a disable. The pin is host configuration rather than session state — only a new pin replaces it. Precedence at startup is the daemon's own `-client-ip` flag, then a persisted pin, then discovery — and because a pin is adopted immediately while the flag only wins at the next restart, `/enable` refuses a pin that differs from the flag, naming it, rather than accepting one the restart would silently overrule. A pin equal to the flag's address is not a conflict. A pin that predates the flag stays on disk, dormant and logged, and is restored — re-validated — if the flag is removed from the unit file.
- A restored pin is re-checked against both of `/enable`'s conditions — globally routable, and held by an interface that is up — before it is used. The check `/enable` made says nothing about now — an address can leave with a DHCP lease, a NIC swap or a re-addressing while the daemon is down — and a host pinned to an address the kernel no longer holds would match no onchain user and build no tunnel, with no un-pin path short of editing the state file. A pin that no longer holds falls back to discovery with a warning and stays on disk, so it applies again once the address returns.
- A daemon that refuses the pin fails the connect loudly. An ordinary enable is a no-op when the reconciler is already on, but a pin is not, so a refusal is no longer written off as a redundant call — it would otherwise leave the operator waiting on a tunnel that could never be built for a user already created at that address. The bare path sends the pin with its up-front enable, so a refusal stops the run before anything is created onchain.
- The flag now parses as `Ipv4Addr` rather than `String`, so a malformed value is rejected instead of silently accepted.

## Security decision this PR makes

The waiver moves the trust boundary from "the verification service observed this address" to "every `ACCESS_PASS_ADMIN` **and every tenant administrator** names addresses correctly".

`set.rs` authorizes `SetAccessPass` on `is_privileged || is_tenant_admin`, where `is_tenant_admin` is just `tenant.administrators.contains(payer)`, and nothing constrains `value.client_ip`. So a tenant administrator can issue a pass at `(someone_else's_ip, their_own_payer)` and then create a user at that address with no proof — a path RFC-27 previously closed for them.

This is accepted deliberately, as the price of making pinned passes usable, and it is called out here rather than left in a code comment because it wants explicit sign-off. A per-pass waiver flag set by a narrower authority is the lever if it needs tightening; see Follow-up.

## Diff Breakdown

| Core logic   |    13 | +845 / -162  |  +683 |
| Tests        |     8 | +1639 / -98  | +1541 |
| Scaffolding  |     4 | +45 / -2     |   +43 |
| Docs         |     2 | +57 / -11    |   +46 |
| Config/build |     2 | +4 / -0      |    +4 |
| Generated    |     1 | +1 / -0      |    +1 |
| **Total**    |    26 | +2591 / -273 | +2318 |

Test-heavy: roughly two thirds of the diff is tests. Core logic is +845 new lines, above the ~500-line guideline; the onchain half (`ip_proof.rs`, `create_core.rs`, the RFC amendment and `user_ip_proof_test.rs`, ~+370 including its tests) is the clean split if a reviewer would rather take it separately. The Rust files appear in both Core logic and Tests, since their tests are inline; the totals row counts each file once.

<details>

- [`smartcontract/programs/doublezero-serviceability/src/ip_proof.rs`](https://github.com/malbeclabs/doublezero/pull/4341/files#diff-8312444b55ec335fc6388307318b5dbbef6834a8ca3c57df6ea807fe4b60ecda) — `accesspass_is_ip_bound` as a second exemption alongside `payer_is_sentinel`, and why it exists
- [`smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs`](https://github.com/malbeclabs/doublezero/pull/4341/files#diff-b825218e6fdd5871d82b5623d95b74b2460b480dd4c0e3f2cb150b98623ff342) — the proof check moved below the access-pass checks, and the predicate derived from the pass
- [`crates/doublezero-daemon-cli/src/connect.rs`](https://github.com/malbeclabs/doublezero/pull/4341/files#diff-59e335d4527da129d614de48572856bacb13b173e1248afe4313f32ceca553e2) — `resolve_connect_client_ip`: the local-possession and exact-pass gates, the flag flowing through to `enable`, and the refused-pin handling in `user_activated`
- [`client/doublezerod/internal/manager/manager.go`](https://github.com/malbeclabs/doublezero/pull/4341/files#diff-27424874932194e8d91a18571b7c6a3bf1f1bfa057986e442db5324e6fced881) — `clientIP` and the pin behind an `RWMutex`, `reconcilerCmd` on the enable channel, pin delivered even when already enabled, tunnel-src cache dropped on change
- [`crates/doublezero-daemon-cli/src/clientip.rs`](https://github.com/malbeclabs/doublezero/pull/4341/files#diff-eedfbc3f5bb6102d04d659a2a4f4aa063bd2195c43df14d26371430ebd1e2d86) — new: validates a caller-supplied address against `is_global` and this host's up+running interfaces, with a named rejection per case
- [`client/doublezerod/internal/manager/http.go`](https://github.com/malbeclabs/doublezero/pull/4341/files#diff-5db980b5d3ad21e383edbd8d3ecb5a1f9c9db61590b1cf0f2364f7ea4df60f80) — `/enable` accepts an optional `client_ip`, validated and re-checked before anything is written; both handlers persist the pin in effect
- [`client/doublezerod/internal/runtime/run.go`](https://github.com/malbeclabs/doublezero/pull/4341/files#diff-015946533bc566ab31d4a5fc0c2d1640b01589a508bda3c6fe5fbedcda6586ff) — startup precedence, and `restoreClientIPPin` re-validating a pin loaded from disk
- [`client/doublezerod/internal/manager/state.go`](https://github.com/malbeclabs/doublezero/pull/4341/files#diff-f6c2b2a1081502a978d5d7b0186f8e7f7958f3876207a2b5ce3825f961422eda) — `State.ClientIP` persisted; `LoadOrMigrateState`/`WriteState` carry a `State` instead of a bool
- [`smartcontract/sdk/rs/src/commands/accesspass/get.rs`](https://github.com/malbeclabs/doublezero/pull/4341/files#diff-28bb8c01da17ef8520582393dc0efdb7f9f4bb5fc87696749a6c275852adf8f9) — `GetExactAccessPassCommand`, which never falls back to the dynamic pass
- [`rfcs/rfc27-ip-verification.md`](https://github.com/malbeclabs/doublezero/blob/main/rfcs/rfc27-ip-verification.md) — the amendment: the exemption, what it does and does not cover, and the trust boundary it moves

</details>

## Testing Verification

- Onchain, the waiver: a creation with no proof is accepted on a pass pinned to exactly this address, with the flag on — on the IBRL path and on `create_subscribe_user`, since both share `create_user_core`. It stays rejected for a pass at the wildcard PDA and for a pinned pass flagged `allow_multiple_ip`, on first creation and on the idempotent rerun path alike.
- Onchain, the waiver's boundary: a pass pinned to this address does not excuse a proof that *fails* validation — one naming a different address, payer, user type or epoch is still rejected. The waiver covers absence only.
- Rust: the flag is honored for an exactly pinned pass and refused when only a dynamic pass exists (the case `get_accesspass` would have admitted), refused for an exact pass flagged `allow_multiple_ip` with its own message rather than the absent-pass one (the case that would otherwise fail late onchain), refused for an address this host does not hold and for a non-global one, and ignored entirely when absent — the ledger is never consulted about a pin in that case.
- Rust, end-to-end through `connect`: three combinations that must all create a user — dynamic pass with a discovered and certified address, pinned pass with a discovered address and no proof, and pinned pass with `--client-ip` and no proof. The last asserts the pinned address is what reaches provisioning, and that every downstream pass lookup is keyed on it rather than the discovered one.
- Rust, refused pins: a refusal is not excused by an already-enabled reconciler, *is* excused when the daemon already reports that exact address, and leaves the unpinned lenient path untouched. A bare `connect --client-ip` against a daemon that refuses the pin stops with no `create_user` call at all.
- Go: `/enable` pins and reports the address on `/v2/status` immediately; a body-less request leaves both the address in use and the persisted pin alone; a malformed or non-local address changes nothing (no state write, no pin); an interface-enumeration failure fails the request; a pin on an already-enabled daemon is still delivered; the daemon and the state file agree at every step of pin → disconnect → reconnect-without-the-flag; `State.ClientIP` round-trips through the state file.
- Go, the `-client-ip` flag: a pin that differs from the flag is refused with the flag named and nothing written; a pin equal to it is adopted and persisted; a body-less enable is unaffected by the flag.
- Go, restart: a restored pin the host still holds is used; one it no longer holds, one that is not globally routable even though the host holds it, and one that is unparseable or IPv6, all fall back to discovery; an enumeration failure keeps the pin rather than dropping it, since it is a failure to check rather than a failed check.
- Rust, the SDK's exact lookup: a missing account, an account owned by another program and a transport failure are told apart — the first two are absence, the third propagates — with no error text matched.
- E2E not run locally. Several e2e tests already pass `--client-ip` to `connect` (`e2e/device_maxusers_rollover_test.go`, `e2e/multi_client_ibrl_allocated_ip_test.go`), where it was previously ignored and now takes effect. They pin the AccessPass to the exact CYOA address and that address is assigned to the client container, so both gates should pass, but this wants a CI run to confirm.

## Follow-up

- **A narrower waiver.** The exemption is available to any issuer `SetAccessPass` authorizes, tenant administrators included (see Security decision above). Recording at issuance whether the issuer held `ACCESS_PASS_ADMIN` — a bit in the existing `flags` byte, so no layout change — and waiving only for those passes is the contained way to tighten it. Deliberately not in this PR: it would make tenant-issued pinned passes require a proof, which is a product call about how tenants provision rather than a correctness fix.
- **No explicit un-pin.** Returning a host to discovery means pinning a different address, or clearing `client_ip` from the state file and restarting; an `--client-ip=`-style clear would need the daemon to re-run discovery on demand, which is worth doing separately if operators ask for it.
- **`--client-ip` is refused while the payer holds a dynamic pass.** User creation attaches the dynamic pass, so the flag cannot be honored against the pinned one without changing which pass is charged for the seat — a larger change than this flag should make. The gate says so up front instead of promising something the transaction does not deliver.
- **The client-side mismatch abort is unchanged.** When a verifier is configured and the address it observes disagrees with the one being claimed, `connect` still treats that as fatal, before the onchain waiver is ever reached. A host that needs the waiver gets there through the address the daemon is configured with or discovers.


client/doublezerod/internal/manager/http.go
```diff
@@ -145,22 +147,70 @@ func (n *NetlinkManager) ServeStatus(w http.ResponseWriter, r *http.Request) {
...
+	// Persist the pin as given: an absent client_ip clears a previous one, so an operator
+	// returns to auto-discovery by connecting again without the flag.
+	if err := WriteState(n.stateDir, State{ReconcilerEnabled: true, ClientIP: req.ClientIP}); err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
 		json.NewEncoder(w).Encode(map[string]string{"status": "error", "description": err.Error()}) //nolint:errcheck
 		return
 	}
-	n.SetEnabled(true)
-	w.Header().Set("Content-Type", "application/json")
+	n.SetReconcilerState(true, clientIP)
```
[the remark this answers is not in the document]
juan-malbeclabs: Confirmed and fixed in b27b360a5.

The handler was persisting the request rather than the pin in effect, so a body-less `/enable` — an ordinary `connect` with no flag, or `doublezero enable` — blanked `client_ip` while `SetReconcilerState(true, nil)` deliberately left the running daemon on the pinned address. As you say, the split only surfaces at the next restart.

The manager now tracks `pinnedClientIP` alongside the address in use (they legitimately differ: the daemon's own `-client-ip` flag outranks a restored pin, leaving one recorded but dormant), and every state write persists that instead of `req.ClientIP`.

One knock-on: `/disable` had the same bug in the other direction. It cleared the pin from disk while the live daemon kept using it, and there is no discovered address to fall back to without probing for one again — so the next body-less `/enable` would have re-created the same divergence. The pin now outlives a disable, which is also the more defensible semantics: it describes which address this host presents to DoubleZero, not anything about one session. `TestServeDisable_ClearsPin` became `TestServeDisable_KeepsPin` to match, and `state.go` documents that only a new pin replaces it.

The third location you flagged, `enable.rs:25`, is fixed too, though it needed the opposite change: `ensure_reconciler_enabled` hardcoded `enable(None)` and returned early when the reconciler was already on, so the bare `connect` path sent an unpinned enable first and only pinned later, from `user_activated`. It now takes the pin and skips the early return when one is present.

New coverage: `TestServeEnable_NoBodyKeepsPersistedPin`, `TestServeEnable_NoBodyKeepsRestoredPin`, `TestServeEnable_PinSurvivesDisableEnableCycle` (asserts the daemon and the state file agree at every step of pin → disconnect → reconnect), `TestServeDisable_KeepsPin`, and two in `enable.rs` for pin delivery to an already-enabled daemon.

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -2231,7 +2287,7 @@ impl Connect {
         spinner.inc(1);
 
         // Enable the reconciler (no-op if already enabled).
-        if let Err(e) = daemon.enable().await {
+        if let Err(e) = daemon.enable(self.client_ip).await {
```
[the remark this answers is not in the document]
juan-malbeclabs: Confirmed and fixed in b27b360a5.

The `already_enabled` escape hatch is right for an unpinned enable — that call only has to flip a flag, and an already-on reconciler means it was flipped — but it is wrong for a pin, where adopting the address *is* the request. A rejected pin (not locally assigned, enumeration failure, state write error) was read as a redundant enable, and the connect then waited out `poll_for_daemon_provisioned` on a tunnel the daemon was never going to build, for a user that already existed onchain at that address.

`user_activated` now branches on whether a pin was requested. With one, only a daemon whose `/v2/status` reports that exact address makes the failure moot; otherwise it bails with the daemon's own reason attached, so the leg is reported as failed rather than provisioned. Without a pin the lenient path is unchanged.

The bare `connect` path also sends the pin with its up-front `ensure_reconciler_enabled`, and a refusal there is fatal — nothing is created onchain for an address the daemon just declined. That is the better place to fail: the previous ordering created the user first and only discovered the problem afterwards.

Three new tests in `connect.rs`: `a_refused_pin_is_not_excused_by_an_enabled_reconciler`, `a_refused_pin_is_excused_when_the_address_is_already_adopted` (the one case where it genuinely is moot), `an_unpinned_failure_still_carries_on_when_already_enabled` (guards the lenient path), plus `test_connect_bare_with_client_ip_stops_when_the_daemon_refuses_the_pin`, which asserts the run stops with no `create_user` call at all.

elitegreg (commented):
Reviewed the full diff. Code quality is high — the comments explain *why* rather than
*what*, the concurrency work in `manager.go` is disciplined, and onchain test coverage is
strong. My concerns are mostly about scope and disclosure rather than craft.

### 1. The description says the onchain waiver is not in this PR, but it is (High)

The Follow-up section states Part 2 of #4333 "is not in this PR. It is an onchain change
in `validate_ip_ownership_proof` plus an amendment to RFC-27, and the issue's open
question ... is undecided."

That change *is* here: `ip_proof.rs`, `create_core.rs`, `user_ip_proof_test.rs` (+167)
and a 34-line RFC-27 amendment. Neither Summary of Changes nor Testing Verification
mentions it. The highest-risk change in the PR is described as absent in the text a
reviewer reads first. It also makes the Diff Breakdown table stale (19 files / +1588 vs.
an actual 24 / +1845).

Could you update the description, or split the onchain half out?

### 2. The waiver is reachable by any tenant administrator (High)

The logic itself is sound — I checked both ways it could break and neither does.
`accesspass_is_ip_bound` is only reachable with `accesspass.client_ip == client_ip`,
because the guard above rejects a mismatch, and the `client_ip` field cannot diverge from
the PDA seed (`set.rs:116` derives from `value.client_ip`, `:192`/`:253` store it).
Moving the `validate_ip_ownership_proof` call below the access-pass checks is what makes
that hold, and the tests cover `allow_multiple_ip` and wildcard-PDA passes still
requiring a proof, on both first creation and rerun.

The issue is the trust boundary. In `set.rs:142-163` a caller is authorized if
`is_privileged` **or** `is_tenant_admin` (just `tenant.administrators.contains(payer)`),
and nothing constrains `value.client_ip`. So any tenant administrator can issue a pass at
`(someone_elses_ip, their_own_payer)` and then create a user at that address with no
proof — the squatting RFC-27 exists to close.

The doc comment and RFC amendment both acknowledge this and accept it, pointing at a
per-pass waiver flag as the future lever. Given the PR body says the waiver option is
still an open question, this reads like a design decision that got settled in-flight.
Worth an explicit call rather than a code comment — and the narrower `ACCESS_PASS_ADMIN`-only
waiver flag may be worth doing now rather than later.

### 3. A restored pin is never re-validated (Medium)

`/enable` gates a pin on IPv4 parse plus `isLocallyAssigned`. The restart path applies
neither: `run.go` passes the persisted pin into `DiscoverClientIP`, whose explicit branch
(`runtime/clientip.go:84-96`) only parses, with no local-assignment check and no fallback
to discovery.

So a host whose pinned address goes away (DHCP change, NIC swap, re-addressing) comes back
up pinned to an address the kernel does not hold, matches no onchain user, and builds no
tunnel. With no un-pin path, recovery is hand-editing the state file and restarting.
Re-running `IsLocallyAssigned` on the restored pin and falling back to discovery with a
warning would close it.

### 4. The waiver's justification contradicts the client gate (Medium)

The RFC amendment justifies the waiver because an attested address "is not always one the
host can originate from — under asymmetric routing, behind NAT". But the client gate this
PR adds requires the address be assigned to a local up+running interface. Behind NAT the
host does not hold its public address, so `connect --client-ip` rejects exactly the hosts
the onchain waiver exists to serve. The two halves address disjoint sets of hosts; the
amendment should probably say so.

### 5. The CLI and program gates test different predicates (Low)

The CLI checks "an exact-PDA pass exists"; the program checks
"exact PDA **and** `!allow_multiple_ip()`". A pass at `PDA(X, payer)` with
`allow_multiple_ip` set passes the CLI gate, so `connect --client-ip X` proceeds, but
onchain `accesspass_is_ip_bound` is false and a proof is still required. For the flag's
target case (no obtainable proof) that fails late at `create_user` with
`IpOwnershipProofRequired` instead of early with the clear "No AccessPass is pinned to X"
message. Fails safe, so UX rather than a hole, but the predicates should match.

### 6. The daemon does not check globally-routable (Low)

The description says "Both sides verify the address is globally routable and assigned to
an interface that is up and running." `IsLocallyAssigned` is possession-only and there is
no `is_global` equivalent in `http.go`. The CLI and the program both catch it, so this is
just a description inaccuracy.

### Things I specifically liked

- `GetExactAccessPassCommand` — the pre-existing wildcard-first resolution in
  `GetAccessPassCommand` would have admitted a payer holding both passes on the strength
  of the dynamic one. Returning `Ok(None)` for an `UNSPECIFIED` query rather than
  resolving it is a good call.
- Adopting the address synchronously in `SetReconcilerState` and sending only
  `clientIPChanged` over the channel, so `tunnelSrcCache` stays owned by the reconciler
  goroutine while `/v2/status` can report the pin the moment `/enable` returns.
- Snapshotting `clientIP := n.ClientIP()` once per `reconcile()` pass, so a concurrent
  pin cannot match a user against one address and provision against another.
- "Persist the pin *in effect*, not the request being served" — the subtlest invariant
  here, and correct. A body-less `/enable` blanking a live pin would only surface at the
  next reboot.

juan-malbeclabs:
Thanks — all six are addressed. Point by point:

**1. Description claimed the onchain waiver was absent (High).** Correct, and the worst of the six. The description is rewritten: the waiver is in the opening paragraph, the Summary, Testing Verification and Key files, and the Follow-up section no longer claims it is separate work. The table is recomputed from the real diff (25 files, +1988/−219). I kept the two halves together — either alone leaves the flag unusable for the host that asked for it — but the size note now names the onchain half as the clean split if you would rather take it separately. The CHANGELOG carried the same stale sentence and is fixed too.

**2. Waiver reachable by any tenant administrator (High).** Accepted, and now an explicit call rather than a code comment: the description has a **"Security decision this PR makes"** section naming `set.rs`'s `is_privileged || is_tenant_admin` authorization, the unconstrained `value.client_ip`, and the resulting `(someone_else's_ip, their_own_payer)` path. It wants a conscious sign-off, not a silent merge.

The narrower lever is recorded in Follow-up with the shape it would take: stamp a bit in the existing `flags` byte at issuance when the issuer held `ACCESS_PASS_ADMIN`, and waive only for stamped passes — no account layout change. Deliberately not in this PR, because it would make tenant-issued pinned passes require a proof, which is a product call about how tenants provision rather than a correctness fix. Happy to land it first if you would rather not merge the wider boundary at all.

**3. Restored pin never re-validated (Medium).** Fixed — `restoreClientIPPin` in `run.go` repeats the local-assignment check at startup. Two choices worth flagging, both slightly different from your suggestion:

- A rejected pin is **not** erased. It stays in the state file, so it applies again once the address returns and an interface that is merely late coming up does not cost the operator their configuration.
- An enumeration *failure* keeps the pin rather than dropping it: that is a failure to check, not a failed check. Only a definite "this host does not hold it" falls back to discovery, with a warning.

Worth noting this was inherited rather than introduced — the daemon's own `-client-ip` flag has never been locally validated either — but the pin is machine-written, so it is the one that deserves the fallback.

**4. Justification contradicts the client gate (Medium).** Right, and the NAT example was weaker than either of us said: behind plain NAT the service observes the same address discovery finds, so a proof is obtainable. The amendment is rewritten around the cases that actually hold — a host reaching the service from an address other than the attested one, or unable to reach it from that source at all — and now states the thing that resolves the contradiction: the waiver is reached through the address the daemon is configured with or discovers, never through `connect --client-ip`, which independently requires local possession. The same wording in `ip_proof.rs` and in the test doc comment is fixed.

**5. CLI and program test different predicates (Low).** Fixed — the gate filters on `!accesspass.allow_multiple_ip()`, matching `accesspass_is_ip_bound`, and the error names the flag. New test `refuses_an_exact_pass_flagged_allow_multiple_ip`.

**6. Daemon does not check globally-routable (Low).** Fixed as the description inaccuracy it was, rather than by adding the check: globality is a property of the address rather than of the host, so it cannot change between the CLI's check and the daemon's, and `User::validate` enforces it onchain for every path in. The `/enable` comment now says which of the CLI's two checks it repeats and why it does not repeat the other.

**One thing outside the review.** The merge of main at `63ee88c17` resolved `CHANGELOG.md` by keeping this branch's `Unreleased` section wholesale, which dropped four entries main had added — the feed-lifecycle SDK commands and three RFC-28 serviceability bullets. Main has since cut v0.41.0 and moved them there, so a later merge would have conflicted or lost them again. Rebuilt from main's file with this branch's entries under `Unreleased`; the diff against main is now purely additive.

juan-malbeclabs:
A second self-review pass turned up six more, five of which I'd rank above some of the first round. All fixed and pushed; each has a regression test that fails without the fix.

**1. The waiver was reachable by a legacy wildcard pass.** `accesspass_is_ip_bound` read the stored `accesspass.client_ip` field, which is not on its own an attestation. Between #1608 (`3b13effb5`) and #3859 (`2550270fa`), `create_user` wrote the first address a dynamic pass connected from into that field and serialized it back, and #3859 removed the lock-in without migrating the accounts. Such a pass still sits at the **UNSPECIFIED PDA** while its field names an address its own holder picked — so the waiver accepted an address no authority ever vouched for, reached through exactly the wildcard passes RFC-27 names as its threat. Latent until `RequireIpOwnershipProof` is set, which is the moment it matters.

The waiver is now keyed on the PDA seed, which is what an issuing authority chose. A legacy pass still authorizes the creation; it only stops waiving the proof. `test_missing_proof_is_rejected_for_a_legacy_self_bound_wildcard_pass` fabricates that account shape directly, since no instruction can produce it today, and fails without the fix.

Worth a check against live ledger state to see whether the population is non-empty — I could not verify that from here.

**2. The `--client-ip` gate did not predict what the transaction evaluates.** It asked `get_accesspass_exact`, but `CreateUserCommand` resolves the account with `GetAccessPassCommand`, which prefers the dynamic PDA whenever one exists. A payer holding both passes cleared the gate on the exact one and then had the dynamic one sent, failing with `IpOwnershipProofRequired` — the late failure the gate exists to prevent. It now refuses up front and says why. Letting the pin choose which pass is charged for the seat is a bigger change than this flag should make, so that is stated rather than done.

**3. The single-read invariant stopped halfway.** `reconcile` snapshots `clientIP` under a comment saying a concurrent pin must not split the pass between matching one user and provisioning against another — but `buildProvisionRequest` re-read `n.ClientIP()`, and for plain IBRL that value becomes the GRE tunnel source verbatim. Now passed down like `devicesByPK` and `allPrefixes` already are.

**4. `/enable` accepted a non-routable pin.** I argued last round that the daemon needn't repeat the CLI's `is_global` check because `create_user` enforces it onchain. That was wrong: the program rejects the onchain *user*, while what `/enable` writes is the daemon's *pin*, and a pin no user can exist at is the worst shape to hold — it matches nothing, tears down the host's services, persists, and survives every restart. `IsPublicIPv4` moved to the manager package so the discovered address and the pinned one share one definition.

**5. A restored pin the host had lost was still held as the pin in effect.** Falling back to discovery but keeping it in memory meant the next state write put it back, and a later restart could apply it once the address happened to return. That is a silent address migration, and it contradicts the invariant this PR claims — that the daemon and the state file cannot disagree. Dropped instead, with the warning saying to re-pin. @elitegreg, this is your original suggestion; I traded it away last round for boot-race tolerance and the trade was wrong.

**6. Two tests were vacuous, one newly so.** `test_sentinel_payer_may_create_without_a_proof_while_the_flag_is_set` provisioned a pinned pass, which the new waiver satisfies on its own — I confirmed it passed with the `payer_is_sentinel` branch deleted entirely. It uses an `allow_multiple_ip` pass now and fails without the exemption. And `accepts_addresses_this_host_actually_holds` filtered to globally routable addresses, so on any NAT'd host — every CI runner — the loop had no iterations.

Two diagnostics went with them: `GetExactAccessPassCommand` folded every error into `Ok(None)`, so an unreachable or wrong-cluster ledger read as "this payer holds no pass"; and the daemon client discarded the `/enable` response body, collapsing four distinct rejections into "400 Bad Request".

Still open and unfixed, flagged rather than silently carried: version skew (a daemon predating this PR 200s the pin and discards it, so a CLI upgrade without a daemon restart exits 0 with an orphan onchain user); `poll_for_daemon_provisioned` matching on `user_type` alone, so a stale service can satisfy the wait; the pin reaching the daemon before any onchain write only on the bare `connect` path; a read-modify-write window in `ServeEnable`; flag-versus-pin precedence differing between boot and runtime; and `IFF_RUNNING` rejecting addresses on a carrier-less bridge or bond. Happy to take any of them here or in a follow-up — say which.

client/doublezerod/internal/runtime/run.go
```diff
@@ -80,17 +115,31 @@ func Run(ctx context.Context, sockFile string, routeConfigPath string, enableLat
...
+	slog.Info("reconciler: loaded state", "enabled", state.ReconcilerEnabled, "client_ip", state.ClientIP)
+
+	// Precedence: the daemon's own -client-ip flag, then an address pinned by a previous
+	// `connect --client-ip`, then discovery. The flag comes from the unit file and is the
+	// operator's standing configuration for this host, so it outranks a pin left by a
+	// connection; without it, restoring the pin is what keeps a restart from reverting the
+	// host to its discovered address and tearing the tunnel down.
+	effectiveClientIP := clientIP
+	pinnedClientIP := state.ClientIP
+	if clientIP == "" && state.ClientIP != "" {
```
ben-malbeclabs, line 131: Important — the persisted pin is only consulted when `-client-ip` is unset, and the manager is never told the flag was set, so `/enable` accepts and persists a pin that the next restart overrides with the flag's address, leaving the host on an address no onchain user matches. Pass the flag's presence through to the manager and have `ServeEnable` refuse a differing pin, naming the flag.

CHANGELOG.md
```diff
@@ -8,6 +8,21 @@ All notable changes to this project will be documented in this file.
...
...gainst it is the address squatting RFC-27 exists to prevent. The flag is also refused while the payer holds a dynamic pass, because user creation attaches that one and the gate would otherwise promise something the transaction does not deliver. The address must also be globally routable and assigned to an interface that is up on this host — possession, not ownership, which is what the local configuration needs and all the CLI can check; RFC-27 proofs remain the answer to who holds an address on the public internet. The value now parses as an IPv4 address rather than a string, so a malformed one is refused instead of silently ignored. A daemon that refuses the pin fails the connect: on the bare path the refusal stops the run before anything is created onchain, and elsewhere it is reported rather than written off as a redundant enable, which would otherwise leave the operator waiting on a tunnel that could never be built for a user already created at that address. A pass flagged `allow_multiple_ip` is refused for the same reason a dynamic one is, and for the same predicate `create_user` applies, so the refusal arrives up front rather than as a late `IpOwnershipProofRequired`. (#4333)
```
ben-malbeclabs, line 12: Important — `--client-ip` goes from accepted-and-ignored to a hard failure, so an existing invocation by a payer holding a dynamic pass now aborts the connect. This belongs under the empty `### Breaking` heading above.

CHANGELOG.md
```diff
@@ -8,6 +8,21 @@ All notable changes to this project will be documented in this file.
...
...r_payer)` means an issuing authority asserted that this payer may use this address. The flag and the pass together left a host unable to connect at all when no proof was obtainable for its pinned address — the service signs the source it observes, so a host reaching it from some other address gets a proof it cannot use, and one that cannot reach it from that source gets none. The waiver covers *absence* only; a supplied proof is still validated in full, and a proof naming another address, payer, user type or epoch is rejected as before. Wildcard passes and passes flagged `allow_multiple_ip` still require one, including a legacy wildcard pass whose stored field names an address: neither has an authority's seed behind the address being claimed, which is the squatting RFC-27 exists to close. The trust boundary moves with it, onto issuance: `SetAccessPass` authorizes tenant administrators as well as `ACCESS_PASS_ADMIN` holders, so a tenant administrator can pin a third party's address to their own payer and create a user there with no proof. Accepted deliberately; a per-pass waiver flag set by a narrower authority is the lever if it needs tightening. RFC-27 is amended to match. (#4333)
```
ben-malbeclabs, line 18: Minor — these four entries run 200-400 words each and carry `(#4333)`. The convention here is a TL;DR with no issue or PR IDs.

client/doublezerod/internal/manager/manager.go
```diff
@@ -433,12 +515,25 @@ func (n *NetlinkManager) StartReconciler(ctx context.Context) error {
...
 			return nil
-		case enabled := <-n.enableCh:
-			n.enabled.Store(enabled)
-			if enabled {
+		case cmd := <-n.enableCh:
+			// A changed address invalidates the tunnel-src cache, whose entries were
+			// resolved for the old source. Cleared here because the cache belongs to
+			// this goroutine.
+			if cmd.clientIPChanged {
+				n.tunnelSrcCache = make(map[string]net.IP)
```
ben-malbeclabs, line 523: Minor — entries are `ResolveTunnelSrc(tunnelDst)`, a function of the routing table alone, and are read only for `IBRLWithAllocatedIP` and `Multicast`, where `clientIP` is not the source; a pin change neither invalidates them nor is described by the comment above. Drop the clear, or state what it actually protects.

client/doublezerod/internal/runtime/run.go
```diff
@@ -30,6 +30,41 @@ const (
...
+// A rejected pin is dropped rather than kept dormant. Holding an address the daemon is not using
+// is the worse failure: it would be reported as the pin in effect, written back by the next state
+// write, and applied again on some later restart once the address happened to return — a silent
+// address migration triggered by nothing the operator did, and a state file that disagrees with
+// the running daemon in the meantime. The warning says so, and re-pinning is one `connect
+// --client-ip` away.
+//
+// An enumeration failure keeps the pin: that is a failure to check, not a failed check.
+func restoreClientIPPin(pinned string, isAssigned func(net.IP) (bool, error)) string {
+	ip := net.ParseIP(pinned)
```
ben-malbeclabs, line 51: Minor — a restored pin is re-checked for parseability, IPv4-ness and local assignment but not `IsPublicIPv4`, which `/enable` applies, so a non-global pin is adopted and matches no onchain user: the lockout this function exists to prevent.

crates/doublezero-daemon-cli/src/connect.rs
```diff
@@ -172,6 +175,100 @@ enum FeedJoinUser {
...
+        return Err(eyre::eyre!("--client-ip {client_ip} cannot be used here"));
+    }
+
+    let pinned = ledger
+        .get_accesspass_exact(client_ip, ledger.get_payer())?
+        .filter(|accesspass| !accesspass.allow_multiple_ip());
+    if pinned.is_none() {
+        writeln!(
+            out,
+            "❌  No AccessPass is pinned to {client_ip} for UserPayer: {}",
```
ben-malbeclabs, line 232: Minor — an exact pass flagged `allow_multiple_ip` reaches this line through the `filter` above, and the message says no pass is pinned to the address when one is.

rfcs/rfc27-ip-verification.md
```diff
@@ -345,6 +350,26 @@ IP, and that `InitGlobalState` seeds `sentinel_authority_pk` to whoever initiali
...
+`SetAccessPass` is permissioned, so a privileged party asserted that this payer may use this
+address. The exemption exists because an attested address is not always one a proof can be
+obtained for: the verification service signs the source it observes, so a host that reaches it
+from some other address — multi-homed, asymmetrically routed, or egressing through a NAT other
+than the one the pass names — gets a proof for an address it cannot use, and a host that cannot
+reach the service from that source at all gets none. The pass and the flag together left such a
+host unable to connect. It waives the *requirement* only; a supplied proof is still validated in
+full.
+
+This exemption is reached through the address the client daemon is configured with or discovers,
```
ben-malbeclabs, line 363: Minor — nothing onchain enforces the local-interface check, so `doublezero user create --client-ip` and any SDK caller reach the waiver without it; this states a program property that only the daemon CLI provides.

smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs
```diff
@@ -238,6 +217,46 @@ pub fn create_user_core(
...
+    // names this address and this owner. Still before the already_exists return below, so a rerun
+    // is covered too.
+    //
+    // Bound to `effective_owner`, not the transaction payer: the owner-override path lets the
+    // sentinel or a USER_ADMIN holder create a user owned by somebody else, and it is that owner
+    // who must have demonstrated control of `client_ip` — which is also the identity the AccessPass
+    // is keyed on (`accesspass.user_payer` above). On the ordinary path the two are the same
+    // account.
+    //
+    // A pass at the UNSPECIFIED PDA authorizes any address, and `allow_multiple_ip` says the same
```
ben-malbeclabs, line 234: Minor — the `assert!` above restricts `accesspass_account` to `PDA(client_ip)` or `PDA(0.0.0.0)`, so a pass stored at a specific address can only ever be presented for that address and `allow_multiple_ip` cannot make it authorize another; keep the clause as an issuer opt-out, but the stated reason is wrong. Same claim in `ip_proof.rs`, `tests/user_ip_proof_test.rs:236` and `connect.rs:191`.

smartcontract/sdk/rs/src/commands/accesspass/get.rs
```diff
@@ -41,6 +41,55 @@ impl GetAccessPassCommand {
...
+            // `GetAccessPassCommand` can fold an error into `None` because a second lookup
+            // follows it; here the lookup *is* the answer, so an unreachable or misconfigured
+            // ledger would otherwise render as "this payer holds no pass" and send an operator
+            // off to have one reissued. Only a genuinely absent account is `None`. The RPC
+            // reports that as `AccountNotFound` and offers no typed form of it through this
+            // trait, so the string is what there is to match; misreading one as absent is the
+            // behaviour this replaces, and the transport errors worth retrying have already
+            // been retried by the client.
+            Err(err) => {
+                if format!("{err:#}").contains("AccountNotFound") {
```
ben-malbeclabs, line 81: Minor — absence is decided by matching upstream error text, so a message change turns the ordinary no-pinned-pass case into a hard error and breaks `connect --client-ip`. `DoubleZeroClient::get_multiple_accounts` already returns `Vec<Option<Account>>` for typed absence.

ben-malbeclabs (commented):
**Verdict: 2 important, 7 minor.** The two important ones want addressing before merge.

Reviewed at `8f8d28c`. CI is green on that commit including all six e2e shards, so the `connect --client-ip` invocations the description calls out as unverified do pass: the e2e CYOA subnet is `9.128.0.0/9`, which `is_global` accepts, and the passes are issued at the exact PDA with no dynamic pass alongside.

### Findings

| # | Sev | Location |
|---|-----|----------|
| 1 | Important | `runtime/run.go:131` — a runtime `/enable` pin loses to `-client-ip` on the next restart |
| 2 | Important | `CHANGELOG.md:12` — behaviour change is breaking, filed under Changes |
| 3 | Minor | `CHANGELOG.md:18` — entry length and PR IDs |
| 4 | Minor | `manager/manager.go:523` — tunnel-src cache clear has no reachable effect |
| 5 | Minor | `runtime/run.go:51` — restored pin skips the globality check |
| 6 | Minor | `connect.rs:232` — rejection names the wrong condition |
| 7 | Minor | `rfc27-ip-verification.md:363` — states a program property the program does not enforce |
| 8 | Minor | `create_core.rs:234` — `allow_multiple_ip` rationale is false against the assert above it |
| 9 | Minor | `accesspass/get.rs:81` — absence decided by matching error text |

### Rationale without a single anchor

The onchain waiver itself checks out. `accesspass_is_ip_bound` keys on `accesspass_account.key == &accesspass_pda`, which is seeded on the `client_ip` being created, so a legacy pass whose stored field its own holder chose cannot buy the waiver, and the test for that shape is present. Moving `validate_ip_ownership_proof` below the access-pass checks weakens no other path: every check that ran before still runs, only the error a bad caller sees changes.

The tenant-administrator escalation is accurately described. `add_administrator` is foundation-gated so administrators are appointed, but they hold no `ACCESS_PASS_ADMIN` and nothing constrains `value.client_ip` in `set.rs`. One property the description omits: the waiver is unconditional once `RequireIpOwnershipProof` is set, so no configuration enforces RFC-27 strictly any more. If the follow-up per-pass flag slips, that is what was lost.

Finding 1 is the one worth blocking on. The changelog states the previous workaround was `-client-ip` in the packaged unit, so the hosts carrying that flag are exactly the population this feature targets; on them the pin works until the next restart and then reverts silently. `/v2/status` reports the flag address while `PinnedClientIP()` reports the dormant pin, and `restoreClientIPPin` is skipped in that branch, so the dormant pin is never validated either.

elitegreg:
Re-reviewed at `8f8d28c`. Thanks for turning the earlier round around so thoroughly — the
rewritten description, the explicit security-decision section, and `restoreClientIPPin` all
address what I raised, and the second self-review pass caught real problems, the legacy
wildcard-pass bypass in particular.

I agree with Ben's review. Both of his important findings match what I see, and I traced
`run.go:131` independently: on a host whose unit file carries `-client-ip`, the whole
restore branch is skipped, so a runtime pin is neither used nor validated on restart, and
`/v2/status` and `PinnedClientIP()` then disagree about which address the host presents.
That is the population the CHANGELOG names as the one the feature is for. I'd also not
discount his RFC point (#7) just because it is filed as minor — the amendment leans on
`connect --client-ip`'s local-assignment check in a section that otherwise speaks only in
program MUSTs, and the program enforces nothing of the kind.

I'm satisfied with the onchain waiver as scoped. The predicate keying on the PDA seed rather
than the stored `client_ip` field is the right call, moving `validate_ip_ownership_proof`
below the access-pass checks is what makes it hold, and the absence-only boundary is well
covered by tests. Signing off on the wider trust boundary knowingly; the per-pass flag stays
the lever if we need it.

One nit nobody has mentioned: `nix` is inserted between `libc` and `log` in the workspace
`Cargo.toml`, out of alphabetical order with the rest of the list.

Leaving this as comments rather than a verdict — Ben's two important items are the ones
worth resolving before merge.

client/doublezerod/internal/runtime/run.go
juan-malbeclabs: Fixed in d6b5eafcf.

Agreed, and the direction you named is the one I took: the flag keeps precedence, and a pin that differs from it is refused up front rather than accepted and reverted at the next restart. `WithClientIPFlag` carries the flag's address into the manager, and `ServeEnable` answers a differing pin with a 409 naming the flag and the address it carries — the operator has to edit the unit file, and nothing else in the message would tell them that. The CLI surfaces it verbatim through `daemon_status_error`, so `connect --client-ip` on such a host now fails with the reason instead of succeeding and reverting.

Two details beyond the suggestion:

- A pin equal to the flag's address is not a conflict. It asks for what the host will do anyway, and refusing it would fail a correct `connect --client-ip`.
- A pin that predates the flag stays on disk rather than being dropped. It is dormant, logged as such at startup, and restored — re-validated — if the flag is later removed from the unit file, which is better than silently reverting that host to discovery. It cannot be created any more, since `/enable` now refuses it.

New tests: `TestServeEnable_RefusesPinConflictingWithClientIPFlag` (409, flag named, nothing written, reconciler untouched), `TestServeEnable_AcceptsPinMatchingClientIPFlag`, `TestServeEnable_NoBodyUnaffectedByClientIPFlag`.

CHANGELOG.md
juan-malbeclabs: Fixed in 8543c5988.

Agreed. There is now a `### Breaking` entry naming both ways an invocation that worked before aborts now: an address this host does not hold, and a payer holding a dynamic pass. The `### Changes` entry keeps what the flag does when it *is* honored.

CHANGELOG.md
juan-malbeclabs: Fixed in 8543c5988 — the four entries are now 48 to 132 words, against a median of 129 in v0.41.0 and 104 in v0.40.0, and the IDs are gone.

One correction on the stated reason, since it will come up again: PR and issue IDs are not against convention here — 234 of the file's 1066 entries carry one, including `(malbeclabs/doublezero#4186)` in v0.41.0 directly above these. Length was the real problem: at 281/403/296/65 words, the second of these was longer than anything else in the file. Dropped the IDs anyway, they cost nothing.

client/doublezerod/internal/manager/manager.go
juan-malbeclabs: Fixed in d6b5eafcf — the clear is gone, not just the comment.

You are right that it is unreachable: `ResolveTunnelSrc(dst)` is a function of the routing table alone, keyed on the destination, and only read for `IBRLWithAllocatedIP` and `Multicast`, where the pin is not the source. `clientIPChanged` stays on `reconcilerCmd`, since the loop still has to reconcile on a pin — the previous address's services are torn down and the new one's provisioned — and the comment there now says that instead.

The CHANGELOG advertised the clear too ("drops the cached tunnel sources resolved for the old address"); that sentence went with it.

client/doublezerod/internal/runtime/run.go
juan-malbeclabs: Fixed in d6b5eafcf — `restoreClientIPPin` applies `manager.IsPublicIPv4` before the assignment check, and a non-global pin falls back to discovery with its own warning.

Same lockout by a different route, as you say, and it is the shape a pin most easily decays into: globally routable when pinned, behind CGNAT after a re-addressing. New case in `TestRestoreClientIPPin`, "non-global pin falls back to discovery even when held" — held by the host is not enough on its own.

One knock-on worth noting: the existing cases in that table used `203.0.113.9`, which is TEST-NET-3 and so not globally routable, and would have started failing for the right reason. They use a routable address now.

crates/doublezero-daemon-cli/src/connect.rs
juan-malbeclabs: Fixed in fef0dba70. The `filter` is gone; the lookup now matches on three arms, so a pass that exists and is flagged gets its own refusal:

```
❌  The AccessPass at <ip> for UserPayer <pk> is flagged allow_multiple_ip
```

The distinction matters beyond accuracy: the two cases send the operator to do different things — obtain a pass for this address, or have the flag cleared on the one they already hold. `refuses_an_exact_pass_flagged_allow_multiple_ip` now asserts the new message *and* that the absent-pass one is not printed.

rfcs/rfc27-ip-verification.md
juan-malbeclabs: Fixed in 1d9bb934d.

You are right, and I'd agree it was worse than minor: the paragraph read as a program guarantee in a section of MUSTs, and `smartcontract/cli/src/user/create.rs` takes `client_ip` as a plain argument with no such check, as does any SDK caller. The paragraph now opens with what the program does enforce — issuance alone, through the PDA seed — states explicitly that the program does not know and MUST NOT be assumed to enforce that the host asking holds the address, and attributes the local-interface check to `doublezero connect --client-ip` as a property of the client daemon's CLI, naming `user create` and SDK callers as the paths that reach the exemption without it.

smartcontract/programs/doublezero-serviceability/src/processors/user/create_core.rs
juan-malbeclabs: Fixed in 2b78e6aaf, in all four places plus the CHANGELOG.

Your reading is right and I had this backwards: the `assert!` admits only `PDA(client_ip)` or `PDA(0.0.0.0)`, and `accesspass_is_ip_bound` already requires the former, so the pass is seeded on the address being claimed whatever the flag says — the flag cannot make it authorize another one.

The clause stays, with the reason it actually has: `allow_multiple_ip` is the issuer marking the pass reusable across addresses, which is a weaker statement about any one of them than a pass pinned to a single address, so the waiver declines it and takes only the stronger. Clearing the flag on the pass is what restores the waiver. Behavior is unchanged — comments, one test doc comment, and the CLI's copy of the same claim.

smartcontract/sdk/rs/src/commands/accesspass/get.rs
juan-malbeclabs: Fixed in daced0bc5 — `get_multiple_accounts(vec![pubkey])`, so absence is the typed `None` in the returned vector and an `Err` is the read failing and propagates. No error text is matched anywhere now.

Two things came with it: an account at that PDA owned by another program is absence too, decided on the owner rather than on a failed decode, and a decode failure on an account that *is* ours is an error rather than a silent `None`. The four tests move to `expect_get_multiple_accounts`, and the transport-error one no longer depends on the wording it asserts against.

juan-malbeclabs:
Thanks both. All nine are fixed and pushed; the threads are resolved, each with its own reply. One commit per finding area, so they can be read separately:

| # | Sev | Fix |
|---|-----|-----|
| 1 | Important | `d6b5eafcf` — `/enable` refuses a pin the `-client-ip` flag would overrule, naming the flag |
| 2 | Important | `8543c5988` — `### Breaking` entry for the abort |
| 3 | Minor | `8543c5988` — entries cut to 48–132 words, IDs dropped |
| 4 | Minor | `d6b5eafcf` — the cache clear is gone, not just its comment |
| 5 | Minor | `d6b5eafcf` — `restoreClientIPPin` applies `IsPublicIPv4` |
| 6 | Minor | `fef0dba70` — a flagged pass gets its own refusal |
| 7 | Minor | `1d9bb934d` — the local-interface check attributed to the client CLI |
| 8 | Minor | `2b78e6aaf` — the `allow_multiple_ip` rationale corrected in all four places |
| 9 | Minor | `daced0bc5` — absence decided on the account, not the error text |

**On #1**, I took the direction you named — the flag keeps precedence and a differing pin is refused with a 409, which the CLI surfaces verbatim. Two decisions inside that are mine rather than yours: a pin equal to the flag's address is not a conflict, and a pin that predates the flag stays on disk dormant (logged at startup, re-validated and restored if the flag is later removed) rather than being dropped. Say if you'd rather it were erased.

**On #4**, the clear was also advertised in the CHANGELOG, so that clause went with it.

**On #7**, agreed it was worse than minor, and @elitegreg's reading is the one I fixed to: the paragraph now says what the program does enforce — issuance alone — states that it does not know the host holds the address, and names `user create` and SDK callers as the paths that reach the exemption without the check.

**On #8**, you are right and I had it backwards. The assert admits only `PDA(client_ip)` or `PDA(0.0.0.0)`, so the flag cannot make a pass authorize another address. The clause stays as an issuer opt-out with the reason it actually has; no behavior change.

**One pushback, on #3.** Length was real — at 281/403/296/65 words those were the longest entries in the file, against a median of 129 in v0.41.0. PR IDs are not against convention, though: 234 of the file's 1066 entries carry one, including `(malbeclabs/doublezero#4186)` in v0.41.0 directly above these. Dropped them anyway.

**Nit**: `nix` is sorted after `mockall` in `86f5710cc`.

The description is updated for three things this round invalidated: the tunnel-src cache clear, the claim that the daemon repeats only the local-assignment half (it repeats both, since the round-2 fix), and the diff table, now +2591/−273 with core logic at +845.

Still open from my last round and unclaimed by either of you — version skew (a daemon predating this PR 200s the pin and discards it), `poll_for_daemon_provisioned` matching on `user_type` alone, the pin reaching the daemon before any onchain write only on the bare path, the read-modify-write window in `ServeEnable`, and `IFF_RUNNING` on a carrier-less bridge or bond. Happy to take any here; otherwise they go to a follow-up issue.

crates/doublezero-daemon-cli/src/clientip.rs
```diff
@@ -0,0 +1,209 @@
...
+//! one the daemon discovered. Nothing onchain attests that address, so before it is used the
+//! CLI checks the two things it can check locally: that the address is one DoubleZero will
+//! accept at all, and that this host actually holds it.
+//!
+//! The second check is what makes the flag safe to honor. An address assigned to an up
+//! interface is one the kernel will let this host source from, so a caller cannot pin an
+//! address belonging to somebody else and have a tunnel built toward it. It is a possession
+//! check, not a proof of ownership: it says the host holds the address now, which is what the
+//! local configuration needs, and says nothing about who holds it on the public internet.
+//! RFC-27 proofs remain the answer to that question.
```
elitegreg, line 13: Does this flag serve the NAT'd host that #4333 names as motivation?

The issue and the RFC amendment both give "egressing through a NAT other than the one the pass names" as a case that needs this. Such a host doesn't hold the public address on any interface, so `NotLocallyAssigned` refuses it here, and the daemon repeats the check fatally. Meanwhile the workaround this replaces, `doublezerod -client-ip B`, has no such check — so the new flag is strictly less permissive than the old one for exactly that population.

For plain IBRL the pinned address becomes the GRE source verbatim, so I can believe a NAT'd host genuinely can't be served and the check is right. If that's the answer, could the `--client-ip` help text and the RFC's "The IP-bound pass exemption" paragraph say so? As written the RFC lists the NAT host among those the exemption exists for, while the client half declines it, and an operator reading both would reasonably expect `connect --client-ip` to work for them.

client/doublezerod/internal/manager/http.go
```diff
@@ -145,22 +147,103 @@ func (n *NetlinkManager) ServeStatus(w http.ResponseWriter, r *http.Request) {
...
+		clientIP = parsed.To4()
+
+		// A pin the next restart would overrule is refused rather than accepted and quietly
+		// reverted. The -client-ip flag outranks a pin at startup (it is the operator's standing
+		// configuration for this host, in the unit file), but SetReconcilerState adopts a pin
+		// immediately, so without this check the two disagree: the pin works, is persisted and is
+		// reported by /v2/status, until a restart puts the flag's address back and the host
+		// matches no onchain user. Naming the flag is the whole value of the message — the
+		// operator has to edit the unit file, and nothing else would say so.
+		if n.flagClientIP != "" && n.flagClientIP != clientIP.String() {
```
elitegreg: This compares the raw unit-file string against a canonicalized `net.IP`. A flag written `1.2.3.04`, `::ffff:1.2.3.4`, or with stray whitespace would 409 an address that is in fact identical to the pin.

Parsing `flagClientIP` once at construction (it's immutable after that, as the field comment notes) and comparing with `net.IP.Equal` would close it. Low severity, but the message tells the operator to go edit their systemd unit, so a false positive here is an expensive one.

CHANGELOG.md
```diff
@@ -6,6 +6,24 @@ All notable changes to this project will be documented in this file.
...
+- Serviceability
+  - With `RequireIpOwnershipProof` set, a user creation carrying no RFC-27 proof is accepted when its AccessPass sits at the PDA of the address being claimed and is not flagged `allow_multiple_ip`. `SetAccessPass` is permissioned, so such a pass is an issuing authority's attestation of that address, and the flag and the pass together otherwise left a host unable to connect when no proof was obtainable for it. The waiver is keyed on the PDA seed rather than the stored `client_ip` field, which a legacy wildcard pass can carry without any authority behind it. It covers absence only: a supplied proof is still validated in full, and wildcard passes still require one. The trust boundary moves onto issuance, which `SetAccessPass` extends to tenant administrators as well as `ACCESS_PASS_ADMIN` holders. RFC-27 is amended to match.
+
+## [v0.41.0](https://github.com/malbeclabs/doublezero/compare/client/v0.40.0...client/v0.41.0) - 2026-09-16
```
elitegreg: Stray empty section — merge artifact. There's already a real, populated `v0.41.0` section further down the file, and this duplicate sits *above* `v0.42.0`, so the file reads v0.41.0 → v0.42.0 → … → v0.41.0.

Deleting lines 23-28 (this heading and its empty `### Breaking` / `### Changes`) restores the order. The `changelog` CI check passes, so nothing will catch it automatically.

elitegreg (approved):
Approving. The design holds up: the waiver is keyed on the PDA seed rather than the stored `client_ip` field (so a legacy self-bound wildcard pass can't buy it), it covers absence only, and the CLI gate predicts the pass the transaction will actually attach rather than the one the operator hoped for. Reading `clientIP` once per reconcile pass and threading it down to `buildProvisionRequest` closes a real race, and "persist the pin in effect, not the pin in the request" is the non-obvious rule that keeps the daemon and the state file from diverging.

Three comments inline, none of them blocking except the changelog one, which should be tidied before merge.

Two things I'd like on the record rather than changed here:

- **The tenant-admin reach of the waiver wants an explicit sign-off.** The PR and the RFC amendment both say plainly that `SetAccessPass` authorizes on `is_privileged || is_tenant_admin` with nothing constraining `value.client_ip`, so a tenant administrator can pin a third party's address to their own payer and create a user there with no proof. The `ACCESS_PASS_ADMIN` bit in the existing `flags` byte is the right containment. I'd want a yes from whoever owns RFC-27 on this rather than a merge-by-default.
- **Size.** Core logic is ~845 lines against the ~500 guideline. The onchain half is the clean split you already identify, and it's the half carrying the security decision, so it would have got a narrower review on its own. Not asking you to unpick it now given the halves are coupled end-to-end.

I checked the e2e concern: the CYOA network allocates from `9.128.0.0/9`, which is globally routable and assigned to the client container's interface, and those tests set the pass at the exact CYOA address with no dynamic pass alongside, so both new gates should pass. Still wants the green run — `rust-test`, `go-test` and the client release validation were all pending when I looked.

crates/doublezero-daemon-cli/src/clientip.rs
juan-malbeclabs: No, it does not serve that host — and you are right that saying so is owed to anyone reading the issue and the RFC together. Documented in 0d1a341ef, in both places.

The answer, for the record: for plain IBRL the pinned address becomes the GRE tunnel source verbatim, so an address the kernel does not hold cannot carry a tunnel whatever the ledger says about it. The NAT'd host reaches its pass the way it always did — through the address the daemon discovers, or `doublezerod -client-ip` — and that path is untouched by this PR. The flag serves a different host: one that *does* hold several addresses and has to say which. The RFC now has a paragraph making that split explicit, and the help text says it with the reason and the two ways out.

On "strictly less permissive than the old workaround": true, and worth being precise about how far. The daemon flag has never been locally validated, so it will still pin an address this host does not hold — that asymmetry is deliberate for now (it is standing operator configuration in a unit file, not a per-connection claim), but it is the thing that makes the old path work for the NAT'd host and the new one not.

One case where the gate is stricter than the tunnel needs, which I would rather record than paper over: `buildProvisionRequest` uses the pin as the source only for plain IBRL. For `IBRLWithAllocatedIP` and `Multicast` it resolves the source from the routing table precisely because those hosts may be behind NAT, so a pin the host does not hold would work there — it is only ever matched against onchain users. The gate refuses it anyway, in the CLI and in `/enable`, because neither knows which service the user will end up with and a pin outlives the connection that set it. Relaxing it per service is a real change, not a wording fix; happy to file it if you think that population exists.

client/doublezerod/internal/manager/http.go
juan-malbeclabs: Fixed in e228977f7 — parsed once in `WithClientIPFlag`, held as a `net.IP`, compared with `Equal`.

Narrowed the reachable set while doing it, since it changes what the test needs to cover: `DiscoverClientIP` parses the same flag at startup and the daemon does not start if it fails, so `1.2.3.04` and stray whitespace never get as far as this handler — `net.ParseIP` rejects both. The IPv4-in-IPv6 form is the one that survives startup and then mismatches on the string, so that is what `TestServeEnable_FlagConflictComparesAddressesNotStrings` pins.

Agreed on the severity reasoning: a false positive here is expensive precisely because the message is "go edit your systemd unit".

CHANGELOG.md
juan-malbeclabs: Fixed in ce073a997 — the six lines are gone and the file reads v0.42.0 → v0.41.0 → v0.40.0 again.

It came in with the merge of main at 4166768b5: this branch had rebuilt its `Unreleased` section from main's file in an earlier round, main then cut v0.42.0, and the merge kept both the real section and the emptied heading the release had left behind. As you say, nothing catches it — the `changelog` check only asks that the file was touched.

juan-malbeclabs:
Thanks for the approval, and for the NAT question — that one was the most useful comment on this PR, because it is the only one that found something the code and the documents disagreed about rather than something the code got wrong.

All three are addressed and the threads are resolved:

| Comment | Fix |
|---|---|
| Does the flag serve the NAT'd host? | `0d1a341ef` — no, and the RFC and the help text now say so, with the reason and the two paths that do serve it |
| Flag compared as a raw string | `e228977f7` — parsed once at construction, compared with `net.IP.Equal`, with a test on the IPv4-in-IPv6 spelling |
| Stray empty v0.41.0 section | `ce073a997` — removed; it arrived with the merge of main |

**On the NAT answer**, the substance is: for plain IBRL the pin becomes the GRE tunnel source verbatim, so a host that does not hold the address cannot be served through this flag, whatever the ledger says. Such a host reaches its pass exactly as it did before this PR — through discovery or `doublezerod -client-ip` — and that path is untouched. The two halves address disjoint populations and the RFC now has a paragraph saying which is which.

I recorded one thing on that thread rather than changing it: for `IBRLWithAllocatedIP` and `Multicast` the source is resolved from the routing table, not from the pin, so a pin the host does not hold would in fact work for those. The gate refuses it anyway, because neither the CLI nor `/enable` knows which service the connection will end up with and the pin outlives it. Relaxing that per service is a real change; say the word and I will file it.

**On the tenant-administrator sign-off**: agreed that it should not merge by default. @ben-malbeclabs, this is the one thing outstanding on this PR — the "Security decision this PR makes" section in the description states it in full, and both reviewers have now asked for an explicit yes rather than silence. @elitegreg, you and Juan are the two names on RFC-27's history; if you would rather the narrower `ACCESS_PASS_ADMIN`-stamped flag land *before* this, I would rather do that than merge the wider boundary and file the containment.

**On size**: agreed, and no argument that the onchain half would have got a sharper review on its own. If the sign-off above goes the other way, splitting it out stops being a size question and becomes the plan.

CI was green at `86f5710cc` (44/44, all six e2e shards) and is running again on `ce073a997`; I will report it rather than assume it carries over.

PR #4342, opened 2026-09-15
serviceability: drive the RFC-28 sequence in order

bgm-malbeclabs:
The first test that reaches `ActivateFeed` from a real `CreateFeed` on a mirror a real `WriteStakeMirror` produced.

## Summary

- **The sequence, driven in order**: the relayer writes a `StakeMirror`, a builder holding no permission creates a feed against it, grants its own publisher the right to send, and an operator activates it. Steps 2 to 5 of the demo.
- **Nothing is seeded that an instruction could write.** The feature flag, the multicast group, the relayer's `STAKE_ORACLE`, the mirror, the feed, the publish grant and the activation all go through their real instructions.
- **Every route into `authorize` is ruled out, not some of them.** `CreateFeed` asks for `FEED_AUTHORITY | FOUNDATION`, and three grants satisfy that: a `Permission` account, the foundation allowlist, and `feed_authority_pk`. All three are asserted. The whole claim is that a bond deploys a feed, and if the builder quietly held any of them the rest would prove nothing.
- **The account lists match what a real caller sends.** The allowlist step sends the optional `user_payer`, so `airdrop_user_credits` runs, and `WriteStakeMirror` passes `globalstate` read-only, as `write_stake_mirror` does.

Composition on its own is not the novelty, and the header says so: `feed_test::test_builder_grants_its_own_publish_rights` already chains create plus two allowlist grants with two signers, and `feed_test::test_a_mirror_update_keeps_the_claiming_feed` drives `CreatePermission` to `WriteStakeMirror` to `CreateFeed` without seeding either. What is new is the reach through to activation.

## What it does not cover

Written into the test's header, because a green test invites more belief than it earns:

- **The bond.** `PostBond` is a different program on Solana, so the mirror is written from values rather than read from a real `BuilderStake`. The relayer decoding a bond a program produced is C2.
- **Datagrams.** Nothing in a `ProgramTest` sends multicast. That needs a cluster, and it is G2.
- **The refusals.** Only the path that succeeds runs here. `feed_lifecycle_test` holds the guards, and the header names them.

## Testing Verification

`cargo test -p doublezero-serviceability --tests`: **all green**, the walkthrough among them.

Each step now fails if what it claims stops being true:

- **Deleting the `AddMulticastGroupPubAllowlist` step fails the test.** Verified by deleting it. Before this, it asserted nothing, so the step could be removed and every remaining assertion stayed green while the header still claimed the builder granted its publisher the right to send. It now reads the access pass back and pins `mgroup_pub_allowlist` and `user_payer`.
- **Disabling A7's gate fails the create.** `split_stake_mirror` consumes the mirror by address, so the `authorize` candidate list is empty and it falls to the legacy path, where the builder matches none of the three grants. `NotAllowed`.
- **The `Permission` check reads the account directly.** It went through a helper that returns `None` both for an absent account and for one that stopped deserializing. That assertion is what makes the rest of the test mean anything, so it must not pass for the second reason.
- **`feed_authority_pk` was the untested third route.** It held only because `initialize` writes `Pubkey::default()` there, so a fixture change would have opened it quietly.
- **Both committed rates come from `StakeTier::UpTo1Gbps.max_rate_bits_per_sec()`**, so the coverage check stays at the tier boundary if that ceiling moves rather than sitting comfortably under it.

## Known gap, not fixed here

`rfc26_builders_test.rs` exists to catch drift between the instruction builders and the processors, and its coverage stops at device, link, user and topology. Nothing runs the feed or stake-mirror builders against the program, which is how both account-shape divergences above survived. This test hand-rolls its account lists like its siblings; driving it through `doublezero-serviceability-instruction` instead would close the gap for good, and that is a change worth making on its own.


bgm-malbeclabs:
## Self-review

One finding, fixed in `835868fe0`, and it is the one that decides whether this test is worth having.

**The test ruled out one route into `authorize`, not both.** It asserted the builder held no `Permission` account and stopped there. `authorize` has two ways in: a `Permission` account, and the legacy foundation allowlist. A builder sitting in that allowlist would have been admitted by it, the create would have succeeded for entirely the wrong reason, and this test would still have gone green while claiming a bond deployed the feed.

It happens not to be true of this fixture — `init_globalstate` makes the ProgramTest payer the sole allowlist member, and the builder is `test_payer()` — but the test relied on that staying true rather than saying so. The only thing that actually caught it was the mutation run, where disabling A7's gate produced `NotAllowed`, and mutations do not run in CI.

It now asserts both. Verified the new assertion can fail by pointing it at the foundation key, which fails with "the builder must not be a foundation key, or the create above proves nothing".

**Checked and sound**, rather than assumed:

- Nothing is seeded: `init_test_with_accounts(program_id, &[])`, and every account the walkthrough touches is written by the instruction that owns it.
- The `feed_key` assertion on the mirror is what makes "one bond backs one feed" a claim the test holds rather than a sentence in a comment.
- The header's exclusions are accurate: no `BuilderStake` is read, because `PostBond` is a different program on another chain, and no datagram is sent, because `ProgramTest` has no network.

**One thing I did not add.** The demo's claim is that an `Active` feed sells seats, and this asserts the status rather than driving `SubscribeFeed`. The status is what A5's gate reads, so the assertion is the right proxy, but it is a proxy. Driving a subscription would make it end-to-end through the subscriber as well, and that is G2's job against a cluster rather than this test's.

ben-malbeclabs (approved):

PR #4343, opened 2026-09-15
cli: an operator can run a feed's lifecycle

bgm-malbeclabs:
## Summary

- `doublezero feed` gains `activate`, `halt`, `resume`, `retire` and `finalize-retirement`. Every feed instruction added since the catalog verbs was unreachable from the CLI: `feed` offered create, update, list and delete only, so moving a feed to `Active`, stopping it publishing, or starting its retirement notice all meant building a transaction by hand. No test caught that, because tests call instructions directly, which is the blind spot a missing command sits in.
- `activate` and `resume` re-read the feed's stake mirror, and each derives that address from the feed's own `stake_ref` rather than taking it as a flag. The feed already records which stake backs it, so an argument would only be a way to name a different one. A catalog feed sends no mirror, because there is no stake to re-read and sending one asks the program for an account it refuses.
- `finalize-retirement` appends no `Permission` account, unlike the other four. Finalizing is permissionless: the clock decides rather than a key, and sending one would ask the program for an account it does not read.
- The retirement help said thirty days flat. The processor sets `retires_at` to now for a `Pending` feed, because the notice exists for seat holders and a feed that never published has none, so the help and the doc comments now say both.
- Five SDK commands and five `CliCommand` methods behind them.

The first commit said halt, resume and retire belonged in a separate change. That was wrong about the size: the five verbs share the argument group, the feed lookup and the mirror derivation, and the whole set is 478 lines of non-test code. A reviewer who has read one verb has read all five.

## Testing Verification

- **The mirror reaches the program.** `test_commands_feed_activate_forwards_the_stake_mirror` and its `resume` twin build the expected instruction with `Some(&mirror)` and match the sent transaction against it. Replacing `self.stake_mirror.as_ref()` with `None` in `activate.rs` fails that test and passes every other, which is the drift these two exist to catch: a staked feed would otherwise reach the program with no stake to read and be refused as `StakeMirrorMissing`, with nothing naming the cause.
- **A catalog feed sends no mirror.** Four tests, CLI and SDK, on a feed whose `builder` is the default pubkey. The CLI pair pins `stake_mirror: None` in the command; the SDK pair pins an instruction one account shorter.
- **The permission account is appended, or not.** `halt`, `retire`, `activate` and `resume` each assert the `Permission` PDA lands last, after the accounts the instruction builder names, which is where `authorize` reads it. `finalize_retirement` asserts the opposite, with no `get_multiple_accounts` expectation set: the mock panics if the command ever starts looking one up.
- **The right feed is read and the right command is sent.** Every CLI test pins the `GetFeedCommand` it issues and the SDK command it builds with `predicate::eq`, so a verb wired to its neighbour's command fails rather than passing on a shared shape.
- **Output.** The signature line is matched exactly, not by substring.
- 444 CLI and 219 SDK tests pass. `cargo clippy --all-targets -- -Dwarnings` is clean on both crates.

Implements the CLI half of RFC-28's feed lifecycle.


smartcontract/sdk/rs/src/commands/feed/activate.rs
```diff
@@ -0,0 +1,29 @@
...
+impl ActivateFeedCommand {
+    pub fn execute(&self, client: &dyn DoubleZeroClient) -> eyre::Result<Signature> {
+        let mut ix = activate_feed(
+            &client.get_program_id(),
+            &client.get_payer(),
+            &self.pubkey,
+            self.stake_mirror.as_ref(),
+        );
+
+        append_payer_permission_account(client, &mut ix)?;
```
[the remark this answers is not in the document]
bgm-malbeclabs: Half stale, half taken. The SDK tests you are asking for landed after this thread was written: `activate` and `resume` each have `forwards_the_stake_mirror` and `sends_no_stake_mirror_for_a_catalog_feed`, and `halt` and `retire` have `appends_the_payer_permission_account`. Each builds the expected instruction and matches it with `predicate::eq`, so account ordering and the appended Permission PDA are both pinned. Replacing `self.stake_mirror.as_ref()` with `None` fails one of them and nothing else, which is the regression the CLI tests could not see.

The both-cases point was worth taking, but not everywhere. The no-permission branch lives in the shared `append_payer_permission_account`, and `create.rs` and `update.rs` already pin it, so repeating it per command tests the helper rather than the command. A missing call to the helper is already caught by the with-permission test. Where it does say something new is halt and resume: under RFC-28 a builder holds no `Permission` and those two are its levers, so the no-permission instruction is the normal path for them rather than an edge. Added for both. Activate and retire keep one case each, because an operator is the only caller either has.

smartcontract/cli/src/cli/feed.rs
```diff
@@ -13,6 +16,16 @@ pub enum FeedCommands {
...
+    /// Admit a feed that was waiting on a conformance verdict
+    Activate(ActivateFeedCliCommand),
+    /// Stop a feed publishing
+    Halt(HaltFeedCliCommand),
+    /// Put a halted feed back to publishing
+    Resume(ResumeFeedCliCommand),
+    /// Start a feed's retirement notice, thirty days once it has published
+    Retire(RetireFeedCliCommand),
+    /// End a retirement once its notice has elapsed
+    FinalizeRetirement(FinalizeFeedRetirementCliCommand),
```
ben-malbeclabs, line 28: Five verbs drive a lifecycle no CLI surface shows: `FeedDisplay` in `feed/list.rs` omits `status`, `halted_by` and `retires_at`, and there is no `feed get`. An operator cannot tell `Pending` from `Halted` from `Retiring`, or when `finalize-retirement` will stop failing.

smartcontract/cli/src/feed/halt.rs
```diff
@@ -0,0 +1,120 @@
...
+    validators::{validate_code, validate_pubkey, validate_pubkey_or_code},
+};
+use clap::{ArgGroup, Args};
+use doublezero_cli_core::{print_signature, require, CliContext, RequirementCheck};
+use doublezero_sdk::commands::feed::{get::GetFeedCommand, halt::HaltFeedCommand};
+use std::io::Write;
+
+/// Stop a feed publishing. Its own builder may sign this, or an operator.
+#[derive(Args, Debug)]
+#[clap(group(ArgGroup::new("target").args(&["pubkey", "code"]).required(true)))]
```
ben-malbeclabs: The target arg struct, its `ArgGroup`, the `require!` and the resolve-exchange-then-`get_feed` block are copied into all five new verbs; halt, retire and finalize-retirement differ by six lines out of fifty-three, and the group now sits in seven files. Flatten one `FeedTargetArgs` with a `resolve` method and have each verb call it.

ben-malbeclabs (commented):
Two findings, both Important. The wiring itself checks out: `stake_mirror_of` uses the same condition and seed as `split_stake_mirror`, the mirror-then-permission account order is handled by the program's address-based split, and `finalize-retirement` correctly appends no `Permission` account.

1. `smartcontract/cli/src/cli/feed.rs` — the lifecycle is not observable from the CLI.
2. `smartcontract/cli/src/feed/halt.rs` — the five verbs are copies of one another.

smartcontract/cli/src/feed/halt.rs
bgm-malbeclabs: Done. `FeedTargetArgs` in `feed/resolve.rs` holds the three arguments, the `ArgGroup` and a `resolve` method; all seven verbs flatten it and call `self.target.resolve(client)?`. The group is gone from seven files and the resolve-exchange-then-`get_feed` block from seven copies to one.

One thing it costs, so you can veto it: the per-verb help text goes. "Feed pubkey to delete" becomes "Feed pubkey" everywhere, since a flattened struct carries one help string. The verb is already on the command line, so I took that as a fair trade rather than passing the noun in.

Checked that clap still enforces the group through the flatten, since it configures at runtime and a bad group panics rather than failing to compile: `feed halt` with neither argument gives `the following required arguments were not provided: <--pubkey <PUBKEY>|--code <CODE>>`, and with both gives `the argument '--pubkey <PUBKEY>' cannot be used with '--code <CODE>'`.

smartcontract/cli/src/cli/feed.rs
bgm-malbeclabs: Done for `feed list`: `FeedDisplay` now carries `status`, `halted_by` and `retires_at`. A zero key and a zero timestamp render empty rather than thirty-two ones or 1970, which read as an answer instead of an absence. `retires_at` formats as `2026-10-01 00:00:00 UTC`, matching `slot_to_datetime` in `helpers.rs`.

`halted_by` was the one worth having beyond visibility: an operator's halt is not the builder's to lift, so which of them holds it is the difference between `resume` working and not. A new test drives a halted feed and a retiring one and pins all three columns for each.

I did not add `feed get`. It is a new verb rather than a fix to an existing surface, and `list --code <code> --exchange <metro>` already narrows to one feed with the same columns. Say the word if you want it in this PR and I will add it; otherwise it seems like its own change.

Flagging the size while I am here: this PR is now about 595 lines of non-test code, over the ~500 the guide asks for. The clean split is the `feed list` change, which touches only `list.rs` and the changelog and does not depend on the five verbs. Happy to lift it into its own PR if you would rather review it separately.

ben-malbeclabs (approved):

PR #4346, opened 2026-09-15
ci: start the offchain fork loader from its built binary

### LINKED ISSUE #4345: ci: offchain fork job timeouts are undiagnosable, and the loader races the build
## Summary

The two `offchain-local-validator` fork jobs can fail with `Solana fork did not start within 60 seconds`, and when they do the logs contain nothing that identifies why. The loader's output is sent to `/dev/null`, so a loader that failed for its own reasons is indistinguishable from one that was merely slow. Separately, the way the loader is started makes it structurally likely to be slow on a cold cache.

Observed on PR #4341 in run [35029844666](https://github.com/malbeclabs/doublezero/actions/runs/35029844666); **both jobs passed on a re-run**, so the failure is transient rather than deterministic.

## The diagnostic gap

```yaml
run: cargo run -p doublezero-solana-fork-cli --bin doublezero-solana-fork -- -um --reset ... > /dev/null 2>&1 &
```

Nothing the loader prints is retained. The only signal is the generic timeout from the wait loop, which cannot distinguish:

- the loader crashing or rejecting a flag,
- the loader failing to clone accounts from mainnet-beta,
- the loader still compiling when the window expired.

In the run above, the third is plausible on timing (below), and the second is plausible because `sdk-compat-test` failed in the same window with RPC `503 Service unavailable`, which points at upstream mainnet RPC being degraded at 22:14–22:21 UTC. **The logs cannot tell these apart**, which is the primary problem: the job is unactionable when it fails.

## The structural race

The loader is backgrounded as `cargo run`, then the next step runs a foreground `cargo build`:

```yaml
- name: Start Solana mainnet-beta fork in background
  run: cargo run -p doublezero-solana-fork-cli --bin doublezero-solana-fork -- ... &
- name: Build `doublezero-solana`
  run: cargo build -p doublezero-solana-cli --bin doublezero-solana
```

Started that way the loader must compile before it can serve, and both cargo invocations contend for the same `target/` lock, so the loader blocks behind the foreground build. The readiness wait then starts counting at the moment that build finishes. Timeline from the failing run:

| Time (UTC) | Event |
| --- | --- |
| 22:14:21 | `Swatinem/rust-cache` reports `No cache found.` |
| 22:14:40 | Fork step launches `cargo run … &` |
| 22:14:45 | `Build doublezero-solana` starts compiling |
| 22:19:19 | Foreground build finishes (4m39s) |
| 22:19:19 | Readiness wait begins |
| 22:21:20 | Wait gives up |

This leaves the loader ~2 minutes to finish compiling and boot. It is a real defect regardless of what caused this particular failure: the readiness budget is spent on compilation that could have happened earlier, and the margin shrinks exactly when the cache is cold.

`Swatinem/rust-cache` is keyed on `Cargo.lock`, so a PR changing `Cargo.lock` compiles the whole graph cold and has the least margin. That is also the only reason the workflow runs on such a PR: its path filter is `offchain/**`, `solana/**`, `Cargo.lock`, `Cargo.toml`, and `Cargo.lock` is often the sole match, with nothing under `offchain/` touched.

## Smaller problems

1. **The timeout message is wrong.** Both scripts loop `for i in {1..60}` with `sleep 2`, i.e. ~120s, but report `did not start within 60 seconds`. The observed 22:19:19 → 22:21:20 gap is 2m01s, matching the loop.
2. **Nothing checks the loader is alive.** If the backgrounded process dies immediately, the scripts still spin the full window before reporting a timeout.

## Proposed fix

- Compile the loader in the build step and start it from `target/debug/doublezero-solana-fork`, removing both the lock contention and the compile inside the readiness window.
- Send the loader's output to a log file and dump it when readiness fails, so the next failure is diagnosable.
- Record the loader's pid and fail fast if it exits.
- Derive the timeout message from the actual window.

## Affected files

- `.github/workflows/offchain.local-validator.yml`
- `offchain/sh/test_doublezero_solana_fork.sh`
- `offchain/sh/test_validator_debt_fork.sh`

juan-malbeclabs:
## Summary

- Captures the fork loader's output to `solana-fork.log` and dumps it when readiness fails, so a fork-job timeout stops being unactionable.
- Starts the loader from its built binary instead of a backgrounded `cargo run`, removing a race in which the loader competed with the foreground build for the `target/` lock and then had to compile inside the readiness window.
- Reports a loader that exited immediately rather than spinning out the full window.
- Derives the timeout message from the actual wait, which was 180s of `sleep 2` while reporting "60 seconds".

Fixes #4345.

## Why

The loader was backgrounded as `cargo run … > /dev/null 2>&1`, with a foreground `cargo build` in the next step.

**Nothing it printed was kept.** The wait loop's timeout was the only signal, and it cannot distinguish a loader that crashed, one that could not clone accounts from mainnet-beta, and one that was still compiling. This is the main thing the PR fixes: the job could fail without leaving evidence of why.

**Started via `cargo run`, the loader had to compile before serving**, and both cargo invocations contend for the same `target/` lock, so it blocked behind the foreground build while the readiness wait began at the moment that build finished:

| Time (UTC) | Event |
| --- | --- |
| 22:14:21 | `Swatinem/rust-cache` reports `No cache found.` |
| 22:14:40 | Fork step launches `cargo run … &` |
| 22:14:45 | Foreground build starts compiling |
| 22:19:19 | Foreground build finishes (4m39s) |
| 22:19:19 | Readiness wait begins |
| 22:21:20 | Wait gives up |

That leaves ~2 minutes to compile and boot. The margin is thinnest on a cold cache, which is exactly when a PR changing `Cargo.lock` runs — `rust-cache` is keyed on `Cargo.lock`, and that is usually the only path filter such a PR matches, with nothing under `offchain/` touched.

## Scope note

The failure that prompted this (run [35029844666](https://github.com/malbeclabs/doublezero/actions/runs/35029844666) on #4341) **passed on a re-run**, and it coincided with `sdk-compat-test` failing on RPC `503 Service unavailable` in the same window, which suggests upstream mainnet RPC was degraded at the time. So this PR does not claim to have fixed that specific failure — with the loader's output discarded, its cause cannot be recovered. It removes a genuine race and makes the next occurrence diagnosable.

## Testing Verification

- Verified the wait/report mismatch: the loop is 90 × `sleep 2` = 180s, and the observed 22:19:19 → 22:21:20 gap is 2m01s, matching the loop rather than the "60 seconds" message.
- Confirmed the failing run built cleanly (`Finished dev profile in 4m 39s`, no compile errors) and failed only on fork readiness, with `No cache found.` logged before the cold build — i.e. the loader, not the tests, was the blocker.
- `cargo metadata` confirms `doublezero-solana-fork-cli` resolves from the root workspace and its `[[bin]]` is `doublezero-solana-fork`, so the loader lands at `target/debug/doublezero-solana-fork`.
- Both scripts pass `bash -n`; the workflow parses as YAML with both jobs' steps intact.
- The liveness check is gated on the pid file existing, so the manual `cargo run` flow documented in the script's own comments keeps working unchanged.
- This PR edits the workflow, which is in its own path filter, so both fork jobs run here. Note they will run with a **warm** cache (it leaves `Cargo.lock` untouched), so a green result exercises the new start-from-binary path but not the cold-cache margin the race depends on.


offchain/sh/test_doublezero_solana_fork.sh
```diff
@@ -4,18 +4,35 @@ GENESIS_DZ_EPOCH=31
 
 set -eu
 
-# Wait for Solana fork to start. Only try for 60 seconds.
-for i in {1..60}; do
+# Wait for the Solana fork to become reachable. The caller starts the loader in
+# the background; it clones accounts from mainnet-beta, so it is slower to serve
+# than a bare validator boot.
+FORK_WAIT_SECONDS=180
```
elitegreg: This raises the wait window from 120s to 180s, and the PR description doesn't present it that way.

The old loop was `for i in {1..60}` with `sleep 2`, so 120 seconds, while printing "did not start within 60 seconds". #4345 states that correctly. But the PR's Testing Verification says:

> the loop is 90 x `sleep 2` = 180s, and the observed 22:19:19 -> 22:21:20 gap is 2m01s, matching the loop rather than the "60 seconds" message

2m01s matches the old 120s loop, not 180s. That paragraph describes the new loop as if it were the pre-existing one, so the timing evidence reads as supporting a claim it doesn't support. Same issue in the summary bullet: "derives the timeout message from the actual wait, which was 180s of `sleep 2`" -- the actual wait was 120s.

Raising the budget to 180s is fine on the merits given the cold-cache margin, but please call it out as its own change rather than folding it into the message-accuracy fix, and correct the arithmetic in the body. (Applies equally to `test_validator_debt_fork.sh`.)

offchain/sh/test_doublezero_solana_fork.sh
```diff
@@ -4,18 +4,35 @@ GENESIS_DZ_EPOCH=31
...
+
+for _ in $(seq 1 $((FORK_WAIT_SECONDS / 2))); do
     if solana cluster-version -u l > /dev/null 2>&1; then
         echo "Solana fork is ready."
         break
     fi
-        sleep 2
+    # A loader that died is reported now rather than after the whole window: the
+    # timeout alone cannot distinguish a crash from a slow start.
+    if [ -f "$FORK_PID_FILE" ] && ! kill -0 "$(cat "$FORK_PID_FILE")" 2>/dev/null; then
```
elitegreg: Nothing ever removes `solana-fork.pid` -- not this script, not the workflow -- so a stale one breaks the manual flow this PR says it preserves.

The description argues:

> The liveness check is gated on the pid file existing, so the manual `cargo run` flow documented in the script's own comments keeps working unchanged.

That holds on a clean checkout. But after one local run the file survives with a dead pid, and the documented flow (the comment block around line 237) starts the loader with `cargo run` and writes no pid file. The next invocation finds the stale file, `kill -0` fails on the dead pid, and the script exits 1 with "Solana fork loader exited before becoming ready" while the loader is starting normally -- a confusing failure in exactly the local path the gating was meant to protect.

Adding `rm -f "$FORK_PID_FILE"` on the readiness path would keep it from outliving the run.

While here, two small ones in this block:

- `[ -s "$FORK_PID_FILE" ]` rather than `-f`: an empty pid file makes this `kill -0 ""`, which fails and produces the same false "loader exited".
- The manual-flow comment at line 237 still reads `cargo run --bin doublezero-solana-fork`. Per the workflow header's own note about `default-members = []`, a bare `--bin` with no `-p` selects nothing from the root workspace, so that command probably doesn't work as written. Pre-existing, but this PR is in the neighborhood.

elitegreg (changes requested):
Two things to address. The change itself is the right shape -- moving the loader's compile into the build step removes real `target/` lock contention, and capturing its output is what actually makes the next failure diagnosable. Both fork jobs run and pass here. The requests are about the description's accuracy and one cleanup gap.

offchain/sh/test_doublezero_solana_fork.sh
juan-malbeclabs: You are right on both counts, and the arithmetic error was mine in the description rather than in the script.

The old loop is `for i in {1..60}` with `sleep 2` = **120s**, printing "60 seconds" — wrong by a factor of two, not three, exactly as #4345 states. The 2m01s gap is evidence for the old 120s loop, and I had written it up as if it were evidence for a 180s one, which it is not.

Description rewritten. The wait loop now gets its own section separating the two changes:

1. the message is derived from `FORK_WAIT_SECONDS` so the two cannot drift apart again, and
2. the budget goes 120s → 180s, called out as a deliberate increase.

On (2) — worth being explicit that it is belt-and-braces, not load-bearing. Moving the compile into the build step should make the old 120s sufficient on its own, so if you would rather keep this PR to the race and the diagnostics, say so and I will put it back to 120s. The message fix stands either way.

offchain/sh/test_doublezero_solana_fork.sh
juan-malbeclabs: All three fixed in `1dd4d5d28`.

**Stale pid file.** You are right that the "keeps working unchanged" claim only held on a clean checkout, which makes it the wrong claim to have made about the local path. I went slightly further than the readiness path and used a trap, so the file cannot outlive the run on *any* exit — including the two failure exits, which are the ones most likely to leave a stale file behind for the next local attempt:

```bash
trap 'rm -f "$FORK_PID_FILE"' EXIT
```

Nothing reads the pid file after the script (the workflow only writes it), so clearing it unconditionally is safe.

**`-s` rather than `-f`.** Done, in both scripts. Verified the old behaviour was reachable: an empty pid file gives `kill -0 ""`, which fails, producing the same false "loader exited before becoming ready". Checked empirically — empty file no longer takes the branch, dead pid still does, and the trap removes the file afterwards.

**The manual-flow comment.** Also done, both occurrences (lines 254 and 326). You are right that it does not work as written — and the workflow header three lines up documents precisely that trap for `default-members = []`, so the comment was contradicting its own file. Now `cargo run -p doublezero-solana-fork-cli --bin doublezero-solana-fork -- …`.

Both scripts pass `bash -n`. `shellcheck -S warning` flags three things, all pre-existing and none in the changed lines — including a shebang on line 2 of `test_validator_debt_fork.sh` (the file opens with a blank line), which I left alone as unrelated.

juan-malbeclabs:
Both findings are fixed in `8e2ac826a`, answered on their threads.

- **Wait-loop arithmetic.** You were right: the old loop is 120s (`{1..60}` × `sleep 2`), not 180s, and the observed 2m01s is evidence for *that* loop. The description now separates the message fix from the budget increase, and says plainly that 120s → 180s is belt-and-braces rather than load-bearing — say the word and I will drop it back to 120s.
- **Stale pid file.** Fixed with a `trap ... EXIT` rather than only on the readiness path, so the file cannot outlive a failed run either. Plus `-s` rather than `-f`, and both manual-flow comments now pass `-p doublezero-solana-fork-cli` (a bare `--bin` selects nothing under `default-members = []`, which the workflow header three lines up already documents).

**Also rebased onto main.** The branch was 12 commits behind and the e2e shards were failing on `doublezero feed create --chain` — the flag #4353 added, which the shreds e2e suite calls and this branch did not have. Nothing to do with the change here; the shards should go green now. Both fork jobs were already passing on the pre-rebase head.

elitegreg (commented):
Two structural suggestions; neither blocks. Details inline.

.github/workflows/offchain.local-validator.yml
```diff
@@ -82,13 +93,17 @@ jobs:
...
-      - name: Build the admin and validator-debt CLIs
+      - name: Build the admin and validator-debt CLIs and the fork loader
         run: |
           cargo build \
             -p doublezero-revenue-distribution-admin-cli --bin doublezero-revenue-distribution-admin \
             -p doublezero-solana-validator-debt --bin doublezero-solana-validator-debt \
-            -p doublezero-solana-cli --bin doublezero-solana
+            -p doublezero-solana-cli --bin doublezero-solana \
+            -p doublezero-solana-fork-cli --bin doublezero-solana-fork
+      - name: Start Solana mainnet-beta fork in background
```
elitegreg: Starting from the built binary is right, but note what the reorder costs: the loader used to start *before* the build, so even with the `target/` lock contention it got some overlap once the lock released. Now it starts after, and the whole mainnet account clone — the network-bound part, and the part that was plausibly degraded in the original failure — sits wholly inside the readiness window. That makes the 120s → 180s bump more load-bearing than "belt-and-braces" suggests.

There's a strictly better ordering that's nearly free:

1. `cargo build -p doublezero-solana-fork-cli --bin doublezero-solana-fork`
2. start `target/debug/doublezero-solana-fork &` — no cargo, so no lock contention
3. `cargo build` the remaining CLIs
4. run the tests

The loader clones while step 3 compiles. This job builds three more CLIs after the start point, so that's several minutes of head start currently left on the table, and with it 120s would plausibly hold on its own. Same applies to `test-doublezero-solana-fork` above, with a smaller payoff since it only builds one other CLI.

offchain/sh/test_doublezero_solana_fork.sh
```diff
@@ -4,18 +4,40 @@ GENESIS_DZ_EPOCH=31
...
+    sleep 2
 done
 
-# If not ready after 60 seconds, bail out.
 if ! solana cluster-version -u l > /dev/null 2>&1; then
-    echo "Solana fork did not start within 60 seconds." >&2
+    echo "Solana fork did not start within ${FORK_WAIT_SECONDS} seconds." >&2
+    if [ -f "$FORK_LOG" ]; then
+        tail -50 "$FORK_LOG" >&2
+    fi
```
elitegreg: This block is byte-identical in `test_validator_debt_fork.sh` — I diffed them, no drift yet, but that's 37 duplicated lines that now have to be edited in lockstep. And `test_full_debt_flow.sh:110` already carries a third implementation of the same wait (`wait_for_solana`, parameterized by `FORK_STARTUP_WAIT`/`FORK_STARTUP_INTERVAL`), so this directory ends up with three ways to wait for the fork.

Worth extracting to a sourced helper — something like `offchain/sh/lib/wait_for_fork.sh` exposing `wait_for_fork`, with the log and pid-file paths as parameters. `test_full_debt_flow.sh` could adopt it later; it doesn't have to move in this PR.

elitegreg (approved):

.github/workflows/offchain.local-validator.yml
juan-malbeclabs: Good catch on what the reorder cost — I had moved the start point past the build without accounting for the overlap it used to get for free, and you are right that it made the 180s more load-bearing than I was presenting it as.

Reordered as you suggested, in `de05b7b86`. Both jobs are now:

1. `cargo build -p doublezero-solana-fork-cli --bin doublezero-solana-fork`
2. start `target/debug/doublezero-solana-fork &` — no cargo, no `target/` lock
3. `cargo build` the remaining CLIs
4. run the tests

So the mainnet clone overlaps with step 3 rather than sitting wholly inside the readiness window. `test-doublezero-solana-validator-debt-fork-current` gets three CLIs' worth of head start, `test-doublezero-solana-fork` one, as you noted.

The workflow header now says why the steps are in this order, not just why the loader starts from its binary — the ordering is the load-bearing part and a future edit that recombines the two `cargo build` steps would silently undo it.

**On the budget.** With the head start back, 120s would probably hold — but the wait loop exits the moment the fork answers, so the extra 60s costs nothing on a passing run and only ever buys margin on a failing one. I have left it at 180s. Say the word if you would still rather it went back.

offchain/sh/test_doublezero_solana_fork.sh
juan-malbeclabs: Extracted in `a81cf4c99`. New file `offchain/sh/lib/wait_for_fork.sh` exposing `wait_for_fork`, parameterized as you suggested:

```bash
wait_for_fork [log_path] [pid_path] [wait_seconds]   # defaults: solana-fork.log, solana-fork.pid, 180
```

Both callers drop from 37 lines to two:

```bash
# shellcheck source=lib/wait_for_fork.sh
source "$(dirname "${BASH_SOURCE[0]}")/lib/wait_for_fork.sh"
wait_for_fork
```

`${BASH_SOURCE[0]}` rather than `$0` so the path resolves whichever way the script is invoked; both are run as `bash offchain/sh/<name>.sh` from the repo root.

I left `test_full_debt_flow.sh` alone, as you said it needn't move here. Its `wait_for_solana` is genuinely a different shape — it has the `log_*` helpers, its own `FORK_STARTUP_WAIT`/`FORK_STARTUP_INTERVAL` knobs and a `SKIP_FORK_START` path, and it does not write a pid file at all. The three parameters are there so it can adopt this later without a signature change, but folding in its logging would have to be part of that.

**Two deliberate deltas from the block being replaced**, neither behavioral in the cases that matter:

- The trap now names a module-scoped `_WAIT_FOR_FORK_PID_FILE` instead of interpolating the path into the trap string, so a path with a quote in it cannot break the trap.
- The old code `break`-ed out of the loop and then re-probed, which meant the success path probed twice and the timeout path got one extra attempt after the final `sleep 2`. The helper returns on success, and keeps the extra post-loop probe so the timeout path still gets it. Same total wait, one fewer redundant RPC call on success.

**Verified against a stubbed `solana`** rather than by reading it, since the whole point is the failure paths:

| Case | Result |
| --- | --- |
| ready immediately | returns 0, script continues, pid file gone |
| dead pid | reports "exited before becoming ready", dumps the log, exit 1 |
| **empty** pid file | does *not* take the liveness branch (the `-s` fix), times out instead |
| no pid file (manual flow) | times out cleanly, message carries the real budget |

The pid file is removed on all four, including both failure exits.

`shellcheck -S warning` on all three files reports the same three pre-existing findings as before and nothing in the new one; the `source=` directive means it follows the include rather than emitting SC1091.

juan-malbeclabs:
Both structural suggestions are in, answered on their threads.

- **Step ordering** (`de05b7b86`) — you were right that starting from the built binary moved the start point past the build and put the whole mainnet clone inside the readiness window. Both jobs now build only the loader, start it, then build the remaining CLIs, so the clone overlaps that compile. The workflow header documents the ordering, since recombining the two `cargo build` steps would silently undo it.
- **Shared wait helper** (`a81cf4c99`) — `offchain/sh/lib/wait_for_fork.sh` exposing `wait_for_fork [log_path] [pid_path] [wait_seconds]`; both callers drop from 37 duplicated lines to two. `test_full_debt_flow.sh` is left alone as you said it could be — the parameters are there so it can adopt this later without a signature change.

With the head start restored I left the budget at 180s: the loop exits as soon as the fork answers, so it costs nothing on a passing run. Happy to drop it to 120s if you would still rather.

Description updated to describe the ordering as its own change rather than folding it into the start-from-binary one.

.github/workflows/offchain.local-validator.yml
juan-malbeclabs: Both fork jobs are green on the reorder, and the step timings confirm the mechanism rather than just the outcome:

| | `test-doublezero-solana-fork` | `…-validator-debt-fork-current` |
| --- | --- | --- |
| Loader starts | 17:11:33 | 17:06:22 |
| Remaining CLIs build | 17:11:33 → 17:14:23 | 17:06:22 → 17:10:07 |
| **Head start the clone gets** | **2m50s** | **3m45s** |
| Test step begins | 17:14:23 | 17:10:07 |
| `Solana fork is ready.` | 17:14:23 (**+0.09s**) | 17:10:07 (**+0.49s**) |

The fork answered on the **first probe** in both jobs, so the readiness window consumed ~0s of its budget. That is the overlap you described being left on the table — 3m45s of it in the debt job, as predicted from it building three more CLIs.

It also settles the 120s-vs-180s question in a more satisfying way than my "costs nothing" argument did: on a passing run the budget is not drawn down at all, so the number only ever matters on a failing one. Still happy to take it back to 120s if you prefer — just say so.

One caveat on how much this proves: the PR leaves `Cargo.lock` untouched, so these ran with a warm cache (the loader built in 1m23s / 1m43s). It exercises the new ordering end-to-end but not the cold-cache margin the original race depended on.

PR #4350, opened 2026-09-16
solana: a CLI for builder-stake, so a bond can be posted

bgm-malbeclabs:
## Summary

- **Nothing outside the `builder-stake` program could call it.** `initialize_builder_stake`, `post_bond`, `withdraw` and the admin instructions were reachable only from the program's own tests, so posting a bond meant writing Rust. This adds `offchain/crates/solana-admin-cli/builder-stake`, a fourth sibling to `passport`, `revenue-distribution` and `sol-conversion`.
- `initialize`, `post-bond`, `withdraw` and `show` for a builder. `initialize-program`, `set-admin`, `set-tier-parameters`, `show-config` and `set-paused` for an admin.
- **`initialize-program` sends `SetAdmin` in the same transaction.** `InitializeProgram` sets only `bump_seed` and pauses the program, leaving `admin_key` zero, and `ProgramConfig::try_require_admin` refuses a zero admin. A deployment initialized on its own could never be configured, and there was no recovery path. `set-admin` exists for the same reason.
- **Amounts are taken in 2Z and scaled by `DOUBLEZERO_MINT_DECIMALS`**, and printed back in both. Taking base units on the command line makes a bond a hundred million times too small a typo nobody notices, because the coverage check refuses it with an error that never names the amount.
- **`post-bond` and `withdraw` refuse before sending what the program would refuse anyway**, naming which reason applies. The token program rejects a short balance with an error naming neither the account nor the amount, and `withdraw` has two distinct reasons the program reports as one.

491 lines of non-test code.

## The 2Z mint

Compiled in, feature-gated the way the program is, so `--features development` talks to the devnet and localnet deployments. There is deliberately no `--mint-2z` flag: `post_bond` carries no mint, the stake's token account is created on the compiled one, and a flag whose only legal value is a constant is a flag that can only produce an error.

## Testing Verification

**Verified end to end against live Solana devnet**, which is where the token programs live for both DZ environments. A real bond, not a simulation:

```
$ doublezero-builder-stake-admin initialize --committed-rate-bits-per-sec 1000000000
Builder stake: CFaimJrTH7Eew3W675shU2tUnY772ny9BV9wvC5MPTi2
$ doublezero-builder-stake-admin post-bond --amount 1
Paying 1 2Z (100000000 base units) from FQFG1yb6EKC7StpFf6ZCPiPBnvTkYdk7JKio462TQy2y
$ doublezero-builder-stake-admin show
bonded       : 1.00000000 2Z (100000000 base units)
required     : 1.00000000 2Z (100000000 base units)
covers       : true
```

`show-config` read the live tier table and matched the values decoded by hand from the raw account: 1, 2 and 5 2Z.

**Both new guards were driven to their refusals**, not just written:

```
$ ... withdraw --amount 1
Error: nothing is withdrawable yet: the hold on CFaim... runs to 1805120566
$ ... post-bond --amount 99
Error: FQFG1... holds 4 2Z and the bond is 99 2Z
```

**Six unit tests over `to_base_units`**, which is the one piece of real logic. The `checked_add` test was verified to bite: replacing `checked_add` with `+` fails `test_the_fraction_cannot_push_the_sum_past_u64` and nothing else. `u64::MAX` is exactly `184467440737.09551615` 2Z at eight decimals, so the boundary is not a round number and the test says why.

The three admin verbs that need the program's upgrade authority are not exercised against devnet, because the deployment there is already initialized and configured. `initialize-program` is the one that changed most in review and is the least covered; it is worth running on a fresh cluster before it is trusted.

## Not in this change

- **`hold_expires_at` is documented as always zero.** `BuilderStake`'s comment says "Always zero for now: nothing writes it. The hold and `Withdraw` are B3." The live stake came back with `1805120566`, so something does write it and that comment is stale. Not touched here.
- Nothing that moves bonds outside the program's rules. `admin_key` configures the program and cannot reach the vault.


bgm-malbeclabs:
Closing: this duplicates tooling that already exists, and it is in the wrong repo.

An admin CLI for `builder-stake` already exists with `initialize`, `set-admin`, `set-tiers`, `pause`, `unpause` and `show`. I searched this repository for callers of `post_bond`, found none, and concluded there was no operator tooling without checking the other repository the work spans.

What is genuinely new here is the builder-facing half, `post-bond`, `withdraw` and a per-stake `show`, plus taking amounts in 2Z rather than base units and pairing `initialize` with `set-admin`. Those are being ported into the existing crate instead, where they stay private until the audit gate.

Nothing here is lost: the branch stands if any of it is wanted later.

Issue #4337, opened 2026-09-15
internet-latency-collector: lossy RIPE Atlas circuits get their sample timestamps compressed onchain


nikw9944:
## Summary

The RIPE Atlas exchange export path records only results that carried an RTT. The telemetry program stores samples append-only at `next_sample_index` and reconstructs each sample's time as `start_timestamp + index × sampling_interval`. A missed ping therefore consumes no slot, and a lossy circuit's timeline is compressed by its loss ratio. On mainnet-beta, Columbus (`cmh`) circuits at ~17% success have their RTTs recorded up to ~30 hours earlier than they happened.

## Evidence (mainnet-beta, 2026-09-15)

- `cmh → dub`, epoch 218, from `fact_dz_internet_metro_latency`: `event_ts` advances by exactly 10 minutes per `sample_index` (14:10:42, 14:20:42, 14:30:42 …) while `ingested_at` for those same 40 samples runs from 2026-09-12 14:27 to 2026-09-14 05:24. Sample index 39 is recorded at 2026-09-12 20:40 and was ingested at 2026-09-14 05:24 — it happened roughly 32 hours after its recorded time.
- Per epoch (212 through 219), cmh-as-origin ripeatlas samples span roughly 14 hours of `event_ts` against roughly 44 hours of real time — a ~3:1 compression that matches the measurement's success rate in the collector's own export logs (`valid_latencies` / `raw_results`, steady at 13–26% per hour).
- Ingest lag `ingested_at − event_ts` over 2026-09-13/14: cmh-as-origin ripeatlas p50 828 min, p90 1345 min, max 1954 min. Every other metro: p50 3–14 min. Control: **wheresitup rows for the same cmh circuits lag 56 min at p50** — same metro, same table, same indexer, so this is the ripeatlas collector's encoding, not the pipeline.
- The lag is uniform across all 27 cmh circuits regardless of their sample count, confirming it is the write encoding rather than anything per-circuit downstream.

## Mechanism

- `controlplane/internet-latency-collector/internal/ripeatlas/collector.go`, `exportSingleMeasurementResults`: records are appended only `if latency > 0`; a result whose ping array has no `rtt` produces no record.
- `controlplane/internet-latency-collector/internal/exporter/submitter.go`, `SubmitSamples`: each batch is written as `Samples []uint32` with `StartTimestampMicroseconds = min(batch timestamps)`; there are no per-sample timestamps.
[...]

PR #4351, opened 2026-09-17
internet-latency-collector: drop expired wheresitup jobs from polling and send a one-hour expiry

nikw9944:
## Problem

WheresItUp discards a job's results at `expire_after` and thereafter returns the job with an empty `response.complete` and a populated `response.in_progress`. In the fields this collector reads, an expired job is a running one, so it was polled every cycle and never yielded anything.

## Why

WheresitUp had a service outage that caused jobs to be stuck in an in_progress state indefinitely. This caused the "jobs to be collected" list to grow, and the polling pass grew along with it, and the cycle stretched past the ten-minute expiry, and every new job was first polled after it had already expired, read as in progress, and was retained in turn — a loop that sustained itself on its own polling cost. 

## Fix

- `ExportJobResults` drops tracked jobs past the cutoff instead of polling them, and is the only place expiry happens: `State.Save()` deliberately does not prune, since creation saves a batch shortly before each export pass and a prune there would evict jobs before the pass could count them.
- One `JobExpireAfter = time.Hour` constant feeds the job payload, the creation log and the state-file cutoff, which previously disagreed with each other. The steady-state poll set is `ceil(MaxJobAge/interval)` batches, about 11 at the 6-minute cadence.
- Missing samples are counted on every cycle, as `ripeatlas` already does, and dropped jobs increment `WheresitupExpiredJobsTotal`: an expired job leaves the pending gauge, so a total stall read as a falling pending count.
- Every completed job is still exported, sorted oldest-first before the write. Each one stands for the interval it was created in, and the ledger derives a sample's timestamp from its position in the account.

## Testing Verification

- Jobs aged 1 min, 30 min, 90 min and 3 h: the two inside the window are polled newest-first and the completed one is exported, the two past it never reach `GetJobResults`, and the circuit that produced nothing is still reported missing.
- A job saved by `AddJobIDsWithCircuits` while already past the cutoff survives that save and is then dropped and counted by the export pass.
- A cycle where every tracked job has expired reports owed circuits missing.
- Three completed jobs for one circuit are all exported and reach the exporter in ascending timestamp order.
- A cancelled context stops the pass instead of walking the remaining jobs issuing calls that can only fail.
- Each test fails when its fix is reverted.


controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -18,9 +18,31 @@ import (
 const (
 	RequestTimeout         = 30 * time.Second // Timeout for job requests
 	CreditWarningThreshold = 10000
-	ExpireAfter            = "10 minutes" // Relative time like "1 hour" - https://wheresitup.com/docs/?shell#creating-jobs
+
+	// JobExpireAfter is how long WheresItUp is asked to retain a job's results. Past it the
+	// API discards them and reports the job with an empty "complete" and a populated
+	// "in_progress", indistinguishable from a job still running, so this has to comfortably
+	// exceed one collection cycle.
+	JobExpireAfter = time.Hour
```
juan-malbeclabs: **A one-hour retention lets several jobs for the same circuit complete in one export pass, and the downstream sample timestamps are positional.**

A job is removed from tracking as soon as it returns `complete`, so each job yields at most one record — but the number of jobs per circuit that can *simultaneously* still have results is `retention / interval`. Under the old 10-minute `expire_after` that was ~2 (ages 30s and 6m30s); at one hour it is ~10.

The realistic trigger is the scenario this PR describes: the vendor stalls, ~10 batches accumulate reading `in_progress`, then it recovers. The next pass reads all ~10 jobs for circuit C as complete and appends 10 records for it, so `circuitActualSamples[C] = 10` and 10 samples land in one `PartitionKey`. Downstream reconstructs each sample's time positionally:

```go
// controlplane/telemetry/internal/data/internet/latencies.go:215
timestampMicros := startTimestampMicroseconds + uint64(i)*samplingIntervalMicroseconds
```

and `start_timestamp_microseconds` is set exactly once per account (`smartcontract/programs/doublezero-telemetry/src/processors/telemetry/write_internet_latency_samples.rs:123`). Nine extra samples therefore shift every later sample in that account forward by nine 6-minute slots, and `..._samples_per_collection_interval_actual` exceeds `_expected`. The account also fills `MAX_INTERNET_LATENCY_SAMPLES` proportionally sooner.

A second trigger: `WriteRecords` failing at line 620 `return`s before `RemoveJobIDs`, so every completed job stays tracked and is re-read next cycle — previously capped by the 10-minute retention, now re-readable for an hour.

Worth deciding explicitly whether the export should keep at most one sample per circuit per pass (newest job wins, drop the rest).

controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -466,7 +487,24 @@ func (c *Collector) ExportJobResults(ctx context.Context, jobIDsFile string) err
 	if err := state.Load(); err != nil {
 		return err
 	}
-	jobIDs := state.GetJobIDs()
+
+	// Polling an expired job can only ever return in_progress, and the time it costs is what
+	// pushes the next batch past their own expiry, so dropping them stops that loop forming.
+	expiryCutoff := time.Now().Add(-MaxJobAge)
```
juan-malbeclabs: **This partition mostly duplicates a prune that already ran ~30-60s earlier in the same cycle, so `expired_count` will read 0 on most cycles.**

`ExportJobResults` is only called from `Run` (line 803), always straight after `RunJobCreation` ends with `AddJobIDsWithCircuits` -> `Save()`, and `Save()` drops everything older than `MaxJobAge` (state.go:89-97) — silently, with no log or count. Because the two cutoffs are now *identical*, a batch can only be caught here if its age crosses the cutoff inside the gap between that prune and this line: `jobWaitTimeout` (30s by default) plus `getLocationsFunc` + `buildLocationMapping`.

With the default 6-minute interval that gap is decisive, because `MaxJobAge mod interval` = 61 mod 6 = **1 minute**: at export the oldest surviving batch is 60m + gap, so it is partitioned here only when the gap exceeds 60s, and otherwise `Save()` removes it at 66m before export ever sees it. So:

- `expired_count` is 0 whenever the gap is under a minute, and the new diagnostic then cannot confirm the fix works or detect a recurrence;
- either way each batch is still polled once at ~60m30s, i.e. after the vendor discarded its results;
- in the `Save()` case batches leave tracking at 66m with nothing logged or counted.

The load-bearing part of this PR is the `MaxJobAge` 2h -> 1h1m constant change. This loop is also skipped on the `WriteRecords` error path (line 622 returns before the `RemoveJobIDs` at 663), and on the paths where `RunJobCreation` returns nil without saving (`len(locations)==0`, all-empty job IDs) it is the only thing that drops anything.

Suggest making `State` the single owner of expiry (e.g. `PruneExpired(now) []string` returning what it dropped), logging and counting at that one site, so the count is emitted whichever of the two paths evicts the job.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -477,7 +515,7 @@ func (c *Collector) ExportJobResults(ctx context.Context, jobIDsFile string) err
 		slog.Int("circuit_count", len(circuitExpectedSamples)),
 		slog.Any("circuits", state.Circuits))
 
-	if len(jobIDs) == 0 {
+	if len(jobIDs) == 0 && len(expiredJobIDs) == 0 {
```
juan-malbeclabs, line 513: **This makes the all-expired cycle a first-class path, but the missing-samples accounting below is still gated on `len(records) > 0`, so a cycle that produces nothing reports zero missing.**

`LatencySamplesPerCollectionIntervalExpected` is incremented per circuit at job creation (line 260) unconditionally, but `LatencySamplesPerCollectionIntervalMissing` only increments inside the `if len(records) > 0` block (line 645). In a cycle where every job is expired or `in_progress`, `records` is empty: `_expected += N`, `_actual += 0`, `_missing += 0`. Counters never decrease, so the `expected ≈ actual + missing` invariant breaks permanently — over the described 22-hour / 34-cycle outage by tens of thousands of counts.

This is the detection hole that let the incident run 22 hours, and it is the one thing that would have paged in hour one. The sibling provider already does it right — `ripeatlas/collector.go:529-538` runs the identical loop unconditionally, outside any record-count guard.

```suggestion
	if len(jobIDs) == 0 && len(expiredJobIDs) == 0 {
		c.log.Info("No tracked jobs found to export")
		return nil
	}
```
(leaving the early return as-is is fine; the change that matters is dedenting the missing-samples loop at 643-656 out of the `len(records) > 0` guard so it matches ripeatlas.)

juan-malbeclabs (commented):
Review pass over the full changed files. 15 comments, most-significant first: the positional-timestamp/epoch consequences of raising retention from 10 minutes to an hour, the new drop-expired path largely duplicating the prune `State.Save()` already performs at the same cutoff (so `expired_count` will read 0 on most cycles), and the missing-samples metric still not firing on a zero-sample cycle.

Verified locally: `go build ./controlplane/internet-latency-collector/...` and `go test ./controlplane/internet-latency-collector/internal/wheresitup/...` both pass, and both new tests genuinely fail when `JobExpireAfter` is widened (checked with a `-overlay` copy setting it to 4h), so neither passes vacuously.

Nothing here is a hard blocker — the findings are design and observability questions about the retention increase, not build or test breaks.

Two comments were corrected after posting: the reachability claim on the partition loop (it is gap-dependent, not unreachable) and the claim that the all-expired path destroys the expected-circuit baseline (it does not — `RunJobCreation` rewrites `Circuits` before every export).

controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -18,9 +18,31 @@ import (
 const (
 	RequestTimeout         = 30 * time.Second // Timeout for job requests
 	CreditWarningThreshold = 10000
-	ExpireAfter            = "10 minutes" // Relative time like "1 hour" - https://wheresitup.com/docs/?shell#creating-jobs
+
+	// JobExpireAfter is how long WheresItUp is asked to retain a job's results. Past it the
+	// API discards them and reports the job with an empty "complete" and a populated
+	// "in_progress", indistinguishable from a job still running, so this has to comfortably
```
juan-malbeclabs: **A six-fold longer retention also raises the maximum staleness of an exported record from ~6.5 minutes to ~60 minutes, which feeds epoch selection and the account's base timestamp.**

`record.Timestamp` is the job's start time (`time.Unix(results.Request.StartTime, 0)`, line 607), and the exporter derives the partition epoch from it:

```go
// internal/exporter/ledger.go:163
epoch, err := e.cfg.EpochFinder.ApproximateAtTime(ctx, record.Timestamp)
```

Two consequences of a record now being readable for an hour rather than ten minutes:

1. If an epoch boundary fell inside that window, the sample is written into the *previous* epoch's account. `InitializeInternetLatencySamples`/`WriteInternetLatencySamples` have no current-epoch check, so an account can be re-opened or created for an epoch that `offchain/crates/contributor-rewards` may already have processed.
2. `StartTimestampMicroseconds` is the batch minimum (`submitter.go:124-137`) and the program keeps the first value it ever sees (`write_internet_latency_samples.rs:123`). One stale sample in the first batch written to a fresh epoch account sets the base up to an hour early, so every reconstructed timestamp in that account is wrong for the rest of the epoch.

Bounding how stale a record may be before it is exported (rather than only how long a job may be tracked) would close both.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -618,19 +656,25 @@ func (c *Collector) ExportJobResults(ctx context.Context, jobIDsFile string) err
...
+				slog.Int("job_count", len(removedJobIDs)),
 				slog.String("error", err.Error()))
 		}
 	}
 
 	// Calculate failure rate and log appropriately
 	totalJobs := len(jobIDs)
 	pendingJobs := totalJobs - len(completedJobIDs) - failedCount
-	failureRate := float64(failedCount) / float64(totalJobs)
+	var failureRate float64
```
juan-malbeclabs, line 689: **Excluding expired jobs from `totalJobs` (line 672) removes the only metric that visibly moved during the incident, and nothing replaces it.**

Before: a stuck job stayed in `jobIDs`, was polled, hit the `len(Complete)==0 && len(InProgress)>0` branch, and so landed in neither `completedJobIDs` nor `failedCount` — it counted in `totalJobs` and therefore in `pendingJobs`, so `doublezero_internet_latency_collector_wheresitup_pending_jobs` climbed into the thousands. After: those jobs are removed before `totalJobs` is computed, so the gauge is capped at roughly one cycle's in-flight set by construction and *falls* as data is discarded uncollected.

An expired job that never yielded a sample is lost data, and it now leaves no metric behind at all — `internal/metrics/metrics.go` has only `WheresitupJobCreationRuns/Failures`, `CreditBalance`, `PendingJobs` and `APIResponseDuration`; `expired_count` exists solely as an slog attribute. Re-run the incident and Prometheus sees `pending_jobs ≈ N` (healthy-looking), `_actual` flat, `_missing` flat (see the comment on line 518), and nothing alertable.

A `WheresitupExpiredJobsTotal` counter incremented alongside the gauge write would cost two lines and mirrors `RipeatlasSourcesWithoutSamples` (metrics.go:112-115), whose doc comment already states the principle.

Minor, same block: with the new `if totalJobs > 0` guard, the `&& totalJobs > 0` in the `failureRate > 0.10` condition below is now dead.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -18,9 +18,31 @@ import (
...
+	// certainly gone.
+	ExpiryGrace = time.Minute
 )
 
+// expireAfterParam renders a duration as the relative time string the job API accepts,
+// e.g. "1 hour" - https://wheresitup.com/docs/?shell#creating-jobs
+func expireAfterParam(d time.Duration) string {
+	unit, n := "hour", int64(d/time.Hour)
+	if d%time.Hour != 0 {
+		unit, n = "minute", int64(d/time.Minute)
```
juan-malbeclabs: **`expireAfterParam` truncates toward zero, so several inputs render as strings the vendor would reject or read as an immediate expiry — and the new table test pins none of them.**

Running the function verbatim:

| input | output |
|---|---|
| `0` | `"0 hours"` |
| `30 * time.Second` | `"0 minutes"` |
| `90 * time.Second` | `"1 minute"` (30s dropped) |
| `time.Hour + 30*time.Second` | `"60 minutes"` (unit flips, 30s dropped) |
| `-30 * time.Second` | `"0 minutes"` (sign lost) |
| `-time.Hour` | `"-1 hours"` |

`JobExpireAfter` is the only value any caller passes, so the whole range beyond `"1 hour"` is unexercised in production and unverified against the API — and `TestInternetLatency_Wheresitup_ExpireAfterParam` pins `{time.Minute, "1 minute"}` and `{90 * time.Minute, "90 minutes"}`, forms the docs linked in the comment above never show, while omitting every input that renders wrongly. A future author reads that table as the specification, sets a sub-minute or non-minute-aligned value, the table still passes, and `expire_after: "0 minutes"` ships — which either fails job creation or is read as immediate expiry, i.e. exactly the silent no-samples mode this PR exists to fix.

Since `JobExpireAfter` is a `const`, only one string can ever be produced. A pair of adjacent constants gets the same drift protection with no branch that cannot run:

```suggestion
// JobExpireAfter is how long WheresItUp is asked to retain a job's results, and
// jobExpireAfterParam is the same value in the relative-time form the job API accepts
// (https://wheresitup.com/docs/?shell#creating-jobs). Keep the two in sync.
JobExpireAfter      = time.Hour
jobExpireAfterParam = "1 hour"
```

That also hoists 435 `fmt.Sprintf` calls per cycle out of the O(n²) creation loop at line 328, and makes the log-vs-payload agreement unconditional rather than test-asserted.

controlplane/internet-latency-collector/internal/wheresitup/state.go
```diff
@@ -11,7 +11,9 @@ import (
 	"github.com/malbeclabs/doublezero/controlplane/internet-latency-collector/internal/collector"
 )
 
-const MaxJobAge = 2 * time.Hour
+// MaxJobAge bounds how long a job stays in the state file. It matches ExportJobResults'
+// poll cutoff on purpose: an independent value here could never fire.
+const MaxJobAge = JobExpireAfter + ExpiryGrace
```
juan-malbeclabs, line 18: **Welding the poll cutoff to the vendor-retention knob collapses two settings that want opposite directions, and the quantity that actually decides when a job first gets polled is the sampling interval.**

`MaxJobAge` has two jobs here: *don't leak unbounded state* and *don't waste a poll*. Neither is a function of how long the vendor keeps results:

- Lower `JobExpireAfter` to 10 minutes to save credits and the state file starts discarding jobs at 11 minutes — under the 6-minute interval a job can be pruned before the cycle that created it finished `buildLocationMapping`, yielding zero samples with no error anywhere.
- Raise it to 2 hours to widen the safety margin and the poll ceiling silently doubles from `ceil(61/6) × 435` = 4,785 to `ceil(121/6) × 435` = 8,700 jobs per pass — back above the pre-PR number.

Relatedly, the bound is not what the PR claims. With `defaultWheresitupSamplingInterval = 6m` (cmd/collector/main.go:36) and 61-minute retention, the stuck-vendor steady state holds `ceil(61/6)` ≈ **10-11 batches**, not "roughly one cycle's worth of jobs" (CHANGELOG.md:12 and the PR body). At mainnet-beta's 30 exchanges that is 435 jobs/cycle × 11 = ~4,785 polled serially — a 45% cut from the old `ceil(120/6)` = 20 batches, not a one-cycle bound. The comment's "an independent value here could never fire" is true only *because* the cutoff was set equal to the expiry.

`Run` already receives `interval` and `collector.Config.WheresitupSamplingInterval` already holds it. Deriving the cutoff from the cadence — e.g. `maxJobAge = 2 * interval` — keeps the ceiling at ~870 jobs whatever expiry is later chosen, fires on every pathological cycle, and leaves `JobExpireAfter` free to be generous. Nothing currently enforces the doc comment's requirement that the expiry "comfortably exceed one collection cycle" against that configurable interval.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -466,7 +487,24 @@ func (c *Collector) ExportJobResults(ctx context.Context, jobIDsFile string) err
...
 	}
-	jobIDs := state.GetJobIDs()
+
+	// Polling an expired job can only ever return in_progress, and the time it costs is what
+	// pushes the next batch past their own expiry, so dropping them stops that loop forming.
+	expiryCutoff := time.Now().Add(-MaxJobAge)
+	var jobIDs []string
+	var expiredJobIDs []string
+	for _, job := range state.Jobs {
+		if job.CreatedAt.Before(expiryCutoff) {
```
juan-malbeclabs: **The premise that an expired job is indistinguishable from a running one is not actually true — the vendor's authoritative expiry is already decoded in three places and thrown away.**

- `JobResultResponse.Request.Expiry.Sec` (client.go:137) arrives in the *same* payload as `Response.InProgress`, so the poll result itself answers "expired or in progress". It is decoded and never read — only `Request.StartTime`, at line 607.
- `JobResponse.Expires` (client.go:60) comes back at creation and is only debug-logged (line 372).
- `JobDetails.Expiry.Sec` from the jobs listing is already rendered by `ListJobs` (line 415), and `GetAllJobs` is already on `clientInterface` — one call per cycle yields every tracked job's true expiry.

`expire_after` is a request, not a guarantee. If WheresItUp clamps it (a plan retention cap), rejects the exact string form, or renames the parameter, the payload keeps saying `1 hour` while results are discarded at ten minutes — and the collector would again poll every job for a full hour past that point with no signal, because the one field that would reveal it is discarded. This PR's own changelog documents that request and reality already diverged unnoticed once.

Storing the vendor's `Expires` in `JobEntry` at creation (or gating on `Request.Expiry.Sec` in the poll response) makes the test definitive and retires the local clock, `ExpiryGrace`, and the coupling between the payload string and the state-file cutoff. At minimum, warn when the granted expiry is materially shorter than `JobExpireAfter`.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -466,7 +487,24 @@ func (c *Collector) ExportJobResults(ctx context.Context, jobIDsFile string) err
...
+			expiredJobIDs = append(expiredJobIDs, job.JobID)
+			continue
+		}
+		jobIDs = append(jobIDs, job.JobID)
```
juan-malbeclabs: **`CreatedAt` is one `time.Now()` stamped after the whole creation pass finishes, so the first jobs of a batch have their age understated by the pass duration — which is larger than `ExpiryGrace`.**

`CreateJobsBetweenLocations` issues one API call per pair with a 100ms `CallDelay` between them (line 386), and `AddJobIDsWithCircuits` then takes a single `now` for the entire batch (state.go:139-145). At mainnet-beta's 30 exchanges that is 435 calls → ≥45s of sleeps alone, plus 435 round trips (~90s at 200ms), and far more when the vendor is degraded — i.e. precisely when this bound matters.

The vendor's `expire_after` clock starts at its own creation time, so the job created first is polled until a *true* age of `1h1m + pass duration`, and every poll in that overrun returns the empty-`complete`/populated-`in_progress` shape this PR is trying to stop paying for. `ExpiryGrace = time.Minute` documents itself as the whole slack ("a job is dropped only once its results are certainly gone") when the batch skew already provides several times that — and there is no protection in the other direction if a batch is ever created faster than the grace.

Stamping `CreatedAt` per job as each response comes back (or storing the vendor's `Expires`, per the comment on line 497) removes the skew and the need to calibrate a grace against an unmeasured pass duration.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -466,7 +487,24 @@ func (c *Collector) ExportJobResults(ctx context.Context, jobIDsFile string) err
...
+	}
```
juan-malbeclabs: **The mechanism that actually drove the incident is untouched: the export pass has no wall-clock bound, and this partition preserves append order so the newest batch is polled last.**

`state.Jobs` is append-ordered (state.go:126, 141) and the loop above preserves it, so `jobIDs` is oldest-first — at a multi-batch steady state the stale `in_progress` entries are polled *before* the newest batch, the only one whose results are certainly still within the vendor retention. Any truncation of the pass (SIGTERM, an OOM restart, a future deadline) therefore sheds exactly the polls that would have produced samples. Iterating in reverse is one line and makes a slow pass degrade stale jobs instead of fresh ones.

There is also no deadline. `Run` detects the overrun and only warns (line 816), and `time.Ticker` buffers one tick, so the next cycle starts immediately and the 6-minute cadence silently becomes however long a pass takes. Solving `T = fixed + jobs_per_cycle × p` for `T = 3600s` gives `p ≈ 8s` — once the vendor answers `GetJobResults` in about eight seconds (a quarter of the 30s HTTP timeout at client.go:172), the cycle reaches an hour and every newly created job is again first polled after its own expiry, reads `in_progress`, and is dropped having produced nothing. Same loop, now triggered by vendor slowness rather than vendor `in_progress`.

`Run` already holds `interval` and `c.jobWaitTimeout`; `context.WithTimeout(ctx, interval - c.jobWaitTimeout)` passed into `ExportJobResults` plus an `if ctx.Err() != nil { break }` at the top of the poll loop gives a bound independent of vendor behaviour — and also fixes the shutdown case, where the loop currently walks every remaining entry issuing immediately-failing calls and then logs `failure_rate 1.0` at Error level.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -466,7 +487,24 @@ func (c *Collector) ExportJobResults(ctx context.Context, jobIDsFile string) err
 	if err := state.Load(); err != nil {
 		return err
 	}
-	jobIDs := state.GetJobIDs()
+
+	// Polling an expired job can only ever return in_progress, and the time it costs is what
```
juan-malbeclabs: **Both of the calls at the top of this function (lines 479-484) are unconditional and duplicate work `RunJobCreation` did seconds earlier in the same cycle — and on the nothing-to-poll and new all-expired paths every byte of the result is discarded.**

Per export pass, before any decision about whether there is anything to do:

- `c.getLocationsFunc(ctx)` → `collector.GetLocations` → a serviceability `GetProgramAccounts` scan over the whole program, borsh-deserialising thousands of accounts to read ~30 exchange codes. `Run` already fetched exactly this at line 789 and passed it to `RunJobCreation`.
- `c.buildLocationMapping` → `GetNearestSourcesForLocations` → `GetNearestSources` per location → **`GetAllSources` per location, with no cache** (client.go:246), i.e. N identical `GET /sources` of the global source list plus N × 100ms of `CallDelay` sleeps (client.go:307). `RunJobCreation` made the same call at line 173.

At 30 locations that is 60 full source-list fetches and ~6s of pure sleep per cycle — 14,400 fetches and 24 minutes of sleeping per day at the 6-minute cadence — on the critical path of the very cycle-duration budget whose overrun produced this outage. `locationMap` is consumed only at line 578, so on the line-520 return and on the all-expired fall-through it is entirely wasted.

Cheap fixes, in increasing order of value: move lines 479-484 below the line-518 early return; have `Run` pass its already-fetched `locations`/`locationMatches` into `ExportJobResults`; memoize `GetAllSources` on the `Client` with a short TTL, collapsing 60 fetches per cycle to one.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -618,19 +656,25 @@ func (c *Collector) ExportJobResults(ctx context.Context, jobIDsFile string) err
 		}
 	}
 
-	if len(completedJobIDs) > 0 {
-		if err := state.RemoveJobIDs(completedJobIDs); err != nil {
+	removedJobIDs := make([]string, 0, len(completedJobIDs)+len(expiredJobIDs))
+	removedJobIDs = append(removedJobIDs, completedJobIDs...)
+	removedJobIDs = append(removedJobIDs, expiredJobIDs...)
+	if len(removedJobIDs) > 0 {
+		if err := state.RemoveJobIDs(removedJobIDs); err != nil {
```
juan-malbeclabs: **The `expiredJobIDs` half of this argument is redundant with the prune `Save()` already performs, and the warning below now names the wrong set.**

`RemoveJobIDs` ends in `Save()`, which unconditionally drops every job older than `MaxJobAge` with a cutoff computed marginally *later* than `expiryCutoff` — a strict superset of `expiredJobIDs`. Enumerating those IDs into the remove set buys nothing except forcing the write on cycles where `completedJobIDs` is empty, while making `expired_count` and `removed_job_count` count the same entries twice and leaving a reader to work out which of two mechanisms actually evicts an expired job.

Note this call is also skipped entirely when `WriteRecords` fails (line 622 returns first), so the drop is not robust to an exporter outage.

Minor, in the lines just above: `slices.Concat(completedJobIDs, expiredJobIDs)` replaces the `make` plus two appends and the hand-computed capacity. And the warn message below still says "failed to remove **completed** job IDs" though the set now includes expired ones.

(An earlier version of this comment claimed the all-expired path destroys the expected-circuit baseline via the `len(Jobs) == 0` -> `Circuits = nil` branch at state.go:170. That is wrong: `Run` always calls `RunJobCreation` -> `AddJobIDsWithCircuits` before each export, which rewrites `Circuits` from scratch, so the nil never survives into the next pass.)

controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -466,7 +487,24 @@ func (c *Collector) ExportJobResults(ctx context.Context, jobIDsFile string) err
...
+		if job.CreatedAt.Before(expiryCutoff) {
+			expiredJobIDs = append(expiredJobIDs, job.JobID)
+			continue
+		}
+		jobIDs = append(jobIDs, job.JobID)
+	}
+	if len(expiredJobIDs) > 0 {
+		c.log.Info("Wheresitup - Dropping expired jobs from tracking without polling",
+			slog.Int("expired_count", len(expiredJobIDs)),
+			slog.Duration("expire_after", JobExpireAfter))
```
juan-malbeclabs: **This line reports the wrong cutoff, and emits `expire_after` with a different type than the creation log does.**

(a) The jobs were dropped because `job.CreatedAt.Before(time.Now().Add(-MaxJobAge))`, i.e. at **1h1m**, but the line says `expire_after=1h`. An operator triaging the next recurrence reads the wrong threshold off the one line that exists for that triage, and the grace is invisible. Log `MaxJobAge` — the cutoff genuinely applied.

(b) `internal/collector/logging.go` uses `slog.NewJSONHandler` for every non-debug level, and `slog`'s JSON handler encodes `KindDuration` as an integer nanosecond count. So this line emits `"expire_after":3600000000000` while line 206 emits `"expire_after":"1 hour"` from `slog.String`. Same service, same key, number vs string — a log index that types fields by key gets a mapping conflict and drops or rejects one of the two lines, and any dashboard filtering `expire_after="1 hour"` never matches this one.

```suggestion
		c.log.Info("Wheresitup - Dropping expired jobs from tracking without polling",
			slog.Int("expired_count", len(expiredJobIDs)),
			slog.Duration("max_job_age", MaxJobAge))
```

controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -466,7 +487,24 @@ func (c *Collector) ExportJobResults(ctx context.Context, jobIDsFile string) err
...
 		return err
 	}
-	jobIDs := state.GetJobIDs()
+
+	// Polling an expired job can only ever return in_progress, and the time it costs is what
+	// pushes the next batch past their own expiry, so dropping them stops that loop forming.
+	expiryCutoff := time.Now().Add(-MaxJobAge)
+	var jobIDs []string
+	var expiredJobIDs []string
+	for _, job := range state.Jobs {
```
juan-malbeclabs: **This loop re-implements the cutoff filter `State.Save()` already owns, with an inverted boundary predicate and a second clock reading — and it retires the last production caller of `State.GetJobIDs()`.**

`Save()` computes the same `time.Now().Add(-MaxJobAge)` and keeps `job.CreatedAt.After(cutoffTime)` (state.go:89-96); this keeps `!job.CreatedAt.Before(expiryCutoff)`. A job sitting exactly on the cutoff is polled here but dropped there, so the two copies are already inconsistent at the boundary and will drift further whenever `MaxJobAge` semantics change (e.g. pruning by a stored `ExpiresAt` instead of `CreatedAt` would have to be made in both places).

More reachably, the two cutoffs come from different `time.Now()` calls minutes apart: `expiryCutoff` here, and `Save()`'s inside `RemoveJobIDs` after the whole poll pass. A job aged `MaxJobAge - 10s` at this point is kept, polled, found in-progress, counted in `pendingJobs` and published to the `WheresitupPendingJobs` gauge — then silently deleted by `Save()` at the end of the same call, appearing in neither `expired_count` nor `removed_job_count`.

The sibling package already puts this on the state type: `MeasurementState.PruneExpiredUnresponsiveProbes()` (ripeatlas/state.go:274) returns the pruned count and owns the single expiry predicate. A `State.PartitionByAge(maxAge) (active, expired []string)` next to `GetJobIDs()` — reused by `Save()` — would give one predicate, one clock read, and one place to log.

`State.GetJobIDs()` is now called only from `_test.go` files; worth deleting or unexporting it rather than leaving it looking like live API.

controlplane/internet-latency-collector/internal/wheresitup/collector_test.go
```diff
@@ -1187,3 +1188,173 @@ func TestInitializeCreditBalance(t *testing.T) {
...
+	var polled []string
+	c := &Collector{
+		client: &MockWheresitupClient{
+			GetJobResultsFunc: func(ctx context.Context, jobID string) (*JobResultResponse, error) {
+				mu.Lock()
+				polled = append(polled, jobID)
+				mu.Unlock()
+				// Still running: no completed results yet, so the job stays tracked.
+				results := &JobResultResponse{}
+				results.Response.InProgress = []any{"in progress"}
```
juan-malbeclabs: **A few things about the new test that make it weaker than it reads:**

1. **`exporter` is left nil.** Every other `Collector` literal in this file that drives `ExportJobResults` (lines 437, 941, 1060) sets it. This one passes only because the mock always returns in-progress, so `len(records) > 0` is never true. The obvious next extension — have one job return a completed result so the test also asserts a live job *is* exported — nil-panics inside the production function rather than failing. `env` is likewise `""`, which would register an empty-label series on `LatencySamplesPerCollectionIntervalMissing`.

2. **The mock's response shape cannot occur in production.** `Complete` is `map[string]ServiceResult` and `InProgress` is `[]any` (client.go:143-144), and the comment at collector.go:538-549 records the vendor's dual encoding: a running job sends `"complete": []` and `"in_progress": {…}`. *Both* fields then fail to unmarshal, so `GetJobResults` returns an error and production classifies in-progress via the string match at line 550. Reaching the line-563 branch requires `complete` to arrive as an object *and* `in_progress` as a non-empty array simultaneously. The expiry assertions still hold (the partition happens before any poll), but the "still running, so the job stays tracked" premise — and the guarantee that a genuinely in-progress job is retained — is never covered against the real shape; the test would pass with the line-550 branch deleted.

3. **`recordingHandler` duplicates `testLogHandler` at line 45** in this same file (a stored `slog.Record` carries both `Message` and `Level`, so it strictly subsumes it), and byte-identical copies already exist at `controlplane/telemetry/internal/telemetry/pinger_test.go:1049` and `controlplane/controller/internal/controller/metrics_test.go:151`. Two handlers in one file with different safety properties means the next test picks whichever it finds first; this is the moment to hoist one shared helper rather than add a fourth copy.

4. **`WithAttrs`/`WithGroup` silently discard.** Every other test here builds its logger as `logger.With("test", t.Name())`; an author copying that gets `attr()` returning `(zero, false)` and a misleading failure. `WithGroup` fails in the passing direction: if the summary were later emitted inside `slog.Group(...)`, the real JSON nests the key and any consumer keyed on a top-level `expired_count` breaks, while this assertion still finds the ungrouped attr and passes.

5. **`attr()` returns the first matching record**, so reusing this handler across two `ExportJobResults` calls (or a two-tick `Run` test, as line 1060 already drives) silently asserts on cycle 1.

6. Minor: `require.Equal(t, expireAfterParam(JobExpireAfter), options["expire_after"])` at line 1339 (and 1343) compares the code's output to the same expression the code used, so it passes for any `JobExpireAfter`. The literal `"1 hour"` pin on line 1338 is the assertion doing the work; asserting the literal directly at the point of observation would be both shorter and stronger.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
juan-malbeclabs: Worth adding to this thread, because the same two calls carry a consequence that this PR's longer retention amplifies: they fail open, and the export then deletes completed jobs whose samples were silently discarded.

`GetLocations` returns `[]LocationMatch{}` rather than an error on any ledger-RPC failure or empty exchange list (`internal/collector/location.go:107-119`), and `RunJobCreation` treats `len(locations) == 0` as success (`collector.go:166-169`), so `Run` still reaches the export. With an empty `locationMap`, every job parses as `Unknown -> Unknown` (`collector.go:747-757`); `Record.Validate()` passes because both codes are non-empty; the exporter warns `Source exchange not found, skipping record` and `continue`s past each one, returning nil. But `completedJobIDs` was already appended at `collector.go:612`, and `removedJobIDs` at `659-663` deletes those jobs unconditionally.

So the samples are gone and not retried. Note the asymmetry: had `WriteRecords` returned an error, the early return at `621` would have preserved the jobs for the next cycle. The fail-open path is strictly worse than the error path.

The retention change is what sets the blast radius. At a 10-minute expiry only the current cycle's jobs could be in flight; at `JobExpireAfter = 1h` with a 6-minute interval, roughly ten cycles of paid jobs can be resident, and one bad ledger fetch during a single export pass consumes all of them. It also leaves a permanent bogus `LatencySamplesPerCollectionIntervalActual{circuit="Unknown -> Unknown"}` series, since the actual-samples map is built from `records` before the exporter drops them.

Guarding `len(locationMap) == 0` with an early return before the poll loop would keep the jobs tracked until the mapping is available again.

controlplane/internet-latency-collector/internal/wheresitup/collector_test.go
```diff
@@ -1187,3 +1188,173 @@ func TestInitializeCreditBalance(t *testing.T) {
...
+			if a.Key == key {
+				value, found = a.Value, true
+				return false
+			}
+			return true
+		})
+		if found {
+			return value, true
+		}
+	}
```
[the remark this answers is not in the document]
nikw9944: Fixed in 74c6ca67. The helper is now `only(t, message)`, which fails if the message is absent *or* logged more than once and returns its position in the stream, so a duplicate line cannot satisfy an assertion. The expiry test pins the order explicitly with `require.Less(t, droppedAt, summaryAt)` rather than a hardcoded index, so a control-flow change that emits the expiry log after the cycle summary fails.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -618,19 +656,25 @@ func (c *Collector) ExportJobResults(ctx context.Context, jobIDsFile string) err
...
 	}
 
-	if len(completedJobIDs) > 0 {
-		if err := state.RemoveJobIDs(completedJobIDs); err != nil {
+	removedJobIDs := make([]string, 0, len(completedJobIDs)+len(expiredJobIDs))
+	removedJobIDs = append(removedJobIDs, completedJobIDs...)
+	removedJobIDs = append(removedJobIDs, expiredJobIDs...)
+	if len(removedJobIDs) > 0 {
+		if err := state.RemoveJobIDs(removedJobIDs); err != nil {
 			c.log.Warn("Wheresitup failed to remove completed job IDs",
```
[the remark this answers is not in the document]
nikw9944: Fixed in 74c6ca67, by making the warning true rather than rewording it. Expired jobs are now pruned and persisted before any polling, so the removal at the end of the pass really does carry only completed jobs. The expiry drop has its own warning ("Wheresitup failed to drop expired job IDs") and its own counter, and it no longer depends on the export pass succeeding.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Fixed in 74c6ca67: the export pass now keeps at most one sample per circuit, newest job wins. Jobs are polled newest-first, so the first record for a circuit is its freshest; later ones are still removed from tracking but not exported, and are counted in a new `superseded_count`. `TestInternetLatency_Wheresitup_ExportJobResults_OneSamplePerCircuitPerPass` covers it and fails if the dedupe is removed.

The second trigger you name is fixed too: the `WriteRecords` early return no longer skips the expiry cleanup, since the drop is persisted before polling starts.

What this does **not** fix is the staleness of the one sample that does get exported — an hour-old record can still select the previous epoch and set that account's base timestamp. That needs a decision about discarding stale samples outright rather than exporting them late, which I would rather not fold into this PR; replied on that in thread 4039417177.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Fixed in 74c6ca67. `State.PruneExpired(now)` is now the only expiry predicate in the package: `Save()` calls it on every write, and `ExportJobResults` calls it to get both the active set and the dropped IDs from one reading of the clock. So there is no longer a gap in which `Save()` can silently evict a job the export pass had counted — whichever call drops it, it is logged and counted at that one site, and `WheresitupExpiredJobsTotal` makes it visible in Prometheus rather than only in an slog attribute.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Fixed in 74c6ca67: the missing-samples loop is dedented out of the `len(records) > 0` guard, matching `ripeatlas/collector.go`. `TestInternetLatency_Wheresitup_ExportJobResults_AllExpiredReportsMissingSamples` asserts a cycle where every job is expired still reports both circuits missing, and it fails if the guard is put back.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Not fixed, deliberately, and worth its own change: bounding how stale a record may be before export means discarding a paid-for sample, and the right threshold (one interval? one epoch?) is a data-quality decision rather than a cleanup to this PR.

What 74c6ca67 does remove is the multiplier. Only one sample per circuit is exported per pass, newest first, so the "nine extra samples shift every later sample" case is gone and `_actual` can no longer exceed `_expected`. A single hour-old record can still land in the previous epoch and set that account's base timestamp, which is the part left open — I will file it as a follow-up unless you would rather it block here.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Both fixed in 74c6ca67. `WheresitupExpiredJobsTotal` is a counter in `internal/metrics`, incremented wherever the drop happens, with a doc comment recording exactly your point: an expired job leaves the pending gauge, so a vendor stall that discards everything reads as a *falling* pending count. Re-run the incident and the expired counter climbs while `_missing` also now moves every cycle (thread 4039417152), so there are two alertable signals instead of none.

The dead `&& totalJobs > 0` is gone, with a one-line comment saying why it was implied.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Taken as suggested in 74c6ca67: `expireAfterParam` is deleted and replaced by the adjacent `JobExpireAfter` / `jobExpireAfterParam` constant pair, so there is no branch that cannot run and no per-job `Sprintf` in the creation loop. The table test is gone with it; the consistency test now asserts the literal `"1 hour"` at both points of observation, the payload and the log line.

controlplane/internet-latency-collector/internal/wheresitup/state.go
nikw9944: The claim is fixed, the derivation is not. The `MaxJobAge` comment and the changelog now say the steady-state poll set is `ceil(MaxJobAge/interval)` batches — about 11 at the 6-minute cadence, against 20 under the previous 2-hour cutoff — instead of "roughly one cycle's worth".

Deriving the cutoff from the interval means threading `interval` into `ExportJobResults` and into `State` (which `Save()` also needs, since it now shares the one predicate), and it changes what the state file means across a config change. That is a better design than a constant and I would rather do it as its own change than bury it here; noting it as a follow-up. `JobExpireAfter` is a `const` today, so the two failure directions you describe both require a code change that this comment now warns against.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Not changed, and the comment that overstated it is fixed: it now says the expired shape is the same as a running job "as far as the fields this collector reads are concerned", which is the accurate claim.

Using the vendor's own expiry properly means storing `Expires` in `JobEntry` — a state-file schema change with a migration path for entries written by the running version — or an extra `GetAllJobs` call per cycle on the pass whose duration caused the incident. Both are more than this PR should carry. The cheap half (warn when the granted expiry is materially shorter than requested) is worth doing on its own and I will file it; it is the check that would have caught the ten-minute/one-hour divergence at the time.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Fixed in 74c6ca67: `AddJobIDsWithCircuits` takes a `createdAt`, and `RunJobCreation` passes `jobCreationStart` — the time the pass began, not the time it ended. Ages are now overstated by at most the pass duration rather than understated by it, so a job is never polled past its real expiry, which is the direction that costs polls. The doc comment on the method records why the argument exists so it does not get "simplified" back to `time.Now()`.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Half fixed in 74c6ca67. The pass now iterates newest batch first, so a truncated pass sheds the stale end rather than the fresh one, and the poll loop checks `ctx.Err()` and stops — which fixes the shutdown case you describe, where it walked every remaining entry issuing calls that could only fail and then logged `failure_rate 1.0` at Error. `TestInternetLatency_Wheresitup_ExportJobResults_StopsOnContextCancellation` covers it.

The `interval - jobWaitTimeout` deadline is not in. It is the right bound, but it needs `interval` plumbed into `ExportJobResults` and a decision about what a deliberately truncated pass should report (a truncated pass is not a failure, and the failure-rate branch would currently call it one). Same follow-up as thread 4039417202, which needs the same plumbing.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Fixed the first of the three in 74c6ca67: `getLocationsFunc` and `buildLocationMapping` moved below the early return *and* behind `len(jobIDs) > 0`, so both the nothing-to-poll and the all-expired paths now do zero work — the mapping only labels records, which is now stated in a comment at the call site.

Passing `Run`'s already-fetched locations in, and memoizing `GetAllSources` on the `Client`, are both worth doing and neither is specific to expiry; leaving them for a follow-up. The memoization is the big one — 60 full source-list fetches per cycle is on the same critical path as the overrun that caused this.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Fixed in 74c6ca67 by restructuring rather than by rewording. The expiry drop is now applied and persisted before polling begins, so: the removal at the end carries only `completedJobIDs` (no double counting, and the warning's wording is correct again), the drop no longer depends on `WriteRecords` succeeding, and `slices.Concat` is not needed because there is nothing to concatenate. Expiry lives in `State.PruneExpired` as the single predicate — see thread 4039417141.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Taken as suggested in 74c6ca67, both parts: the line logs `slog.Duration("max_job_age", MaxJobAge)`, so it reports the cutoff actually applied, and the `expire_after` key is left to the one place that emits it as a string. The test asserts the attribute equals `MaxJobAge`, so the two cannot drift apart again.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Fixed in 74c6ca67: `State.PruneExpired(now)` owns the predicate and `Save()` calls it, so there is one boundary rule (`CreatedAt.After(cutoff)`), one clock read per pass, and one place to log — which also closes the "kept here, silently deleted by `Save()` at the end of the same call" case.

`GetJobIDs()` is staying for now: `state_test.go` exercises it as the public read path alongside `AddJobIDs`, which is in the same position (production-unused) and predates this PR. Deleting both, and having the tests read `state.Jobs`, is a reasonable cleanup but is churn unrelated to expiry.

controlplane/internet-latency-collector/internal/wheresitup/collector_test.go
nikw9944: Points 1-6 addressed in 74c6ca67.

- The collector literal now sets `exporter` and `env`, via a shared `exportTestCollector` helper, and the test actually exports a completed job so the exporter is on the path.
- The in-progress mock returns the production shape: `inProgressResults()` returns the decode error the real client produces when `complete` arrives as an array, so the test now exercises the string-match branch rather than the unreachable one. The premise "still running, so the job stays tracked" is asserted against that shape.
- `recordingHandler` is gone. There is one handler in the file, and it honours `WithAttrs`/`WithGroup` (nesting attrs under open groups as a real handler would) instead of discarding them — every new test builds its logger as `slog.New(handler).With("test", t.Name())` to keep that exercised. I did not hoist a shared helper across `pinger_test.go` and `metrics_test.go`; that is a fair point but touches three packages for no behaviour change, so it wants its own PR.
- `only(t, message)` replaces first-match `attr`: it fails on a duplicate as well as an absent line, and returns the position so order can be asserted.
- The `expire_after` assertions compare against the literal `"1 hour"` at both observation points; `expireAfterParam` itself no longer exists (thread 4039417191).

nikw9944:
Addressed the review in 74c6ca67. Thanks — the two structural findings (the partition duplicating `Save()`'s prune, and missing samples being gated on `len(records) > 0`) were the ones worth restructuring for, and the second was the actual detection hole.

**Changed**

- `State.PruneExpired(now)` is the only expiry predicate in the package. `Save()` calls it, `ExportJobResults` calls it — one boundary rule, one clock read per pass, one place that logs and counts. The drop is persisted before polling, so it no longer depends on the exporter being up, and the removal at the end of the pass carries only completed jobs.
- Missing samples are counted on every cycle, as `ripeatlas` already does. A cycle where every job is expired or in progress now reports what it owed instead of reporting nothing.
- `WheresitupExpiredJobsTotal` counter: an expired job leaves the pending gauge, so a stall that discards everything read as a *falling* pending count with nothing alertable.
- One sample per circuit per pass, newest job wins. Downstream derives each sample's timestamp from its position in the account, so a recovery burst of ~10 completed batches for one circuit would have shifted every later sample in that epoch.
- Jobs are stamped with the start of the creation pass, not its end — a 435-pair pass understated the earliest jobs' age by more than `ExpiryGrace`.
- Polls newest batch first and stops on a cancelled context; the location mapping is built only when there is something to poll; `expireAfterParam` is replaced by the `JobExpireAfter` / `jobExpireAfterParam` constant pair; the drop log reports `max_job_age` (the cutoff actually applied) rather than a `Duration` under the same key a `String` uses elsewhere; the dead `&& totalJobs > 0` is gone.
- Tests: one log handler in the file that honours `WithAttrs`/`WithGroup`, `only()` in place of first-match attribute lookup (fails on duplicates, returns position so order is assertable), the in-progress mock returns the shape production actually sees, and `exporter`/`env` are set. Three new tests cover the per-circuit cap, the all-expired cycle reporting missing samples, and cancellation; each fails when its fix is reverted.
- Changelog: dropped the claim that the old two-hour cutoff could never fire and the "roughly one cycle's worth" bound. It is `ceil(MaxJobAge/interval)` batches, about 11 at the 6-minute cadence, against 20 before.

**Not changed, with reasons in-thread**

- Bounding how stale a record may be before export, which is the epoch-selection and base-timestamp half of the retention increase. Discarding a paid-for sample is a data-quality decision, not a cleanup.
- Deriving the poll cutoff from the sampling interval, and the `interval - jobWaitTimeout` export deadline. Both need `interval` plumbed into `ExportJobResults` and `State`, and the deadline needs a decision about what a deliberately truncated pass reports.
- Storing the vendor's granted `Expires` per job (state-file schema change). The cheap half — warn when the granted expiry is materially shorter than requested — is worth filing on its own; it is the check that would have caught the ten-minute/one-hour divergence.
- Memoizing `GetAllSources`, and passing `Run`'s already-fetched locations into the export. Real wins, unrelated to expiry.
- Hoisting the test log handler shared with `pinger_test.go` and `controller/metrics_test.go`.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Correcting my earlier reply on this thread: it was wrong. `PruneExpired` centralised the predicate but `Save()` still called it and dropped the returned IDs, so creation's save — about 30s before each export pass — evicted a batch that crossed the cutoff between two passes, and the export pass found zero. `expired_count` and the new counter could not fire in practice, which is exactly the hole you described.

Fixed in 9582082d: `Save()` no longer prunes (it still clears `Circuits` when `Jobs` empties), so `ExportJobResults` is the only caller and expiry is counted, logged and persisted in one place before polling. `TestInternetLatency_Wheresitup_ExportJobResults_SaveDoesNotEvictBeforeExpiryIsCounted` saves an already-expired job, asserts it survives the save, and asserts the export pass then drops and counts it; it fails if the prune is put back into `Save()`.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Reverted the per-circuit cap in b909a44d — it was the wrong call, and worse than what it replaced. Under the positional encoding each completed job stands for the interval it was created in, so the N jobs for a circuit that complete in one pass after a stall are the N intervals the stall left empty; writing all of them moves the account's implied clock back toward real time, while keeping only the newest discards N-1 real measurements and leaves the account N-1 slots further behind for the rest of the epoch, on every circuit. That is malbeclabs/doublezero#4337 made worse.

Every completed job is exported again, and `records` is now sorted by timestamp ascending before `WriteRecords`, since polling is newest-first and the exporter preserves insertion order without sorting. `TestInternetLatency_Wheresitup_ExportJobResults_ExportsEveryCompletedJobInTimeOrder` covers it.

On the second trigger you raised: `WriteRecords` failing is a retry rather than a duplicate, because nothing was written — and the expiry cleanup no longer sits behind that early return.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: This one is not addressed, and it should be. The fail-open chain is real: `GetLocations` returns an empty slice on an RPC failure, `RunJobCreation` treats that as success, and an empty `locationMap` then turns every record into `Unknown -> Unknown` — dropped by the exporter, which returns nil, while the jobs were already in `completedJobIDs` and get removed. One bad ledger fetch consumes every resident job, and this PR's retention is what makes that roughly ten cycles' worth rather than one.

Your guard is the right shape: an early return before the poll loop when `len(locationMap) == 0`, leaving the jobs tracked, plus the missing-samples accounting for the cycle so the gap is still visible.

I could not land it this turn. The sandbox lost access to the Go module proxy (`github.com/klauspost/compress@v1.18.6` has only its `.mod` in the local cache, and `storage.googleapis.com` times out), so nothing in this module compiles here — including the pre-rebase commit that CI passed. Rather than push a change I cannot type-check or test, I have left it for the next turn and pushed only the rebase onto main (37afb1bb), whose code diff is byte-identical to 10232d72.

nikw9944:
Rebased onto main (37afb1bb) — the base had moved by one commit (malbeclabs/doublezero#4331, ripeatlas). The rebase was clean; the only overlap was adjacent bullets in the `Unreleased` changelog section. This PR's code diff is byte-identical to the previous head, so nothing in the review above changed under it.

One review item is still open: the fail-open `GetLocations` / empty-`locationMap` path in juan-malbeclabs's follow-up on thread 4039417226. The finding is accepted and the guard he suggests is the right fix; I did not land it because this sandbox lost access to the Go module proxy and nothing in the module compiles here right now, including the commit CI passed earlier. I would rather leave it for a turn that can run `go test` than push an unverified change onto a branch under review. Details in that thread.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Landed in ebcb3aad. The pass now returns before the poll loop when `len(locationMap) == 0`, leaving the jobs tracked, and calls the missing-samples accounting for the cycle first so the gap still shows. It returns an error rather than nil, so the cycle also counts in `CollectionFailuresTotal{provider="wheresitup"}` instead of reading as a clean pass that exported nothing.

`TestInternetLatency_Wheresitup_ExportJobResults_KeepsJobsWhenLocationMappingIsEmpty` asserts the job is never polled, survives in the state file, and that both owed circuits are reported missing; it fails with the guard disabled. One existing sub-test (`JobResults In Progress`) had been relying on an empty mapping to reach the poll loop and now gets a real one.

The module-proxy outage is worked around rather than fixed: `GOPROXY=direct` reaches github.com from here, while `storage.googleapis.com` still times out. Full `go build`, `go vet`, `go test ./controlplane/internet-latency-collector/...` and `golangci-lint` all pass on the rebased branch.

nikw9944:
The last open review item is closed. Head is now ebcb3aad on top of the rebase onto main.

**Fixed** — the fail-open `GetLocations` path from juan-malbeclabs's thread on collector.go:491. `ExportJobResults` stops before the poll loop when the location mapping is empty, leaving the jobs tracked for the next cycle, reports the cycle's missing samples, and returns an error so the cycle counts in `CollectionFailuresTotal` rather than reading as a clean pass. Without it, one failed ledger fetch labelled every record `Unknown`, the exporter dropped them and returned nil, and the jobs were removed as completed — a loss this PR's retention would have multiplied to about ten cycles' worth of jobs.

No other review feedback is outstanding; the earlier threads were answered on the previous heads and this change does not disturb them.

**Verification** — the module-proxy outage that blocked the previous turn is worked around with `GOPROXY=direct` (github.com is reachable from the sandbox; `storage.googleapis.com` still times out). On the rebased branch: `go build`, `go vet`, `go test ./controlplane/internet-latency-collector/...` and `golangci-lint run` all pass, including main's new ripeatlas code. The new test fails with the guard disabled.

controlplane/internet-latency-collector/internal/wheresitup/collector_test.go
```diff
@@ -1187,3 +1261,360 @@ func TestInitializeCreditBalance(t *testing.T) {
 		require.Contains(t, err.Error(), "failed to get Wheresitup credit balance")
 	})
 }
+
+// writeJobState writes a state file directly so tests can control job ages, which
+// State.Save() cannot express (it prunes old entries on write).
```
juan-malbeclabs: Stale after `5ab9ebd`: `Save()` no longer prunes on write — that is the whole point of the ownership move. The helper is still needed, but for a different reason: `AddJobIDs`/`AddJobIDsWithCircuits` stamp `CreatedAt` from a caller-supplied or current clock and there is no API for backdating an arbitrary mix of ages in one file.

Suggest: `// writeJobState writes a state file directly so tests can set arbitrary per-job ages, which the Add* helpers cannot express.`

Leaving it as-is documents the opposite of what the PR establishes, which is the comment most likely to mislead the next reader of this package.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
```diff
@@ -477,20 +482,80 @@ func (c *Collector) ExportJobResults(ctx context.Context, jobIDsFile string) err
...
+	// the next batch past its own expiry, so dropping them stops that loop forming. Persisted
+	// here, not with the completed jobs below, so it survives an early return on write failure.
+	expiredJobIDs := state.PruneExpired(time.Now())
+	if len(expiredJobIDs) > 0 {
+		metrics.WheresitupExpiredJobsTotal.Add(float64(len(expiredJobIDs)))
+		c.log.Info("Wheresitup - Dropping expired jobs from tracking without polling",
+			slog.Int("expired_count", len(expiredJobIDs)),
+			slog.Duration("max_job_age", MaxJobAge))
+		if err := state.Save(); err != nil {
+			c.log.Warn("Wheresitup failed to drop expired job IDs",
```
juan-malbeclabs, line 494: When this save fails the counter has already been incremented, and `RemoveJobIDs` at the end of the pass starts with `Load()` — so it re-reads the unpruned file, drops only the completed jobs, and writes the expired ones back. Next cycle prunes and counts them a second time, so `WheresitupExpiredJobsTotal` over-reports by the number of resident expired jobs for every failed save.

Rare (it takes a write error on the state file) and it self-heals, so not worth restructuring. But since this counter exists precisely to be the quantity an operator trusts during the next stall, consider either incrementing it only on a successful save, or noting the double-count in the counter's doc comment in `metrics.go` alongside the pending-gauge rationale already there.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 ### Changes
 
 - Collector
+  - The Wheresitup collector drops tracked jobs whose results WheresItUp has already discarded instead of polling them, and asks for a one-hour expiry rather than ten minutes. Past `expire_after` the API keeps returning a job with an empty `complete` and a populated `in_progress`, the shape of a job still running, so expired jobs were polled every cycle: a vendor stall stretched the pass until every new job was first polled after its own expiry and was retained in turn, which left mainnet-beta's wheresitup provider reporting no samples for about 22 hours. Missing samples are now counted on every cycle rather than only on cycles that exported a record, which is the signal that was absent throughout. Dropped jobs increment a counter of their own, since an expired job leaves the pending gauge.
```
juan-malbeclabs: Consider a sentence for `ebcb3aad`. It is a separate user-visible change from the expiry work: the export pass now returns an error and increments `CollectionFailuresTotal` when the location mapping comes back empty, where it previously completed silently and consumed the jobs' results. An operator reading this entry to explain a new `wheresitup` failure-rate alert would not find it here.

Something like: "A cycle whose location mapping comes back empty — the shape of a failed ledger fetch — now fails instead of exporting every record as `Unknown` and dropping the jobs, so the samples survive to the next cycle."

juan-malbeclabs (approved):
Approving. Every thread from the previous round is either fixed in the code or answered with a reason I accept, and the two that changed direction after my comment (the per-circuit cap reverted in `b909a44d`, the `Save()`-owns-expiry claim corrected in `5ab9ebd`) both landed on the better answer.

What I checked on this revision:

- **The prune is now the only expiry predicate and it is reachable.** `State.Save()` no longer filters, so the creation-pass save ~30s ahead of the export pass can no longer evict a job before `ExportJobResults` counts it. `PruneExpired` reads the clock once, persists before polling, and `TestInternetLatency_Wheresitup_ExportJobResults_SaveDoesNotEvictBeforeExpiryIsCounted` pins exactly that ordering. `AddJobIDs` is test-only, so dropping the prune from `Save()` leaves no production path that grows the state file unbounded.
- **The ordering fix is real, not decorative.** Removing the `SortStableFunc` fails `ExportsEveryCompletedJobInTimeOrder` on the actual record order (`16:23, 16:17, 16:11`). It matters: `defaultWheresitupSamplingInterval` is what the ledger exporter is handed as this provider's sampling interval, so position *is* the timestamp, and newest-first polling without the sort would have written each catch-up run backwards.
- **The fail-open chain is genuinely closed.** `GetLocations` returns `[]LocationMatch{}` on an RPC error, and `BufferedLedgerExporter.WriteRecords` `continue`s past an unmapped exchange code and returns `nil` — so before `ebcb3aad` one bad ledger fetch really did consume every resident job's results with no error anywhere. Stopping before the poll loop is the right trade at this retention.
- **The retention no longer runs away.** At 435 jobs/batch and a 61-minute cutoff, a full vendor stall settles where `N² · 435 · t_poll ≈ 61min`; at ~200ms/poll that is ~6 batches and a ~9-minute cycle, so a new job is first polled far inside its one-hour expiry. Under the old ten-minute expiry that same equilibrium sat just past the cutoff, which is the loop the incident describes. The fix removes the positive feedback rather than damping it.
- `go test -race -count=2 ./controlplane/internet-latency-collector/...` and `golangci-lint` are clean.

Three nits below, none blocking.

One residual worth knowing but not worth holding this PR for: a mapping that is non-empty but *missing one source* still labels that job `Unknown`, the exporter drops the record and the job is removed as completed. It is now at least visible, because the circuit shows up in the missing-samples counter on the same cycle — which it would not have before this change.

controlplane/internet-latency-collector/internal/wheresitup/collector_test.go
nikw9944: Fixed in 07e5fc29 — the comment now reads:

```go
// writeJobState writes a state file directly so tests can set arbitrary per-job ages, which
// the Add* helpers cannot express: they stamp every job in a call from one clock.
```

You're right that it documented the opposite of what the PR establishes: `Save()` stopped pruning in 5ab9ebd3, and the helper's actual reason is that the `Add*` helpers stamp `CreatedAt` once per call, so a mix of ages in one file can't be built through them.

controlplane/internet-latency-collector/internal/wheresitup/collector.go
nikw9944: Took the first option in 07e5fc29 — the counter now increments only on a successful save:

```go
if err := state.Save(); err != nil {
    c.log.Warn("Wheresitup failed to drop expired job IDs", ...)
} else {
    // Counted only once the prune reaches disk. RemoveJobIDs reloads the file at the
    // end of the pass, so a failed save leaves these jobs tracked to be pruned again
    // next cycle, and counting here too would report them twice.
    metrics.WheresitupExpiredJobsTotal.Add(float64(len(expiredJobIDs)))
}
```

Confirmed the mechanism before changing it: `RemoveJobIDs` opens with `Load()` (state.go:147), so a failed save is fully discarded and the jobs come back on the next cycle to be pruned and counted then — exactly once overall.

No test for it. Forcing a `Save()` failure needs a read-only state file, and `go test` runs as root in the CI container, where 0444 doesn't block the write — the test would pass without exercising the path. The `.json` suffix check isn't a usable seam either, since `Load()` fails on it first and the pass returns before `PruneExpired`.

CHANGELOG.md
nikw9944: Added in 07e5fc29:

> A cycle whose location mapping comes back empty, the shape of a failed ledger fetch, now fails instead of exporting every record as `Unknown` and dropping the jobs, so the results survive to the next cycle.

Agreed it belonged here — it's the sentence an operator would need to explain a new wheresitup failure-rate alert, and it wasn't derivable from the expiry text.

nikw9944:
Addressed all three comments from the latest pass in 07e5fc29, and rebased onto `main` (69839695) to pick up #4339.

- **Expiry counter double-count** — `WheresitupExpiredJobsTotal` now increments only after the prune is saved. `RemoveJobIDs` reopens with `Load()`, so a failed save was discarded and the same jobs were re-pruned and re-counted next cycle.
- **`writeJobState` comment** — was stale since `Save()` stopped pruning in 5ab9ebd3; it now gives the real reason (the `Add*` helpers stamp one clock per call, so they can't build a mix of ages).
- **Changelog** — added the empty-location-mapping sentence, which was a separate user-visible change with no coverage in the entry.

Not changed: no test for the failed-save path. It needs a read-only state file, and `go test` runs as root in the CI container where 0444 doesn't block the write, so the test would pass without exercising the branch. Reasoning is in the thread.

`go build`, `go vet`, `gofmt` and `go test -race ./controlplane/internet-latency-collector/...` all clean locally after the rebase.

Separately, the `go-test` failure on the previous run was a testcontainers reaper flake in `telemetry/gnmi-writer` (`TestIntegration/SystemState`), unrelated to this branch — the identical failure is on `main` at run 34265066380. It passed on re-run. Worth its own PR: ClickHouse setup there has no retry loop while Redpanda does, and `isRetryableContainerStartErr` matches `"timeout"` but not the `"connection timed out"` the reaper actually returns.

PR #4353, opened 2026-09-17
serviceability: add the feed chain to feed accounts

martinsander00:
Resolves: https://github.com/malbeclabs/infra/issues/2654

## Summary

- Wrote a new field on the Feed account called chain, which stores Unspecified, Solana, or Hyperliquid.
- `doublezero feed create --chain solana` writes Solana. `doublezero feed update --chain hyperliquid` changes a stored chain.
- An existing account still decodes as Unspecified. The migrate issue writes the first real chain.

## Testing

- Added unit tests


bgm-malbeclabs (commented):
Reviewed the byte layout by hand and empirically: `feed.bin` grew by exactly one appended `0x02`, all four decoders (Rust, Go, Python, TS) read the tail in the same field order as the Rust struct, the Go decoder against the existing `feed_legacy.bin` still lands on `status=Active, chain=0`, and `borsh-incremental` only substitutes a field default at true EOF, so an out-of-range chain discriminant is rejected rather than silently zeroed.

**The correctness core is sound.** No wrong condition, no off-by-one, no dropped error, no broken caller. Four low-severity notes inline.

Checked and clean: field order matches the Rust struct in all four decoders; `status_index = len - 1 - 32 - 8 - 1` is right; account growth on first update is handled; no fixed-offset or `dataSize` feed decoder exists outside the SDKs; no Go, Python, or TypeScript consumer builds a `Feed` literal the new field would break; `e2e/feed_crud_test.go` parses `feed list --json` into a partial struct, so the new column is additive; `FeedChain` missing from `enum_strings.json` matches the existing `FeedStatus` gap.

One cross-PR question, since #4354 depends on the answer: see the CHANGELOG comment below. If `Unspecified` cannot tell a legacy feed from a deliberately-unset one, then #4354's once-only migrate guard is guarding a state it cannot identify.

smartcontract/programs/doublezero-serviceability/src/state/feed.rs
```diff
@@ -46,6 +46,43 @@ impl fmt::Display for FeedStatus {
     }
 }
 
+/// The chain a feed publishes for. Monthly rewards copy this onto each `FeedDistribution`.
```
bgm-malbeclabs: **low** — This doc comment promises something no code does: "Monthly rewards copy this onto each `FeedDistribution`."

`FeedDistribution` belongs to the feed-subscription program, is a 120-byte bytemuck Pod (`sdk/shreds/go/state.go`, pinned by `TestStructSizes`), and has no chain field. Nothing in this PR copies anything onto it.

The risk is a reader believing the chain already reaches monthly distributions and building per-chain reward reporting on a field that is never populated. Either drop the sentence or mark it as planned.

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ All notable changes to this project will be documented in this file.
...
+- CLI
+  - `doublezero feed create --chain` and `doublezero feed update --chain` take `solana`, `hyperliquid`, or `unspecified`. Create without the flag writes `Unspecified`. Update without the flag leaves the stored value. `doublezero feed list` prints the chain after the name. (#4353)
+- SDK
+  - `CreateFeedCommand` and `UpdateFeedCommand` carry the chain. Go, Python, and TypeScript decoders read the trailing byte and decode a short account as `Unspecified`. The feed fixture ends with `Hyperliquid` (2). (#4353)
+- Serviceability
+  - `Feed` carries `feed_chain`: `Unspecified` 0, `Solana` 1, `Hyperliquid` 2. Monthly rewards copy this onto each `FeedDistribution`. A feed written before the field decodes as `Unspecified`, so a later migrate can tell it from a feed that already chose a chain. `CreateFeed` writes the caller's value and accepts `Unspecified`. `UpdateFeed` changes it only when the argument is set. (#4353)
```
bgm-malbeclabs: **low** — This line is not true as written: "A feed written before the field decodes as `Unspecified`, so a later migrate can tell it from a feed that already chose a chain."

`feed create` defaults `--chain` to `Unspecified` (`smartcontract/cli/src/feed/create.rs:26-28`) and `CreateFeed` accepts it, so a feed created **after** the upgrade with no `--chain` is byte-identical to a legacy feed. The variant's own doc comment says as much ("A feed written before this field, **or created without a chain**"), which contradicts this entry.

Why it matters for #4354: operators create feeds without `--chain`, then a migrate keyed on "Unspecified means legacy" rewrites feeds whose owners deliberately left the chain unset. If distinguishability is the goal, `feed create` has to require a chain, or the field needs a separate "set" marker.

smartcontract/cli/src/feed/create.rs
```diff
@@ -23,6 +23,9 @@ pub struct CreateFeedCliCommand {
     /// Multicast group pubkey or code joinable in this metro (repeatable)
     #[arg(long = "group", value_parser = validate_pubkey_or_code, num_args = 1..)]
     pub groups: Vec<String>,
+    /// Chain this feed publishes for
+    #[arg(long, default_value_t = FeedChain::Unspecified)]
+    pub chain: FeedChain,
```
bgm-malbeclabs, line 28: **low, release ordering** — During a rollout where the CLI ships before the program upgrade, `doublezero feed create --chain solana` returns a success signature and writes a feed with no chain. The program decodes instruction args through `borsh-incremental`'s generated `TryFrom<&[u8]>`, which does not check for trailing bytes, so an un-upgraded program silently ignores the chain byte.

`feed update --chain solana` alone fails loudly (the old two-field args decode to `default` and `process_update_feed` rejects that), but `feed update --name X --chain solana` renames and silently drops the chain.

This is the established pattern for every incremental arg here, so it is a deploy note rather than a code defect: program first, then CLI. Or have the CLI check the program version.

smartcontract/sdk/go/serviceability/fixture_test.go
```diff
@@ -240,6 +240,20 @@ func TestFixtureFeed(t *testing.T) {
 	require.Len(t, feed.Groups, 2)
 	assert.Equal(t, byte(0xE2), feed.Groups[0][0])
 	assert.Equal(t, byte(0xE3), feed.Groups[1][0])
+	assert.Equal(t, serviceability.FeedStatusRetiring, feed.Status)
+	assert.Equal(t, int64(-1_764_547_200), feed.RetiresAt)
+	assert.Equal(t, serviceability.FeedChainHyperliquid, feed.FeedChain)
+}
+
+func TestFixtureFeedWithoutChainByteDefaultsUnspecified(t *testing.T) {
```
bgm-malbeclabs, line 248: **low, coverage gap** — This PR adds the *entire* RFC-28 tail read to the Go decoder (nine fields plus the `hasRfc28Tail` -> `FeedStatusActive` fallback), not just the chain byte, yet the only backward-compatibility test truncates one byte off the modern fixture. That never exercises the `hasRfc28Tail == false` branch.

Python (`test_legacy_deserialize`) and TypeScript both assert against the shared `feed_legacy` fixture; Go does not, although `sdk/serviceability/testdata/fixtures/feed_legacy.bin` already exists.

I ran it by hand and the path is correct today, so this is coverage, not a live bug. But it is the one path where a future edit to `DeserializeFeed` would regress Go alone and silently report every legacy catalog feed as `Pending`, pulling it out of service. Adding a `feed_legacy` case is three lines.

smartcontract/sdk/rs/src/commands/feed/update.rs
```diff
@@ -111,11 +159,131 @@ mod tests {
...
+            err.to_string(),
+            format!(
+                "{} feeds exceed the {MAX_FEEDS_PER_TRANSACTION}-feed transaction limit; send one transaction per chunk",
+                MAX_FEEDS_PER_TRANSACTION + 1
+            )
+        );
+    }
+
+    #[test]
+    fn max_feed_batch_fits_transaction_size() {
```
bgm-malbeclabs, line 269: **The 8-feed guard is a count, but the failure is a size — and this test only proves the easy case.**

`chain_args()` is the smallest payload the command accepts: no name, no groups. I worked the serialization by hand for that case — 14 unique accounts, 5 metas per `update_feed`, 5 bytes of args — and get roughly 670 bytes. Comfortable.

But the CLI writes `--name` and `--group` to *every* feed in the batch, and those bytes repeat in all 8 instruction payloads:

- a 64-char name (`MAX_FEED_NAME_LEN`) adds ~576 bytes across the batch
- one group adds ~288
- `MAX_FEED_GROUPS = 20` adds ~5100

So `feed update --pubkey x8 --name <64 chars> --group g1` is already over 1232 and fails at send time with an opaque error, while `MAX_FEEDS_PER_TRANSACTION` says it is fine.

Two ways out. Build the message and chunk on measured size, which defines the error out of existence. Or, at minimum, change this test to use worst-case args — it should fail today, and that failure is the point.

(My byte counts are hand-computed, not run. The direction holds either way: the test's args are the cheapest possible and the command's are not.)

smartcontract/cli/src/feed/update.rs
```diff
@@ -58,23 +76,55 @@ impl UpdateFeedCliCommand {
...
         if let Some(new_groups) = &groups {
-            unsubscribe_orphans(
-                client,
-                out,
-                &pubkey,
-                &feed.code,
-                new_groups,
-                self.force_unsubscribe,
-            )?;
+            for (pubkey, feed) in &feeds {
```
bgm-malbeclabs, line 71: **`unsubscribe_orphans` runs for every feed before any update is sent.**

Its own errors read "the feed change was not submitted" and "nothing was changed on the feed" (`guard.rs:99`, `guard.rs:128`). That contract holds for one feed. In a batch it does not: feed 1's orphan unsubscribes are already on the ledger when feed 3 bails, and those feeds are left holding groups nothing is subscribed to.

It also runs a full `scan(client)` per feed, so 8 feeds means 8 full ledger scans.

smartcontract/cli/src/feed/update.rs
```diff
@@ -2,17 +2,32 @@ use crate::{
...
+        get::GetFeedCommand,
+        update::{UpdateFeedCommand, MAX_FEEDS_PER_TRANSACTION},
+    },
+    Feed, FeedChain,
+};
+use solana_sdk::pubkey::Pubkey;
+use std::{collections::HashSet, io::Write};
 
 #[derive(Args, Debug)]
+#[clap(group(ArgGroup::new("target").args(["pubkey", "code"]).required(true)))]
```
bgm-malbeclabs: **This is now a field-for-field copy of `FeedTargetArgs`** — same three arguments, same `ArgGroup`, same validators — and `resolve_feeds` below rebuilds a `FeedTargetArgs` for the non-`--pubkey` path anyway.

That struct's own doc says feed selection is load-bearing and warns that copies drift. Making its `pubkey` a `Vec<String>` and giving it a `resolve_many` keeps the decision in one place, and every other feed verb keeps working unchanged.

smartcontract/cli/src/feed/update.rs
```diff
@@ -58,23 +76,55 @@ impl UpdateFeedCliCommand {
...
+                code: self.code.clone(),
+                exchange: self.exchange.clone(),
+            }
+            .resolve(client)?]);
+        }
+
+        let mut seen = HashSet::new();
+        let mut feeds = Vec::new();
+        for arg in &self.pubkey {
+            let (pubkey, feed) = client.get_feed(GetFeedCommand {
```
bgm-malbeclabs: One `get_feed` round trip per pubkey, and it happens *before* the dedup check, so a repeated pubkey costs a full call just to be discarded.

`get_feeds` in the sibling `resolve.rs` already batches feed keys into a single `getMultipleAccounts` — its doc comment describes exactly this case.

smartcontract/cli/src/feed/update.rs
```diff
@@ -58,23 +76,55 @@ impl UpdateFeedCliCommand {
...
+            }
         }
 
-        let signature = client.update_feed(UpdateFeedCommand {
-            pubkey,
-            name: self.name,
-            groups,
-        })?;
+        let pubkeys: Vec<Pubkey> = feeds.iter().map(|(pubkey, _)| *pubkey).collect();
+        for chunk in pubkeys.chunks(MAX_FEEDS_PER_TRANSACTION) {
```
bgm-malbeclabs, line 84: Minor, but it feeds the sizing problem above: `--pubkey x8 --name X` writes one name onto eight feeds. The argument help says so, so it is deliberate, but it is a quiet footgun and it is also the main payload driver in the transaction-size note.

Refusing `--name` together with more than one `--pubkey` would remove both.

smartcontract/programs/doublezero-serviceability/src/state/feed.rs
bgm-malbeclabs: Still open from the last round: "Monthly rewards copy this onto each `FeedDistribution`" describes something no code does.

`FeedDistribution` lives in the feed-subscription program, is a 120-byte bytemuck Pod pinned by `TestStructSizes`, and has no chain field. `CHANGELOG.md:24` now repeats the same sentence, so it is in two places.

Drop it, or mark it as planned.

bgm-malbeclabs (commented):
Re-reviewed at `9c4ba969`. Two of my four are fixed:

- **CHANGELOG** — the "so a later migrate can tell it from a feed that already chose a chain" claim is gone. Good.
- **Go legacy coverage** — `TestFixtureFeedLegacy` added, and the modern fixture now asserts the RFC-28 tail too. That was the fix I wanted.

`create.rs` release ordering needed no code change; it stays a deploy note (program first, then CLI). The `FeedDistribution` doc comment is still there, in two places now — inline below.

**The PR also grew.** It now absorbs batched `feed update` (`--pubkey` repeatable, dedup, `MAX_FEEDS_PER_TRANSACTION = 8`, `send_instructions`). That is close to the loop I suggested on #4354 instead of a dedicated `MigrateFeed` instruction, so it is worth deciding whether #4354 still needs to exist before it merges.

New surface, five new notes inline. The first one I would call blocking: the 8-feed limit is a count, but the failure it guards against is a size, and the test that checks it uses the smallest args the command accepts.

smartcontract/sdk/rs/src/commands/feed/update.rs
martinsander00: Refused `--name` and `--group` when more than one feed is named, so an 8-feed batch stays chain-only and stays under the size cap. https://github.com/malbeclabs/doublezero/commit/69b4d3fe7

smartcontract/cli/src/feed/update.rs
martinsander00: Same refuse. `--group` on more than one feed no longer runs the orphan guard across a batch. https://github.com/malbeclabs/doublezero/commit/69b4d3fe7

smartcontract/cli/src/feed/update.rs
martinsander00: `FeedTargetArgs.pubkey` is now a list. `resolve` still takes one pubkey. Update uses `resolve_many`. https://github.com/malbeclabs/doublezero/commit/69b4d3fe7

smartcontract/cli/src/feed/update.rs
martinsander00: `resolve_many` dedups first, then reads the keys in one `getMultipleAccounts`. A missing key fails. https://github.com/malbeclabs/doublezero/commit/69b4d3fe7

smartcontract/cli/src/feed/update.rs
martinsander00: Treated this as the same case as the size note. `--name` with more than one `--pubkey` is refused. https://github.com/malbeclabs/doublezero/commit/69b4d3fe7

smartcontract/programs/doublezero-serviceability/src/state/feed.rs
martinsander00: Dropped the sentence from the rust doc and the changelog. https://github.com/malbeclabs/doublezero/commit/69b4d3fe7

bgm-malbeclabs (approved):

